PR #16, opened 2013-08-06
Submitted a spec for message guarantee

migzai:
Submitted a spec for guaranteed message delivery


DarthGandalf:
`echo-message` supersedes this.

PR #58, opened 2014-03-11
Client-to-client inquiries

kythyria:
This proposes a replacement for the parts of CTCP that are not messages intended for human consumption and therefore are somewhat strange to place in a PRIVMSG.

### DIFF
--- a/extensions/inquiries
+++ b/extensions/inquiries
@@ -49,6 +49,9 @@ the ERRMSG command used in CTCP.
 When receiving an INQUIRE directed at a channel, clients should respond directly
 to the sender, not via the channel.
 
+If the ircd implements a channel mode that blocks non-ACTION CTCP messages, that
+mode MUST also block INQUIRE, IQRESULT, and IQERROR.
+
 ## Known commands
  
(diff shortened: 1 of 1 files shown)


kaniini:
I think that this should be merged into 3.3 intents.  Thoughts?

clokep:
I think this looks pretty reasonable and would likely fit some use cases I have! Are there any thoughts on whether an `INQUIRY` can result in multiple `IQRESULT` messages?

extensions/inquiries
```diff
@@ -0,0 +1,73 @@
...
+command. The arguments are whatever response is appropriate for the command.
+ 
+## Sending to channels
+
+Clients MAY send INQUIRE messages with channels as the target, subject to normal
+access control for the channel in question. As with PRIVMSG, the server forwards
+the message to all other users in the channel. A client receiving such a message
+SHOULD respond in the same manner as if the inquiry had been sent directly to it.
+
+Clients MUST NOT send IQRESULT messages to channels. Servers MUST respond to any
```
SaberUK: Would it not be better to replace `<Target>` with `<Nick>` instead of explicitly rejecting channels?

jwheare:
Closing via #270

PR #59, opened 2014-03-14
First draft for using SRV records with IRC

Argure:
This is a pull request to keep track of the proposal and open a discussion platform. Please do not merge it until there is a higher level of consensus on this proposal.

### DIFF
--- a/extensions/srv-records-tcp-3.2
+++ b/extensions/srv-records-tcp-3.2
@@ -1,0 +1,52 @@
+Using SRV records with IRC (TCP protocol)
+=========================================
+
+Copyirhgt (c) 2014 Patrick Godschalk <argure@argure.nl>
+
+Unlimited redistribution and modification of this document is allowed
+provided that the above copyright notice and this permission notice
+remains in tact.
+
+This specification defines the recommended behaviour of servers and
+clients for using SRV records.
+
+The SRV RR is defined in RFC 2782:
+<https://tools.ietf.org/html/rfc2782>
+
+Servers
+-------
+IRC servers that listen using the tcp protocol, SHOULD have a SRV
+record for the domain name being advertised.
+
+For example, an IRC server with the advertised domain name of
+<irc.example.org> listening on 194 for irc, and 994 for ircs:
+
+    _irc._tcp.irc.example.org    IN    SRV    0 5 194 irc.example.org
+    _ircs._tcp.irc.example.org   IN    SRV    0 5 994 irc.example.org
+
+When listening on multiple ports, a SRV record MUST be assigned for
+each of these.
+
+For example, the above IRC server also listening on 6667 for irc, and
+6697 for ircs:
+
+    _irc._tcp.irc.example.org    IN    SRV    0 5 194 irc.exapmle.org
+    _irc._tcp.irc.example.org    IN    SRV    1 5 6667 irc.example.org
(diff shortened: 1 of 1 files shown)


extensions/srv-records-tcp-3.2
```diff
@@ -0,0 +1,52 @@
+Using SRV records with IRC (TCP protocol)
+=========================================
+
+Copyirhgt (c) 2014 Patrick Godschalk <argure@argure.nl>
```
clokep: Just happen to be reading this and noticed there's a typo in copyright. Cool stuff though! :)

SaberUK:
@Argure Are you still working on this?

extensions/srv-records-tcp-3.2
```diff
@@ -0,0 +1,52 @@
...
+    _ircs._tcp.irc.example.org   IN    SRV    1 5 6697 irc.example.org
+
+
+Clients
+-------
+
+Clients that attempt to connect using the tcp protocol, SHOULD attempt
+to look up the SRV record(s) of the domain name attempting to connect
+to. The client MUST attempt to look up records for both irc and ircs
+to define if the server supports connections for irc, ircs, or both.
```
SaberUK: If there is a SRV record for both irc and ircs we should probably either require or strongly recommend that the ircs record is used.

Argure:
@SaberUK Yes, and now that 3.2 is out I will continue working on it.

dequis:
I think you'll have to reopen the PR, it says it's from "unknown repository" (because the original fork that had the branch is missing) and I suspect it might not magically pick it up if you fork this repo again (also change the filename while you're at it)

extensions/srv-records-tcp-3.2
Mikaela: :+1:, it would fit STS and the goal of deprecating plain text IRC. But if there is invalid certificate there probably should be fallback to irc?

Mikaela:
Is this PR still active? It doesn't seem that anything has been done recently, even the things that were suggested earlier or the typo fix.

extensions/srv-records-tcp-3.2
grawity: With invalid certificates, it's impossible to reliably distinguish between lazy sysadmins and actual attacks. Automatic downgrade in such cases would be a really bad thing.

Mikaela: Automatic downgrade would be more what STS would also do.

1. Connect with plain text and see the STS cap.
2. Connect to 6697 or specified port using TLS.
3.
    1. if certificate isn't valid to the OS or Mozilla's list, return to plain text.
    2. if the certificate is valid to the OS or Mozilla's list, connect using TLS.

Mikaela: I just think different specifications shouldn't conflict each other.

grawity: That seems quite the opposite of what I'd expect from STS...

dequis:
This PR can't be active, it says `unknown repository`, so there's no way to push new commits.

Mikaela:
So is @Argure or anyone going to make a new one or is this suggestion just dead without anything ever going to happen?

attilamolnar:
I'm closing this as it's still in early stages and it wasn't updated for more than 18 months.

tacerus:
Hi,

are there any updates on this? I would highly advocate the use of SRV records for IRC(s).

PR #92, opened 2014-06-25
Possibility of multiplexing servers.

TrainerGuy22:


attilamolnar:
While an interesting concept there have been no updates on this from the authors or from the community for more than one year so I can safely conclude that right now there is no interest in specifying or implementing multiplexing.

PR #100, opened 2014-09-29
Introduce echo-message capability (3.3)

jpnurmi:


extensions/echo-message-3.3.md
```diff
@@ -0,0 +1,63 @@
...
+The capability is useful for clients to get an acknowledgement that a
+message was delivered. It can be also used for measuring lag between
+client and server, that is, the time it takes from sending a message
+to receiving it back.
+
+Furthermore, the capability is useful for users that have multiple
+clients attached to a bouncer. In this scenario, when users send
+messages from one client, the messages get automatically relayed to
+other attached clients. This allows all attached clients to display
+full conversation.
```
dequis: Isn't this covered by self-message? I mean, you can already show these self messages to all other clients except the one that sends it, which is what this spec would dd.

kaniini: This is apparently a more flexible version of `self-message`.  It should replace `self-message` in 3.2, I created a clarification request (#135) allowing that to be implemented using the argument that `self-message` presence would be an errata since it is to be replaced in 3.3 with this one...

PR #105, opened 2014-12-15
extensions: add register-and-verify extension

kaniini:
The `REGISTER` and `VERIFY` commands provide a unified, standardized interface for account creation and validation to the network's authentication layer.  This allows for a network to provide a common interface regardless of what authentication layer they have chosen for their network to operate, i.e. traditional services, a different central authority, or a decentralized model similar to [IRCX][ircx].

   [ircx]: http://en.wikipedia.org/wiki/IRCX

We believe that the benefit of adding a common interface for account creation and validation, is helpful to the end user in discovering the capabilities offered by the authentication and authority layer of the network. Further, we believe it is helpful to client authors, so that they may provide guided account creation and validation for their users, regardless of network configuration.


wilhelmy:
Quakenet also has an accounts system that isn't linked to nicknames, so it definitely would make sense imho.

kaniini:
Well the goal is to supplement the SASL component of IAL (the IRC implementation of SASL) with registration-related commands, so that a user can connect to the network and then register their account using a standardized interface.  This allows us to streamline registration on networks where it's applicable, as clients can provide whatever UI they want for it.

wilhelmy:
Yes, I was just pointing out another big, real-world network that might be interested in having support for decoupled nick- and account names as part of the standard.

kaniini:
This pull request is screwed, I will open a new one once we start working on this.

PR #107, opened 2015-02-06
specification/metadata-3.2: Remove channels.

Aerdan:
This commit also completes the monospaced formatting for commands and
numeric labels.

### DIFF
--- a/specification/metadata-3.2.md
+++ b/specification/metadata-3.2.md
@@ -23,9 +23,6 @@ that services-driven metadata will be encapsulated via `PRIVMSG`.
 ## METADATA
 
 All metadata subcommands will flow through the outgoing `METADATA` verb.
-Metadata may apply to channels, as well; in that case, an optional
-argument is provided prior to the subcommand, as outlined in the format for
-each, if supported.
 
 This specification adds the `metadata-notify` capability for notifications.
 Please see the 'Metadata Notifications' section for more information.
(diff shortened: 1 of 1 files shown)


M2Ys4U:
Why is channel-related metadata being removed?

kaniini:
It's only for 3.2 that it's being scrapped, as we determined it had shortfalls when trying to implement it in mammon:

```
2015-02-06 16:50:51     kaniini Aerdan: is there a serious issue with metadata?
2015-02-06 16:51:49     Aerdan  no, but it makes mention of support for channel metadata, while the MONITOR spec specifically doesn't have such support.
2015-02-06 16:53:01     Aerdan  and there's no specific hive for channel keys, either.
2015-02-06 16:53:14     kaniini send them a pull-request fixing it :P
2015-02-06 16:54:47     Aerdan  so, I'm not sure which direction I want to go in: patching MONITOR to allow channels for metadata-notify and METADATA to include a channel namespace and e.g. channel.topic as a default key, or else scrub metadata for channels.
2015-02-06 16:55:51     kaniini scrub it for now
```

It's his own specification, so there shouldn't be a problem modifying it, right?

kaniini:
Instead this should be closed and we will do some last minute cleanup on `METADATA` notifications.

Aerdan:
This PR is no longer needed; a new one will be issued shortly.

PR #108, opened 2015-02-08
index: add Limnoria as IRCv3.1 compliant bot

Mikaela:
I hope I put it into correct spot following alphabetical order.


Mikaela:
The force-push just now was accidentally writing 2014 instead of 2015.

PR #109, opened 2015-02-08
Add a markup specification

embolalia:
One of the places where IRC falls short is in formatting messages. The control codes are a hack, and do not provide a very good user experience. They also don't allow for italics or links. The codes are foreign to nearly everyone, since they have little use outside of IRC, and their meaning is not obvious without rendering them. A Markdown-like syntax would address these problems, since it was originally designed specifically to follow conventions already in place, and to look sensible in plain text. With that in mind, I've written up a specification for the use of Markdown in IRC messages.

Mostly, this is just a subset of normal Markdown. I've left out any multi-line things, since they obviously won't have use in a single-line protocol. I've also added a color marker, of the form `@@(color)text@@`. `@@` was chosen simply for being a typeable string that isn't in common use; if there's another one that would be better (or is in use for a similar purpose in another MD flavor), it should be used instead. I haven't yet added background color support, but it would be easy enough to do. There are other places where the spec should be tweaked, but I'd like to get some input and see if there's buy-in to the general idea before getting too deep into it.

As a proof of concept, I've tweaked qwebirc to support this spec, and am running a server [here](http://eagle.embolalia.com:9090/?channels=#formatting). This is just a quick patch, so it probably has some bugs. It is not meant to be a reference implementation, just something to show off the general idea.

### DIFF
--- a/extensions/markup-3.2.md
+++ b/extensions/markup-3.2.md
@@ -1,0 +1,131 @@
+# IRC Message Markup
+
+Copyright (c) 2015 Edward Powell <me@embolalia.com>.
+
+Unlimited redistribution and modification of this document is allowed
+provided that the above copyright notice and this permission notice
+remains in tact.
+
+## What message markup attempts to solve
+
+Currently, styled text in IRC messages is achieved through the use of control
+codes. These codes are bytes within the range of unprintable ASCII characters,
+with color formatting followed by printable characters. Clients must know what
+these codes are and handle them, even if they don't support the functionality,
+in order to avoid showing garbage to the user. Few, if any, keyboards natively
+support the entry of these codes, essentially requiring key bindings for entry.
+Moreover, there is no formal specification of how these codes should work,
+instead leaving the definition of this behavior to what is supported by clients
+*de facto*.
+
+## Overview of markup elements
+
+Message markup is defined through the use of a number of common characters. For
+the purposes of the below, text is considered "wrapped" by a string if it is
+immediately preceded and immediately followed by the given string, without
+containing the string.
+
+* Text which is wrapped by one or more asterisks (`*`) or underscores
+  (`_`) SHOULD be rendered as emphasized. If a client is unable to support one
+  form of emphasis, another form MAY be supported.
+  * Where the text is wrapped by a single asterisk (`*`), it should be rendered as
+    italic.
+  * Where the text is wrapped by two asterisks (`**`), it should be rendered as
+    bold.
+  * Where the text is wrapped by a single underscore (`_`), it should be
+    rendered as underlined.
+* Text wrapped by square brackets (`[` and `]`) which is immediately followed
+  by text which is wrapped by parentheses (`(` and `)`) SHOULD be rendered as a
+  link.
+  * The text of the link MUST be the text within the square brackets.
+  * The target of the link MUST be the text within the parentheses.
+  * If the target of the link contains a colon (`:`) or period (`.`), it should
+    be treated as a URI.
+  * If the target of the link contains neither a colon nor a period, it should
+    be treated as either a nickname or a channel on the network from which the
+    message was sent.
+  * If the target of the link is being treated as a URI, any characters which
+    are not valid in a URI SHOULD be escaped appropriately. Such parsing MUST
+    NOT escape URIs which are already valid.
+  * If the target of the link contains a period, but not a colon, it MUST be
+    prepended with `http://`.
+  * Clients MAY limit rendering URI links to those with certain URI schemes. If
+    this is done, the full text MUST be displayed instead, including the URI.
+  * If the link appears within a block which is otherwise formatted, the text
+    of the link SHOULD be simmilarly formatted. The text of a link MUST NOT
+    contain another link.
+* Text wrapped by two "at" symbols (`@@`) SHOULD be rendered in a different
+  color.
+  * This color SHOULD be consistent with colors used for other important text,
+    for example messages containing the user's nickname.
+  * If the initial pair of symbols is immediately followed by text wrapped in
+    parentheses, the text in the parentheses should be parsed as a color.
+* Text wrapped by a single backtick (`\``) SHOULD NOT be formatted by any of
+  the above rules. It SHOULD be rendered in a monospaced font or, where
+  monospaced font is already the default, it SHOULD be differentiated from the
+  rest of the text (for example, by a slightly different background color).
+  * Text wrapped by three backticks (`\`\`\``) should also be rendered as such,
+    noting that single backticks within this text will be rendered literally
+    along with all other formatting characters.
+* Where any formatting mark is immediately preceeded by a backslash (`\\`)
+  which is not, itself, immediately preceeded by a backslash, the character
+  (but not the backslash) should be rendered, and the formatting rule should
+  not be considered to have been matched.
+
+
+## Colors
+
+Where the color is specified as one of the numbers or names below, it SHOULD be
+rendered in a color simmilar to the one with the RGB color code specified.
+Where a two-digit number starting with 0 is used, it MUST be treated as the
+appropriate one digit number. A color given as a number MUST be treated the
+same as the corresponding name. Color names MUST be case-insensitive.
+
+Clients SHOULD prefer using color names to numbers or codes, especially in user
+interfaces. Clients MAY allow users to enter locale-specifc names for colors,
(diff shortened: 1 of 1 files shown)


extensions/markup-3.2.md
```diff
@@ -0,0 +1,131 @@
...
+
+* Text which is wrapped by one or more asterisks (`*`) or underscores
+  (`_`) SHOULD be rendered as emphasized. If a client is unable to support one
+  form of emphasis, another form MAY be supported.
+  * Where the text is wrapped by a single asterisk (`*`), it should be rendered as
+    italic.
+  * Where the text is wrapped by two asterisks (`**`), it should be rendered as
+    bold.
+  * Where the text is wrapped by a single underscore (`_`), it should be
+    rendered as underlined.
```
grawity: This seems different from the current practice everywhere (`*foo*` for bold, `_foo_` for italic).

extensions/markup-3.2.md
```diff
@@ -0,0 +1,131 @@
...
+* Text wrapped by square brackets (`[` and `]`) which is immediately followed
+  by text which is wrapped by parentheses (`(` and `)`) SHOULD be rendered as a
+  link.
+  * The text of the link MUST be the text within the square brackets.
+  * The target of the link MUST be the text within the parentheses.
+  * If the target of the link contains a colon (`:`) or period (`.`), it should
+    be treated as a URI.
+  * If the target of the link contains neither a colon nor a period, it should
+    be treated as either a nickname or a channel on the network from which the
+    message was sent.
```
grawity: Channel detection should be done earlier (before URI detection), based on whether the first character is a valid channel type. It's valid for channel names to have periods, for example.

extensions/markup-3.2.md
embolalia: I wanted to match as many features of the current system as possible, which includes underline. I agree, using it other than for links isn't a good UX choice, but people do.

I've seen `*` used for italics and bold. I think Markdown is probably the most familiar plaintext-readable markup with support for links, so I wanted to stick with it where possible. Markdown, being HTML-focused, maps it to `<em>`, which doesn't *have* to be italics, but generally is.

SaberUK:
I honestly don't think that this is feasible. We would be better off having a robust specification for what is currently in use and maybe working from there.

kythyria:
If you really want markdown-style formatting entry, do it in your client of choice. Also, this is something I'd intentionally *not* implement fully on a mobile client, because then you can't see link targets.

extensions/markup-3.2.md
embolalia: I considered this. The main reason I didn't is that it would require the `CHANTYPES` value from `ISUPPORT` to be available within the parsing code, which may be difficult in some implementations. It's certainly open to discussion, though.

extensions/markup-3.2.md
grawity: Regardless of how familiar Markdown is on the web, it should not have priority over existing conventions on IRC.

kythyria:
@SaberUK [This](https://gist.github.com/kythyria/5692490) is what Konversation does.

extensions/markup-3.2.md
kaniini: :+1: to what grawity said, we should maintain convention that already exists if we suggest how IRC messages should be interpreted.  A better approach if you wanted *explicit* bold and italic would be to add additional inline codes for italic (`0xa`?).

extensions/markup-3.2.md
kaniini: :-1: to this part.

embolalia:
@SaberUK: The issue with what's currently in use is that, beyond not providing some features, it's a horrible system. The characters are meaningless, and not human-friendly. They aren't particularly bot-friendly either. As the document @kythyria posted points out, there isn't even really one common way as it stands. Different clients use different variations, and any standardization effort would require asking some clients to drop their variants. Given that this proposal is written with the intention that unparsed text is just as readable, and the intention just as clear, as parsed text, I would argue that this is at least as feasible as standardization of the current system.

---

@kythyria: Doing it just in one client without standardization would lead to the same kind of hodge-podge we have today. It would address the input accessibility issue, yes, though not the rest of the problems. And while links may not be desirable in your client, they would be highly useful in others. If a client were to implement an input mechanism, it wouldn't matter, since there would still be no standard over the wire. 

Not implementing this fully is totally acceptable (after all, I don't know that any terminals support links, so you won't see it fully in weechat and the like). This is written, unlike the existing system, with the express intent that it fall back gracefully with no effort from the client. None of the features are mandatory. What I'm proposing here is to make sure that *if* you want a feature in your client, you know exactly what the protocol is for it.

wilhelmy:
mIRC has some control code for italics.
https://smuxi.im/issues/show/463 (according to this bug report apparently 0x1D)

:-1: on markdown support, having two conflicting standards for text formatting doesn't help anyone, IMO.

SaberUK:
@embolalia

The problem is that as flawed as the current system is there is a rough consensus about the core features even if some things are different. Most clients have their own method of inserting colour codes (e.g. `#red,#blue hello world #default` in Linkinus) so the fact that the control characters are obscure is not really a problem. We would be better off working towards unifying what is currently in use than going off on our own tangent.

There are also some implementation concerns about this new system. Due to compatibility reasons it will require servers to do conversion between legacy and modern formats for clients that do and do not support the new system. This is going to require servers to implement a full parser for all markup formats which could cause an intolerable amount of server resources to be consumed. Also, I have some concerns about your link proposal. A lot of IRC use involves CLI and mobile clients which do not have the ability to hover over links. This is means that a malicious user could post something malicious like `[https://private-information.com/](http://totally-not-malware.com/)` and fool users into opening it.

extensions/markup-3.2.md
DarthGandalf: There is already an inline code for italic, though it's implemented in less number of clients than bold and underline.

DarthGandalf:
And we lose ability to send code over IRC. (perl, c++, etc)

clokep:
> One of the places where IRC falls short is in formatting messages. The control codes are a hack, and do not provide a very good user experience.

I've never found IRC formatting to fall short. Granted the "specs" for providing colors over IRC are a bit wonky and incompatible, but the other formatting seems to work OK (although it could certainly use standardization!). I believe the control codes are all non-printable characters; thus they shouldn't impact user experience at all (unless they're parsed and displayed, which is a positive to user experience IMO). (I guess this isn't true if you're printing hex or something for non-printables, but that seems less than ideal.)

I don't see how adding an additional competing way to do formatting will help IRC. Additionally, to reiterate other people's comments: this specification conflicts with generic plaintext formatting that is in common usage: `*bold* /italic/ _underline_ -strike- |preformatted|`.

kaniini:
yeah, please just reject this pull request now.

embolalia:
@SaberUK, as I said, the point is to create a standard which does not need to be parsed at all in order to be understood. There would be no need for translation, as the intent would still be understood. There would be no issue with clients choosing not to implement the link feature, as the purpose of this proposal (and Markdown in general) is that it is still understood as plaintext.

@DarthGandalf it's worth noting that systems like Slack and HipChat (as well as websites like GitHub) which use similar markup do not have any issue with sharing code, because it can simply be placed in backticks. It's a slight change in habit, granted, but it is not the removal of an ability.

---

It seems pretty clear that there won't be buy-in for a full Markdown-style syntax. This spec wouldn't be difficult to tweak to describe the existing de-facto system, and I'm willing to do so. I'd like to see code blocks and links stay in it in some form. They will add a worthwhile feature for some clients and, as proposed (which, to reiterate, is in a form which is understood by a human reader even if no parsing whatsoever is performed), would be at worst neutral to the rest. Would there be interest in this?

SaberUK:
@embolalia So what you're proposing is we have new clients sending formatted messages which old clients can't understand? That seems very unfriendly for users.

kythyria:
Full markdown is impossible in IRC in any case, as it relies heavily on newlines and spacing.

I'm not fond of things overloading graphic characters on the wire; this is a proposal which will mangle the text of anyone not using it, unless server side conversion takes place. That's something I turn off in any client, and hate in services that don't give you the choice.

This also makes colour blocking modes more resource intensive to implement; particularly if you want to eat colours alone.

kaniini:
I'll say it again, someone please just reject this specification.

grawity:
Pretty much nobody here seems to be in favor of this. Closing.

Issue #112, opened 2015-02-20
Have commands supported by server in ISUPPORT and send those to server automatically


Mikaela:
What was talked about  https://github.com/ircv3/ircv3-specifications/issues/111 at IRC.

Someone else can probably also explain this issue better. Have something in ISUPPORT to list the commands that the server supports and their syntax, so clients can know that server supports those commands and send those commands to server directly without `/quote`. Primarily `/nickserv` etc.

Mikaela:
NOTE: I still think that client command should take priority over server commands. E.g. `/set`, `/connect`...

Mikaela:
Commands that come available as you oper/deoper? Could cap-notify be used here?

Mikaela:
IRC discussion:

```
[05:36:55] <Mikaela> was there some IRCv3 thing that allows user who has already connected to network see that SASL is supported?
[05:37:29] <SaberUK> CAP ls
[05:37:38] <Snerf> yup, CAP LS
[05:38:01] <Mikaela> I am just woondering as Limnoria's documentation on SASL is based on assumption that the netwoork uses Atheme and Anope doesn't have SASLSERV or anything visible.
[05:38:03] <Mikaela> I think my ZNC captures that :(
[05:38:16] <Snerf> ZNC does capture it
[05:39:06] <Snerf> CAP LS is when you start connecting, which the bouncer would handle
[05:39:07] <Attila> thats broken, should look for cap sasl
[05:39:16] <Attila> saslserv has nothing to do with sasl its an arbitrary client
[05:39:33] <Attila> sasl can be done perfectly fine without it
[05:39:45] <Mikaela> saslserv being visible says that the services support sasl, but Anope again doesn't have it
[05:39:56] <Attila> <Attila> thats broken, should look for cap sasl
[05:40:24] <Mikaela> I see, that again doesn't work with bouncer users or average user
[05:40:46] <Snerf> an average user doesnt know what sasl even is :)
[05:41:09] <Snerf> they think its ssl
[05:41:12] <Mikaela> with help from freenode or https://freenode.net/sasl they might learn
[05:41:27] <Snerf> ya, but they are "average" they dont want to know
[05:41:27] <Kyth> Average user doesn't know what cap ls is either, or how to send it, I should think
[05:41:30] <Snerf> and dont care
[05:41:41] <Snerf> they just wanna chat
[05:41:52] <Attila> why is a bouncer a problem here?
[...]

PR #114, opened 2015-02-23
Command discovery/description

kythyria:
This is quite draft; I'm not quite sure some of the syntax is right. It tries to address #112 without requiring that the banner sent by servers be EVEN LONGER.

There's no cap, because `ISUPPORT` is used instead.

### DIFF
--- a/extensions/command-discovery-3.3.md
+++ b/extensions/command-discovery-3.3.md
@@ -25,8 +25,8 @@ machine-readable, in-band manner.
 
 A server supporting this specification must announce `CMDLIST=<token>` in
 its `RPL_ISUPPORT` banner. The `<token>` may be any valid string provided
-that it changes whenever the reply to `LISTCMDS` does, such as a sequence
-number or hash of the reply.
+that it becomes a different sequence of octets whenever the reply to
+`LISTCMDS` changes.
 
 ## Requesting a command list
 
@@ -38,15 +38,12 @@ the end with `RPL_CMDLISTEND`.
 ## Format of `RPL_CMDLIST`
 
 ```
-RPL_COMMANDLIST <command> <context> <parameter1> <parameter2> ...
+RPL_CMDLIST <command> <context> <parameter1> <parameter2> ...
 ```
 
-If there are more than 13 parameters, they are to be placed as a space-separated
-list in the final parameter, with a leading colon indicating the beginning of a
-final parameter containing spaces:
-```
-RPL_COMMANDLIST FOO nick o:a o:b o:c o:d o:e o:f o:g o:h o:i o:j o:k o:l :o:m o:n :label:Frob 15 things
-```
+Clients must be able to parse an unlimited number of parameters in `RPL_CMDLIST`,
+as a command with 15 parameters will be represented as a `RPL_CMDLIST` with
+17 total arguments.
 
 `command` is the verb used by this command (so, the same value might recur
 if subcommands are involved).  
@@ -61,9 +58,10 @@ nick        | Any nick
 nickchan    | Nick in a channel
 target      | Nick or channel, or possibly a targetlist
 self        | The user issuing the command
-hetwork     | The network the user is connected to
+network     | The network the user is connected to
 other       | Something else
 ```
+Non-standard contexts must be vendor-prefixed, as with capabilities: `example.org/group`.
 
 ### Parameters
 Parameters are described in the format
@@ -79,7 +77,8 @@ defined, `c` and `o`. `o` indicates a parameter is entirely optional, while `c`
 indicates that the user should not be required to provide a value explicitly if
 the client can determine it from the context in which the command is invoked.
 
-The following types are defined:
+The following types are defined, and as with contexts, non-standard types may
+be used if they have a vendor prefix.
 ```
 sc          | Subcommand; literal text to include to invoke this one.
 chan        | Channel name
@@ -87,7 +86,7 @@ nick        | Nickname
 user        | Services username (which is usually a nick)
 target      | Nick or channel, or possibly a targetlist.
 pattern     | Pattern such as a wildcarded hostmask
-o           | Some other parameter
+other       | Some other parameter
 text        | Long text (must be the last non-label parameter!)
 label       | Not actually parameter. UI text to display for this command.
 ```
@@ -112,7 +111,7 @@ S: 298 MODE channel chan:c:channel sc:+b pattern:target label:Ban
 S: 298 MODE nickchan chan:c:channel sc:+o nick:c:target :label:Give op
 S: 298 MODE nickchan chan:c:channel sc:-o nick:c:target :label:Take op
 S: 298 ZNC channel sc:detach chan:c:channel
-S: 298 NS self sc:register o::password o::email :label:Register this nickname
+S: 298 NS self sc:register other::password other::email :label:Register this nickname
 S: 299 :End of command list.
 ```
 
@@ -126,7 +125,3 @@ It also indicates that sending `MODE <channel> +o <target>` performs an operatio
 labelled `Give op`, that this should be added to the list of operations that may
 be performed on users in the same channel as oneself, and that `<channel>` and
 `<target>` are readily inferred from context.
-
-## Notes
-If a command is presented as a GUI element, it is recommended that a client
-nonetheless permit the user to enter it textually. 
--- a/extensions/command-discovery-3.3.md
+++ b/extensions/command-discovery-3.3.md
@@ -1,0 +1,132 @@
+# IRCv3 Server Command Discovery
+
+Copyright (c) 2015 Kythyria Tieran <kythyria@berigora.net>
+
+Unlimited redistribution and modification of this document is allowed
+provided that the above copyright notice and this permission notice
+remains intact.
+
+## Rationale
+
+Many servers (and bouncers, which are servers from the point of view of a
+client) implement non-standard commands which it is nevertheless entirely
+reasonable for a client to expose in a user interface in some manner more
+intelligent than passing unrecognised `/command`s on to the server.
+
+For instance, ZNC's "detach" facility, an enormous number of the commands
+in a typical services implementation, or commands only meaningful for one
+network or rare server configuration, are unlikely to be standardised, or
+be explicitly supported by clients.
+
+This document specifies a way to list the commands a server supports in a
+machine-readable, in-band manner.
+
+## Indicating support
+
+A server supporting this specification must announce `CMDLIST=<token>` in
+its `RPL_ISUPPORT` banner. The `<token>` may be any valid string provided
+that it changes whenever the reply to `LISTCMDS` does, such as a sequence
+number or hash of the reply.
+
+## Requesting a command list
+
+The client sends a `LISTCMDS` message to the server, with no parameters.
+
+The server responds with a sequence of `RPL_CMDLIST` messages, indicating
+the end with `RPL_CMDLISTEND`.
+
+## Format of `RPL_CMDLIST`
+
+```
+RPL_COMMANDLIST <command> <context> <parameter1> <parameter2> ...
+```
+
+If there are more than 13 parameters, they are to be placed as a space-separated
+list in the final parameter, with a leading colon indicating the beginning of a
+final parameter containing spaces:
+```
+RPL_COMMANDLIST FOO nick o:a o:b o:c o:d o:e o:f o:g o:h o:i o:j o:k o:l :o:m o:n :label:Frob 15 things
+```
+
+`command` is the verb used by this command (so, the same value might recur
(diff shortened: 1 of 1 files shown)


SaberUK:
:+1:

extensions/command-discovery-3.3.md
```diff
@@ -0,0 +1,132 @@
...
+The last three are of somewhat special nature: `text` must be the last parameter
+other than `label`, since spaces are permitted in this--the payload of, say,
+`PRIVMSG` is a `text` parameter--and `label` does not define a parameter at all,
+but rather suggests a name to use when displaying the command in a GUI, such as
+if it is presented as a menu item.
+
+## Example
+Suppose `RPL_CMDLIST` and `RPL_CMDLISTEND` are `297` and `299` respectively.
+```
+S: 005 EXCEPTS INVEX CMDLIST=571
```
SaberUK: There's probably no point in having a value for the CMDLIST token. If a client supports it they will know what numeric it uses anyway.

robotoad:
How would this information be localized?

kaniini:
@robotoad brings up a good point actually, not everyone speaks english, and not everyone on a predominantly spanish language server for example, wants their commands in spanish etc.

SaberUK:
I was working on a proposal a while back for a language cap. If anyone is interested I can finish that and submit it.

grawity:
I wonder if the server should provide a command ID (to let clients override the label or just recognize it) as well as a few primary translations (including an `en` translation is strongly recommended).

(Or, gettext-style, just English + translations.)

kaniini:
Adding type and context as separate things seems unintuitive to me.  Why can't context *be* type?

kythyria:
> Adding type and context as separate things seems unintuitive to me. Why can't context be type?

@kaniini Because context applies to the entire command, type applies to individual parameters, and some of the values for each don't make sense for the other (eg, `sc` doesn't make sense as a context, and `self` doesn't as a type).

Consider `KICK`. It has a context of `nickchan`, and parameters of `channel`, `nick`, and `text`. `nickchan` doesn't make sense as a type, because why would you put both those things in one parameter. But it certainly works as a context, because `KICK` removes a specific user from a specific channel.

kaniini:
@spb mentions that the format provided by this may not map well to some esoteric things such as charybdis' `SCAN` command.  this is probably not of much relevance however, since the odds of charybdis adopting 3.3 is nil.  but there may be other commands like it.

a better study, thinking about it, would be to consider how this would map to IRCv3.3 `XINFO`, which has similar properties to `SCAN`.

spb:
More than that, this is entirely the wrong approach to implementing and documenting extensions to the protocol.

As the two examples above illustrate, *any* machine-readable description of command syntax that you can come up with will have shortcomings and somebody will want to implement a command that can't be described in that way. More than that, it can't describe to the client what the effect of that command will be.

In the general case, and I suspect in the majority of real-world cases, adding a sensible user interface for non-standard commands requires an understanding of what the command *does*, not just what its argument types are. Any sensible GUI for a particular command requires that the client actually understands the command, and hasn't just taken it out of a list of supported stuff.

To that end, it'd be far better to provide a means for a server to document which feature extensions it supports, which can then be documented either inside or outside the IRCv3 spec, and clients can then add support for those features as and when they want to. Trying to have the server define the client's UI will only end up in reinventing the web browser. That means might or might not look a lot like ISUPPORT.

SaberUK:
>To that end, it'd be far better to provide a means for a server to document which feature extensions it supports

As previously discussed, every major server implementation allows users to define custom aliases. You can't write a specification for that.

spb:
You also can't auto-generate a GUI for it.

kythyria:
@SaberUK You can, sort of. It will look a lot like this one. Compare [XEP-0050](http://xmpp.org/extensions/xep-0050.html) and [XEP-0004](http://xmpp.org/extensions/xep-0004.html) which add up to an attempt to solve a similar problem. It works reasonably well, actually, since the latter's data model is quite simple and so doesn't demand much from a GUI.

kythyria:
@spb
> You also can't auto-generate a GUI for it.

You can for some simple yet common cases, since very little GUI is needed, eg, for `CS KICK`. Channel and nick are implied by, eg, selecting the user from a list of users in the channel, and then prompting for the reason field is easy.

My intention was as much about doing optional parameters the way most clients do for `/part` as any sort of GUI.

spb:
Is the intention to write an IRC specification that works "for some simple yet common cases"?

Because that'd be a pretty bad specification.

kythyria:
There's not much complex you can fit in an IRC message, but if there were a ton of non-context fields in a command that nonetheless gets exposed in a GUI, you could do that by presenting a simple tabular layout, like XEP-0004 clients do. A client could also organise subcommand-heavy verbs, such as `CS`, using a submenu approach.

And this won't ever cover every nuance of what a client can send to the server. That would require a much bigger hunk of data that would be a pain to fit in-band, since you'd need to cover the values of enumerated fields, probably group subcommands explicitly, and I'm not even going to touch modes.

M2Ys4U:
I agree with @spb.

Any auto-generated GUI will have an awful UX. You may be able to encode *syntactic* information but you'll be missing *semantic* information.

----

I still fail to see any compelling rationale for this.

What server vendors should be doing is *specifying* any non-standard commands. If they're any good they'll probably be adopted by other vendors (perhaps even published by the IRCv3 WG, who knows!). Publishing a spec also helps to reduce namespace collisions.

When there's a specification, clients can be notified by including an `RPL_ISUPPORT` token for the spec and toggle any helpful UI.

SaberUK:
@M2Ys4U I recommend reading the entire comment thread for information why that is not feasible.

kaniini:
I think before we really move on this, we need to see something in action as a PoC so we can weigh how useful it really is.  I plan to submit an RFC which requires that large changes be accompanied with a PoC.

M2Ys4U:
@SaberUK I have, I don't see why it's not feasible.

kythyria:
@M2Ys4U You seriously think it's feasible for every single server developer, server *module* developer, and network adminstrator to a) write a proper spec of the extensions introduced, and b) convince the developers of a large number of clients to implement something so niche?

kaniini:
@kythyria Right now the proposed example is usually some ZNC detach channel thing.  In that case I'm not convinced, why can't ZNC make a specification for detaching and reattaching channels instead?  (In fact, a high quality specification documenting bouncer-specific extensions to IRC would be a wonderful contribution to IRCv3.)  The second big example is `NS` and `NICKSERV` aliases, both of which are then proposed as a hint to `IDENTIFY`, which is deprecated by SASL for over 4 years now, so that's kind of pointless to document.

This is why I say a PoC is necessary -- we need to get into a realistic mindset about this functionality in order to properly evaluate it.  Otherwise it becomes a "wouldn't it be nice if..." scenario.  Lots of things are nice, but less things are practical, and spending time on impracticalities slows the IRCv3 process.

kythyria:
@kaniini The other thing is `CHANSERV`. Part of the reason using it is obnoxious compared to `MODE` and `KICK` and so forth is because you *have* to manually specify the channel each time, as that context doesn't travel over the wire.

M2Ys4U:
@kythyria: It's not like I'm asking developers to write an Internet Draft and submit it to the IETF for publication as an RFC on the Standards Track.

It'll be enough to publish a document stating:

1. How the command works (the syntax),
2. What the command does, and
3. Why the command exists.

That should be enough information for any client that wants to implement a UI for it to do so in a way that makes sense. This current spec only deals with `1`

If something is "so niche" that nobody would write a spec for it then surely it's "so niche" that clients don't need a badly integrated UI for it.

Besides, isn't the entire point of the IRCv3 WG to specify, come to a consensus on, and standardise extensions to the IRC protocol?

SaberUK:
@M2Ys4U Here is a specification for arbitrary server aliases as supported by every major server implementation:

Servers may implement any command. Any number and type of arguments are possible.

kythyria:
@SaberUK @M2Ys4U We can extend that:

1. Servers may implement any command. Any number and type of arguments are possible.
2. These commands do *something*. This may or may not correspond to what the name implies, and may or may not happen when you try to use them.
3. It exists because someone thought it was a good idea. Or was drunk. Or a furry.

(I'm guilty of that last one, although because this spec doesn't exist it's triggered by `CTCP-ACTION` starting with a particular string)

clokep:
I still don't see what specific problem this is attempting to solve. This solution seems to be conflating a whole bunch of unrelated issues, as @M2Ys4U alludes to. Additionally, at best you'll end up with a confusing UI that talks in terms of protocol level stuff instead of things the user actually cares about.

In terms of comments for the actual proposal: I strongly dislike the idea of having a label passed along with a command. The major issue I see with this is localization, which has been mentioned before. Even if multiple localized versions are provided, you'll likely run into the issue that IRC doesn't really have a defined character encoding, although most people seem to use UTF-8 now.

One of the arguments that keeps coming up is that it isn't possible since a server can implement arbitrary aliases. How does that behoove a server if a client can't interact without manual commands? Is an overly complicated spec like this really the answer? What aliases are pretty really declaring? Would it make more sense to declare a list of aliases and the "real" command they point to?

Regardless, it seems like some of the commands (e.g. NICKSERV) couldn't even be described in this format since the syntax of each sub-command that can be sent to NICKSERV is fairly arbitrary. If there's an ulterior motive of wanting *specific* functionality...we should concentrate on that (I think the idea of having a bouncer spec is a great idea!)

robotoad:
Following are some issues from the perspective of client developers (please don't take my objections as anything more than a healthy part of the vetting process):

* Localization of user-visible text. @SaberUK has offered a solution for this.
* Native functionality only possible with knowledge of the meaning of a command. For example, typing "/msg " in an input box might display a popover of recently messaged nicknames. Automatic command discovery couldn't possibly supply enough contextual information for every possible interface.
* Poor user experience via auto-generated GUIs that couldn't possibly support all environments.
* Inability to display command help without an active connection to a server that has supplied the information. It could be cached, but it would need to be organized by server and would scale poorly for large lists in terms of usability.
* A move away from codifying extensions toward automation of arbitrary whim without intervention or knowledge on the part of client authors or the greater community, which might benefit from standardization of a useful feature.

A counterargument has been made that it isn't feasible for servers to document their additions or for client authors to track them. As I said before, if servers are to go through the trouble of supplying command syntax and localized help descriptions, it stands to reason the information could be placed in a published document since it's practically a specification already. If it's difficult to imagine client authors tracking published specifications, why would it be easier to imagine them implementing automated command parsing and GUI generation?

In the face of IRCv3's attempts at standardization, this feels like a reversal of direction by encouraging servers to go their own way. In my opinion, this proposal won't be treated as a bridge toward standardization; it will be a crutch used to avoid standardization.

kaniini:
@robotoad that is basically my concern with it, we need to be very careful because we're playing with fire :)

attilamolnar:
> In the face of IRCv3's attempts at standardization, this feels like a reversal of direction by encouraging servers to go their own way. In my opinion, this proposal won't be treated as a bridge toward standardization; it will be a crutch used to avoid standardization.

I agree. I support the idea of documenting some of the currently existing major features (and the commands that go with them) such as callerid, SA* commands and silence so clients can properly integrate them.
On the other hand, I don't think it's realistic to document every feature (command) because there are lots of oper-only features that opers should be expected to be able to use without any assistance from the client. For example the `LOADMODULE`, `NICKLOCK`, `MKPASSWD` commands are of no interest to most users. Some oper-only features could use a proper UI, such as displaying the output of a clone check but I don't think we should worry about things like that for now.

extensions/command-discovery-3.3.md
```diff
@@ -0,0 +1,127 @@
...
+```
+sc          | Subcommand; literal text to include to invoke this one.
+chan        | Channel name
+nick        | Nickname
+user        | Services username (which is usually a nick)
+target      | Nick or channel, or possibly a targetlist.
+pattern     | Pattern such as a wildcarded hostmask
+other       | Some other parameter
+text        | Long text (must be the last non-label parameter!)
+label       | Not actually parameter. UI text to display for this command.
```
ProgVal: Could you add an integer type?

ProgVal:
I think it would be nice to make it possible for bots (and services) to advertise their commands this way.
Do you think it should be in this spec, or in an other one once this PR is merged?

DanielOaks:
I think this would be a very bad type of reply to standardise right now. As others have mentioned, I think this would hurt the standardisation of new IRC features and commands much more than help it. I think this feature directly opposes the working group's goals, and would be used to stifle the standardisation of client commands between different server software. Unless shown otherwise, I don't support this feature in any way.

attilamolnar:
Closing due to lack of support. Moreover it was mentioned that this should be prototyped and demoed before moving forward which have not happened.

PR #115, opened 2015-02-25
sasl-3.2 updates per #103

### LINKED ISSUE #103: Offer SASL capability again if user gets unidentified
I am not sure if the title says anything, but discussion started at https://github.com/atheme/atheme/pull/415#issuecomment-66869694 and there it was said that this is ircv3 thing.

I open this issue, because no one else seems to do it.

kaniini:
Per this evening's discussion on `#ircv3`, we:

 * document the reauthentication ability
 * document the integration with `cap-notify`, and
 * explicitly allow a server to disconnect a user for excessive SASL authentication failures

The latter is to clarify that it is allowed in case anyone was confused about that.

All of this closes #103.

### DIFF
--- a/extensions/sasl-3.2.md
+++ b/extensions/sasl-3.2.md
@@ -2,6 +2,8 @@
 
 Copyright (c) 2014 Attila Molnar <attilamolnar@hush.com>
 
+Copyright (c) 2015 William Pitcock <nenolod@dereferenced.org>
+
 Unlimited redistribution and modification of this document is allowed
 provided that the above copyright notice and this permission notice
 remains in tact.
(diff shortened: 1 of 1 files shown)


extensions/sasl-3.2.md
```diff
@@ -27,6 +31,23 @@ negotiation protocol that allows values (i.e. 3.2).
...
+Servers MUST advertise availability of the `sasl` capability to any clients which have
+requested the `cap-notify` notification.
+
+## SASL Reauthentication
+
+Servers MUST allow a client, authenticated or otherwise, to reauthenticate by
+sending a new `AUTHENTICATE` message at any time.
+
+Servers MAY disconnect ANY client at any time as a result of failed authentication,
+including both unregistered and registered clients.
```
DarthGandalf: But before that it must report the error via the proper numeric, right?

kaniini: Ideally yes.  I will update the spec.

DarthGandalf:
Should sasl capability be not available during netsplit? If so, why would client know that it needs to wait until it will be available (e.g. if it doesn't want to join channels unauthed)?

kaniini:
Yes, I agree, we should hide the sasl cap if it's going to be unavailable in practice.  Many 3.1 implementations already do this anyway, so we'll be documenting the status quo.

DarthGandalf:
It is a question, you can't really agree with a question :P 
freenode doesn't hide sasl during netsplit.
If I don't want to join channels unauthed, earlier I could tell client to use sasl, and if client sees sasl cap, it tries to auth, maybe fails with timeout, but it doesn't join channels. If sasl is hidden when not available, client doesn't see its support upon connect, and gladly joins.

kaniini:
freenode does not hide sasl during netsplit because of a charybdis bug.  In my opinion a client should abort connection if it absolutely wants to pre-auth and SASL is unavailable.

What language do you propose for that?

attilamolnar:
> If sasl is hidden when not available, client doesn't see its support upon connect, and gladly joins.

That is broken behavior. If you want auth and you can't, then the connection should fail instead of continuing. Similar to STARTTLS.

kaniini:
I added this language:

> If a client requires pre-authentication and is unable to obtain the `sasl` capability,
> then the client MUST disconnect and MAY retry the connection.

Rationale: if a client is connecting to a round robin, they may be able to connect to the other side of the netsplit, and thus have access to authentication by cycling the connection.

kaniini:
I *think* this is everything for sasl now.

DarthGandalf:
@attilamolnar yes, makes sense.

PR #118, opened 2015-02-25
Push intents back to 3.3 and mark as experimental in 3.2.

kaniini:
We discuss `intents` in 3.2 as somewhat of a wakeup call for client and server authors,
that way they are fully aware that the `intents` framework is coming and that they hopefully,
may choose to participate in the discussion concerning it.


dequis:
Would be nice to have the text of the commit message (or equivalent text) written somewhere in the markdown files, explaining what the "experimental extensions" section means

What happened to `account-tag`?

kaniini:
`account-tag` was both in core and extension section.  I moved it to `extensions` because that is where it lives in the source tree.

PR #120, opened 2015-02-25
cap-notify changes for #119

### LINKED ISSUE #119: Remove ambiguity concerning `cap-notify` between 3.1 and 3.2 servers
On 3.1 servers, `cap-notify` MUST be shown in CAP LS, because it's a 3.1 server.  However, on 3.2 this is presently optional.  There is no obvious benefit to making the token optional in 3.2, even though it is implied as a 3.1 client may want to take advantage of the feature as it migrates toward adding full 3.2 support.

Requested change:

Remove from `extensions/cap-notify-3.2.md` lines 24-25, the following text:

> In that case cap-notify MAY be presented in response to CAP LS,
> but the server MAY choose not to.

If consensus is reached, I will submit a PR doing this.

kaniini:
this change implements the requested action in the #119 proposal.

closes #119


attilamolnar:
LGTM

PR #121, opened 2015-02-25
metadata-3.2: Major rework.

Aerdan:
Other changes:

  * Remove services section; services should be receiving `METADATA` commands anyhow and metadata should be implicitly persistent.
  * Reword the daemon section to be more explicit.
  * Miscellaneous formatting fixes for enhanced readability.

### DIFF
--- a/specification/metadata-3.2.md
+++ b/specification/metadata-3.2.md
@@ -13,27 +13,26 @@ several mechanisms for doing this, but they typically rely on the presence of
 services and aren't really suitable for transient metadata such as a user's
 current location.
 
-This proposal aims to codify two coexisting mechanisms for working with
-metadata: first, persisting metadata may be configured through services via
-NickServ or an appropriate user authentication agent; second, transient
-metadata may be configured through a client --> server event created
-specifically for this purpose. Both mechanisms will behave identically, save
-that services-driven metadata will be encapsulated via `PRIVMSG`.
+This proposal aims to codify one mechanism for working with metadata: metadata
+may be configured through a client-to-server event created specifically for
+this purpose.
 
 ## METADATA
 
 All metadata subcommands will flow through the outgoing `METADATA` verb.
-Metadata may apply to channels, as well; in that case, an optional
-argument is provided prior to the subcommand, as outlined in the format for
-each, if supported.
 
 This specification adds the `metadata-notify` capability for notifications.
 Please see the 'Metadata Notifications' section for more information.
 
-Targets specified by the following subcommands MUST be an asterisk (`*`) when
-used for a user's own metadata, a valid nickname, or a valid mask as determined
-by the IRC server. Invalid targets MUST be responded to with
+For the purposes of this specification, 'targets' are entities for which
+metadata may be set. Such entities MUST be a valid nickname or channel. As a
+convenient shorthand, an asterisk (`*`) MAY be specified to indicate that the
+target is the client itself. Invalid targets MUST be responded to with
 `ERR_TARGETINVALID`.
+
+Implementations MUST provide a mechanism for controlling the ability to set or
+clear metadata on channels. Such mechanisms SHALL respond to unauthorized
+attempts with `ERR_KEYNOPERMISSION`.
 
 ### METADATA LIST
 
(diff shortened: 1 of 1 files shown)


specification/metadata-3.2.md
```diff
@@ -56,140 +59,147 @@ format of `METADATA SET` MUST be as follows:
...
 
-## IRC Daemons
+Setting metadata for channel:
 
-IRCds MAY have a blacklist or whitelist and may have an option to enforce
-keys against either or neither of them. Implementations may block keys which
-might result in impersonation. It is an error for user-set metadata to have
-any effect on server operations.
+    METADATA #example SET website :http://www.example.com
+    761 #example website :htp://www.example.com
```
attilamolnar: Typo after `:`

kaniini:
Also, the specification says:

> Targets specified by the following subcommands MUST be an asterisk (*) when used for a user's own metadata, a valid nickname, or a channel name. Invalid targets MUST be responded to with ERR_TARGETINVALID.

Does this mean `METADATA kaniini LIST` is invalid if my client is nicknamed kaniini?

kaniini:
@attilamolnar, @grawity, @DarthGandalf i think we're good here (finally).

specification/metadata-3.2.md
```diff
@@ -45,151 +44,161 @@ of `METADATA LIST` MUST be as follows:
...
+extends `MONITOR` behaviour to include subscribing users to notifications for
+those users they are currently monitoring. Clients are also subscribed to
+notifications for channels they join. Clients may discontinue notifications
+for users by issuing a disabling `MONITOR` command, and for channels by
+parting the channel. Clients are automatically subscribed to notifications for
+their own metadata.
+
+Notifications use the `METADATA` event, the format of which is as follows:
 
 `METADATA <Source> <Target> <Key> :<Value>`
```
attilamolnar: Any reason this is not using the usual `:<source> METADATA <target>...` format?

specification/metadata-3.2.md
```diff
@@ -45,151 +44,161 @@ of `METADATA LIST` MUST be as follows:
...
-## WHOIS
+If `METADATA` is supported, it MUST be specified in `RPL_ISUPPORT` using the
+`METADATA` key.
 
-Metadata MUST appear in `WHOIS` output using `RPL_WHOISKEYVALUE` if any is defined
-for the user.
+## Numerics
 
-## Services
+The numerics 760 through 769 MUST be reserved for metadata, carrying the
```
attilamolnar: Should reserve a few more just in case, what do you think?

attilamolnar:
There are no examples at all for `metadata-notify`, please provide a few or let me know if you want me to do it.

specification/metadata-3.2.md
Aerdan: The source could be any entity capable of making a change, not just legal prefixes.

specification/metadata-3.2.md
Aerdan: That range is pretty much what I was given to work with back when I wrote the original specification. If you want to reserve more, we can.

grawity: Not entirely sure, what's the point of reserving numerics without actually defining them? AFAIK the ID ranges have never had any specific meaning or structure (other than `0xx`).

kaniini: So the idea was to reserve numerics ahead of time to hopefully allow migration to numerics which are grouped by topic.

But, now days it seems we want to just move to informational verbs and away from numerics in general.  So I think just claiming the numerics which are actually defined is good for now, since informational verbs in this way won't be coming for a while.

specification/metadata-3.2.md
```diff
@@ -45,151 +44,174 @@ of `METADATA LIST` MUST be as follows:
...
-This specification extends `MONITOR` to enable metadata notifications using the
-above verb. Clients MUST handle all `METADATA` verbs, regardless of
-subscription status. Notifications MUST be sent for the client's own keys,
-regardless of subscription status.
+Metadata propagates to clients automatically under certain conditions:
+
+1. Upon authentication and successful negotiation of `metadata-notify`, clients
+   MUST have their non-transient metadata propagated to them. If none exists,
+   servers MUST send `ERR_METADATAEND`.
+2. Clients SHOULD have current metadata propagated to them.
```
DarthGandalf: When client sets its metadata, server will reply with this?
```
> METADATA nick SET foo :bar
< :server 761 nick!ident@host foo :bar
< :server 762 :end
< :server METADATA nick!ident@host foo :bar
```

specification/metadata-3.2.md
```diff
@@ -45,151 +44,174 @@ of `METADATA LIST` MUST be as follows:
...
+The numerics 760 through 769 MUST be reserved for metadata, carrying the
+following labels and formats:
 
-Metadata SHOULD be presentable via NickServ or similar service in order to
-allow for persistent metadata; the format for encapsulating `METADATA` events
-for persistence MUST be as follows:
+| No. | Label                 | Format                                     |
+| --- | --------------------- | ------------------------------------------ |
+| 760 | `RPL_WHOISKEYVALUE`   | `<Target> <Key> [<Visibility>] :<Value`    |
+| 761 | `RPL_KEYVALUE`        | `<Target> <Key>[ <Visibility>][ :<Value>]` |
```
DarthGandalf: Let's not break IRC messages even more than how they are already broken. There should no parameter, which meaning depends on whether it has `:` before it or not.

kaniini: Yes, please make `<Visibility>` it's own field.

Aerdan: ... It _is_ its own field.

    :irc.example.com 760 user1 account :user1
    :irc.example.com 760 user1 account 0 :user1

are identical except that the visibility is explicitly public in the second line.

    :irc.example.com 760 user1 account
    :irc.example.com 760 user1 account 0

are equally identical and both say that the key was cleared.

specification/metadata-3.2.md
Aerdan: I'm puzzled as to why this comment was applied to this portion of the diff, but clients are only sent metadata notifications for changes they themselves didn't make.

specification/metadata-3.2.md
grawity: But how do you distinguish between

    :irc.example.com 760 user1 account :user1

which sets metadata without visibility, and

    :irc.example.com 760 user1 account 0

which clears metadata with visiblity?

Aerdan:
argh shit, fuck git so much. :|

specification/metadata-3.2.md
```diff
@@ -45,151 +44,174 @@ of `METADATA LIST` MUST be as follows:
...
-IRCds MAY have a blacklist or whitelist and may have an option to enforce
-keys against either or neither of them. Implementations may block keys which
-might result in impersonation. It is an error for user-set metadata to have
-any effect on server operations.
+Setting metadata for channel:
 
-If `METADATA` is supported, it MUST be specified in RPL_ISUPPORT using the
-`METADATA` key.
+    METADATA #example SET url :http://www.example.com
+    :irc.example.com 761 #example url :http://www.example.com
```
DarthGandalf: Visibility field missing

specification/metadata-3.2.md
DarthGandalf: Uhm, I can't find where it is stated that "clients are only sent metadata notifications for changes they themselves didn't make"

DarthGandalf: "Clients are automatically subscribed to notifications for their own metadata." in combination with this line seems to imply that server sends both 761 and metadata notification

PR #123, opened 2015-03-10
metadata-3.2: Fix up numerics, add visibility to notifications.

Aerdan:
This addresses a few issues which were noted just prior to the recent metadata-3.2 work being merged.


kaniini:
I think this can be merged now.

PR #124, opened 2015-03-10
server-time-3.2: Fix ircv3.atheme.org links to be relative

dequis:
Trivial change. This is how all the other specs do it.


attilamolnar:
Thanks!

PR #126, opened 2015-03-10
metadata-3.2: Non-normative example for visibility field.

Aerdan:


specification/metadata-3.2.md
```diff
@@ -218,3 +218,14 @@ External server updates metadata on a channel:
 External server sets metadata on a user:
 
     :irc.example.com METADATA user1 account * :user1
+
+### Non-normative
+
+The following examples describe how an implementation might use certain
+features. Unlike previous examples, they are in no way intended to guide
+implementations' behaviour.
```
DarthGandalf: Which makes the examples useless :(
Because that's what examples are usually for.

Aerdan: No...? The point is to demonstrate usage while disclaiming any policy directives that might be inferred by the demonstration.

PR #127, opened 2015-03-15
Rewrite METADATA LIST

DarthGandalf:


DarthGandalf:
Though, probably 2 different commands would be better for this.

kaniini:
Perhaps `METADATA GET`? :)

DarthGandalf:
Almost made it SHOW. But now changed it to GET, ok...

kaniini:
`GET` is definitely the right verb to use here, it fits in with other protocols too (HTTP GET for example).

Otherwise it looks okay.  :+1: from me.

Aerdan:
Personally, I feel it should be `GET` → `SEARCH`, `LIST` → `GET`, if we're splitting functionality off of `LIST` like this.

DarthGandalf:
`GET` is counterpart for `SET`, and `SET` works with only 1 key.

Issue #130, opened 2015-03-25
Metadata: ERR_NOMATCHINGKEYS discussion


attilamolnar:
@DanielOaks commented in #129:

> ERR_NOMATCHINGKEYS is also used in METADATA GET to show that a requested key could not be retrieved, could maybe modify it to ERR_NOMATCHINGKEY or something, but we'd need to leave one of those in for that subcommand unless we rework it.

DanielOaks:
Well, given that `METADATA GET` is the only thing that uses that error, and `METADATA GET` can only return a single key using that numeric, it could make sense to change it to this:
```
| 766 | `ERR_NOMATCHINGKEY`   | `<Key> :no matching key`                 |
```

I'm wondering whether any extensions later on would make use of `ERR_NOMATCHINGKEYS` to refer to multiple keys, similar to the rejected change in #129, but I don't think we need to take that into consideration.

Also semi-related, we should remove `ERR_NOMATCHINGKEYS` from the *Errors* line at the bottom of `METADATA LIST`

kaniini:
I concur with @DanielOaks here

DanielOaks:
This can be closed, superseded by #131

PR #131, opened 2015-03-30
Fixes and updates for METADATA spec

DanielOaks:
Just implements the changes in #130, makes the error name and string more sensible with the new rewrite of `METADATA`.

### DIFF
--- a/specification/metadata-3.2.md
+++ b/specification/metadata-3.2.md
@@ -142,8 +142,9 @@ Metadata propagates to clients automatically under certain conditions:
 
 1. Upon authentication and successful negotiation of `metadata-notify`, clients
    MUST have their non-transient metadata propagated to them. If none exists,
-   servers MUST send `ERR_METADATAEND`.
-2. Clients SHOULD have current metadata propagated to them.
+   servers MUST send `RPL_METADATAEND`.
+2. Clients SHOULD have current metadata of targets propagated to them upon
+   subscription.
 3. Clients who enable `metadata-notify` after issuing `MONITOR` commands to
    subscribe to users SHOULD have current metadata propagated to them for those
    users.
--- a/specification/metadata-3.2.md
+++ b/specification/metadata-3.2.md
@@ -123,6 +123,9 @@ for users by issuing a disabling `MONITOR` command, and for channels by
 parting the channel. Clients are automatically subscribed to notifications for
 their own metadata, excluding changes made by the clients themselves.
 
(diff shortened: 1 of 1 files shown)


kaniini:
@DanielOaks while you're here can you fix the other quirk we found when implementing METADATA since it's a minor change?  Might as well get both of them done at once.

DanielOaks:
Added a suggested error clarification , and a channel commonality clause for `MONITOR` from @Aerdan , discussed in the channel.

DanielOaks:
I think this should be right to merge (and then I can get the website PR going). Could someone please have a look and merge if it's good?

attilamolnar:
a2e19be and 2b44c68 should be squashed

specification/metadata-3.2.md
```diff
@@ -111,6 +123,9 @@ for users by issuing a disabling `MONITOR` command, and for channels by
 parting the channel. Clients are automatically subscribed to notifications for
 their own metadata, excluding changes made by the clients themselves.
 
+Client subscriptions to a channel includes notifications for other users in
+the channel, regardless of `MONITOR` state.
```
attilamolnar: I don't think this should be the case

DanielOaks: I believe the use case given for that addition is for clients that would very much want to use the `avatar`/`info`/etc set by users in their display, where otherwise they would have to resort to monitoring each of the users joined to common channels separately.

@jwheare wanna weigh in on this?

attilamolnar:
I believe that 805c220a is a major change, if implemented METADATA commands that alter the client's metadata may need heavy rate limiting to prevent floods. What do others think? Was there a discussion about this change and is the majority ok with it or...?

jwheare:
There was a discussion where I raised this issue. Aerdan pointed out that as initially worded, I would have to MONITOR each channel member to get avatar metadata notifications. kaniini said that the wording should be changed. Rate limiting and/or BATCHing these notifications may be sensible, but that could well be an implementation detail.

kaniini:
The spec should document mechanism, ratelimiting is already handled via `RPL_LOAD2HI` hybrid extension or some other form of pacing such as fakelag, while a good practice I do not think it is noteworthy for the core specification.

DanielOaks:
Alright, I think that's about all I've found for the `METADATA` spec. Those two commits have also been squashed together.

Do we wanna get that last thing worked out (if it's still an issue) so we can merge this?

Issue #135, opened 2015-04-14
self-message vs inspircd.org/replay-message vs 3.3 echo-message


kaniini:
why do we have `self-message` when we plan to replace it with something else?

why not skip this and just rip out self-message and put the replacement in 3.2?  i don't like the idea of ratifying something that we *know* is going to have no future...

kaniini:
to whoever made this decision: please do not reply with some excuse like "it was too late to change the spec".  the impact of this change is minimal at best and i would prefer to have it instead of shipping a spec with intentional errata, because if the plan is to replace `self-message`, then its presence is an errata.

kaniini:
some cursory review shows that with the exception of `echo-message` supporting 3.3 pushed tags, it is functionally equivilant to `inspircd.org/replay-message`.  thusly, i recommend we merge `echo-message` with the 3.3-specific clause allowing `@id` pushed tags removed.

PR #136, opened 2015-04-14
Modified echo-message for 3.2

kaniini:
This pull request provides a modified `echo-message` capability for 3.2.  The modification removes the example which takes advantage of 3.3's labeled replies proposal.

Closes #100, #135.

### DIFF
--- a/extensions/echo-message-3.2.md
+++ b/extensions/echo-message-3.2.md
@@ -1,0 +1,56 @@
+echo-message client capability specification
+--------------------------------------------
+
+Copyright (c) 2014 Attila Molnar <attilamolnar@hush.com>
+Copyright (c) 2014 J-P Nurmi <jpnurmi@gmail.com>
+
+Unlimited redistribution and modification of this document is allowed
+provided that the above copyright notice and this permission notice
+remains in tact.
+
+## Description
+
+This client capability MUST be named `echo-message`.
+
+If enabled, servers MUST send `PRIVMSG` and `NOTICE` messages back to
+the client that sent them. If servers apply any modifications to these
+messages, they MUST send the final version of the message back to the
+originating client.
+
+For clients, receiving a message with themselves as the sender acts as
+an acknowledgement that the message has been delivered to the server.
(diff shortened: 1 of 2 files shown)


jpnurmi:
:+1:

kaniini:
What language would you prefer?

dequis:
Since I needed a diff to realize what was removed in this version, including it here for the record (it's just the last paragraph/example)

```diff
--- extensions/echo-message-3.3.md
+++ extensions/echo-message-3.2.md
@@ -54,10 +54,3 @@
 
     --> PRIVMSG #ircv3 :back from \02lunch\0F
     :example!ex@example.com PRIVMSG #ircv3 :back from lunch
-
-A client may add an `id` tag to correlate between sent messages and
-echoed ones (which might have been modified by the server). If an `id`
-tag is present, the server must echo it.
-
-    --> @id=cd87d2 NOTICE jpnurmi :important note
-    @id=cd87d2 :example!ex@example.com NOTICE jpnurmi :important note
```

kaniini:
@dequis The example removed depends on 3.3's labelled replies.

kaniini:
I propose replacing the entire paragraph with:

> If enabled, servers MUST send `PRIVMSG` and `NOTICE` reply with the
> message that was sent after applying any relevant modifications to
> the message.

I believe this, in combination with the following paragraph, resolve the concern with bouncers, since bouncers are required to treat it as a client message, which to me, would mean redistribute it to the other clients which support `echo-message`.

@grawity @jpnurmi thoughts?

jpnurmi:
I like the simplified version of the paragraph. I don't mind having some additional bouncer specific mentions to clarify things, but the motivation of this capability was to replace self-message with something that is not so specific to bouncers. :)

DarthGandalf:
self-message spec needs to be deleted then?
If server is allowed to send modified line back, it makes it harder for client to show line in pending state, because it can't know if received line is the sent line or not. It should be addressed by labelled-replies later though.

kaniini:
@DarthGandalf @grawity already deleted the broken `self-message` one.  this is the "replace" part of our repeal and replace strategy ;)

PR #139, opened 2015-04-25
core/capability-negotiation-3.1: remove CAP CLEAR text per #134

### LINKED ISSUE #134: Deprecate and eventually phase out CAP CLEAR
Hello,

CAP CLEAR is of dubious utility to client authors and of very dubious utility to users. It is not desirable that features be able to be turned off (just do not request CAPs you do not want), and may in fact harm user experience by being in the spec - it just screams "don't push this button" to anyone who comes across reading it. It also is limited in utilty by sticky caps, which cannot be removed once requested.

The only use case that even came up in discussion in #ircv3 was for bouncers - but servers cannot even send CAP CLEAR, making it useless for even this purpose. A well-written bouncer should be able to translate caps, anyway, or mandate them if needs be.

Anecdotally, the only reason I can find that it's even in the spec was apparently because in the past, it was desirable to demonstrate that everything was optional and could be turned off at any time for political reasons. Such political reasons have long since passed, as CAP is now a thing of the present and implemented by most major clients - sans CAP CLEAR in any of them known to me :).

kaniini:
this removes CAP CLEAR from the specification as an errata update for 3.1 and newer.  it closes #134.


core/capability-negotiation-3.1.md
```diff
@@ -117,10 +116,6 @@ In the first usage, acknowledging a REQ subcommand, the ACK subcommand has a sin
 consisting of a space separated list of capability names, which may optionally be preceded with
 one or more modifiers. 
 
-The second usage, acknowledging a CLEAR subcommand, is similar to the first usage. When a CLEAR
-subcommand is issued, all non-"sticky" capabilities are disabled, and a set of ACK subcommands
-will be generated by the server with the disable modifier preceding each capability. 
-
 The third usage is when, in the preceding two cases, some capability names have been preceded
```
DarthGandalf: first, third

PR #144, opened 2015-04-27
Rework links to be relative rather than absolute.

Aerdan:
This commit also fixes a couple links that weren't corrected prior to the site going live.


DanielOaks:
:+1: Looks good to me.

PR #149, opened 2015-05-15
Specify the behavior of "MONITOR +" with duplicates.

ProgVal:


kaniini:
This is OK for merging.

grawity:
Hmm, I thought it is already bloody obvious that it's a unique set, but I guess caution won't hurt

PR #150, opened 2015-05-16
Language negotiation.

sadiepowell:
This is the initial draft of my language negotiation specification which I posted in #ircv3 a while back.

[Rendered view available here.](https://github.com/SaberUK/ircv3-specifications/blob/master%2Blanguage-negotiation/extensions/language-negotiation-3.3.md)

### DIFF
--- a/extensions/language-negotiation-3.3.md
+++ b/extensions/language-negotiation-3.3.md
@@ -1,0 +1,75 @@
+---
+title: IRCv3.3 Language Negotiation
+layout: spec
+copyrights:
+  -
+    name: "Peter Powell"
+    period: "2014-2015"
+    email: "petpow@saberuk.com"
+---
+
+## About
+
+This specification allows clients to request that messages from the server are sent in a specified language. It does not define any translation behaviour for `NOTICE` or `PRIVMSG` except in the case of network pseudoclients such as services.
+
+### Capability
+
+Available languages are sent using the `languages` capability. The value of the capability is a comma delimited list of tokens. The first token is an positive integer which specifies the number of language codes which can be requested. All tokens after that are BCP 47 language codes that the server supports in the order they are used by default.
+
+### Command
+
+Languages are requested with the `LANGUAGE <code>[ <code>]+` command.
+
+It is recommended that clients request similar language codes (e.g. requesting `es` as well as `es-419`) in case the server does not support phrases in the preferred language.
+
+If a language request is successful then the server should respond with the `RPL_YOURLANGUAGESARE` numeric. It is recommended that the user's language preferences be broadcast to other servers on the network.
+
+If a language request fails due to too many language codes being requested then the server should respond with `ERR_TOOMANYLANGUAGES` numeric.
+
+If a language request fails due to the client requesting one or more unsupported language codes then the server should respond with the `ERR_NOLANGUAGE` numeric.
+
+### Numerics
+
+#### RPL_YOURLANGUAGESARE
+
+`:<server-name> 687 <nick> <code>[ <code>]+ :Language preferences have been set.`
+
+#### RPL_WHOISLANGUAGE
+
+`:<server-name> 690 <nick> <code>[ <code>]+ :can speak these languages.`
+
+#### ERR_TOOMANYLANGUAGES
+
+`:<server-name> 981 <nick> <count> :You specified too many languages.`
+
+#### ERR_NOLANGUAGE
+
+`:<server-name> 982 <nick> <code>[ <code>]+ :Languages are not supported by this server.`
+
+## Example Traffic
+
+### Valid language negotiation
+
+    [S] :irc.example.com CAP * LS :languages=5,en-GB,en-US,fr-CA,de,nl
+    [C] CAP REQ language
+    [C] LANGUAGE en-GB en-US
+    [S] :irc.example.com 687 NickName en-GB en-US :Language preferences have been set.
+
+### Client requested too many languages
+
+    [S] :irc.example.com CAP * LS :languages=2,en-GB,en-US,fr-CA,de,nl
+    [C] CAP REQ language
+    [C] LANGUAGE fr-CA en-GB en-US
+    [S] :irc.example.com 981 NickName 2 :You requested too many languages
+
+
+### Client requested an invalid language
+
+    [S] :irc.example.com CAP * LS :languages=3,en-GB,de,nl
+    [C] CAP REQ language
+    [C] LANGUAGE fr-CA en-GB en-US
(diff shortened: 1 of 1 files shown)


grawity:
Um, exactly which messages does this affect? Server statuses & errors? Translation of PRIVMSGs?

SaberUK:
@grawity The former.

extensions/language-negotiation-3.2.md
```diff
@@ -0,0 +1,73 @@
...
+    email: "petpow@saberuk.com"
+---
+
+## About
+
+This specification allows clients to request that messages from the server are sent in a specified language. It does not define any translation behaviour for `NOTICE` or `PRIVMSG` except in the case of network pseudoclients such as services.
+
+### Capability
+
+Available languages are sent using the `languages` capability. The value of the capability is a comma delimited list of tokens. The first token is an unsigned 8-bit integer which specifies the number of fallback languages which can be requested. All tokens after that are a list of BCP 47 language codes that the server supports in the order they are used by default.
```
DanielOaks: Should that integer be the maximum number of *fallback* language codes that can be requested, or the maximum number of language codes that can be requested in total?

*Fallback* implies that the first requested language doesn't count towards the limit (and if that's the case, example 2 needs to have another language added to the client request for it to be over). Might make sense to either clarify that or switch it to mean the total number of languages.

SaberUK: Fixed.

kaniini:
Why can't this build ontop of the metadata framework?  Isn't language preference a type of metadata?

Mikaela:
I don't understand the point of this language negotiation and I agree with it being type of website.

Usually everyone speaks English if the network is international, and channels speak whatever language they do and it doesn't seem like specific language negotiations are needed.

Charset might make a little more sense, but there is also question why as everyone should use UTF-8 (even if some networks/channels accept others such as latin1 or latin15).

SaberUK:
>@kaniini: Why can't this build ontop of the metadata framework? Isn't language preference a type of metadata?

Metadata does not (from my reading) support arrays of data. Also, it does not seem to have a method of informing the user that their `METADATA SET` request is invalid or otherwise malformed.

>@Mikaela: I don't understand the point of this language negotiation 

 As it says at the top of the specification:

>This specification allows clients to request that messages from the server are sent in a specified language. It does not define any translation behaviour for NOTICE or PRIVMSG except in the case of network pseudoclients such as services.

This means that the IRCd could translate messages like:

    :irc.example.com 251 NickName :There are 4 users and 405 invisible on 5 servers

To this (assuming es-419):

    :irc.example.com 251 NickName :Hay 4 usuarios y 405 invisibles en 5 servidores

DarthGandalf:
I think it's job of IRC client to render the number of invisible people (or whatever) in the user's language in client-specific way (e.g. in a window title), not of IRC server.

However, the message in `251` needs to be machine-readable.

Mikaela:
Oh, sorry, I missed that entirely.

SaberUK:
>I think it's job of IRC client to render the number of invisible people (or whatever) in the user's language in client-specific way (e.g. in a window title), not of IRC server.

I really don't think it's feasible for every single client to make a translation database of every single message sent by every single server implementation (this is IRC don't forget so a large amount of them will be conflicting). It's a lot easier for this to be done server side. Especially considering the reality which is that a majority of clients show the messages directly sent by the server.

kaniini:
If the point is to select what languages you want the human-readable part of server messages translated to, why does it need to be an array?  Why can't the fallback be the server default locale?

SaberUK:
>Why can't the fallback be the server default locale?

Because if I can comprehend German and French better the server's default language of English it would probably be better for the server to fall back to French when a phrase isn't available in German rather than the server's default language.

robotoad:
> I really don't think it's feasible for every single client to make a translation database of every single message sent by every single server implementation (this is IRC don't forget so a large amount of them will be conflicting). It's a lot easier for this to be done server side. Especially considering the reality which is that a majority of clients show the messages directly sent by the server.

Clients show messages to users because messages like 251 carry data in trailing sentences that vary in format, which forces some form of "status window" for dumping messages that can't be parsed. In contrast, the WHOIS numerics have a parseable format that can be presented in an inspector window:
screen shot 2015-05-26 at 10 51 56 am

SaberUK:
I am withdrawing this due to lack of interest.

extensions/language-negotiation-3.3.md
```diff
@@ -0,0 +1,75 @@
...
+    email: "petpow@saberuk.com"
+---
+
+## About
+
+This specification allows clients to request that messages from the server are sent in a specified language. It does not define any translation behaviour for `NOTICE` or `PRIVMSG` except in the case of network pseudoclients such as services.
+
+### Capability
+
+Available languages are sent using the `languages` capability. The value of the capability is a comma delimited list of tokens. The first token is an positive integer which specifies the number of language codes which can be requested. All tokens after that are BCP 47 language codes that the server supports in the order they are used by default.
```
thatcher-gaming, line 17: Is there an actual use case for limiting the number of languages that can be sent? (other than limiting it to the number of languages that are supported by the server)

extensions/language-negotiation-3.3.md
```diff
@@ -0,0 +1,75 @@
...
+
+### Command
+
+Languages are requested with the `LANGUAGE <code>[ <code>]+` command.
+
```
thatcher-gaming, line 22: I think it would be worth explicitly allowing this to be sent before registration has been completed. There are many error messages that might be sent as a result of failed registration, so having those localised would be helpful

extensions/language-negotiation-3.3.md
```diff
@@ -0,0 +1,75 @@
...
+It is recommended that clients request similar language codes (e.g. requesting `es` as well as `es-419`) in case the server does not support phrases in the preferred language.
+
+If a language request is successful then the server should respond with the `RPL_YOURLANGUAGESARE` numeric. It is recommended that the user's language preferences be broadcast to other servers on the network.
+
+If a language request fails due to too many language codes being requested then the server should respond with `ERR_TOOMANYLANGUAGES` numeric.
+
+If a language request fails due to the client requesting one or more unsupported language codes then the server should respond with the `ERR_NOLANGUAGE` numeric.
+
+### Numerics
```
thatcher-gaming, line 31: This obviously predates them, but I think it would be good to update this to use standard replies instead

thatcher-gaming (commented):
I am interested in implementing this!! 
I've just added some suggestions. I'm happy to enact them myself if others agree. I have no idea how to do that on github.com, but I'll figure it out.

PR #152, opened 2015-06-06
extensions: add register-and-verify extension

kaniini:
The `REGISTER` and `VERIFY` commands provide a unified, standardized interface for account creation and validation to the network's authentication layer.  This allows for a network to provide a common interface regardless of what authentication layer they have chosen for their network to operate, i.e. traditional services, a different central authority, or a decentralized model similar to [IRCX](http://en.wikipedia.org/wiki/IRCX).

We believe that the benefit of adding a common interface for account creation and validation, is helpful to the end user in discovering the capabilities offered by the authentication and authority layer of the network. Further, we believe it is helpful to client authors, so that they may provide guided account creation and validation for their users, regardless of network configuration.

### DIFF
--- a/extensions/reg-core-3.3.md
+++ b/extensions/reg-core-3.3.md
@@ -34,7 +34,7 @@ authentication layer.  It is similar to current methods of signalling that inten
 
 A `REG CREATE` command consists of the following format:
 
-`REG CREATE <accountname> <callback> <cred_type> :<credential>`
+`REG CREATE <accountname> [callback_namespace:]<callback> [cred_type] :<credential>`
 
 A passphrase `credential` MAY have spaces in it.
 
@@ -42,12 +42,14 @@ The `callback` field designates an opaque value which indicates where a verifica
 will be sent.  The callback field is implementation-specific, with namespaces being listed using the
 `REGCALLBACKS` ISUPPORT token.  An invalid callback parameter should result in the `ERR_REG_INVALID_CALLBACK`
 error.  In the event that a callback is not provided, the client SHOULD send `*` to indicate that
-they are not providing a callback resource.
+they are not providing a callback resource.  If a callback namespace is not explicitly provided, the IRC
+server MAY choose to use `mailto` as a default.
 
 The `cred_type` field designates a value which indicates the type of supplied credential.  The accepted
 values of the `cred_type` field is implementation-specific, with credential types being listed using the
 `REGCREDTYPES` ISUPPORT token.  An invalid credential type parameter should result in the `ERR_REG_INVALID_CRED_TYPE`
-error.  The client MUST supply a `cred_type` value.
+error.  The client SHOULD supply a `cred_type` value, however if one is not provided the IRC server SHOULD
+use `passphrase` as a default credential type.
 
 The IRC server MAY forward the `REG CREATE` command to a central authority, or process it locally, however
 the authentication layer SHOULD NOT send any legacy messages in reply to a `REG CREATE` command.
@@ -59,7 +61,7 @@ Upon success, the IRC server MUST send the `RPL_REGISTRATION_SUCCESS` numeric, w
 If the server requires a verification token, it MUST also reply with the `RPL_REG_VERIFICATION_REQUIRED`
 numeric, which looks like:
 
-    :<server> 927 <user_nickname> <accountname> <callback> :A verification token was sent
+    :<server> 927 <user_nickname> <accountname> <callback_namespace:><callback> :A verification token was sent
 
 Upon error, the IRC server MUST send an error code that is relevant.  We suggest these numerics:
 
@@ -134,6 +136,9 @@ A sample 005 reply indicating that no verification callback methods are supporte
 
     :irc.example.com 005 kaniini REGCALLBACKS= :are supported by this server
 
+The first callback type SHOULD be the implementation-defined default if default callback types
+are supported.
+
 ## The `REGCREDTYPES` `RPL_ISUPPORT` token
 
 Implementations which provide support for this framework SHOULD specify the allowed credential
@@ -152,6 +157,9 @@ The following credential types are defined:
 A sample 005 reply indicating support for credential types is:
 
     :irc.example.com 005 kaniini REGCREDTYPES=passphrase,certfp :are supported by this server
+
+The first credential type SHOULD be the implementation-defined default if default credential types
+are supported.
 
 ## Examples
 
--- a/extensions/reg-core-3.3.md
+++ b/extensions/reg-core-3.3.md
@@ -173,23 +173,23 @@ A sample 005 reply indicating support for credential types is:
     S: 923 kaniini rabbit :Account verification successful
     S: 903 kaniini :Authentication successful
 
-### Registering the account with an unsupported verification method:
+### Registering the account "rabbit" with an unsupported verification method:
 
     C: REG CREATE rabbit 1vBjNBdhjWFFbbbbVBHJEWBHJWcfbbvjkhbea passphrase :testpassphrase123
     S: 929 kaniini rabbit 1vBjNBdhjWFFbbbbVBHJEWBHJWcfbbvjkhbea :Callback token is not valid
 
-### Registering the account but providing an invalid verification token:
+### Registering the account "rabbit", but providing an invalid verification token:
 
     C: REG CREATE rabbit mailto:kaniini@example.com passphrase :testpassphrase123
-    S: 920 kaniini kaniini :Account registered
-    S: 927 kaniini kaniini kaniini@example.com :A verification code was sent
-    C: REG VERIFY kaniini 3qw4tq4te4gf34
-    S: 925 kaniini kaniini :Invalid verification code
+    S: 920 kaniini rabbit :Account registered
+    S: 927 kaniini rabbit kaniini@example.com :A verification code was sent
+    C: REG VERIFY rabbit 3qw4tq4te4gf34
+    S: 925 kaniini rabbit :Invalid verification code
 
-### Registering the account where a verification token is not required:
+### Registering the account "rabbit" where a verification token is not required:
 
     C: REG CREATE rabbit * passphrase :testpassphrase123
-    S: 920 kaniini kaniini :Account registered
+    S: 920 kaniini rabbit :Account registered
     S: 903 kaniini :Authentication successful
 
 ### Registering an account which already exists:
--- a/extensions/reg-core-3.3.md
+++ b/extensions/reg-core-3.3.md
@@ -1,0 +1,209 @@
+---
+title: "The REG command framework"
+layout: spec
+copyrights:
+  -
+    name: "William Pitcock"
+    period: "2014-2015"
+    email: "nenolod@dereferenced.org"
+---
+
+The `REG` command framework provides a unified, standardized interface for account creation and
+validation to the network's authentication layer.  This allows for a network to provide a common
+interface regardless of what authentication layer they have chosen for their network to operate,
+i.e. traditional services, a different central authority, or a decentralized model similar to [IRCX][ircx].
+
+   [ircx]: http://en.wikipedia.org/wiki/IRCX
+   
+We believe that the benefit of adding a common interface for account creation and validation,
+is helpful to the end user in discovering the capabilities offered by the authentication and
+authority layer of the network.  Further, we believe it is helpful to client authors, so that
+they may provide guided account creation and validation for their users, regardless of network
+configuration.
+
+## The `REG` command
+
+The `REG` command is the main command used by the account registration subsystem.  It provides
+several subcommands listed in the `REGCOMMANDS` ISUPPORT type.  The core sub-commands are listed
+below:
+
+## The `REG CREATE` sub-command
+
+The `REG CREATE` command signals the intent of a client to register an account in the network's
+authentication layer.  It is similar to current methods of signalling that intent.
+
+A `REG CREATE` command consists of the following format:
+
+`REG CREATE <accountname> <callback> <cred_type> :<credential>`
+
+A passphrase `credential` MAY have spaces in it.
+
+The `callback` field designates an opaque value which indicates where a verification code, if any,
+will be sent.  The callback field is implementation-specific, with namespaces being listed using the
+`REGCALLBACKS` ISUPPORT token.  An invalid callback parameter should result in the `ERR_REG_INVALID_CALLBACK`
+error.  In the event that a callback is not provided, the client SHOULD send `*` to indicate that
+they are not providing a callback resource.
+
+The `cred_type` field designates a value which indicates the type of supplied credential.  The accepted
+values of the `cred_type` field is implementation-specific, with credential types being listed using the
+`REGCREDTYPES` ISUPPORT token.  An invalid credential type parameter should result in the `ERR_REG_INVALID_CRED_TYPE`
+error.  The client MUST supply a `cred_type` value.
+
+The IRC server MAY forward the `REG CREATE` command to a central authority, or process it locally, however
+the authentication layer SHOULD NOT send any legacy messages in reply to a `REG CREATE` command.
+
+Upon success, the IRC server MUST send the `RPL_REGISTRATION_SUCCESS` numeric, which looks like:
+
+    :<server> 920 <user_nickname> <accountname> :Account created
+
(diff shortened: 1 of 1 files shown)


extensions/register-and-verify-3.3.md
```diff
@@ -0,0 +1,154 @@
...
+numeric, which looks like:
+
+    :<server> 927 <user_nickname> <accountname> <callback> :A verification token was sent
+
+Upon error, the IRC server MUST send an error code that is relevant.  We suggest these numerics:
+
+`ERR_ACCOUNT_ALREADY_EXISTS`:
+    `:<server> 921 <user_nickname> <accountname> :Account already exists`
+`ERR_REG_INVALID_PARAMS`:
+    `:<server> 922 <user_nickname> <accountname> :Invalid params: <descriptive_text>`
```
SaberUK: I feel like this is a bit generic. It would be useful to have error messages that the client can actually process.

M2Ys4U:
I have a couple of issues with this:

1. There's no way to tell a priori which callback mechanisms are available for the client to use. A client may prefer to use either SMS over email but the chances of successfully registering with e-mail are much, much greater. The client will either have to first try SMS and then email or just opt for the most likely (email) and use that every time.

2. This presupposes that pass phrases are required by all auth methods - a client (or server) may wish to authenticate with, say, a TLS client certificate *only* (cf `SASL EXTERNAL`). One might use an empty passphrase for this, but again there's no a priori way for a client to determine if that's acceptable to the server.

extensions/register-and-verify-3.3.md
```diff
@@ -0,0 +1,154 @@
+`REGISTER` and `VERIFY` command specification
+---------------------------------------------
+
+Copyright (c) 2014-2015 William Pitcock <nenolod@dereferenced.org>.
+
+Unlimited redistribution and modification of this document is allowed
+provided that the above copyright notice and this permission notice
+remains in tact.
```
DonteMayers: This should be yaml front matter instead.

extensions/register-and-verify-3.3.md
kaniini: The whole point of 922 is that it's generic.  Please propose specific numerics you would like added explicitly, such as for example, with 926.

kaniini:
> There's no way to tell a priori which callback mechanisms are available for the client to use. A client may prefer to use either SMS over email but the chances of successfully registering with e-mail are much, much greater. The client will either have to first try SMS and then email or just opt for the most likely (email) and use that every time.

The specification does not SPECIFY *ANY* callback mechanisms.  Would you like to propose a change to provide specification of some callback mechanisms?

> This presupposes that pass phrases are required by all auth methods - a client (or server) may wish to authenticate with, say, a TLS client certificate only (cf SASL EXTERNAL). One might use an empty passphrase for this, but again there's no a priori way for a client to determine if that's acceptable to the server.

Fair 'nuff, but in most cases are going to use a passphrase.  The only reasonable alternative I see right now would be certfp, and since it depends on hash brute-force resistance I don't think it's really much better.  Should `REGISTER` provide a field to specify the credential type?

extensions/register-and-verify-3.3.md
```diff
@@ -0,0 +1,165 @@
...
+Upon error, the IRC server MUST send an error code that is relevant.  We suggest these numerics:
+
+`ERR_ACCOUNT_ALREADY_EXISTS`:
+    `:<server> 921 <user_nickname> <accountname> :Account already exists`
+`ERR_REG_INVALID_PARAMS`:
+    `:<server> 922 <user_nickname> <accountname> :Invalid params: <descriptive_text>`
+`ERR_REG_INVALID_CALLBACK`:
+    `:<server> 926 <user_nickname> <accountname> <callback> :Callback token is invalid`
+
+The server MAY send additional informative text upon registration success or failure using `RPL_TEXT` or `NOTICE`.
```
prawnsalad: This could warrant a numeric itself. There is no way to tell what services may require or warn about, so for clients processing these it would be handy to detect any arbitrary text/warnings relating to registration.

extensions/register-and-verify-3.3.md
```diff
@@ -0,0 +1,196 @@
...
+    :<server> 927 <user_nickname> <accountname> <callback> :A verification token was sent
+
+Upon error, the IRC server MUST send an error code that is relevant.  We suggest these numerics:
+
+`ERR_ACCOUNT_ALREADY_EXISTS`:
+    `:<server> 921 <user_nickname> <accountname> :Account already exists`
+`ERR_REG_UNSPECIFIED_ERROR`:
+    `:<server> 922 <user_nickname> <accountname> :<descriptive_text>`
+`ERR_REG_INVALID_CALLBACK`:
+    `:<server> 926 <user_nickname> <accountname> <callback> :Callback token is invalid`
```
SaberUK: Numeric 926 collides with an existing numeric used by InspIRCd.

kaniini: @SaberUK can you provide a list of numerics in the area that are already used so I can assign one that does not have a collision?

tilal6991:
Overall I'm in favour of this addition but I find the use of the 920 numeric with "Account registered" as the message a bit misleading if the server still needs the user to verify. Surely the account would not actually be registered unless the verification process was also successful?

kaniini:
@tilal6991 the 920 numeric means that the registration was accepted, if 927 is sent then it's in limbo until the verification process is done.  so perhaps "Account reserved" is better since we send 903 anyway on login.

tilal6991:
"Account reserved" sounds much better to me.

kaniini:
I went with "Account created" which seems to cover both cases.  But really clients should just render their own message instead of using the default text imo.

kaniini:
I moved numeric 926 to 929 because it collided with InspIRCd's forbidden channel numeric.

440 is also defined/recycled as the "authentication layer is unavailable" numeric.  Is there any other specific cases which should have explicit numerics defined?

extensions/reg-core-3.3.md
```diff
@@ -0,0 +1,209 @@
...
+    C: REG CREATE rabbit mailto:kaniini@example.com passphrase :testpassphrase123
+    S: 920 kaniini kaniini :Account registered
+    S: 927 kaniini kaniini kaniini@example.com :A verification code was sent
+    C: REG VERIFY kaniini 3qw4tq4te4gf34
+    S: 925 kaniini kaniini :Invalid verification code
+
+### Registering the account where a verification token is not required:
+
+    C: REG CREATE rabbit * passphrase :testpassphrase123
+    S: 920 kaniini kaniini :Account registered
```
DarthGandalf: rabbit or kaniini?

jwheare:
I think we probably need to allow an optimisation for the vastly most common likely use case of email and passphrase and accept a default of:

`REG CREATE kaniini kaniini@example.com :testpassphrase123`

extensions/reg-core-3.3.md
```diff
@@ -0,0 +1,209 @@
...
+`ERR_ACCOUNT_ALREADY_EXISTS`:
+    `:<server> 921 <user_nickname> <accountname> :Account already exists`
+`ERR_REG_UNSPECIFIED_ERROR`:
+    `:<server> 922 <user_nickname> <accountname> :<descriptive_text>`
+`ERR_REG_INVALID_CALLBACK`:
+    `:<server> 929 <user_nickname> <accountname> <callback> :Callback token is invalid`
+`ERR_REG_INVALID_CRED_TYPE`:
+    `:<server> 928 <user_nickname> <accountname> <cred_type> :Credential type is invalid`
+`ERR_REG_UNAVAILABLE`:
+    `:<server> 440 <user_nickname> <subcommand> :Authentication services are currently unavailable`
```
dequis: The rendered markdown of this section joins all the lines as one. Same thing with the numerics list of the next section.

Checked the other specs with lists of numerics, all of them do it differently, so I guess anything that is readable will do.

dequis:
@jwheare what is it that you're 'optimizing' for there? Seems like it's better for humans who type the command manually (which would make sense if this was intended to be used without client support - but is it?), but slightly more complex for code (how do you know it's the short version? counting parameters? checking if it starts with `mailto:`?)

jwheare:
@dequis it's intended for client support but will inevitably still be used by humans (same reason for `REG` instead of `REGISTRATION`. It's quite easy to check by counting parameters, especially if documented. Also, avoids having to parse and read ISUPPORT, which may not seem like a big issue for most clients, but client authors are expert at cutting corners, I expect that initial client support will only handle email/passphrase UI anyway and assume that form.

Are there even any examples of sms verification in the wild? And I know there's certfp instead of passphrase, but it's not particularly popular (mainly due to poor support from lazy client authors, myself included).

kaniini:
@jwheare 

I don't like the idea of special-casing the e-mail part because it will discourage client authors from doing the right thing.  How do you propose to mitigate that "corner cutting", especially considering that it is a general desire to move away from passphrases as a primary auth credential?  Even if they only implement UI for the *common* case, at least they will have the *plumbing* for when the *common* case changes.  As for sms verification in the wild -- I do not know of any particular network which does it, but it seems like a bad idea to restrict the protocol to cover only mail because really, mail usage is starting to kind of fall off.

Clients should be reading ISUPPORT anyway, it is required for compliance with the new core spec anyway.

extensions/reg-core-3.3.md
kaniini: I fixed that in the latest commit, thanks for catching it.

jwheare:
It's mainly to help humans. I just think until clients have good built in support for it, there will be a lot of support staff explaining to people how to do reg and if we can reduce any potential confusion there it'll be a more welcome change for people running IRC servers. But I can see why giving client authors less of an excuse to cut corners is desirable.

To be honest, the main barrier to supporting e.g. certfp from a client point of view is not going to be in implementing ISUPPORT parsing or the full verbed REG commands, but in accepting/generating and storing client certs in the first place. Once a client has decided to support it, using the fully specced version of REG commands is going to be trivial.

jwheare:
Potential confusion example:

```
<noob117> so how do i register my nickname?
<admin> /reg create <yournick> <mailto:you@example.com> passphrase :<passphrase>
<noob117> so like this? /reg create <noob117> <mailto:noob117@hotmail.com> hunter2 :<hunter2>
<admin> no, you don't need the angle brackets
<noob117> oh ok, like this? /reg create noob117 mailto:noob117@hotmail.com hunter2 :hunter2
<admin> yeah but the first passphrase is the word "passphrase" not your actual passphrase
<noob117> oh, er, ok
<peanut_gallery> wow that seems way more complicated than when i registered my nick
```

kaniini:
@jwheare et al. is the change I pushed acceptable to you guys for that usability situation?

jwheare:
Yup, great! :+1:

extensions/reg-core-3.3.md
```diff
@@ -0,0 +1,217 @@
...
+below:
+
+## The `REG CREATE` sub-command
+
+The `REG CREATE` command signals the intent of a client to register an account in the network's
+authentication layer.  It is similar to current methods of signalling that intent.
+
+A `REG CREATE` command consists of the following format:
+
+`REG CREATE <accountname> [callback_namespace:]<callback> [cred_type] :<credential>`
```
DarthGandalf: Optional non-last parameter makes it more difficult to parse correctly.
Also this means that empty `<credential>` will look like `:\r\n`?

kaniini:
@DarthGandalf I made "null" credentials `*` in latest commit.  Beyond that, assuming you are using a compliant parser (which should be turning IRC messages into an intermediate representation due to, for example, tags), then you would just check the length of the argument vector to see which form of the command to use.

jwheare:
Does this mean a passphrase of "*" is not allowed? Are users expected to know this, and are clients expected to explain it?

jwheare:
Or does * only mean null for e.g. client cert credentials?

grawity:
>  are clients expected to explain it?

If that's the case, I suggest the following explanation: "Passphrases must be at least 6 characters long."

kaniini:
I think `*` should only be allowed for credential types which allow null credentials.  Passphrase obviously would require non-null credential types.

jwheare:
Makes sense, could be clarified further and documented with examples, then.

kaniini:
I'm sorry but revisiting this 6 months later I am really confused as to what we need to do.  Can someone fill me in on the details.

jwheare:
Afaict from re-reading the actual diff, nothing, just me misunderstanding the discussion I think.

An explicit example stating that * is a valid passphrase wouldn't go amiss, but is non-essential.

extensions/reg-core-3.3.md
```diff
@@ -0,0 +1,218 @@
...
+server MAY choose to use `mailto` as a default.
+
+The `cred_type` field designates a value which indicates the type of supplied credential.  The accepted
+values of the `cred_type` field is implementation-specific, with credential types being listed using the
+`REGCREDTYPES` ISUPPORT token.  An invalid credential type parameter should result in the `ERR_REG_INVALID_CRED_TYPE`
+error.  The client SHOULD supply a `cred_type` value, however if one is not provided the IRC server SHOULD
+use `passphrase` as a default credential type.
+
+The IRC server MAY forward the `REG CREATE` command to a central authority, or process it locally, however
+the authentication layer SHOULD NOT send any legacy messages in reply to a `REG CREATE` command.
```
rburchell: "legacy messages"?

doesn't this also conflict somewhat with the comment on line 80 about RPL_TEXT/notice being OK?

DanielOaks: I read this as "don't send the old `NickServ` registration privmsgs/notices" (all the notices services packages currently send on account registration), and send proper numerics instead of it.

So don't send stuff like:   `-NickServ- A passcode has been sent to coolguy@example.com, please type /msg NickServ confirm <passcode> to complete registration`

If you do have legitimate extra required information for the client you can send it to them using `RPL_TEXT`/notice no problem. You just shouldn't send the bunch of notices you get right now in response to `/NS REGISTER`, because we replace them with real numerics.

kaniini: What @DanielOaks said.

rburchell: Yes, of course, I can infer this. However, a specification is supposed to be relatively unambiguous so as to make life easier for implementors (not all of which might have the same level of experience & knowledge) to do the right thing. The point I was making was that it isn't clear in its current form, and could do with clarification.

DanielOaks: Possibly something like this?  "That is, do not send legacy `NICKSERV REGISTER` responses to the `REG` command, and instead use the numerics defined in this specification."

DanielOaks:
I think it's basically that legacy note, formatting the numeric lists to use markdown lists so they get separated properly on the website, and the clarification that credential types other than `passphrase` may also contain spaces.

prawnsalad:
Just saw this draft now. Wouldn't it be cleaner and using much less numerics if all the error responses were combined? Instead of eating 5 numerics and to keep the error options open for addition in future without the need for extra numerics being assigned, something along the lines of...
~~~
ERR_REG:
      :<server> 921 <user_nickname> <accountname> account_exists:<Human readable error text>
~~~

DanielOaks:
I think it's alright as-is. Using different numerics can let clients better show where errors occur in their interfaces, without having to rely on text prefix parsing or something like that.

dboyz:
In short:
Are there any IRC services out there that allows you to register an account-name that is not your current nickname? If no, I think the `<accoutname>` parameter is essentially an unnecessary parameter which can be removed.

Extended explanation:
From my experience with both Atheme and Anope, the registration command has no `accountname` parameter:
`MSG NICKSERV REGISTER <password> <email>`
The account-name to registered will be your current nickname.

kaniini:
Yes, there is.  There is QuakeNet's `Q` bot, and Undernet's CService (`X`).

grawity:
Q on QuakeNet only allows registering your current nickname, too.

X lets you register anything, although the only way to do that is via a web form, so there's _no_ current nickname at that time.

staticfox:
> Are there any IRC services out there that allows you to register an account-name that is not your current nickname?

On GameSurge, we don't do nickname registrations what-so-ever, so it's purely account based. i.e. `/msg AuthServ@Services.GameSurge.net REGISTER <account> <password> <email>`

ChickenNuggers:
```
18:35:48 <dan-> as an aside, any chance we could get https://github.com/ircv3/ircv3-specifications/pull/152 closed and pointed to #276?
```

attilamolnar:
Resubmitted in #276.

Issue #77, opened 2014-03-22
Let servers advertise to clients that they are secure-preferred


attilamolnar:
Raised by @jwheare

voltagex:
@wilhelmy I agree, how about secure-by-default?

attilamolnar:
"Secure by default" is a good sounding phrase but conveys none of the actions to take to achieve security. STS with preload lists are one step in that direction, it provides a working upgrade path. The next step could be restricting new features to secure connections for example. Doing only the latter requires users to reconfigure all servers they use manually instead of the client doing it.

synandro:
Maybe extend RPL_REDIR(010 numeric) a little to support saying to clients to direct to server/port and somehow say that it's an SSL port.  

*** RPL_REDIR ***/             ":%s %d :Please use this SSL Server/Port instead" perhaps?

Or simply just don't listen on a plaintext port.  Not everything needs to be over-engineered ;)

attilamolnar:
The specification covering the basics has been merged. I have opened new issues for the next TODO items.

PR #155, opened 2015-06-22
Add Strict Transport Security specification

attilamolnar:
Added as core for 3.3, the idea is that this helps migrating from insecure connections to secure ones and paves the way for mandating secure connections for new features (if that ever happens).

The spec is trying to solve two problems at once, one is redirecting clients to a TLS port and the enforcement of it for all future connections. This is different to HSTS because browsers can be redirected to a secure site while there is no standardized way to do the equivalent for IRC clients.

### DIFF
--- a/core/sts-3.3.md
+++ b/core/sts-3.3.md
@@ -1,0 +1,195 @@
+---
+title: IRCv3.3 Strict Transport Security
+layout: spec
+copyrights:
+  -
+    name: "Attila Molnar"
+    period: "2015"
+    email: "attilamolnar@hush.com"
+---
+
+## Description
+
+Strict Transport Security (STS) is a mechanism which allows servers to
+communicate that complying clients should only connect to them over a secure
+connection.
+
+The policy is communicated to the client via the `sts` capability and should
+be processed by the client at capability negotiation time.
+
+The value of the capability specifies the duration during which the client
+MUST only connect to the server securely (using TLS, aka. SSL), and the port
+number to use for it.
+
+Clients MUST use sufficient certificate verification to ensure the connection
+is secure without any errors.
+
+## Details
+
+The capability value, which MUST always be present, is a comma (`,`) separated
+list of tokens. Each token is a key which may have a value attached. If there
+is a value attached, the value is separated from the key by a `=`. That is,
+`<key>[=<value>][,<key2>[=<value2>][,<keyN>[=<valueN>]]]`. Keys specified in
+this document MUST only occur at most once.
+
+Clients MUST ignore every token with a key that they don't understand.
+
+An `sts` capability having at least a `duration` key expresses an STS policy.
+
+See [capability negotiation 3.2](capability-negotiation-3.2.html) for more
+information about capabilities with values.
+
+### Mechanism
+
+When the client sees the `sts` capability advertised over a non-secure
+connection, it MUST first establish a secure connection and confirm that the
+`sts` capability is still present.
+
+Once the client has confirmed that the `sts` capability is indeed offered over
+a secure connection, it then MUST only attempt secure connections to the
+server from now on until the policy expires (see the `duration` key).
+It MUST refuse to connect if a secure connection cannot be established with the
+server for any reason during the lifetime of the policy.
+
+However, if the client fails to connect securely for any reason, the connection
+attempt SHOULD be considered a failure, similar to a network error.
+The client SHOULD retry the secure connection next time it receives the `sts`
+cap with the appropriate keys over a non-secure connection.
+
+If the secure connection succeeds but the STS policy is not present, the client
+SHOULD continue using the secure connection for that session.
+
+Clients MUST NOT request this capability (`CAP REQ`).
+Servers MUST reply with an appropriate `CAP NAK` message if a client requests
+this capability.
+
+### The `duration` key
+
+This key indicates how long from receiving the advertisement clients MUST use a
+secure connection when connecting to this server. This MUST be specified in
+seconds as the value of this key and as a single integer without a prefix or
+suffix.
+
+Servers MUST send this key to securely connected clients.
+Servers MAY send this key to non-securely connected clients, but it will be
+ignored.
+
+If the client receives the `duration` key on a non-secure connection, it is
+invalid and MUST be ignored.
+
+Clients MUST reset the duration "time to live" every time they receive a valid
+policy advertisement having a `duration` key from the server.
+
+A duration of 0 means the policy expires immediately. This method can be used
+by servers to remove a previously set policy.
+
+### The `port` key
+
+This key indicates the port number that clients can use when connecting over
+TLS. This key's value MUST be a single port number.
+
+If the client is not already connected securely, it MUST close the insecure
+connection and reconnect securely on the stated port.
+
+Servers MUST send this key to non-securely connected clients.
+Servers MAY send this key to securely connected clients, but it will be
+ignored.
+
+## No user recourse
+
+User recourse should not be possible if the client fails to establish a
+secure connection for any reason with a server which the client has a STS
+policy for. For example the user should not be presented with a dialog that
+allows them to proceed with the connection. A failure to establish a secure
+connection should be treated similarly to a server error that the user is
+unable to do anything about except retrying at a later time.
+
+This is to ensure that it is not possible to fool users into allowing a
+man-in-the-middle attack.
+
+## Security
+
+### STS policy injection
+
+It is possible for attackers to inject the STS policy into a plaintext
+connection. This causes the client to reconnect over TLS and either fail,
+or connect successfully over TLS but discover that the `sts` cap is not
+present.
+
+### Policy expiration
+
+It is possible that a client successfully receives the STS policy from a server
+but later on attackers begin to block all secure connection attempts from the
+client to the server until the policy expires. At that time, the client will
+revert back to a non-secure connection. Servers SHOULD advertise a long enough
+duration which makes this scenario less likely to happen.
+
+## Examples
+
+### Redirection to a secure port
+
+Server tells a client connecting non-securely to connect securely on port 6697.
+
+    Client: CAP LS 302
+    Server: CAP * LS :sts=port=6697
+
+After the exchange, the client disconnects and reconnects securely to the same
+server on port 6697 and proceeds as it normally would.
+
+### Setting STS policy
+
+Server tells a client that is already connected securely that the client must
+only use secure connections for roughly 6 months.
+
+    Client: CAP LS 302
+    Server: CAP * LS :sts=duration=15552000
+
+Until the policy expires:
+* The client will use the port it is currently connected to in the future to
+connect securely.
+* It will not make any non-secure connection attempts to the same server.
+
+### Ignoring invalid request
+
+Server tells a client that is connected non-securely that the client must
+use secure connections for roughly 6 months. There is no port advertised.
+
+    Client: CAP LS 302
+    Server: CAP * LS :sts=duration=31536000
+
+The client ignores this because it received the STS policy over a non-secure
+connection and the `sts` cap contains no token with key `port`.
+
+### Handling tokens with unknown keys
+
+Server tells a client that is already connected securely that the client must
+use secure connections for roughly a year, but the value of the `sts` capability
+also contains some tokens whose keys the client does not understand.
+
+    Client: CAP LS 302
+    Server: CAP * LS :sts=unknown,duration=31536000,foo=bar
+
+The client ignores the keys it does not understand and until the policy
+expires:
+* The client will use the port it is currently connected to in the future to
+connect securely.
+* It will not make any non-secure connection attempts to the same server.
+
+### Handling 0 `duration`
+
+Server tells a client that is already connected securely to remove the STS
+policy now.
+
+    Client: CAP LS 302
+    Server: CAP * LS :duration=0
+
+If the client has any STS policy stored for the server it clears the policy.
+Future connections should use whatever settings (port, secure/non-secure) the
+client used before it received the STS policy.
+
+## TODO
+
+* SNI integration, `sub` key: [issue #176](https://github.com/ircv3/ircv3-specifications/issues/176)
+* Ensure that clients with long lived connections renew the policy: [issue #178](https://github.com/ircv3/ircv3-specifications/issues/178)
+* Advertise hosts the server has a valid certificate for (?)
+* Mention STARTTLS


attilamolnar:
This is a rough draft, I'll update it as comments come in (you should comment btw.)

attilamolnar:
Ideas for discussion:

* Subdomains

Currently the spec makes it possible that a policy from example.net affects server1.example.net, server2.example.net etc. This is of limited use, as it doesn't affect users connecting to irc.example.net to server1.example.net

There should be a possibility to make it so connecting successfully to a server on the network results in all servers on the network (domain) using secure connections in the future regardless whether the user enters `server1.example.net`, `example.net` or `irc.example.net`. 

Possible solution: the `sub` key could take a value, the highest level domain whose subdomains the policy should affect, limited by the scope of the cert. 

* Disabling STS

How to handle the case when a server no longer wishes to support TLS. Domains may change owners anytime. Should disabling STS be allowed? 

* Preload list

Slightly off topic to the spec, but useful in general.

Mikaela:
> How to handle the case when a server no longer wishes to support TLS? Domains may change owners anytime. Should disabling STS be allowed? 

I think there should be some way to confirm that the domain owner has really changed or the client should offer list of STS domains where it could be removed in this case.

Possibly offtopic for this, but what should happen when the client connects using plain text and enables STS and encounters invalid SSL certificate? Should the client author understand to verify the certificate manually (from where/how)? I think this might be a little too much to expect from average user.

core/sts-3.3.md
```diff
@@ -0,0 +1,94 @@
...
+
+Once the client has confirmed that the `sts` capability is indeed offered over
+a secure connection, it then MUST only attempt secure connections to the
+server from now on. It MUST refuse to connect if a secure connection cannot
+be established with the server for any reason.
+
+If the client cannot establish a secure connection for any reason, it MAY
+continue using the insecure connection.
+
+Clients SHOULD NOT request this capability (`CAP REQ`).
```
Mikaela: Shouldn't this be in ISUPPORT then or why capability that may not be used?

attilamolnar: I've explained this thoroughly in #77 (link: https://github.com/ircv3/ircv3-specifications/issues/77#issuecomment-43907984).

attilamolnar:
> Possibly offtopic for this, but what should happen when the client connects using plain text and enables STS and encounters invalid SSL certificate?

Good question, the spec intentionally leaves out the definition of what is a "secure connection". For example one client may only consider connections as secure when a server presents a certificate signed by a CA that it trusts, another client may be ok with self signed certificates after being accepted by the user, or a combination of these.

core/sts-3.3.md
Mikaela: I see. I think that cap might not be correct place for it, but I don't see any other place either.

Mikaela:
I took

> If the client cannot establish a secure connection for any reason, it MAY continue using the insecure connection.

as answer and based my strikethrough on it. I think secure connection is connection that has been established successfully and the certificate is trusted by client either by being from trusted CA or having the fingerprint manually accepted. Clients that don't perform certificate checking probably accept everything.

I don't know if client should think invalid certificate as failure immediately and reconnect with plain text or hang the connection until user appears to accept the failed certifiacet though.

core/sts-3.3.md
```diff
@@ -0,0 +1,94 @@
...
+is a value attached, the value is separated from the key by a `=`. That is,
+`<key>[=<value>][<key2>[=<value2>][<keyN>[=<valueN>]]]`. Keys specified in
+this document MUST only occur at most once.
+
+See [capability negotiation 3.2](capability-negotiation-3.2.html) for more
+information about capabilities with values.
+
+When the client sees the `sts` capability advertised over a non-secure
+connection, it MUST first establish a secure connection and confirm that the
+`sts` capability is still present.
```
Mikaela: I am not sure on this part with further thinking, if `sts` capability is there with plain connection and disappears with secure connection, doesn't that mean that there is MITM attack going on and the client should use SSL to avoid it?

core/sts-3.3.md
```diff
@@ -0,0 +1,94 @@
...
+
+Once the client has confirmed that the `sts` capability is indeed offered over
+a secure connection, it then MUST only attempt secure connections to the
+server from now on. It MUST refuse to connect if a secure connection cannot
+be established with the server for any reason.
+
+If the client cannot establish a secure connection for any reason, it MAY
+continue using the insecure connection.
```
jwheare: Which one is it? `MUST refuse to connect` or `MAY continue using the insecure connections`

core/sts-3.3.md
```diff
@@ -0,0 +1,94 @@
...
+
+Strict Transport Security (STS) is a mechanism which allows servers to
+communicate that complying clients should only connect to them over a secure
+connection.
+
+The policy is communicated to the client via the `sts` capability and should
+be processed by the client at capability negotiation time.
+
+The value of the capability specifies the duration during which the client
+MUST only connect to the server securely (using TLS, aka. SSL), and the port
```
DarthGandalf: Or via starttls?

core/sts-3.3.md
```diff
@@ -0,0 +1,94 @@
...
+number to use for it.
+
+## Details
+
+The capability value, which MUST always be present, is a comma (`,`) separated
+list of tokens. Each token is a key which may have a value attached. If there
+is a value attached, the value is separated from the key by a `=`. That is,
+`<key>[=<value>][<key2>[=<value2>][<keyN>[=<valueN>]]]`. Keys specified in
```
DarthGandalf: `,` missing

core/sts-3.3.md
jwheare: sts obsoletes starttls imo.

core/sts-3.3.md
DarthGandalf: If client sees this in 005, client can upgrade to TLS using STARTTLS immediately, without reconnect.
But on the next connect, client will remember (using sts) that server supported tls, and will upgrade connection (or use another port) before registration

core/sts-3.3.md
jwheare: If there is an MITM attack directing you to connect to a specific port, that to me is a red flag that you someone is trying to direct you to e.g. a compromised port. As such, no sts cap advertised on insecure protocols should be trusted, beyond the unavoidable necessity of actually trying to connect to that port.

This is, as hinted in the PR, where HSTS differs from IRCSTS. See http://tools.ietf.org/html/rfc6797#section-8.1
> If an HTTP response is received over insecure transport, the UA MUST ignore any present STS header field(s).

And also see http://tools.ietf.org/html/rfc6797#section-14.5 about the risks of obeying compromised directives.

Basically, there's no guarantee that the secure port you've been connected to is legitimate, so using it shouldn't be the default. This is probably a case where a client should present a warning to the user.

core/sts-3.3.md
DarthGandalf: See comment below, I think they work fine together when sts is done via 005.

core/sts-3.3.md
DarthGandalf: During MITM, certificate will be bad. But it's still good idea to show `sts` on secure connection, to update the timeout in client's sts cache.

core/sts-3.3.md
jwheare: 005 is unsuitable, as explained in #77. Also, see comments about MITM and not trusting anything sent on an insecure connection beyond absolutely necessary.

core/sts-3.3.md
```diff
@@ -0,0 +1,94 @@
...
+
+This can be used to enforce TLS for `irc.example.com` as well if the
+client is connected to `example.com`.
+
+TODO: This only works with SNI since otherwise servers don't know what domain
+is used by the client.
+
+## Security
+
+It is possible for attackers to inject the STS policy into a plaintext
```
DarthGandalf: It is possible, but why would they do that? What is the attack vector here?

DarthGandalf:
> I don't know if client should think invalid certificate as failure immediately and reconnect with plain text or hang the connection until user appears to accept the failed certifiacet though.

That's out of scope. It's up to client to decide what to trust based on certificates and whatnot.

core/sts-3.3.md
jwheare: Similarly to the attack outlined here: http://tools.ietf.org/html/rfc6797#section-14.5 you can force a redirect to a different port on a compromised machine, and stop any legitimate attempts to connect to e.g. irc.freenode.net:6667 in future.

Once the machine has been returned to the control of the owners, that host/port is still tainted in all client's hsts caches, and any attempts to connect to it will forever (or for a long time) be forced to the spoofed port.

DarthGandalf: I see.
But with starttls, this is not an issue.

core/sts-3.3.md
attilamolnar: This paragraph is meant to cover the case when the client has seen the STS advertisement over a plaintext connection but has not succeeded to establish a secure connection.

core/sts-3.3.md
jwheare: starttls avoids the issue of a spoofed port transfer, but is vulnerable to password stealing. A fresh reconnect is not. The correct way to avoid the issue is to just not trust any sts cap details received on an insecure connection, beyond the initial port redirect. You then only set your sts cache according to the cap details received from the fresh secure connection.

core/sts-3.3.md
attilamolnar: I've left out STARTTLS intentionally as it adds non-trivial complexity for a small benefit. This correlates with the long-term goal: phasing out plaintext IRC.

core/sts-3.3.md
attilamolnar: If the client accepts STS policies which came over insecure connections denial of service may happen with or without STARTTLS. Suppose the server does not support TLS and an attacker injects `sts=age=15552000`.

Suppose the client does not verify this over a secure connection.

This is what happens without STARTTLS:
Client tries to connect to port 6697, cannot connect or can connect but the handshake fails. Because the client is supposed to establish a secure connection and it cannot do so it is a DoS.

With STARTTLS:
Client connects to the same port, sends STARTTLS. It is not available, which is a fatal error because the client is supposed to establish a secure connection, fallbacks are not allowed. Same as above.

DarthGandalf: > is vulnerable to password stealing

I agree with this.

> The correct way to avoid the issue is to just not trust any sts cap details received on an insecure connection, beyond the initial port redirect. You then only set your sts cache according to the cap details received from the fresh secure connection.

Attacker is more likely to just remove the whole sts from plain text connection...
But ok.

attilamolnar: > Attacker is more likely to just remove the whole sts from plain text connection...

True, but that's not worse than using a plaintext connection, which happens without STS.

DarthGandalf:
There is a small difference between HTTP and IRC: IRC connections are long.
Let's say duration is set to 6 months. Client connects to server, sees sts, reconnects with tls, saves 6 months duration to the cache, then stays connected for 8 months. By that time, cache already expired, and client connects via plain text again.

I don't know what to do with it though, besides some periodic updates which are not via `CAP`

core/sts-3.3.md
attilamolnar: Replacement paragraph:

> However, if the client fails to confirm whether the `sts` capability is legitimate (i.e. it has not managed to connect securely and get the advertisement over a secure connection for any reason), it MAY continue using non-secure connections.

Do you think this is OK?

attilamolnar:
> There is a small difference between HTTP and IRC: IRC connections are long.
Let's say duration is set to 6 months. Client connects to server, sees sts, reconnects with tls, saves 6 months duration to the cache, then stays connected for 8 months. By that time, cache already expired, and client connects via plain text again.

Good point, multiple solutions exist:

The most obvious one is to raise the duration. Another solution for the client is to periodically run `CAP LS` and see if the STS advertisement is still there. If it is, then the cache can be updated to restart the 6 months from now. Doing this once in a day is reasonable and shouldn't cause any problems for either side. Clients may send the query less often e.g. once per week if the original duration was long enough, e.g. months.

DarthGandalf:
Hm, I've just thought about another solution:
Client starts the timeout not on connect, but on disconnect. If server needs to change the duration for any reason, server can send
```
:server CAP DEL :sts
:server CAP NEW :sts=...
```

attilamolnar:
> Client starts the timeout not on connect, but on disconnect.

But then during the lifetime of the first connection the client must still use secure connections, right?

jwheare:
It doesn't matter how long the session is, how long max-age is, or when you start the timer, you'll always encounter a plain text reconnect at *some* point on expiry, unless the client has a list preloaded.

attilamolnar:
> It doesn't matter how long the session is, how long max-age is, or when you start the timer, you'll always encounter a plain text reconnect at some point on expiry, unless the client has a list preloaded.

Wrong. If the client sees an advertisement, the proper behavior is to reschedule expiration.

core/sts-3.3.md
jwheare: That clarifies what you meant, but I don't think it's right. The client should probably not cache failures. So it should try the redirect and fail every time. The client can choose to not retry automatically, but if the user reconnects manually I think it should use the non-secure connection and still follow the invalid secure-redirect.

This is similar to bad HTTP redirects. You don't just ignore the redirect if you can't follow it correctly, you fail.

jwheare:
Yup, I stand corrected. Setting expiry from the disconnect time makes sense to me.

attilamolnar:
Note that this is a bit tricky, because if the client saves its configuration only on disconnect a power outage means the client forgets about STS and will connect via an insecure channel next time, unnecessarily.

If the client saves its configuration on connect and disconnect there can be a window when the policy is expired but the client has not yet disconnected and the same can happen as above.

Clients must be careful to avoid these.

DarthGandalf:
> Note that this is a bit tricky, because if the client saves its configuration only on disconnect a power outage means the client forgets about STS and will connect via an insecure channel next time, unnecessarily.

```
:server BATCH +foo ...
@batch=foo :server CAP DEL :sts
@batch=foo :server CAP NEW :sts=...
:server BATCH -foo
```

DarthGandalf:
Oh, this is cure for different kind of power outage. But yes, clients must be careful anyway.

jwheare:
To avoid complexity on the client, the logic should probably be to save the max-age value, then set an expiry to `connect_time + max-age` on connect, then on disconnect, read the stored max-age again and update the expiry to `disconnect_time + max-age`. That means you at least have a policy in place in case the disconnect isn't clean. I suppose the client may choose to update the policy on some interval basis too.

If the server wishes to change the policy and take advantage of the max-age expiry but still has long running clients, it can send updated `sts` CAPs to all active connections that have ACKed sts.

jwheare:
As written, this spec doesn't show the CAP being ACKed. Should it?

attilamolnar:
Clients don't request this cap, they either follow it or they don't. They could request it for statistics, but we cannot enforce that.

jwheare:
OK so updated `sts` is sent to all clients then.

jwheare:
How about, if a client is still connected when the expiry is reached, it automatically renews it.

jwheare:
That does require proactively checking for expiry on a timer, whereas only setting it on connect and renewing on disconnect does not, and would be simpler to implement.

jwheare:
From the other items for discussion: disabling STS is done by setting duration to 0, and should be sent to all active clients.

attilamolnar:
> That does require proactively checking for expiry on a timer, whereas only setting it on connect and renewing on disconnect does not, and would be simpler to implement.

Setting a timer is not a big deal, it's better design than pushing `CAP` commands to clients every once in a while to make sure they update their configs. Either way, this should be a suggestion in the specification.

DarthGandalf:
> How about, if a client is still connected when the expiry is reached, it automatically renews it.

That implies that "expiry is reached" always triggers some code to be run. It's not always the case, cache can be consulted only during connection attempt.

DarthGandalf:
> That does require proactively checking for expiry on a timer, whereas only setting it on connect and renewing on disconnect does not, and would be simpler to implement.

On connect duration is set to 6 months, and connection is lost after 5 months 30 days 23 hours 59 minutes 58 seconds. Voila, reconnect will be plain text.

DarthGandalf:
> disabling STS is done by setting duration to 0

Or by deleting it from CAP

DarthGandalf:
> Setting a timer is not a big deal

Correct.
Client can run a daily timer: if it's still connected, update duration.

attilamolnar:
So renew when...
* The server explicitly modifies the policy (this includes the initial update on connection)
* On disconnect
* No renewal occurs for one day

jwheare:
> On connect duration is set to 6 months, and connection is lost after 5 months 30 days 23 hours 59 minutes 58 seconds. Voila, reconnect will be plain text.

The same can be said for a duration set to 24 hours and a daily timer. I think we can agree that *some* plain text reconnects will be unavoidable.

DarthGandalf:
This is actually the same as with HTTP HSTS.
If server sets so low duration, next reconnect (whenever that happens) is more likely to be plain text.

jwheare:
Worth adding a subdomains example: Connecting to chat.example.net with a wildcard cert `CN=*.example.net` the advertisement can be set to any of:

`CAP * LS :sts=duration=31536000,sub=irc.example.net,chat.example.net`
`CAP * LS :sts=duration=31536000,sub=*.example.net`

Any issues with that? I guess they aren't really subdomains, just alternate domains, so maybe `domain=` is a better key

If no wildcard cert is used, do we want to allow setting STS on actual subdomains, e.g. `sts` advertised on root `example.net` with a valueless `sub` will enforce sts on `*.example.net`

`CAP * LS :sts=duration=31536000,sub`

These are different use cases, so maybe different key names.

jwheare:
With regards to dealing with failed secure connection errors, I think there needs to be something written along the lines of the "No user recourse" section from HSTS http://tools.ietf.org/html/rfc6797#section-12.1

> Not doing this, i.e., allowing user recourse such as "clicking
   through warning/error dialogs", is a recipe for a man-in-the-middle
   attack.  If a web application issues an HSTS Policy, then it is
   implicitly opting into the "no user recourse" approach, whereby all
   certificate errors or warnings cause a connection termination, with
   no chance to "fool" users into making the wrong decision and
   compromising themselves.

core/sts-3.3.md
attilamolnar: How about (changed part is in bold):
> However, if the client fails to confirm whether the `sts` capability is legitimate (i.e. it has not managed to connect securely and get the advertisement over a secure connection for any reason), **it MUST ignore the STS advertisement came over the non-secure connection. It SHOULD retry making a secure connection whenever it gets the advertisement again.**

core/sts-3.3.md
```diff
@@ -0,0 +1,94 @@
...
+If this key is present, it MUST have a single port number as its value. If
+this key is not present, it defaults to port 6697.
+
+### The `sub` key
+
+If this OPTIONAL key is present, it indicates that this policy MUST
+be applied to all subdomains of the server as well.
+
+This can be used to enforce TLS for `irc.example.com` as well if the
+client is connected to `example.com`.
```
DarthGandalf: Evil server can poison client cache with `sub=*`, and client won't be able to connect to non-SSL networks anymore.

DarthGandalf: I meant this as reply to @jwheare

attilamolnar:
Do you want to require signed certs?

core/sts-3.3.md
attilamolnar: The certificate scope should limit what can be included.

jwheare:
> Do you want to require signed certs?

Yes, self-signed certs aren't helpful for the general populace. If you already know you trust the server and know the fingerprint of the cert and can verify all that yourself, you'll be using a secure port, and `sts` isn't really for you.

SaberUK:
If you require that then the spec is going to be dead on arrival. Considering that 99% of networks use self signed certificates and have no desire to change we should look at alternate methods (e.g. DANE) which they can use with their existing setups rather than requiring those networks to subscribe to the scam which is certificate authorities.

jwheare:
Pretty sure we can limit this spec to networks that care enough about user trust to get a grown up cert. And for everyone else there'll be https://letsencrypt.org/ soon.

Cronus89:
I agree with @SaberUK I run and help a few networks.  None of us are going to buy certs for this.  I thought this was suppose to push everyone to use TLS connections and get rid of plaintext?  In doing the way you want, this wont happen in the slightest and you'll be right where we are now.

SaberUK:
Calling certificates provided by a certificate authority "a grown up cert" is a quite frankly load of bullshit. We should not be pushing something which is widely considered by security experts (e.g. Moxie Marlinspike) to be be a joke. It is really not that hard to get a valid certificate for a domain you do not own as proven [many](https://en.wikipedia.org/wiki/DigiNotar), [many](https://en.wikipedia.org/wiki/Comodo_Group#Certificate_hacking) times.

jwheare:
@SaberUK Do you and all your network users know how to verify self signed certs when you connect to such networks? Congratulations, sts is not for you, your lack of support for this spec doesn't make it dead on arrival. No incarnation of this spec will be relevant to you because it is simply not needed. Your advanced understanding of public key encryption makes it superfluous, you already know to connect to the right ports.

Now, does the same apply to freenode users, just to take the largest obvious example? No. For them having a trusted authority verify certificates is extremely useful to prevent man in the middle attacks without requiring any understanding of certificates whatsoever.

Is that system flawed? Sure, but it's the best we have at the moment, and the flaws are being addressed.

I'm sorry for casting aspersions previously but self signed certs just do not align with the goals for this particular spec. They may have their uses in providing encryption against casual snooping, but on a hostile network, with no easy way for a user to verify certs, they are incredibly vulnerable to mitm.

Aerdan:
It doesn't matter whether the certs are self-signed or not if the client has a list of signatures to work with, which it should if STS is to work correctly. Obviously that can't help with networks the client wasn't preconfigured to support, but most users aren't going to attempt to connect to networks that aren't in their client's server list anyway, and those that do can follow a tutorial provided by the network.

jwheare:
@Aerdan so you're suggesting clients need to keep an updated list of the current fingerprints of all the servers they connect to? That's not at all needed for STS, it's not how browsers work (they have trusted CA fingerprints only, either bundled or from the OS), and it doesn't scale, even if IRC clients all had the same aggressive update schedule as some browsers do.

Not to mention it's not very, er, open. I should be able to host a brand new server, get a free CA trusted cert, and invite people to join, knowing that after their first connection, they'll always connect securely with a compliant client, and will never send a SASL password over an unprotected channel. That should NOT require certificate pinning to work.

Aerdan:
Yes, I misspoke. It's not required for STS to function, but clients already have most of the information they need if they are SSL-only…and having a local fingerprint store protects against first-connect MitM potential. Also, STS requires you to *already* be connecting over TLS (since otherwise the relevant information would be compromisable via MitM).

jwheare:
The first-connect problem is mitigated with client bundled preload lists. Not for the fingerprint, but for the domain. And they are an enhancement, not a block for getting the basic level of STS protection.

Local fingerprint stores are problematic, what if the server legitimately changes it's cert, after the previous one is revoked or expires? A client that validates with a fingerprint store would have to show a big scary warning, just like SSH does, which users would just learn to ignore and workaround. We don't want to train people to ignore security warnings.

I don't understand your point about requiring to connect over TLS.

Aerdan:
Forgive me if I'm repeating information you already know; I think there is confusion over what STS means.

HSTS works by sending a header that details the length of time the client should trust the current certificate, along with information that e.g. says the client should trust subdomains as well. **This header is ignored on plain-text connections because it has no meaning for such connections.** This header therefore only has meaning if the client is already using TLS. With that length-of-time information, the client knows they should continue to use TLS for that server exclusively. 

None of this instructs the client as to how to connect to the server over TLS, nor what information they need to ensure the server is who they claim to be. I think the consequences are obvious.

jwheare:
HSTS stipulates nothing about trusting a certificate. The client is free to use whatever trust mechanism it chooses for that, whether it's a list of bundled root CAs, or a user provided verification. It's perfectly possible for a server to direct a client to an untrusted certificate using STS.

I understand that STS requires a reconnection over TLS if an insecure one was made initially, I just didn't follow the relevance of that to discussion about self-signed signatures vs CA-signed ones.

Aerdan:
This is true. This is why I suggested including fingerprint information in the server list when I commented earler.

jwheare:
@attilamolnar can you clarify why you asked about self-signed certs initially? Is it to do with trusting wildcard certs for subdomain propagation?

Aerdan:
In any event, I don't even understand the whole point of this.

1. Strict Transport Security only ensures that clients always use TLS. This is solved by not supporting plaintext IRC connections and having only TLS servers in your server list.
2. Any proposal to provide a TLS upgrade mechanism can be MITM'd, which kind of defeats the whole point.
3. Self-signed certificates are inherently insecure because you can't verify them, which you need to be able to do in order to be secure.

jwheare:
1. It provides an upgrade path towards disabling plaintext that is backwards compatible. Killing plaintext is a beautiful dream, but we need to get there by taking realistic, achievable steps. Turning off support in one go is not realistic.
2. Preload lists.
3. Agreed.

Mikaela:
I will ask here too as I didn't fully understand IRC:

Do preload lists mean that clients have to start keeping long list of IRC networks and their fingerprints within their sources?

It was said that preload lists are the server lists what clients already have, but WeeChat and ZNC for example have no servers configured by default.

Aerdan:
1. It can't do that, because see my second point.
2. That's what server lists are. And if you're providing a server list, it's on you to ensure that every server can do TLS and to default *to* TLS. If we ignore self-signed certs, we don't even need fingerprint storage in the server lists.

jwheare:
Preload lists have nothing to do with fingerprints. They are just a list of servers that do TLS and wish to default to TLS.

Many clients have a bundled list of networks, but they might still allow a user to connect manually to those networks insecurely. STS allows those servers to disallow this behaviour.

Typically preload lists are populated from servers that already advertise STS. In browsers, to get on a preload list you have to be correctly serving the HSTS header with an expiry of 1 year. IRC preload lists should be validated the same way before inclusion, and probably tested regularly by the client maintainer. As stated at the start, preload lists are somewhat out of scope of this particular PR, but there should be a more detailed spec about them too.

Mikaela:
And if client doesn't provide a server list, it's incompatible with IRCv3.3 as it cannot have preload lists (and this is supposed to be base extension)?

jwheare:
Preload lists are not part of this spec, and probably would not be a requirement for 3.3 base compatibility.

Mikaela:
So it would be enough for client to check for the existence of `STS`, disconnect and reconnect to the port mentioned, validate the certificate and see that `STS` is there and then use TLS when connecting to the network in the future for the next e.g. 6 months?

jwheare:
Yes, and certificate validation is mostly out of scope. But if the client decides that a certificate is not valid, then it should terminate the connection.

awilfox:
If certificate validation is out of scope, how can this ever hope to work?  All I have to do is DNS poison irc.freenode.net, sign a certificate with

```
openssl genrsa -nodes -out server.key 2048
openssl req -new -key server.key -out server.csr
openssl x509 -req -days 365 -in server.csr -signkey server.key -out server.crt
```

and now I not only have the victim's SASL credentials but also have their trust that they are communicating securely.  It wouldn't even be that hard to write a custom ircd that takes information from a relay bot of some kind to masquerade as the real freenode, while keeping logs of everything the victim(s) say.

Certificate validation needs to be performed and specified as required, or this spec is completely worthless; and no, for what it's worth, [browsers don't allow that in HSTS, either](https://bugzilla.mozilla.org/show_bug.cgi?id=689608) due to the spec [expressly forbidding it](https://tools.ietf.org/html/rfc6797#page-27).

jwheare:
I meant, certificate validation is necessary, but the mechanism of how a client chooses to validate is out of scope and covered by other RFCs.

attilamolnar:
@jwheare 
> @attilamolnar can you clarify why you asked about self-signed certs initially? Is it to do with trusting wildcard certs for subdomain propagation?

Yes, that's the reason exactly. I'm not sure whether it's worth the tradeoff as it will require talking about certificates in the spec opening up a can of worms while the benefits are limited to users who switch servers manually (irc.example.net to server1.example.net). Those users are IMO more likely to be non-casual users (but thinking that most non-casual users who know IRC well use secure connections may be a false assumption).

The best course of action would be to keep the core of the spec in this PR simple and incrementally add new features as we research the topic (e.g. are wildcard certs common? SNI support in clients?).

Mikaela:
Just confirming: if I always use TLS (and ZNC either trusts the CA or forces me to type `/znc addtrustedserverfingerprint SHA-256-CERTIFICATE`) and see the STS cap, I am always using STS and cannot connect with plain text and the plain text connection isn't needed?

>  are wildcard certs common? SNI support in clients?

From my networks freenode & pirateirc have valid wildcard certificate. Mozilla's certificate again appears to be only for `irc.` and same with ChatSpike (which has invalid certificate).

With SNI support, I only know of ZNC which has open issue about it (https://github.com/znc/znc/issues/948) and it doesn't have support at least looking at comments from @DarthGandalf and @teward.

awilfox:
Genuine question: How would SNI even be useful?  You can't possibly hope to run multiple IRC servers on the same TCP port, and I can't imagine someone getting a certificate for both 'foo.my.network' and 'irc.my.network'... I suppose if we revive the third command of USER as servername and then use STARTTLS, we could get /something/ like that?  But I thought this proposal was about completely eliminating plain text registration beyond CAP.

If it means anything, the network on which I am a server admin also has a valid wildcard certificate generated by a trusted CA, and our users seem to be very pleased at not getting a bunch of warnings from HexChat and WeeChat about it.

Mikaela:
>  can't imagine someone getting a certificate for both 'foo.my.network' and 'irc.my.network'... 

if we play weirder version of freenode and I understood correctly what people are after w can't imagine someone getting a certificate for both 'foo.my.network' and 'irc.my.network'... ith SNI, I could connect to irc.freenode.net and rajaniemi.freenode.net without them having wild-card certificate but different name for both.

> If it means anything, the network on which I am a server admin also has a valid wildcard certificate generated by a trusted CA, and our users seem to be very pleased at not getting a bunch of warnings from HexChat and WeeChat about it.

May I ask which network and how much that certificate costed and also amount of users and servers? That was wondered at IRC (not-#IRCv3) earlier when thinking about networks with valid certificates.

glyph:
> Genuine question: How would SNI even be useful? You can't possibly hope to run multiple IRC servers on the same TCP port, and I can't imagine someone getting a certificate for both 'foo.my.network' and 'irc.my.network'... I suppose if we revive the third command of USER as servername and then use STARTTLS, we could get /something/ like that? But I thought this proposal was about completely eliminating plain text registration beyond CAP.

Sure you can.  With SNI support, you can have your TLS terminator do load balancing and direct traffic for different hostnames to entirely different processes.  Not only could you hope to run multiple IRC servers on the same TCP port, you could run multiple IRC servers, a web server, a mail server, and an XMPP server all on the same port and multiplex them just by hostname.

SNI is a basic part of modern TLS infrastructure.  You should always mandate it in anything that specifies interop with TLS.  Given the scarcity of IPv4 and the inconsistent rollout of IPv6, it's often necessary to conserve the scarce resource of address space, especially for smaller sites.  The only reason that people are reluctant to adopt it is the depressingly high market share of antediluvian versions of Android and Windows clients that don't support it.  However, those market shares are dropping every day.

grawity:
> How would SNI even be useful? You can't possibly hope to run multiple IRC servers on the same TCP port

Well that's exactly how SNI would be useful, no?

Personally I see TLS SNI and ALPN as the missing session layer which would allow generic named host & named protocol negotiation without having to add application-layer-specific stuff in every single protocol.

(stunnel, for example, already supports SNI – though not yet ALPN – so you can mix many different services on the same TCP port.)

awilfox:
Okay, to clarify, I do see that the *possibility* exists in the world that you can use the TLS protocol to multiplex a single TCP port to multiple ircds, but I still can't imagine anyone doing it.  Does anyone *truly* want to run multiple networks on even the same IP, let alone the same port?  Sounds like it would just be that much more vulnerable to DoS attacks.  Not to mention the fact that at least some networks want client nodes on isolated machines...

@Mikaela, sure you can: the network is Interlinked, we purchased it from PositiveSSL (each of our opers contributed), and we have four servers for an average of about 100 users.  We also use the certificate for our Web site, our wiki, our qdb, our oper mail server (services emails and opers@ etc), and other various services.  We definitely use it heavily and I think it's quite useful to not have warnings popping up in WeeChat, Firefox, Thunderbird, and all.

We believe strongly in security and encryption, as I believe the community at large does as well.  That's why we use TLS wherever we can.

Elizafox:
There is no actual way to host multiple IRC networks on the same IP. I had proposed a similar proposal involving USER to do a vhost sort of thing and routing, but it had lukewarm reception and I abandoned it.

@glyph There is no use case for this in the IRC world as it stands. For load balancing, DNS round-robin is traditionally used, not IP-based balancing. Networks which do any kind of multiplexing with an IP like you describe use anycast (DALNet comes to mind). It doesn't bring much benefit to have endpoint-based load balancing for IRC; there is no such thing as php-cgi/wsgi like things for IRC, and all servers in a network are linked together in a tree. Just use DNS, seriously.

In addition, vhosts complicate server-to-server links and it is difficult at best to imagine a use case for load balancing due to the above, and impossible to imagine how a proxy could possibly work for virtual hosts in the present protocol.

It is worth noting that I proposed an extension to USER to allow for this, but 1) this does not solve the S2S issue, 2) had a tepid reception at best.

attilamolnar:
Let me clarify this: the use case for SNI here is NOT to run multiple servers or services on the same port, but the following:

* SNI makes it possible for the server to know whether to send the "sub" key
* LetsEncrypt is not planning to sign wildcard certificates but will sign multiple certificates e.g. for irc.example.com, example.com, server2.example.com. Reference: https://letsencrypt.org/faq/.

@glyph @awilfox @Elizafox Can you please edit your comments which contain off-topic discussion to mark them as such? That way future visitors may choose to skip reading the (uninentional) derailment. I believe that me editing your comments is not appropriate nor necessary on this issue tracker. I consider discussion about multiple servers on the same port, modifying `USER`, S2S problems off-topic.

Mikaela:
What is the current status with this PR and are self-signed certificates allowed or not?

Currently it seems that there are:
* three places for free signed certificates:
    * https://buy.wosign.com/free/
    * https://www.startssl.com/
    * https://www.cacert.org/ *even if it's not supported universally on all OSes/distributions ([CAcert Inclusion Status](https://wiki.cacert.org/InclusionStatus))*
* and only irssi supports DANE
    * but others have issues about it
        * https://github.com/znc/znc/issues/784
        * https://github.com/weechat/weechat/pull/121
* SNI again seems to have one issue and one pull request that I am aware of:
    * https://github.com/hexchat/hexchat/pull/1478
    * https://github.com/znc/znc/issues/948

*I am not actively watching all IRC clients, so this list is biased.*

moparisthebest:
I'm the one who submitted the pull request to hexchat for SNI support, I also had [one accepted for Yaaic for SNI support](https://github.com/pocmo/Yaaic/pull/168) as well.  I guess my question for those opposed would be why NOT add SNI support?  It's one line of code in most TLS libraries I've seen, and it lets you do all the things that have been mentioned that are not possible otherwise.

I would personally also suggest setting the ALPN name to 'irc' or something standard where possible for similar reasons.

Quick list of reasons for SNI:
1. Allows server to pick the correct certificate based on the host name, in case there are multiple.
2. Allows server to pick the correct service based on host name, this is technically what ALPN is for but support is low/nonexistent.
3. Allows routing and load balancing without decryption.

I personally have www.mydomain.org and irc.mydomain.org both listening on port 443, and routed to either the irc server or the web server based on SNI name.  At least stunnel/haproxy/sslh and probably more all have support for doing so.

attilamolnar:
Pushed new version for discussion with a new "no user recourse" section as suggested by @jwheare, and many other improvements. Please take your time and read it through.

There's also a new SNI section which we may want to keep here and expand or move to another document if consensus is to require SNI from all 3.3 clients.

attilamolnar:
As far as this document is concerned, SNI is only required for the `sub` key, it is not a hard requirement of this specification but enhances it.

jwheare:
I think for this proposal to move forward it will require some positive momentum from a network operator. Is this something that networks want? Client development doesn't make much sense before we've decided that. Anyone from freenode or elsewhere care to comment? cc: @tomaw

jwheare:
From a freenode staffer on IRC (not an official response):

```
17:08 <evilmquin> if a scheme is arrived at that makes it easy to steer clients to TLS I expect
      we'd consider doing that
17:09 <evilmquin> even more likely if someone makes charybdis do it, and we can pull it in
      when we next merge
17:10 <jwheare> evilmquin: could you comment on the issue please
17:12 <evilmquin> jwheare: I'll let tomaw do that as he's already been pinged, probably better
      to have more than one view on it from our side anyway
17:13 <jwheare> evilmquin: ok thanks
```

attilamolnar:
I'll remove the `sub` key from this for now so we don't have to deal with any problems it may bring. It should be added at a later date.

Mikaela:
From what I understood at IRC it was thought as OK to require certificate to be trusted by the operating system as LetsEncrypt is scheduled to come next week and there are already WoSign and StartSSL and even without wildcard certificate operators could sync one file between hosts that is valid for `irc.example.net`.

jwheare:
Summary of IRC discussion regarding TODOs:

> Should it be possible for the server to explicitly remove the policy after it has been set?

Yes, this should be documented by allowing duration=0

> Should it be clarified what counts as a secure connection?

A note could be added along the lines of:

"Secure connections must be signed with a certificate from a reputable CA chain, using either the OS trust root or a widely trusted list such as the Mozilla Included CA List"

See:
https://mozillacaprogram.secure.force.com/CA/IncludedCACertificateReport
http://www.chromium.org/Home/chromium-security/root-ca-policy

attilamolnar:
Done, ready for merge. Issues #176 and #178 can be used for additional discussion about the remaining TODO items.

DanielOaks:
+1

core/sts-3.3.md
```diff
@@ -0,0 +1,129 @@
...
+The policy is communicated to the client via the `sts` capability and should
+be processed by the client at capability negotiation time.
+
+The value of the capability specifies the duration during which the client
+MUST only connect to the server securely (using TLS, aka. SSL), and the port
+number to use for it.
+
+Secure connections must use a TLS server certificate that is signed with a
+certificate from a reputable CA chain, using either the OS trust root or a
+list such as Mozilla's.
```
ahf: A client can support TLS pinning and/or DNSSEC + DANE verification. No need to put this into a spec.

attilamolnar: Yes, this is why I initially tried to avoid dealing with what should be a "secure" connection in this spec, but realized that it's a better idea to talk about it than to let client devs think it's fine to auto-accept a self-signed cert.

moparisthebest: @ahf what is "TLS pinning" if you aren't talking specifically about DNSSEC + DANE?

And while I think DANE should definitely be supported, it should *probably* be defined somewhere what types IRC should support.  SMTP for example only supports DANE-EE and DANE-TA, and never PKIX-EE or PKIX-TA.

dequis: @moparisthebest that's all out of scope of STS.

The idea is to make this paragraph less specific to *allow* clients to use alternative methods if they consider them better than just traditional CA validation. Which can be a controversial topic, but there's no need to get into those details.

Mikaela:
I opened thread on my previous comment at [LetsEncrypt community](https://community.letsencrypt.org/t/round-robins-and-irc/3662) and it appears that in case of IRC it requires syncing the certificate manually. SRV record was also suggested:

> a) Each server has its own FQDN with own keys srv.example.net

However it's not supported by any client that I am aware and https://github.com/ircv3/ircv3-specifications/pull/59 seems to be open.

jwheare:
Happy for this to be merged.

Mikaela:
It seems that the short definition of securely got removed, but I assume that means it's going to be in the other document that has been talked about?

attilamolnar:
Hopefully. But I think such a document does not belong under the ircv3 umbrella as it is not related to the irc protocol and has no visible consequences on the server side if not followed.

core/sts-3.3.md
```diff
@@ -0,0 +1,189 @@
...
+
+### Ignoring invalid request
+
+Server tells a client that is connected non-securely that the client must
+use secure connections for roughly 6 months.
+
+    Client: CAP LS 302
+    Server: CAP * LS :sts=duration=31536000
+
+The client ignores this because it received the STS policy over a non-secure
```
DarthGandalf: only "duration" is ignored, not presence of sts itself.
I don't want server to send different stuff on plain vs ssl connection.

jwheare: The lack of a port in this sts advertisement makes it useless on a non-secure connection. There's nothing else a client can usefully gain from this.

core/sts-3.3.md
```diff
@@ -0,0 +1,189 @@
...
+An `sts` capability having at least a `duration` key expresses an STS policy.
+
+See [capability negotiation 3.2](capability-negotiation-3.2.html) for more
+information about capabilities with values.
+
+### Mechanism
+
+When the client sees the `sts` capability advertised over a non-secure
+connection, it MUST first establish a secure connection and confirm that the
+`sts` capability is still present.
```
DarthGandalf: What will client do if secure connection didn't use sts?

core/sts-3.3.md
```diff
@@ -0,0 +1,189 @@
...
+It MUST refuse to connect if a secure connection cannot be established with the
+server for any reason during the lifetime of the policy.
+
+However, if the client fails to connect securely and get the STS policy over a
+secure connection for any reason it SHOULD retry the secure connection next
+time it receives the `sts` cap over a non-secure connection.
+
+Clients MUST NOT request this capability (`CAP REQ`).
+Servers MUST reply with an appropriate `CAP NAK` message if a client requests
+this capability.
```
DarthGandalf: For server it's easier to just use the same string in CAP LS, without caring whether client connected with ssl or not.

core/sts-3.3.md
```diff
@@ -0,0 +1,189 @@
...
+### The `duration` key
+
+This key indicates how long from receiving the advertisement clients MUST use a
+secure connection when connecting to this server. This MUST be specified in
+seconds as the value of this key and as a single integer without a prefix or
+suffix.
+
+Servers MUST send this key to securely connected clients.
+Servers SHOULD NOT send this key to non-securely connected clients, because it
+will be ignored.
```
DarthGandalf: For server it's easier to just use the same string in CAP LS, without caring whether client connected with ssl or not.

core/sts-3.3.md
```diff
@@ -0,0 +1,189 @@
...
+
+This key indicates the port number that clients can use when connecting over
+TLS. This key's value MUST be a single port number.
+
+If the client is not already connected securely, it MUST close the insecure
+connection and reconnect securely on the stated port.
+
+Servers MUST send this key to non-securely connected clients.
+Servers SHOULD NOT send this key to securely connected clients, because
+it will be ignored.
```
DarthGandalf: Client will ignore it, yes, but it's not good reason to not send it.

core/sts-3.3.md
jwheare: That's why it's SHOULD NOT instead of MUST NOT. I'd be happy with MAY personally.

core/sts-3.3.md
jwheare: Same again, servers can send it if they want, it's just pointless.

core/sts-3.3.md
jwheare: I don't know what this means. Example?

core/sts-3.3.md
DarthGandalf: It's pointless, but let's switch it to MAY.

core/sts-3.3.md
DarthGandalf: Client connects via plain text and sees:
```
CAP * LS :sts=port=12,duration=34 multi-prefix
```
Client reconnects to port 12 and sees...
```
CAP * LS :multi-prefix
```
Should it now reconnect to plain text again? I don't think so.

jwheare: No it shouldn't.

Mikaela: Did I understand this correctly?

Client connects via plain text and sees STS advertisement for port 12 and connects there with TLS and learns that

1. There is MITM in the plain text connection.
2. There is no proof of there not being MITM on the SSL connection as the advertisement wasn't inserted by the IRCd.

So the client shouldn't connect to MITMed plain text port, should it stay connected to potentially MITMed SSL port? I don't think so.

jwheare: I'm not sure how a client would *learn* that there's a MITM in the plain text connection.

We follow what browsers do. Namely, if they connect to an MITM HTTP site and see a redirect to another site they follow it blindly no matter they see at the redirect target. If there's an error at the target, they show the error state, they never rewind the redirect.

Mikaela: IRC:

```
2015-11-24 14:58:42+0200 < Not-6fa5> [ircv3-specifications] Mikaela reviewed pull request #155 commit - http://git.io/vB3ma
2015-11-24 15:03:20+0200 < Mikaela> Should that just say that the client should disconnect and not reconnect automatically (how does it fit the "don't ask the user"?) and port fixed in the example and STS is ready? Or are there more issues with it?
2015-11-24 16:54:57+0200 < Not-6fa5> [ircv3-specifications] jwheare reviewed pull request #155 commit - http://git.io/vBsIB
2015-11-24 17:06:23+0200 < Mikaela> The client knows that there is a MITM as the STS cap is not visible on the TLS connection
2015-11-24 17:06:46+0200 < jwheare> could just be a badly configured server and no mirm
2015-11-24 17:07:09+0200 < Mikaela> But in that case it's just adding port to example and it's ready for merge?
2015-11-24 17:07:25+0200 < Mikaela> AttilaM?
2015-11-24 17:07:27+0200 < jwheare> in these situations, clients should behave in simple predictable ways without trying to double guess, so that server admins can fix their problems properly
2015-11-24 17:07:55+0200 < Mikaela> Sounds good to me
2015-11-24 17:08:05+0200 < dx> if there's a MITM of the TLS connection, your root certificate store is compromised and you have no security at all
2015-11-24 17:08:22+0200 < dx> that's not a case we should be caring about
2015-11-24 17:09:21+0200 < Mikaela> So port is added to that example and it can be merged?
2015-11-24 19:43:11+0200 < Jobe> <Mikaela> The client knows that there is a MITM as the STS cap is not visible on the TLS connection
2015-11-24 19:43:25+0200 < Jobe> but whats to stop the MITM from listing STS in the cap response?
2015-11-24 19:43:52+0200 < Jobe> I mean if it has an unecrypted access to the C2S steam, it could manipulate what is sent anyway
2015-11-24 20:23:52+0200 < Mikaela> From what I understood that is ignored and it's just fixing the example before it gets ready for merging. But I would like confirmation now maybe three times.              
```

core/sts-3.3.md
Mikaela: Requesting for confirmation again: @attilamolnar adds the port (6697?) here and this finally gets merged?

DanielOaks: Just need to add a note explaining that the lack of the port on a
non-secure connection is why the request gets ignored. It's a nice example,
but at the moment it's a bit unclear why it gets ignored.

PR #156, opened 2015-06-26
Add the CHATHISTORY batch type specification.

SaberUK:
[Rendered view available here.](https://github.com/SaberUK/ircv3-specifications/blob/master+chathistory/extensions/batch/chathistory-3.3.md)

### DIFF
--- a/extensions/batch/chathistory-3.3.md
+++ b/extensions/batch/chathistory-3.3.md
@@ -1,0 +1,44 @@
+---
+title: IRCv3 chathistory Batch Type
+layout: spec
+copyrights:
+  -
+    name: "Peter Powell"
+    period: "2015"
+    email: "petpow@saberuk.com"
+---
+## `chathistory` Batch Type
+
+This document describes the format of the CHATHISTORY batch type.
+
+The CHATHISTORY batch type takes one parameter. This parameter contains the
+target of messages within the batch. The target MUST either be the nick of the
+remote client for private messages or the name of a channel which the
+local client is in for public messages.
+
+When a client which has enabled the [`batch` capability][batch] performs an
+action which requires chat history to be relayed to it (e.g. joining a channel
+with InspIRCd's `chanhistory` mode enabled) the server SHOULD put all PRIVMSGs
+and NOTICEs into a single CHATHISTORY batch. If the client has also enabled the
+[`server-time` capability][server-time] capability then messages inside the
+batch MUST be tagged with the time at which they were originally sent.
+
+## Examples
+
+### Channel Message
+
+    :irc.host BATCH +sxtUfAeXBgNoD chathistory #channel
+    @batch=sxtUfAeXBgNoD;time=2015-06-26T19:40:31.230Z :foo!foo@example.com PRIVMSG #channel :I like turtles.
+    @batch=sxtUfAeXBgNoD;time=2015-06-26T19:43:53.410Z :bar!bar@example.com NOTICE #channel :Tortoises are better.
+    @batch=sxtUfAeXBgNoD;time=2015-06-26T19:48:18.140Z :irc.host PRIVMSG #channel :Squishy animals are inferior to computers.
+    :irc.host BATCH -sxtUfAeXBgNoD
+
+### Private Message
+
+    :irc.host BATCH +sxtUfAeXBgNoD chathistory remote
+    @batch=sxtUfAeXBgNoD;time=2015-06-26T19:40:31.230Z :remote!foo@example.com PRIVMSG local :I like turtles.
+    @batch=sxtUfAeXBgNoD;time=2015-06-26T19:43:53.410Z :local!bar@example.com PRIVMSG remote :Tortoises are better.
+    :irc.host BATCH -sxtUfAeXBgNoD
+
+[batch]: ../batch-3.2.html
+[server-time]: ../server-time-3.2.html


kaniini:
:+1:

Aerdan:
:+1:

grawity:
Could this also handle joins/parts/quits/nicks? These are currently problematic due to being global (a replayed NICK would affect global state).

Mikaela:
> Could this also handle joins/parts/quits/nicks? These are currently problematic due to being global (a replayed NICK would affect global state).

This could possible help [ZNC's buffextras](http://wiki.znc.in/Buffextras) which has some issue that I don't remember, but there are some scripts on bottom of that list hinting that the actions aren't native to the client. The module adds joins/parts/quits/nicks to normal playback when client is attached.

DarthGandalf:
ZNC already supports this, but under name `znc.in/playback`, even the parameter is the same.

> Could this also handle joins/parts/quits/nicks? These are currently problematic due to being global (a replayed NICK would affect global state).

JOINs are problematic too, even though they are not global. In order to send them, server needs to send *outdated* list of users in channel in state how it was at beginning of the playback. It's possible to implement of course, but still harder than just applying `batch` tag to PRIVMSGs.
That's why ZNC's buffextras sends JOINs, QUITs, etc. as PRIVMSGs too, originating from *buffextras, and client-side script can render them as "proper" join instead.

Mikaela:
> ZNC already supports this, but under name znc.in/playback, even the parameter is the same.

Isn't that used by external module [playback](http://wiki.znc.in/Playback)?

DarthGandalf:
@Mikaela no, this batch type is used by ZNC core. Maybe that external module uses it too though.

SaberUK:
>Could this also handle joins/parts/quits/nicks?

I considered this but I decided against it for the reasons that @DarthGandalf has mentioned.

>ZNC already supports this, but under name znc.in/playback, even the parameter is the same.

That's a happy coincidence. Although I did write this specification to be open ended with regards to what could trigger chat history to be sent so that bouncers like ZNC could use it I didn't expect ZNC to have implemented an almost identical feature. :four_leaf_clover:

DarthGandalf:
Oh, there is one difference though: if server-time is not enabled, ZNC
sends timestamp in message itself. Renaming it to CHATHISTORY would break
the spec which says "MUST NOT".

2015-06-26 8:31 GMT+01:00 Peter Powell <notifications@github.com>:

grawity:
>> Could this also handle joins/parts/quits/nicks?
>
> I considered this but I decided against it for the reasons that @DarthGandalf has mentioned.

I'm aware of the reasons, _which is why I'm asking_ – the batch wrapper seems like it could be used as a channel-specific limiter to avoid that.

ariscop:
Not as written, since the intent is to send history regardless of client support. there would need to be some opt-in mechanism for history with quits/nicks

join/part might be handled correctly by if sent before RPL_NAMES, but representing a nick as a part/join doesn't sound acceptable

grawity:
Well yes, it'd have to be all four or nothing. (Cap event-playback?)

DanielOaks:
Add this almost as-is once above is resolved, and then write up an event-playback extension for 3.3?

kaniini:
:+1: for adding this batch type *as-is* and then optionally providing an event-playback extension in 3.3.

the batch type should not depend on some future extension and instead stand on it's own, so lets please concentrate on the semantics of the batch type as it presently stands.  whether event-playback happens should be decided on it's own merits.

DarthGandalf:
```
[26 июн, 19:52:11] <DarthGandalf> SaberUK: do you need server-time to be enabled for chathistory?
[26 июн, 19:53:38] <SaberUK> DarthGandalf: imo yes, because its not very useful for a client to just receive a bunch of historical messages without an attached time
[26 июн, 19:54:16] <DarthGandalf> SaberUK: I won't be able to rename znc.in/playback batch to chathistory... :(
[26 июн, 19:54:28] <SaberUK> D:
[26 июн, 19:55:11] <DarthGandalf> ZNC sends playback with or without client's server-time cap
[26 июн, 19:56:11] <DarthGandalf> And as playback is sent anyway, having it in a batch or not, depending on some other cap is stupid IMHO
[26 июн, 19:57:17] <DarthGandalf> Or, having type of batch different depending on some other cap, the same really
[26 июн, 19:57:45] <DarthGandalf> if (client supports server-time) { batch = "chathistory"; } else { batch = "znc.in/playback"; }
[26 июн, 19:58:47] <DarthGandalf> I'm fine with leaving server-time as recommendation, of course, because chathistory makes much more sense with server-time, but not as requirement
```

extensions/batch/chathistory.md
```diff
@@ -0,0 +1,41 @@
...
+
+When a client performs an action which requires chat history to be relayed to
+it (e.g. joining a channel with InspIRCd's chanhistory mode enabled) the server
+MUST put all PRIVMSGs and NOTICEs into a single CHATHISTORY batch. All
+messages MUST be [tagged][server-time] with the time at which they were
+originally sent.
+
+Clients which have not implemented support for the CHATHISTORY batch type
+SHOULD treat the PRIVMSGs and NOTICEs within the batch using their standard
+message handling method.
```
attilamolnar: This paragraph does not make sense, clients that don't implement CHATHISTORY don't obey this specification.

Note that the batch specification says that
> Client MAY ignore the type and process messages one by one. For example, that happens if the type is unknown to the client.

attilamolnar: One possibility is to remove this paragraph entirely, another one is to change it to be a reminder of the above quote.

SaberUK: Ah, I didn't realise that. 

If that is the case then we should probably remove the equivalent paragraph from the NETQUIT/NETJOIN specification.

SaberUK:
@DarthGandalf Does it look good now? I removed the hard requirement on `server-time` and replaced it with a soft requirement for it to be used if it has been enabled by the client.

extensions/batch/chathistory.md
```diff
@@ -0,0 +1,36 @@
...
+    period: "2015"
+    email: "petpow@saberuk.com"
+---
+## CHATHISTORY Batch Type
+
+This document describes the format of the CHATHISTORY batch type.
+
+The CHATHISTORY batch type takes one parameter. This parameter contains the
+target of messages within the batch. The target MUST either be the nick of the
+receiving client for private messages or the name of a channel which the
```
DarthGandalf: Why would server tell client its own nick? Nick of person who they communicated to would be more appropriate, I think.

SaberUK: Oops, yes. That's a typo. s/receiving/sending/.

SaberUK:
Does anything else need to happen to this?

attilamolnar:
LGTM.

attilamolnar:
@DarthGandalf Since you had some suggestions on this, can you please comment whether you think the current version is good to merge?

dequis:
:+1:

extensions/batch/chathistory.md
DarthGandalf: I still don't understand.
May you add an example of `CHATHISTORY` for a not channel?

DarthGandalf: `znc.in/playback` uses nick of the person who the receiving client was communicating with instead, so for 2 different nicks, it will be 2 different batches.

rburchell:
This sounds great, with some concerns:

* I can imagine that the publishing of a large amount of history from an arbitrary point in time would be a fairly effective way to create a large amount of traffic. Have you considered extending JOIN to somehow pass on the last seen point in the history, somehow?
* Only storing PRIVMSG (and I assume NOTICE) in my opinion limits the usefulness of the spec; I'd prefer to see something that at least allowed for the publication of more than that (even if it didn't require it from an implementer) -- e.g. JOIN/PART/QUIT/TOPIC + PRIVMSG/NOTICE, so, basically a full client-side view of any changes since a given point in time.

The usecase for providing more information has a lot of applications, but two in particular: it fixes a rather common problem in the case where someone is on a flaky connection and disconnects while waiting on a reply from someone else who subsequently leaves. Yes, sure, they'd see this in the nicklist, but in large channels in particular, visually scanning a nicklist isn't a trivial thing to do. So e.g:

    <foo> bar: hey, did you do the thing?
    * foo has quit (Connection reset by peer)
    * bar has quit (Quit: Avoiding foo)
    * foo has joined

.. foo now knows that bar is deliberately trying to not reply, rather than just being able to assume good faith. ;)

rburchell:
Reading up on some of the comments now, I see this was already brought up. I don't understand the concern re: outdated NAMES and such. The historical events are already inside an explicit extension batch that requires client support, so the client knows they aren't regular messages, and all it has to do is to format them - not process them. This is not difficult to do.

So, the client would receive the normal RPL_NAMEREPLY/RPL_TOPIC and so on, and then get a bunch of historical events which they would interleave into their buffer as required.

dequis:
>The historical events are already inside an explicit extension batch that requires client support

Not quite. This gets automatically enabled with only the BATCH capability. That's why a separate cap for event playback was suggested.

rburchell:
> Not quite. This gets automatically enabled with only the batch capability. That's why a separate cap for event playback was suggested.

Yes, but that batch has a type. I'd assume that clients aren't going to magically know how to handle a type they haven't seen before.

Hiding it behind a CAP does seem like a sensible suggestion, though.

dequis:
The part that talks about the target parameter looks completely unambiguous now, which addresses @DarthGandalf's concern. I think this is good to merge now.

DanielOaks:
+1 for merging this as-is, then looking at a separate event-playback cap if we want to afterwards, as discussed above

PR #160, opened 2015-07-12
Specify that CAP NEW can have values, similar to CAP LS.

DarthGandalf:
Current spec doesn't say either "yes" or "no".

Probably this change is too big though, and should go to 3.3?


kaniini:
I think it's fine as an errata clarification for 3.2.

kaniini:
:+1: go ahead and merge it imo.

DarthGandalf:
I need to come up with some better example though, this one doesn't specify type of the batch, and there's no good available type in existing specs.

PR #161, opened 2015-07-12
Make batch types lower case, as are capability names and tag names.

DarthGandalf:
Other examples of batch already use lower case with vendor-specific
types.


DarthGandalf:
If noone objects, I'll merge this tomorrow.

SaberUK:
@DarthGandalf I don't have an objection but is it documented that batch types are case insensitive?

DarthGandalf:
@SaberUK Indirectly, because DNS names are case insensitive.

PR #162, opened 2015-07-12
Add spec for labeled-replies.

### LINKED ISSUE #53: Labeled replies
Client attaches a cookie to the request and get that same cookie back in the response message(s).
Useful for ZNC, when multiple clients are connected, instead of hacky http://wiki.znc.in/Route_replies

See #51

DarthGandalf:
Fix #53

### DIFF
--- a/core/labeled-replies-3.3.md
+++ b/core/labeled-replies-3.3.md
@@ -71,7 +71,7 @@ If a response consists of more than one message, a `batch` MUST be used to group
 
 If no response is required, an empty batch MUST be sent.
 
-When a client sends a message to itself, the label tag MUST NOT be included, except for any `echo-message` acknowledgment.
+When a client sends a message to itself, the label tag MUST NOT be included, except for any `echo-message` acknowledgment. This is because the delivered message is not a response, but a side effect.
 
 #### Tag value
 
--- a/core/labeled-replies-3.3.md
+++ b/core/labeled-replies-3.3.md
@@ -50,7 +50,7 @@ Clients requesting this capability indicate that they are capable of handling th
 
 Servers advertising this capability indicate that they are capable of handling the message tag described below from clients, and will use with the same tag and value in their response.
 
-Servers MUST reject the capability at negotiation time if the `batch` capability is not also negotiated. Additionally, servers supporting `cap-notify` MUST cancel the capability if `batch` is removed at any time.
+If the capability is negotiated, the `batch` capability MUST also be treated as being implicitly negotiated. Servers supporting `cap-notify` MUST cancel the capability if `batch` support is explicitly removed at any time.
 
 ### Batch types
 
--- a/core/labeled-replies-3.3.md
+++ b/core/labeled-replies-3.3.md
@@ -58,7 +58,7 @@ This specification adds the `draft/labeled-response` batch type, described below
 
 ### Tags
 
-This specification adds the `draft/label` message tag.
+This specification adds the `draft/label` message tag, which has one required value.
 
 This tag MAY be sent by a client for any messages that need to be correlated with a response from the server.
 
@@ -67,6 +67,10 @@ For any message received from a client that includes this tag, the server MUST s
 If a response conists of more than one message, a `batch` of type `draft/labeled-response`, starting with the `draft/label` tag MUST be used to group them into a single logical response.
 
 If no response is required, an empty batch MUST be sent.
+
+#### Tag value
+
+The tag value is chosen by the client and MUST be treated as an opaque identifier. The client SHOULD NOT reuse a tag value until it has received a complete response for that value from the server.
 
 ## Server implementation considerations
 
--- a/core/labeled-replies-3.3.md
+++ b/core/labeled-replies-3.3.md
@@ -58,13 +58,13 @@ This specification adds the `draft/labeled-response` batch type, described below
 
 ### Tags
 
-This specification adds the `draft/label` message tag, which has one required value.
+This specification adds the `draft/label` message tag, which has a required value.
 
 This tag MAY be sent by a client for any messages that need to be correlated with a response from the server.
 
 For any message received from a client that includes this tag, the server MUST include the same tag and value in any response required from this message. Servers MUST include the tag in exactly one message.
 
-If a response conists of more than one message, a `batch` MUST be used to group them into a single logical response. The start of the batch MUST be tagged with the `draft/label` tag. The batch type MUST be one of:
+If a response consists of more than one message, a `batch` MUST be used to group them into a single logical response. The start of the batch MUST be tagged with the `draft/label` tag. The batch type MUST be one of:
 
 * An existing type applicable to the entire response
 * `draft/labeled-response`
--- a/core/labeled-replies-3.3.md
+++ b/core/labeled-replies-3.3.md
@@ -40,15 +40,15 @@ Additionally, labeled responses allow bouncers with multiple connected clients t
 
 ### Dependencies
 
-This specification uses the [message tags framework](/specs/core/message-tags-3.2.html) and uses the [`batch`](/specs/extensions/batch-3.2.html) capability which SHOULD be negotiated at the same time.
+This specification uses the [message tags framework](/specs/core/message-tags-3.2.html) and the [`batch`](/specs/extensions/batch-3.2.html) capability which SHOULD be negotiated at the same time.
 
 ### Capabilities
 
-This speciciation adds the `draft/labeled-response` capability.
+This specifiation adds the `draft/labeled-response` capability.
 
 Clients requesting this capability indicate that they are capable of handling the message tag described below from servers.
 
-Servers advertising this capability indicate that they are capable of handling the message tag described below from clients, and will use with the same tag and value in their response.
+Servers advertising this capability indicate that they are capable of handling the message tag described below from clients, and will use the same tag and value in their response.
 
 ### Batch types
 
@@ -71,7 +71,7 @@ If no response is required, an empty batch MUST be sent.
 
 If `batch` has not been enabled, the tag MAY be included on only one of the messages in the response.
 
-When a client sends a message to itself, the server MUST NOT include the label tag, except for any `echo-message` acknowledgment. This is because the delivered message is not a response, but a side effect.
+When a client sends a message to itself, the server MUST NOT include the label tag, except for any acknowledgment sent with the [`echo-message`](/specs/extensions/echo-message-3.2.html) mechanism. This is because the delivered message is not a response, but a side effect.
 
 #### Tag value
 
@@ -88,7 +88,7 @@ When a client sends a private message to its own nick, `echo-message` will resul
 1. Ignore the labeled message, and use any unlabeled message as acknowledgment for all sent messages to clear temporary local messages.
 2. Ignore the unlabeled message.
 
-Both methods assume that the server will acknowledge all successful messages, or return a labeled error response, but differ in their attitude to to the semantics of sending and receiving.
+Both methods assume that the server will acknowledge all successful messages, or return a labeled error response, but differ in their attitude towards the semantics of sending and receiving.
 
 ## Bouncer implementation considerations
 
@@ -96,13 +96,13 @@ This section is non-normative.
 
 Bouncers might choose to restrict routing labeled responses to some or all of their clients. For example, the response to a labeled WHOIS request might only be routed back to the originating client. However, an `echo-message` response might be routed to all clients. The bouncer might also choose to make labeled requests to the server on its own behalf, without routing any response to clients.
 
-To avoid label value clashes from multiple connected clients, a bouncer might choose to modify the client supplied label, for instance, by adding a client id prefix. In this case, the modified label would be returned to its original form before routing the response back to the client.
+To avoid label value clashes from multiple connected clients, a bouncer could choose to modify the client supplied label, for instance, by adding a client id prefix. In this case, the modified label would be converted back to its original form before routing the response back to the client.
 
 ## Examples
 
 This section is non-normative.
 
-1. `echo-message`
+1. `echo-message` showing a message that has been modified by the server to remove formatting
 
 
     ```
@@ -117,7 +117,7 @@ This section is non-normative.
     Server: @label=dc11f13f11 :irc.example.com 401 * nick :No such nick/channel
     ```
     
-3. `WHOIS`
+3. `WHOIS` response using a batch
 
     ```
     Client: @label=mGhe5V7RTV WHOIS nick
--- a/core/labeled-replies-3.3.md
+++ b/core/labeled-replies-3.3.md
@@ -75,12 +75,6 @@ If no response is required, an empty batch MUST be sent.
 
 The tag value is chosen by the client and MUST be treated as an opaque identifier. The client SHOULD NOT reuse a tag value until it has received a complete response for that value from the server.
 
-## Server implementation considerations
-
-This section is non-normative.
-
-TODO
-
 ## Client implementation considerations
 
 This section is non-normative.
--- a/core/labeled-replies-3.3.md
+++ b/core/labeled-replies-3.3.md
@@ -62,7 +62,7 @@ This specification adds the `draft/label` message tag, which has a required valu
 
 This tag MAY be sent by a client for any messages that need to be correlated with a response from the server.
 
-For any message received from a client that includes this tag, the server MUST include the same tag and value in any response required from this message. Servers MUST include the tag in exactly one message.
+For any message received from a client that includes this tag, the server MUST include the same tag and value in any response required from this message. Servers MUST include the tag in exactly one logical message.
 
 If a response consists of more than one message, a `batch` MUST be used to group them into a single logical response. The start of the batch MUST be tagged with the `draft/label` tag. The batch type MUST be one of:
 
@@ -70,6 +70,8 @@ If a response consists of more than one message, a `batch` MUST be used to group
 * `draft/labeled-response`
 
 If no response is required, an empty batch MUST be sent.
+
+When a client sends a message to itself, the label tag MUST NOT be included, except for any `echo-message` acknowledgment.
 
 #### Tag value
 
@@ -81,11 +83,10 @@ This section is non-normative.
 
 In the case of `echo-message` (see example below), a client can use labeled responses to correlate a server's acknowledgment of their own messages with a temporary message displayed locally. The temporary message can be displayed to the user immediately in a pending state to reduce perceived lag, and then removed once a labeled response from the server is received.
 
-When sending messages directed at a client's own nick, `echo-message` will result in duplicate messages being sent by the server, as both sent and received messages. Labeled responses allow clients to deduplicate these messages in one of two ways:
+When a client sends a private message to its own nick, `echo-message` will result in duplicate messages being sent by the server, as both sent and received messages. Labeled responses allow clients to deduplicate these messages in one of two ways:
 
-For private messages that match the clients nickname:
-1. Ignore labeled messages, and use any unlabeled message as acknowledgment for all sent messages to clear temporary local messages.
-2. Ignore unlabeled messages.
+1. Ignore the labeled message, and use any unlabeled message as acknowledgment for all sent messages to clear temporary local messages.
+2. Ignore the unlabeled message.
 
 Both methods assume that the server will acknowledge all successful messages, or return a labeled error response, but differ in their attitude to to the semantics of sending and receiving.
 
--- a/core/labeled-replies-3.3.md
+++ b/core/labeled-replies-3.3.md
@@ -34,6 +34,8 @@ Certain commands sent by a client can result in ambiguous responses from the ser
 
 Labeled responses allow a much simpler way to correlate using a single id attached to a client request and repeated by the server in its response.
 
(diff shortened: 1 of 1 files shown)


DarthGandalf:
Topic for bikeshedding: `label` vs `id` vs [your choice]

SaberUK:
:+1: for `id`.

dequis:
Did someone say bikeshedding? I looooove bikeshedding!

:+1: for id because it's shorter.

kythyria:
:+1: for `id` also. It's shorter and still conveys the right concept.

grawity:
"id" (or "tag") would be better, both because it's shorter, and because I haven't seen any other protocol uses the term "label" for that – it's almost always either a (transaction) ID or a tag. ("Cookie" is also used, but can be easily confused with session-state cookies.)

core/labeled-replies-3.3.md
```diff
@@ -0,0 +1,36 @@
...
+If client requests `labeled-replies`, it MUST also request `batch` capability.
+
+If the capability is enabled, client MAY tag its messages with `label` tag.
+
+For messages from client which are tagged with `label`, server MUST send exactly
+one message tagged with the same `label` value in response to such message.
+If response contains more than one message, `batch` can be used to group them.
+In that case start of the batch MUST be tagged with the label, and the batch type MUST be `labeled-response`.
+
+If the message doesn't require any response, an empty batch MUST be sent.
```
dequis: What does this mean? Is it something like, if `echo-message` isn't enabled, labeling privmsgs would result in something like this?

```
Client: @label=somelabel PRIVMSG #channel :Hello!
Server: @label=somelabel :irc.example.com BATCH +somebatch
Server: :irc.example.com BATCH -somebatch
```

core/labeled-replies-3.3.md
```diff
@@ -0,0 +1,36 @@
...
+title: IRCv3.3 Labeled replies
+layout: spec
+copyrights:
+  -
+    name: "Alexey Sokolov"
+    period: "2015"
+    email: "alexey-irc@asokolov.org"
+---
+Name of this capability is `labeled-replies`.
+If client requests `labeled-replies`, it MUST also request `batch` capability.
```
SaberUK: Might be a good idea to have the server NAK/DEL the cap if batch is not requested/goes away.

core/labeled-replies-3.3.md
```diff
@@ -0,0 +1,36 @@
...
+    Client: @label=pQraCjj82e PRIVMSG #channel :\x02Hello!\x02
+    Server: @label=pQraCjj82e :nick!user@host PRIVMSG #channel :Hello!
+
+2. `WHOIS`
+
+    Client: @label=mGhe5V7RTV WHOIS nick
+    Server: @label=mGhe5V7RTV :irc.example.com BATCH +NMzYSq45x labeled-response
+    Server: @batch=NMzYSq45x :irc.example.com 311 client nick ~ident host * :Name
+    ...
+    Server: @batch=NMzYSq45x :irc.example.com 318 client nick :End of /WHOIS list.
```
SaberUK: It might be a nice idea to have labelled replies deprecate the "end of xxx" numerics.

grawity: Labelled replies can't do that, since you'd expect the entire series to have _the same_ label after all.

Batched replies, on the other hand, could.

core/labeled-replies-3.3.md
DarthGandalf: Yes. Alternatively, I can add a new verb instead:
```
Server: @label=somelabel :irc.example.com EMPTY
```

core/labeled-replies-3.3.md
DarthGandalf: If both caps are enabled, server would need to send nested batches, to be able to provide both batch types...
```
Client: @label=mGhe5V7RTV WHOIS nick
Server: @label=mGhe5V7RTV :irc.example.com BATCH +1 labeled-response
Server: @batch=1 :irc.example.com BATCH +2 whois-without-numeric-end
Server: @batch=2 :irc.example.com 311 client nick ~ident host * :Name
Server: @batch=2 :irc.example.com 671 client nick :is using a secure connection
Server: @batch=1 :irc.example.com BATCH -2
Server: :irc.example.com BATCH -1
```

grawity: Well yes, but without BATCH, how would you know where a `@label=asdfg` series ends?

DarthGandalf: How to specify 2 batch types (labeled-response, and
whois-without-numeric-end) in a single batch?

2015-07-13 8:42 GMT+01:00 Mantas Mikulėnas <notifications@github.com>:

DarthGandalf: On the other hand...
When response contains single line, it's not required to be in a batch at
all, so the batch type isn't necessary.
So just whois-without-numeric-end should probably be enough, without need
for nested batch.

2015-07-13 8:42 GMT+01:00 Mantas Mikulėnas <notifications@github.com>:

grawity: Yes, exactly.

`labeled-response` batches seem completely pointless here (and even everywhere).

core/labeled-replies-3.3.md
kerio92: What's stopping a compliant ircd from answering with `EMPTY` or an empty `BATCH` to a labeled `PRIVMSG` and then sending the echoed `PRIVMSG` that comes from `echo-message`? Shouldn't the spec stress the fact that if a message does, in fact, require a response, then such response MUST be labeled?

DarthGandalf:
Note to self: mention that client SHOULD NOT reuse the same id until it got response for that id from server.

core/labeled-replies-3.3.md
```diff
@@ -0,0 +1,36 @@
...
+---
+Name of this capability is `labeled-replies`.
+If client requests `labeled-replies`, it MUST also request `batch` capability.
+
+If the capability is enabled, client MAY tag its messages with `label` tag.
+
+For messages from client which are tagged with `label`, server MUST send exactly
+one message tagged with the same `label` value in response to such message.
+If response contains more than one message, `batch` can be used to group them.
+In that case start of the batch MUST be tagged with the label, and the batch type MUST be `labeled-response`.
```
DanielOaks: As discussed below, maybe something like:

In that case start of the batch MUST be tagged with the label. The batch type MUST be an existing batch type if it is applicable to the entire response set, or `labeled-response` if one is not.

BlacklightShining:
@DarthGandalf What about messages that generate an error _and_ a (potentially modified) echo? I'm thinking of those IRCds that block configured words in messages, but relay the message to Ops in the channel:

```
Client: @id=4Upxeafq PRIVMSG #channel :A pox on both your houses!
Server: @id=4Upxeafq 000 #channel pox :is a banned word
Server: @id=4Upxeafq :nick!user@example.net PRIVMSG @#channel :A pox on both your houses!
```

BlacklightShining:
Did someone say replace the _end of foo_ numerics? Yes, please.

DarthGandalf:
> What about messages that generate an error and a (potentially modified) echo?

I'd expect them to be in a batch.

jwheare:
Worth keeping in mind how this might interact with server supplied ids #227. That proposal would make more sense as an @id tag, leaving this one as @label or @request-id or @reqid or something else.

jwheare:
Crossposting from https://github.com/ircv3/ircv3-specifications/pull/283#issuecomment-262759960

> The labelled-replies spec should probably instruct bouncers to forward the the label tag to all attached clients. We should consider how that interacts with e.g. znc's route_replies.

jwheare:
A followup point raised in https://github.com/ircv3/ircv3-specifications/pull/283#issuecomment-262763035

If bouncers are to send labelled replies to all clients, their clients should attempt to use non-overlapping labels and ignore all unrecognised labels.

And the general point from that thread is that clients wishing to not show duplicate (sent+received) messages in PMs to self need to adopt the following behaviour:
* Ignore all labelled replies if the source nick matches their own nick
* Clear all temporary local messages displayed

For all other PRIVMSGs, only the label-specified temp message needs to be cleared.

---

It occurs to me that bouncers that use `znc.in/self-message` probably don't need to send labelled replies to all clients.

jwheare:
I rewrote and started fleshing out this spec on a personal branch here:
https://github.com/ircv3/ircv3-specifications/compare/master...jwheare:labeled

I left in some outstanding TODOs for implementation considerations and example explanations.

@DarthGandalf let me know if you're happy for me to push that to this PR.

DarthGandalf:
Comments above have some ideas what else was missing in earlier version of the spec.
I added a few more comments to your commit.

jwheare:
I've pushed some additional changes to my branch to address your commit comments and other points mentioned in this thread.

"Server implementation considerations" still has a TODO in it. Any guidelines we could mention in there?

jwheare:
Some other points raised by @attilamolnar in IRC:

```
<Attila> any reason to use gibberish as labels?
<Attila> should limit to numbers in a given range
```

I don't think we need to restrict the labels beyond being opaque identifiers and the reuse guidelines I added a mention for.

```
<Attila> can you define what is a "response"
<Attila> e.g. what's the response to JOIN, only the JOIN message that comes back or the entire channel burst sequence
<Attila> what's the response to NICK+USER on reg
```

In the absence of a clear use case for those, I'm not sure it matters too much. An empty batch for all, or maybe a labelled JOIN response.

```
<Attila> might make more sense to define for which commands the behavior is required exactly so there's some scope
<Attila> i can think of 2 exact use cases, the first being echo-message and the second is who should znc route the whois to
<Attila> it'd be nice if you could explain why is a completely generic solution needed instead of solving these 2 specfic cases
<Attila> znc has the whois routing problem because whois replies may come out of order or may never arrive, this is not true for other commands such as WHO, MOTD, LUSERS, etc
<Attila> can be true for remote versions of the above commands but still it's only a few commands that allow that
```

route_replies currently has a list of 16 commands it handles and says "this list is far from complete"
https://github.com/znc/znc/blob/master/modules/route_replies.cpp

Should we be defining the behaviour for all of these as well as the JOIN/USER/NICK examples?

Is there a general rule we can enforce that will cover most cases well enough?

A few other non bouncer specific use cases I can think of:

* Tracking automated queries like ISON, WHO to prevent displaying the results to users. While MONITOR, WHOX exist, and channelburst may arrive in future, they're not evenly distributed.
* Tracking which message triggers e.g. unknown nick/channel errors, or other ambiguous errors that can result from multiple different actions.

attilamolnar:
> I don't think we need to restrict the labels beyond being opaque identifiers and the reuse guidelines I added a mention for.

I'd like to see them restricted for simplicity. If they are allowed to be opaque strings then corner cases such as length limiting and truncation can happen which can be avoided by restricting them to e.g. a 32-bit unsigned integer which is both reasonable and avoids servers having to add extra sanity checks to ensure it's not too long.

> route_replies currently has a list of 16 commands it handles and says "this list is far from complete"
https://github.com/znc/znc/blob/2e6c8d90e71c6ba59e41db1074cd863108a5bc89/modules/route_replies.cpp

> Should we be defining the behaviour for all of these as well as the JOIN/USER/NICK examples?

I haven't checked all of those commands but from what I've seen several of those e.g. ISON, WHO, LIST, NAMES, METADATA don't require a labeled-replies like solution to be routed correctly. A reply to ISON is RPL_ISON and the rest have and end numeric. As opposed to e.g. remote WHOIS a bouncer with multiple connected clients could send these commands and expect the replies to come back in the same order as the queries were sent.

jwheare:
PR updated with my changes.

jwheare:
I've implemented this along with echo-message in IRCCloud now. The server side was actually quite straightforward.

Each incoming message has a handler that returns a list of 0 or more output lines in an intermediate record format. If the incoming line has a label, the output is either tagged if it's a single line, or wrapped in a batch for more, or an empty batch is sent for 0.

Caps are then checked once before converting to the wire format, and the tags/batches might get optionally filtered out depending on what the client has negotiated. This saves having to spread cap checks throughout the various different parts that add other tags.

There was no need to add any custom command handling, it's all done at a single entry/exit point.

@attilamolnar can you share any details on how you implemented this on the InspIRCD testnet? We should condense our experiences into the server implementation considerations section.

jwheare:
I dropped the empty "Server implementation considerations" section as I can't think what to put in there that doesn't get lost in unnecessary detail/assumptions about various server architectures. Any objections to merging this as-is?

core/labeled-replies-3.3.md
```diff
@@ -0,0 +1,132 @@
...
+
+## Client implementation considerations
+
+This section is non-normative.
+
+In the case of `echo-message` (see example below), a client can use labeled responses to correlate a server's acknowledgment of their own messages with a temporary message displayed locally. The temporary message can be displayed to the user immediately in a pending state to reduce perceived lag, and then removed once a labeled response from the server is received.
+
+When sending messages directed at a client's own nick, `echo-message` will result in duplicate messages being sent by the server, as both sent and received messages. Labeled responses allow clients to deduplicate these messages in one of two ways:
+
+For private messages that match the clients nickname:
```
attilamolnar: Typo, should be "client's"

core/labeled-replies-3.3.md
```diff
@@ -0,0 +1,132 @@
...
+1. Ignore labeled messages, and use any unlabeled message as acknowledgment for all sent messages to clear temporary local messages.
```
attilamolnar: When this cap is used together with echo-message and the client messages itself then both messages (one from echo-message and the other one being the delivered message) will be labeled because both of those events were generated because of the labeled PRIVMSG the client sent to the server, i.e. both are "replies". This means "unlabeled messages" don't exist in these cases, or do I misunderstand something?

core/labeled-replies-3.3.md
```diff
@@ -0,0 +1,132 @@
...
+
+Servers MUST reject the capability at negotiation time if the `batch` capability is not also negotiated. Additionally, servers supporting `cap-notify` MUST cancel the capability if `batch` is removed at any time.
+
+### Batch types
+
+This specification adds the `draft/labeled-response` batch type, described below.
+
+### Tags
+
+This specification adds the `draft/label` message tag, which has one required value.
```
attilamolnar: Could clarify this, tags either have a value not have one but the wording here implies there may be multiple values for a tag

core/labeled-replies-3.3.md
jwheare: It's a plural, not a possessive.

core/labeled-replies-3.3.md
```diff
@@ -0,0 +1,132 @@
...
+
+### Capabilities
+
+This speciciation adds the `draft/labeled-response` capability.
+
+Clients requesting this capability indicate that they are capable of handling the message tag described below from servers.
+
+Servers advertising this capability indicate that they are capable of handling the message tag described below from clients, and will use with the same tag and value in their response.
+
+Servers MUST reject the capability at negotiation time if the `batch` capability is not also negotiated. Additionally, servers supporting `cap-notify` MUST cancel the capability if `batch` is removed at any time.
```
jwheare: So instead should servers treat `draft/labeled-response` negotiation as implicitly enabling `batch`?

core/labeled-replies-3.3.md
```diff
@@ -38,7 +38,7 @@ Labeled responses allow a much simpler way to correlate using a single id attach
 
 ### Dependencies
 
-This specification uses the [message tags framework](/specs/core/message-tags-3.2.html) and relies on support for the [`batch`](/specs/extensions/batch-3.2.html) capability.
+This specification uses the [message tags framework](/specs/core/message-tags-3.2.html) and relies on support for the [`batch`](/specs/extensions/batch-3.2.html) capability which MUST be negotiated at the same time.
```
MuffinMedic: The `draft/label` tag can be useful in the absence of `batch`. While batches are useful and can aid in keeping relevant content together, it should not be required for all uses of the `draft/labeled-response` message tag. There are instances where it may be necessary to uniquely identify a labeled response for content that does not belong to a batch.

See ircv3/ircv3-ideas#6

core/labeled-replies-3.3.md
```diff
@@ -40,7 +40,7 @@ Additionally, labeled responses allow bouncers with multiple connected clients t
 
 ### Dependencies
 
-This specification uses the [message tags framework](/specs/core/message-tags-3.2.html) and relies on support for the [`batch`](/specs/extensions/batch-3.2.html) capability which MUST be negotiated at the same time.
+This specification uses the [message tags framework](/specs/core/message-tags-3.2.html) and uses the [`batch`](/specs/extensions/batch-3.2.html) capability which SHOULD be negotiated at the same time.
```
MuffinMedic: This specification should be independent of the use of `batch` - if needed, they can be used in tandem and negotiated at the same time. "_... and, if needed, the batch capability should be negotiated at the same time_" 

The `draft/label` tag can be useful in the absence of `batch`. While batches are useful and can aid in keeping relevant content together, it should not be required for all uses of the `draft/labeled-response` message tag. There are instances where it may be necessary to uniquely identify a labeled response for content that does not belong to a batch, such as in ircv3/ircv3-ideas#6.

IRC Discussion
```
<prawnsalad> ideally, for any command created, it should be i send command X, and i expect a response for command X. error, ack, or response. so the client isnt sat there waiting for something that may or may not actually happen
<MuffinMedic> so, response is the batch, error would be thins like NO_CHATHISTORY_FOUND, ACCESS_DENIED, etc....?
<prawnsalad> "CHATHISTORY ERR_NOTFOUND/ERR_NOACCESS" would be my on-a-whim suggestion
<dx> space after ERR
<prawnsalad> "CHATHISTORY ERR <err name>" dx, yea this would be better
<MuffinMedic> that prevents a new issue though
<MuffinMedic> what if they request multiple chat histories? it would need to include a random ID parameter in the chathistory command that the server returns with the error
<dx> MuffinMedic: the labeled replies spec can take care of that
<MuffinMedic> dx: so send "@label=ID SCROLLBACK target.." then return "@label=ID SCROLLBACK ERR msg"?
<dx> yeah
```

jwheare: SHOULD means recommended not required.

jwheare: To clarify further, the spec doesn't say batches must be used every time, only when multiple lines (or none) are in the response and only if batch has been negotiated.

jwheare: Also, you're commenting on outdated changes :)

MuffinMedic: I tried to find the most recent relevant commit that addressed the issue - still finding my way around GitHub in a collaborative sense. I'll always comment on the latest in the future.

> SHOULD means recommended not required.

I'm being a bit nit-picky here and the language seems to suggest that it is always recommended, rather than recommended only in cases where it may be needed.

jwheare: No probs, I find it better to comment on the "files changed" view rather than a specific commit.

SHOULD means exactly that :) Strongly recommended where it makes sense. It's a stronger recommendation for server than clients really, but clients will have a better time if they enable it too.

MuffinMedic:
Looking over this spec again - what advantage / use cases does it have over #285  and #288 ? If we allow for the original sender to include a `draft/msgid` for any event, and allow replies to any event, an additional label tag seems not to add any new functionality - it only serves to add an additional tag that can be used as a unique identifier, which is already covered in the above specifications. 

Since there is no specified format for any of the unique identifiers, as long as it's something that's guaranteed to be unique (as best as possible), a single "message identifier" spec would be ideal. Client sends an event to the server with this event ID, server handles it as appropriate, and forwards it to other clients with the same ID. Likewise, a server-generated event would include an ID that is sent to all clients.

See [comment](https://github.com/ircv3/ircv3-specifications/pull/285/files#r94814237) on #285

jwheare:
This label is only sent back to the original sender, so it doesn't need any sever-wide uniqueness guarantee. These two specs serve very different purposes.

DarthGandalf:
Client can't be trusted to generate an unique ID.

2017-01-05 17:43 GMT+00:00 James Wheare <notifications@github.com>:

grawity:
@MuffinMedic I'd say the two specs work at different layers; they just happen to share the same tag space.

That is, `label` is a low-level feature for just about any _command_, with temporary client-specific IDs acting transaction markers, while `msgid` is high-level metadata for the _message itself_, with global IDs which can be stored or forwarded.

For example, if you sent a labelled request to replay some chat history, the server would be required to both return the current label *and* the original msgid when replaying an old message. If both were merged into one tag, that would become a bit more difficult. (That might not be a very good example, considering that we have batches and stuff, but hopefully you got the point anyway?)

attilamolnar:
> @attilamolnar can you share any details on how you implemented this on the InspIRCD testnet? We should condense our experiences into the server implementation considerations section.

The core logic is this:
Whenever a command that came from a user that has this tag is executed the module remembers which user is the one running the command. Whenever an IRC message is sent during the execution of the command the destination of the message is checked. If the destination is the user executing the command the message is considered to be a response.

There are some extra checks e.g. to ensure something (empty batch) is always sent and remote commands need extra handling.

The new behavior introduced for self PRIVMSGs recently steps on the toes of the above approach because both the echo message PRIVMSG and the other one is sent while the PRIVMSG command from the client is executed so both are considered replies.

attilamolnar:
To clarify, I'm ok with merging this with the current requirements for self PRIVMSGs and will tweak my implementation to support that.

PR #165, opened 2015-07-18
Message Data Framework

jwheare:
Data tags allow a client to attach custom structured key value data to a message that needn't be negotiated beforehand with a server.

Rendered view: https://github.com/jwheare/ircv3-specifications/blob/message-data/core/message-data-3.3.md

### DIFF
--- a/core/message-data-3.3.md
+++ b/core/message-data-3.3.md
@@ -61,7 +61,7 @@ first setting metadata on connect:
 
 And then send tagged messages to the channel:
     
-    @data-favicon=example.com/icon.png :Bot!~Bot_user@example.com PRIVMSG #channel :Example title
+    @data-favicon=https://example.com/icon.png :Bot!~Bot_user@example.com PRIVMSG #channel :Example title
 
 METADATA GET on `Bot` would return
 
--- a/core/message-data-3.3.md
+++ b/core/message-data-3.3.md
@@ -32,7 +32,7 @@ it distributes it to clients supporting the `message-data` capability.
 
 Vendor prefixed tags must also be supported, in the form:
     
-    `<vendor>/data-<key>`
+    <vendor>/data-<key>
 
 ## Defining a data schema
 
--- a/core/message-data-3.3.md
+++ b/core/message-data-3.3.md
@@ -22,13 +22,17 @@ tags.
 
 ## The `data-` prefixed message tag
 
-To send a data tagged message with the key `foo` and the value `bar`, we do the
+To send a data tagged message with the key `foo` and the value `bar`, we send the
 following:
 
     @data-foo=bar PRIVMSG #ircv3 :hello #channel
 
 The IRC server would then attach all `data-` prefixed tags to the message when
 it distributes it to clients supporting the `message-data` capability.
+
+Vendor prefixed tags must also be supported, in the form:
+    
+    `<vendor>/data-<key>`
 
 ## Defining a data schema
 
--- a/core/message-data-3.3.md
+++ b/core/message-data-3.3.md
@@ -1,0 +1,74 @@
+---
+title: IRCv3.3 Message Data framework
+layout: spec
+copyrights:
+  -
+    name: "James Wheare"
+    period: "2015"
+    email: "james@wheare.org"
+---
+This specification concerns adding an optional out-of-band data tag to messages
+using the [message tagging](../core/message-tags-3.2.html) framework, and a
+mechanism for defining such tags using the [metadata](../core/metadata-3.2.html)
+framework.
+
+Data tags allow a client to attach custom structured key value data to a message
+that needn't be negotiated beforehand with a server.
+
+## The `message-data` capability
+
+A client MUST request the `message-data` capability in order to receive data
+tags.
+
+## The `data-` prefixed message tag
+
+To send a data tagged message with the key `foo` and the value `bar`, we do the
+following:
+
+    @data-foo=bar PRIVMSG #ircv3 :hello #channel
+
+The IRC server would then attach all `data-` prefixed tags to the message when
+it distributes it to clients supporting the `message-data` capability.
+
+## Defining a data schema
+
+In order to provide a path for standardisation of data tags, a client SHOULD set
+metadata defining its use of data tags, using the `message-data` metadata key
+and a set of space-separated values.
+
+A client supporting only one set of data tags:
+
+    METADATA * SET message-data :example.com/message-data
+
+A client suppoting more than one set of data tags:
+
+    METADATA * SET message-data :example.com/message-data example.com/extra-data
+
+Specifying the tagset in this way allows clients to interpret them according to
+specifications provided out of band.
+
+### Non-normative
+
+The following examples describe how an implementation might use certain
+features. Unlike previous examples, they are in no way intended to guide
+implementations' behaviour.
+
+A client with nick `Bot` that provides title fetching services for links sent in
+a channel can send additional data including the favicon of the fetched site by
+first setting metadata on connect:
+    
+    METADATA * SET message-data :example.com/favicon-data-tag
+
+And then send tagged messages to the channel:
+    
+    @data-favicon=example.com/icon.png :Bot!~Bot_user@example.com PRIVMSG #channel :Example title
+
+METADATA GET on `Bot` would return
+
+    METADATA Bot GET message-data
+    :irc.example.com 761 Bot message-data * :example.com/favicon-data-tag
+    :irc.example.com 762 :end of metadata
+
+Metadata notifications for `Bot` would include:
+
+    :irc.example.com METADATA Bot message-data * :example.com/favicon-data-tag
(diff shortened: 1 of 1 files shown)


core/message-data-3.3.md
```diff
@@ -0,0 +1,74 @@
...
+
+A client with nick `Bot` that provides title fetching services for links sent in
+a channel can send additional data including the favicon of the fetched site by
+first setting metadata on connect:
+    
+    METADATA * SET message-data :example.com/favicon-data-tag
+
+And then send tagged messages to the channel:
+    
+    @data-favicon=example.com/icon.png :Bot!~Bot_user@example.com PRIVMSG #channel :Example title
```
SaberUK: This URI should probably be fully qualified.

jwheare: Ah, it's a vendor prefix, not a URI, e.g. znc.in/batch etc.

grawity: Really? `favicon=example.com/icon.png` looks very much like a resource location to me...

jwheare: Oh sorry, was thinking you were referring to e.g. example.com/favicon-data-tag. My bad, a keenly observed nitpick, will fix.

kaniini:
Please add a section pointing out that `data-` should be considered insecure and other related security considerations of the namespace.

jwheare:
@kaniini can you provide an example of the sort of security risk you mean?

kaniini:
I suppose the only security considerations would be with an implementation that doesn't namespace the data, which would be noncompliant with this spec.  So nevermind on that.

jwheare:
Is it worth explicitly mentioning that there may in future be an IRCv3 registry of data tags that wouldn't need a vendor prefix when defining the tagset in METADATA? Or would such standardised tags use the ircv3.net/ vendor prefix?

Aerdan:
Assuming that standardised data tags don't require server influence, having them use the ircv3.net vendor prefix would seem appropriate; if they do, for whatever reason, they should be activated through a capability and therefore don't need to be in `METADATA`.

jwheare:
Yeah, makes sense. Also, we can cross that bridge when we get to it.

DarthGandalf:
```
[19 июл, 23:49:32] <Aerdan> well, the intention is that clients will be able to apply tags to messages that will be received by other clients (which is not possible otherwise). client A can send data-foo to client B, but how is client B going to figure out how client A intended for it to be used?
[19 июл, 23:50:13] <Aerdan> with the message-data key in metadata, client B can look at client A's metadata and know that client A is claiming compliance with example.com/foo.
[19 июл, 23:50:58] <DarthGandalf> So a single client can't support both example.com/foo and example.org/foo at the same time?
[19 июл, 23:51:25] <Aerdan> not as currently written, no.
[19 июл, 23:55:01] <prawnsalad> so client A receiving a data- tag from client B, would then have to look up client Bs meta data in order to know what its trying to do?
[19 июл, 23:55:52] <DarthGandalf> Do you really expect any client to look that up? I'd expect most clients to just hardcode example.com/foo and ignore any future presence of example.org/foo
[19 июл, 23:56:45] <Aerdan> Might I suggest commenting on #165? It's not my spec, so I really can't speak authoritatively on it.
```

prawnsalad:
I can't really see clients receiving a data- tag then looking it up via METADATA unless badgered to do so. It is also rather longwinded (and the main reason why I probably wouldn't do so). Why not just recommend that clients experimenting with data- tags name them with suited prefixes? ie. `znc/foo`, `kiwi/foo`. As data- tags become more used then discussions can take place to standardise them if any are obviously overlapping.

M2Ys4U:
I'm also leaning against this proposal, as currently stated.

There's another minor issue, around race conditions.

What happens if there's another message that comes in between receiving the tagged message and receiving the metadata?

For example:

```
-> @data-foo=bar :baz PRIVMSG #channel :quux
<- METADATA baz GET message-data
-> :norf PRIVMSG #channel :Hello, world
-> :irc.example.com 761 baz message-data * :example.com/foo-data-tag
-> :irc.example.com 762 :end of metadata
```

How should the client treat the ordering of the two PRIVMSGs? Should it display the first *before* it gets the metadata, display them in the order in which it does know the proper metadata (e.g. second first), or should it insert the first message *above* the second, but only once it has got the result of the metadata query?

Also, how would this combine with, say, the CHATHISTORY batch type (#156) where the client sending the message may no longer be on the IRC network and therefore we are unable to query for its metadata?

jwheare:
From the metadata spec:

> Clients are also subscribed to notifications for channels they join.

And

> Client subscriptions to a channel includes notifications for other users in the channel, regardless of MONITOR state.

So you will already be notified of metadata when the client joins your channel. No need for a query.

Aerdan:
@jwheare is correct, although then the spec hard-depends on `metadata-notify`. If you want it to, making use of a metadata key is fine. If not, you're still going to have the issues raised above.

jwheare:
I wouldn't say it hard-depends on it, some specific clients could easily function without it, but it does improve it significantly. Enough to warrant a mention though. Will try to add that later today.

jwheare:
@prawnsalad:
> Why not just recommend that clients experimenting with data- tags name them with suited prefixes? ie. znc/foo, kiwi/foo.

According to the current tag spec, tags of the form `data-kiwi/foo` are not allowed, the vendor prefix needs to be a valid DNS name. For this spec to work, a common prefix needs to be used so that servers know to pass it through to clients.

That said, vendor prefixed tags of the form `kiwiirc.com/data-foo` should probably be explicitly allowed with this spec, to allow a client that wants to use two clashing tags from different vendors. But being able to use the non-prefixed form should be allowed, to address concerns that have been raised elsewhere about tags using up too many bytes (e.g. by TwitchIRC).

And on a related note, `data-kiwiirc.com/foo` should probably be explicitly disallowed, the `data-` prefix should only apply to the non-vendored portion of the tag.

jwheare:
@M2Ys4U @prawnsalad @DarthGandalf does this resolve your concerns?

SaberUK:
LGTM.

kaniini:
I think if we merge #170 then we should change this so that data tags are not a capability.

gwillen:
So I chatted with @jwheare some on IRC about the metadata stuff, and I have some thoughts:

I like the way 'xmlns' in XML works, where I can say e.g. xmlns:android="http://schemas.android.com/apk/res/android", and this defines tags prefixed with "android:" but not other tags.

In contrast, I don't like the way metadata in this system works, because when I say "http://schemas.android.com/apk/res/android" as a metadata tag, other clients have no way of knowing what tags might or might not be defined by that metadata tag unless they are already familiar with it.

I think this interacts really badly with metadata tags being a "should" rather than a "must", because if Bob is sending some metadata tag I've never seen before, then I can no longer assume anything about any of the tags he sends -- some of which might be described by the unfamiliar metadata, and some not -- unless the system REQUIRES him to send metadata for every tag (so that I can know for sure which of his tags are the same as mine.)

There are also questions in my mind about overlapping metadata tags -- e.g. if "http://example.com/A" defines data-foo and data-bar, and "http://example.com/B" defines data-bar and data-baz, are we stuck? It should be prohibited, under the current system, to give both metadata tags if they give conflicting definitions of data-bar, but that leaves us no way to reconcile them, without defining a whole new "http://example.com/C" -- but existing clients have no way to know what C is, even if they know about both A and B.

So I think it would make sense -- like how xmlns works -- for metadata tags to be specified for particular tags / sets of tags / prefixes for tags.

Otherwise I think the meaning of tags becomes a hazy and ill-defined thing, and we're going to end up with something like the HTTP user-agent field, where everyone has their own semantics for it, and people just keep nailing on new bits because they can't control how people interpret the old ones.

(@jwheare thinks I worry too much. I think @jwheare doesn't worry enough. :-)

nwf:
I think I am confused a little (or perhaps just dense!) since the proposal seems to permit people to use non-vendor-prefixed tags in ad-hoc, non-standardized ways.  I would have thought, perhaps naively, that all non-vendor-prefixed tags would have had exactly one meaning as set by the ircv3 working group.  Or perhaps the point is that data-* tags are never to be standardized by ircv3 and used exclusively ad-hoc?

That said, for vendor tags at least, where the ad-hoc metadata mechanism seems to make some amount of sense, what if we made the `METADATA SET` arguments explicitly relate keys to schemas?  Something like

    METADATA * SET message-data :example.com/favicon=http://example.com/favicon-data-tag

(We can leave off the `data-` prefix from the tag since it's implied.) For flexibility and byte minimization, we could allow "*" to denote prefixes of tags, something like

    METADATA * SET message-data :example.com/ext1-*=example.com/data-tags-ext1

jwheare:
An update on this issue. I need to find the time to update the PR and it requires a little more thought, but I'm leaning away from the metadata element here, and more towards requiring formally specified tag meanings, which would establish IRCv3 as a home for client-to-client specs too.

I'm also not opposed to dropping the data- prefix requirement if all tags are echoed client to client by the server by default as suggested in #187

nwf:
I agree that the METADATA system may be a little much for a first cut.

Doesn't #187 as written obviate this entire generic effort, making things like the standardization and definition of `subj` in #202 all that's left to do in a world where all tags are relayed to clients?

(Though #187 does seem to imply that there is no room for message tags that are intended to be client-to-server communication.  Perhaps that's right?)

jwheare:
Closing this in favour of #236. Client-to-client tags should be specified individually.

PR #166, opened 2015-07-19
Add a warning about name collision in capabilities and tags.

SaberUK:
Apparently the existing warnings were not strong enough. Added a bigger one to attempt to fix that.

### DIFF
--- a/core/capability-negotiation-3.1.md
+++ b/core/capability-negotiation-3.1.md
@@ -183,6 +183,10 @@ Names for which a corresponding document sits in the IRCv3 Extension Registry.
 
 Names in the IRCv3 Extension Registry are reserved for your capability.
 
+The IRCv3 Working Group reserves the right to reuse names which have not been submitted to the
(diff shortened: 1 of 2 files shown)


core/capability-negotiation-3.1.md
```diff
@@ -183,6 +183,9 @@ Names for which a corresponding document sits in the IRCv3 Extension Registry.
 
 Names in the IRCv3 Extension Registry are reserved for your capability.
 
+The IRCv3 Working Group reserves the right to reuse names which have not been submitted to the
+registry. If you do not wish to do this then you must use a vendor-specific name (see above).
```
DarthGandalf: to do this? to do what? to reuse names?

SaberUK: Good point. I'll clarify it.

kaniini:
:+1: on the latest version.

Aerdan:
:+1: here, too.

SaberUK:
@DarthGandalf Is it OK now?

PR #168, opened 2015-07-28
Add cap-value batch type

### LINKED ISSUE #167: cap-notify-3.2 specifies a vendor-prefixed batch called "example.com/cap-value"
@attilamolnar 

>03:23 < dx> what's up with that "BATCH +cap example.com/cap-value" in http://ircv3.net/specs/extensions/cap-notify-3.2.html
>03:28 < Aerdan> sending both CAP commands as a single transaction.
>03:29 < dx> i mean the example.com/ vendor prefix specifically. seems like it's defining a new batch type and simultaneously making it unusable
>03:30 < Aerdan> ask attila.
>03:30 < dx> hm. i guess the only way to ask attila is through the issue tracker?
>03:31 < dx> he's alive there
>03:32 < DarthGandalf> That can be made a proper batch type, if useful
>03:32 < DarthGandalf> without prefix

DarthGandalf:
Fix #167

### DIFF
--- a/extensions/batch/cap-value-3.3.md
+++ b/extensions/batch/cap-value-3.3.md
@@ -1,0 +1,30 @@
+---
+title: IRCv3 cap-value Batch Type
+layout: spec
+copyrights:
+  -
+    name: "Alexey Sokolov"
+    period: "2015"
+    email: "alexey-irc@asokolov.org"
+---
+## `cap-value` Batch Type
+
+Capability negotiation 3.2 introduced values for capabilities.
+When server needs to change value after negotiation, it can do that using CAP
+DEL followed by CAP NEW, but to make it a single transaction, server SHOULD put
+them in a batch `cap-value`.
+
+Even though it is a single transaction, client MUST use `CAP REQ` again if it
+wants the capability back.
+
+Example:
+
+        Client: CAP LS 302
+        Server: :irc.example.com CAP * LS :sasl=PLAIN batch
+        Client: CAP REQ batch
+        ...
+        Server: :irc.example.com BATCH +cap cap-value
+        Server: @batch=cap :irc.example.com CAP client DEL :sasl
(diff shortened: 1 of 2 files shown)


kaniini:
:+1:

jwheare:
Might be worth discussing whether CAP NEW could be upgraded to work as a CAP UPDATE to avoid the need for this entirely?

jwheare:
Given this disables the CAP on the client either way, what problem does this solve?

DarthGandalf:
This was so long ago...
AFAIR, this is to prevent client from forgetting about certain caps in case
of a power failure.
E.g. if server changes duration of `sts`, it disables DELs it, then NEWs it
with another value. Client might assume that deleting the cap is a signal
that SSL shouldn't to be used anymore. Usually client will receive CAP NEW
after that, and will know again to use SSL, but not in case of
power/network failure.

2016-12-21 17:17 GMT+00:00 James Wheare <notifications@github.com>:

jwheare:
OK. Seems extremely edge case-ish to me. If I understand correctly, in case of a power failure, the connection will close and sts will be renegotiated on the next connection attempt, so is this a real issue? I may be missing something.

DarthGandalf:
Yes, except that the next connection attempt will be without SSL.
Yes, it's a very edge case.

Maybe some other caps can be affected by something similar, not only `sts`.

2016-12-28 10:20 GMT+00:00 James Wheare <notifications@github.com>:

attilamolnar:
> E.g. if server changes duration of `sts`, it disables DELs it, then NEWs it with another value. Client might assume that deleting the cap is a signal that SSL shouldn't to be used anymore. Usually client will receive CAP NEW after that, and will know again to use SSL, but not in case of power/network failure.

That's a wrong assumption for clients to make. sts is disabled by setting the duration to 0, not via CAP DEL. CAP DEL for sts should have no effect on the established policy.

jwheare:
In the absence of a compelling use case, we should probably close this and reopen if needed.

PR #170, opened 2015-08-22
IRCv3.3 Client Capability Negotiation

jpnurmi:


grawity:
…wait, what kind of 'special handling' does echo-message require?

jpnurmi:
> Clients may choose to disable local echoing of sent PRIVMSG and NOTICE messages altogether, or present them in pending state.

grawity:
"may" doesn't _require_ any of that, does it.

jpnurmi:
...or else you'll get double messages.

grawity:
I see, so the new requirement doesn't imply echoing, only requires accepting self-messages.

It occurs to me that this is **exactly what `self-message` was about**, but no, _let's replace it with something entirely unrelated which happens to also do similar stuff._

dequis:
Well, in 1bfba47843c2526707c902034b3395af934713c8 we concluded that every client *should* support it without a CAP.  You could say this PR formalizes that cap-less requirement. Still seems weird to add cap-less requirements to a CAP spec.

Mikaela:
Where else would the requirement for self-message without cap be done? It cannot be in echo-message, because that way clients wouldn't just implement it anyway.

Because of https://github.com/ircv3/ircv3.github.io/issues/17 it looks like to me that CAP is the only thing clients have to implement.

Of course there is issue that what if clients don't implement that part of CAP 3.3, but that is issue with the clients and not the servers that assume client supports it as it requests CAP 3.3.

kaniini:
To clarify this proposal simply makes it no longer a requirement to explicitly opt-in for either `echo-message` behavior or unsolicited tags, yes?

If so I'm not in favor of dropping the opt-in requirement for `echo-message`, but the unsolicited tags seems OK.  I was never in favor of requiring an explicit cap to get a message tag.  Clients are already expected to deal with unexpected server output *anyway*.

kaniini:
Regarding `echo-message`, what we could do is make it so that `CAP LS 303` explicitly enables `echo-message` and makes it sticky so it cannot be removed.  That way it becomes a mandatory cap.

attilamolnar:
There is no need to bump the negotiation protocol version for these changes. It should only be bumped when a backwards incompatible change is made to the negotiation protocol itself, like caps with values or multi line replies.

dequis:
@attilamolnar I don't think the purpose of the version after CAP LS is *that* set on stone. I think you should keep this open unless others agree that this is definitely not the way to do it.

kaniini:
I agree, this, or at least the idea of it, should be reopened.  In my opinion, some of the behaviors of 3.2 should be corrected based on what we've learned implementing them, and 3.3 should be released as a fix for them.

Requiring caps for tags is silly, as clients can just ignore ones they don't care about.

attilamolnar:
Quote from the doc defining the version number:

> The LS subcommand now has an additional argument which is the raw version number of the latest capability negotiation protocol supported by the client.

I do agree with some of the changes, but the changes are unrelated to the negotiation protocol.

kaniini:
Okay, I will open a new issue with an alternate proposal to accomplish the changes.

attilamolnar:
I'll gladly reopen this request if the original author wants to rework it. I closed it because it seemed abandoned and it was going in the wrong direction.

Issue #80, opened 2014-03-23
MONITOR accepting hostmasks


TheSaunders:
There was a discussion on IRC about not allowing MONITOR to accept hostmasks due to the amount of time you may have to spend doing lookups on connect/disconnect/away/account/etc if the server has many different hostmasks being monitored.

How hostmasks differ:
If you are matching nicks then when a user connects you could instantly lookup that nick. There would be no confusion about which MONITOR that nick belongs to.

With hostmasks, since you can do `*!*@example.com`, or `*!ident@example.com`, or even `*!*@*.com`, a connecting user may match multiple different MONITORs and this would mean you would have to loop for every MONITOR to check which one(s) it matches.

This also means that if using away-notify and account-notify with MONITOR integration, you would need to loop for every MONITOR and do hostmask checks when away/account status changes.

Edit :: Added how hostmasks differ from nick lookups

---

I opened this to track the status of this discussion and will update the issue accordingly as new information comes in.

PR #174, opened 2015-09-21
RPL_MONONLINE should include user@host

ariscop:
issue #80 change the wording of target to dissalow hostmasks.
This was to avoid ircd's having to do wildcard matching on every
client registration, but broke the definition of RPL_MONONLINE.

Current implementations send a hostmask with RPL_MONONLINE, the
spec should match that to avoid incompatibilities

### DIFF
--- a/core/monitor-3.2.md
+++ b/core/monitor-3.2.md
@@ -147,4 +147,7 @@ full, so the command failed.  The `<limit>` parameter is the maximum number of
 targets a client may have in their list, the `<targets>` parameter is the
 list of targets, as the client sent them, that cannot be added.
 
+## Errata
 
+* Earlier version of this spec had RPL_MONONLINE only sending nicknames,
+  This did not match existing implementations.
--- a/core/monitor-3.2.md
+++ b/core/monitor-3.2.md
@@ -107,7 +107,7 @@ This numeric is used to indicate to a client that either a target has just
 become online, or that a target they have added to their monitor list is
 online.
 
-The server SHOULD send a hostmask with the target.
+The server MAY send a hostmask with the target.
 
 The server may send "*" instead of the target nick (`<nick>`). (This makes it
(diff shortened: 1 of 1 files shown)


core/monitor-3.2.md
```diff
@@ -101,7 +101,7 @@ online or offline.  All targets that are online will be sent using
 
 ### 730 - `RPL_MONONLINE`
 
-    :<server> 730 <nick> :target[,target2]*
+    :<server> 730 <nick> :target!user@host[,target!user@host]*
```
Aerdan: :-1: This does not permit nick-only responses, which also occur in extant code.

ariscop: fixed by 7dc994d

core/monitor-3.2.md
```diff
@@ -101,12 +101,14 @@ online or offline.  All targets that are online will be sent using
...
 ### 730 - `RPL_MONONLINE`
 
-    :<server> 730 <nick> :target[,target2]*
+    :<server> 730 <nick> :target[!user@host][,target[!user@host]]*
 
 This numeric is used to indicate to a client that either a target has just
 become online, or that a target they have added to their monitor list is
 online.
 
+The server SHOULD send a hostmask with the target.
```
dequis: MAY? Should be less of a change of the existing recommendation.

dequis:
Add an errata section explaining this change? Something along the lines of "earlier version of this spec did not allow hostmasks in the target, this was changed to match existing implementations"

core/monitor-3.2.md
ariscop: Since existing implementations send the hostmask, i thought it best to use stronger language than 'MAY'

attilamolnar: :+1: for MAY.

PR #177, opened 2015-11-11
sasl-3.1: Wrap <mechanism> tag in inline-code backticks

DanielOaks:
Just a quick patch. Fixes the website HTML generation of the SASL 3.1 spec in preparation for https://github.com/ircv3/ircv3.github.io/pull/38


attilamolnar:
Thanks!

PR #184, opened 2015-12-03
STS: Add requirements for long-lived connections

attilamolnar:


attilamolnar:
Discuss:
* interval durations
* wording: suggest intervals or force them?
* minimum interval to avoid clients spamming CAP LS every minute?

grawity:
- Polling itself seems inelegant. At least for policy/config changes cap-notify ought to handle that.

 - If a web server reduces the HSTS expiry, do web browsers adjust accordingly or do they stick with what they've first seen?

 - I wonder if making the expiry time be since last _disconnection_ would make sense.

Mikaela:
> Polling itself seems inelegant. At least for policy/config changes cap-notify ought to handle that.

:+1:

> If a web server reduces the HSTS expiry, do web browsers adjust accordingly or do they stick with what they've first seen?

Yes, if I understand [RFC 6797 HSTS section 5.3](https://tools.ietf.org/html/rfc6797#section-5.3) correctly.

> I wonder if making the expiry time be since last disconnection would make sense.

You mean that the duration starts only counting down when the client disconnects and stays the same as long as client is connected? I cannot think any issues related to that immediately.

attilamolnar:
There are two versions how this could work:

First one is in the PR. The second is that the server pushes policy changes when they occur and the client periodically re-sets the policy itself every X minutes (to work around power failures) and on disconnection. I think this is trickier to implement than what's in the PR.

> Polling itself seems inelegant. At least for policy/config changes cap-notify ought to handle that.

There is no problem with polling in this case. It's done to offload work to the client (by not requiring servers to keep timers) and it's also done very infrequently (once a day).

DarthGandalf:
> The second is that the server pushes policy changes when they occur and the client periodically re-sets the policy itself every X minutes (to work around power failures) and on disconnection. I think this is trickier to implement than what's in the PR.

I think it's easier to implement by both client and server than what's in the PR...

attilamolnar:
> > The second is that the server pushes policy changes when they occur and the client periodically re-sets the policy itself every X minutes (to work around power failures) and on disconnection. I think this is trickier to implement than what's in the PR.

> I think it's easier to implement by both client and server than what's in the PR...

Let's examine both cases.

Poll:

* Changes in client:
  * Add timer to send `CAP LS`, same behavior when a reply arrives containing `sts` as when doing it initially on a secure connection.

* Changes in server:
  * Nothing.

`CAP DEL/NEW` and auto-reset:
* Changes in client:
  * Add timer to re-schedule the expiration, the new expiration is calculated from the current policy and the current time
  * Add handling of CAP NEW for `sts`:  (previously they had nothing to do when `sts` was in CAP DEL or NEW)

* Changes in server:
  * Send CAP DEL/NEW when the `sts` duration changes

attilamolnar:
Either way is fine to me, I just wanted to make sure we are thinking about the same thing first.

DarthGandalf:
> * Changes in client:
>  * Add timer to send CAP LS, same behavior when a reply arrives containing sts as when doing it initially on a secure connection.

Need to modify behavior of what happens upon `CAP LS` coming from server. Now many clients assume that it's initial cap negotiation, and send `CAP REQ` in reply to `CAP LS`, then waiting for `CAP ACK/NAK`.

DarthGandalf:
> Add handling of CAP NEW for sts (previously they had nothing to do when sts was in CAP DEL or NEW)

I don't think there should be any difference in handling any particular cap whether it's coming from `CAP LS` or `CAP NEW`. In this sense `sts` is not any different from any other cap.

attilamolnar:
Updated:
 * Rescheduling is handled by the client
 * Servers should notify compatible clients of policy changes (it is in their interest)
 * Suggested durations are not considered final, suggestions welcome

attilamolnar:
Periodical rescheduling changed from MUST to SHOULD, say that the durations listed are recommendations.

attilamolnar:
This has been idle for a while now, now is the last chance for comments before merge.

jwheare:
I'll try to take a look this week.

PR #185, opened 2015-12-03
Add SNI specification

attilamolnar:


DarthGandalf:
Is this IRC-specific?

2015-12-03 10:38 GMT+00:00 Attila Molnar <notifications@github.com>:

grawity:
> Is this IRC-specific?

Sure, why not? It's normal for an application protocol to require specific features of, and interoperability with, lower-layer protocols.

For example, [RFC 7540: HTTP/2](https://tools.ietf.org/html/rfc7540) also requires it [section 9.2](https://tools.ietf.org/html/rfc7540#section-9.2) (and generally describes in detail which TLS features are required or forbidden), although the previous [RFC 2818: HTTP Over TLS](https://tools.ietf.org/html/rfc2818) didn't mention anything.

Mikaela:
In case this PR is just sitting here and wanting opinions, I say :+1: for merge.

jwheare:
Yup, looks fine to me.

grawity:
hmm, actually, maybe it should be clarified that "clients MUST use SNI when connecting over TLS", rather than "must use TLS when connecting securely"

jwheare:
Are you suggesting it's OK to use SSL?

grawity:
No, and I don't care about your SSL vs TLS pedantry.

But If you feel that IRCv3 should define "connecting securely" to mean using STLSTLSLTSTLL, doing that as a footnote in this spec seems somewhat _inside-out_ to me.

(For example, we could have a "TLS profile for secure IRC connections" spec, describing TLS usage, the 6697 port, requiring SNI, perhaps disallowing export ciphers as HTTP/2 did, etc.)

But as long as it's titled "Server Name Indication", it seems that it should limit itself to requiring SNI when TLS is used, not requiring TLS itself.

jwheare:
Fair point, perhaps a broader spec on the use of TLS called "Secure Connections" is in order along the lines of that HTTP2 spec you linked.

MrBenC:
Two points/comments/questions regarding this proposal:

1. Why wouldn't a wildcard cert be used in lieu of two or more certs provided for server.* and irc.* (and potentially, however many other round-robin hostnames, such as ipv6.* or ipv4.*) ?

2. The other possible use case would be running multiple IRCds on the same IP:port combination, but how often does that really happen or would be desired?

Regarding point one, the argument could be made for SSL cert cost, but many smaller networks have a tendency to use self-signed certificates anyway for precisely this reason.

I'm not convinced of the need or usefulness for this.

Ben (MrC @ Freenode)

grawity:
Plenty of reasons.

 - Wildcard certificates are expensive. Host certificates are free. (And even self-signed certs can make use of SNI.)
 - Wildcard certificates only cover exactly one subdomain level. You need two if you want `*.example.com` and `*.eu.example.com`. Wallet complaining yet?
  - Wildcard certificates allow any server to impersonate _any_ other site under that domain – not only IRC servers, but also forums, DB servers, etc.
  - Wildcard certificates do not cover names outside the domain entirely, e.g. `irc.acc.umu.se` for GIMPnet.

Client-side, enabling SNI is often a one-line matter.

PR #191, opened 2015-12-16
metadata-3.2: Fix METADATA event to allow notification of removed keys

DanielOaks:
This looks like it was just a typo/oversight, as I'm fairly sure being able to notify clients about keys that have been removed is desired. [mammon](https://github.com/mammon-ircd/mammon) actually already implements the `METADATA` event as shown here, sending out an event without the `<Value>` param on removed keys.

Pointed out by @sim642


attilamolnar:
LGTM, please squash the commits.

DanielOaks:
Done

attilamolnar:
Thanks!

PR #195, opened 2015-12-23
There are cases where the client may not follow the STS policy

ProgVal:
According to [RFC 1459](https://tools.ietf.org/html/rfc1459#section-1.2):
>    A client is anything connecting to a server that is not another
>   server.  Each client is distinguished from other clients by a unique
>   nickname having a maximum length of nine (9) characters.

This looks quite old-fashioned, so I'll assume instead a client is a set of code and configuration files, both of which can evolve over time. (I truly started asking myself philosophical questions about the essence of what a “client” is)

But I think raises an interesting point about the feasibility of the STS spec as it currently is.
Indeed, the spec assumes the client can always remember past STS messages, and makes it a requirement. I see different cases where it can not be done:

1. The client has no writeable persistant storage, or no space left on it,
2. The client's storage has been altered (purposively or not),
3. The client is not able to recover the STS policy from its storage, for some reason (eg. an upgrade or a downgrade).
4. The client's system time may changed without the client knowing (eg. while it is stopped)

This pull request changes the verb from MUST to SHOULD to allow the client to not respect this policy such cases.

I also rephrased “a secure connection cannot be established with the server” to “establishing a secure connection with the server failed”, because a client not being able to remember could be interpreted as falling under the former formulation.

### DIFF
--- a/core/sts-3.3.md
+++ b/core/sts-3.3.md
@@ -46,10 +46,11 @@ connection, it MUST first establish a secure connection and confirm that the
 `sts` capability is still present.
 
 Once the client has confirmed that the `sts` capability is indeed offered over
-a secure connection, it then MUST only attempt secure connections to the
+a secure connection, it then SHOULD only attempt secure connections to the
 server from now on until the policy expires (see the `duration` key).
-It MUST refuse to connect if a secure connection cannot be established with the
-server for any reason during the lifetime of the policy.
+It MUST refuse to connect if establishing a secure connection with the server
+failed for any reason during the lifetime of the policy.
+
 
 However, if the client fails to connect securely for any reason, the connection
 attempt SHOULD be considered a failure, similar to a network error.


DanielOaks:
I disagree with this change. I think that all this would do is water down
the spec for the vast majority of clients that will be able to implement it
successfully. MUST tells clients that they MUST do this -- SHOULD is more
of a suggestion, and may end up with clients assuming that they may do
different things as their default behaviour.

I don't think there are enough clients who are going to run into those sort
of issues (ie, that run without persistent storage) and implement this spec
for us to cover it in this spec. In my opinion, the behaviour of those
clients shouldn't be covered by us.

I think the 'no storage space issue' is going to affect the client in more
ways than just the STS policy not getting stored, such as general
configuration and data files not being able to be stored. It's a rare
failure case from the start, so I don't think it's worth mentioning.

The "cannot recover" and "altered" cases are failure and attack cases on
the internal client configuration/data files. I don't think it's worth us
exploring or worrying about anything like that in this specification as
it's not designed to cover those cases.

attilamolnar:
NAK. Clients lacking a writable persistent storage can always connect securely by default to be compliant.

ProgVal:
On 23/12/2015 08:38, Daniel Oaks wrote:
> I disagree with this change. I think that all this would do is water down
> the spec for the vast majority of clients that will be able to implement it
> successfully. MUST tells clients that they MUST do this -- SHOULD is more
> of a suggestion, and may end up with clients assuming that they may do
> different things as their default behaviour.

I meant “SHOULD” with the definition in RFC 2119:

> 3. SHOULD   This word, or the adjective "RECOMMENDED", mean that there
>    may exist valid reasons in particular circumstances to ignore a
>    particular item, but the full implications must be understood and
>    carefully weighed before choosing a different course.

jwheare:
:-1: if your client cannot reliably support STS then you can choose not to implement it. If it's a temporary thing then that's an issue for your client, changing a MUST to a SHOULD for the benefit of protecting the feelings of faulty client authors is a bad practice imo.

PR #199, opened 2015-12-29
server-time: mention echo-message & NTP

Mikaela:
```
2015-12-29 20:43:24+0200 < AttilaM> Mikaela: go ahead, write it
2015-12-29 20:43:28+0200 < AttilaM> then we can decide where it fits
2015-12-29 20:47:07+0200 < Mikaela> sure, I know that it won't be accepted at first and requires heavy fixixng, but at least then you must give constructive comments
```

### DIFF
--- a/extensions/server-time-3.2.md
+++ b/extensions/server-time-3.2.md
@@ -48,3 +48,9 @@ John joined #chan during the leap second of June 2012 (notice the :60)
 
 [cap]: ../core/capability-negotiation-3.1.html
 [message tag]: ../core/message-tags-3.2.html
+
+* * * * *
+
+In case local client is in wrong time, timestamps looking weird can be
+avoided by also supporting the `echo-message` capability. Running NTP
+client is recommended.


SaberUK:
I don't see the point of this. If your clock is wrong then you should expect invalid timestamps.

:-1:

dequis:
I think these issues should be documented *somewhere*, and maybe the server-time spec is not the right place for it, but we can't ignore the issue completely and keep it as something that only those involved in the discussions / bug reports are aware of.

ProgVal:
What if the client could ask the server to give timestamps relative to when messages are sent?

grawity:
Would be simplest to make the server advertise its current time, as a general feature similar to what many s2s protocols have.

This would also be useful for adjusting timestamps in MODE replies, in WHOIS, in ban lists, and so on...Optionally claim that this may also be useful for STS etc.

Clients concerned about incompetent server admins can then determine the offset between server time and local time.

Either redefine the TIME numeric reply to include a machine-readable timestamp parameter which it unfortunately lacks, or define an ISUPPORT token TIME= with the server's current timestamp, or define an (advertise-only) cap time= with the same.

Anything else is useless. Reminds me of "Caution: Hot coffee may be hot" kind of warnings.

DanielOaks:
`CAP dan- LS :server-time=2011-10-19T16:40:51.620Z` ?

grawity:
@DanielOaks: That was also a version I considered (as well as requiring 001 to be time-tagged), but again this might be useful even if the server doesn't support `@time` tagging, so...

DanielOaks:
Hmm, well in that case it will still present the current time to the user on connection, even if they don't explicitly enable it. Interested clients should be able to work it out based on when the `CAP LS` comes from the server anyways, so shouldn't be any need to explicitly time-tag `001`?

We already time-tag `001` in mammon, but I'm not sure how requiring something like that would break other bits of software. If we really need to put this as another numeric or something (like `RPL_TIME` but with a standard, parsable format like above), I think that may work best as another spec or as a 3.3 thing.

grawity:
> Hmm, well in that case it will still present the current time to the user on connection, even if they don't explicitly enable it.

Still implicitly requires tag support on the server though, while a dedicated `time=` cap would be usable even by older ircds.

Mikaela:
The reason I am opening this PR reads in the original comment and in case someone missed it at IRC, I want it to be documented somewhere, so client authors/developers don't go following issues like https://github.com/znc/znc/issues/1144 and remove server-time (even if this only wanted to reduce server-time).

I am open for other ideas too, but I am afraid my skills are limited to this PR unless I get a lot of help.

I think the spec is the best place (when language is fixed), because all developers are going to read it and it's not so big part of the spec.

jwheare:
For reference, here's how we manage clock offsets in IRCCloud's JSON stream api.

On connect to the API stream, a `header` message is sent which includes the server's current time as seconds since the epoch. We use this to calculate the offset between server (bouncer) and client (browser)

`TimeOffset = new Date().getTime() - (header_time*1000)`

Then when it comes to displaying a message to the user (also timestamped by the server) we apply the offset like so:

`timestamp + TimeOffset`

This stops users from complaining that their message times don't match their clock.

One side effect to note is that message timestamps can vary between web client sessions, but only one particularly vocal user has complained about this ever.

I'd be in favour of a note saying client authors MAY implement timestamp offsetting to match the user clock.

jwheare:
Closed in favour of #290

PR #200, opened 2016-01-02
cap 3.1: Use GitHub syntax for referring to #134

Mikaela:
This isn't so important, but it came to mind with weechat/weechat#641.

If the errata is quoted directly to issue/PR of software that is located
at GitHub, this change makes GitHub make it a link. See above mentioned
WeeChat issue for example.


SaberUK:
:-1:

If people want to quote it then thats their problem. The link works fine as-is on the website.

DanielOaks:
I'd agree with @SaberUK, I don't think this is necessary and would mostly just clutter the site a bit more

DarthGandalf:
If people want to quote it to redmine, we also should provide copypasteable text in that format?

-1

PR #201, opened 2016-01-05
metadata-3.2: Clarify METADATA SET

attilamolnar:
- Clarify that only ERR_KEYINVALID is sent in reply when the key name is invalid.
- Clarify that the RPL_KEYVALUE and RPL_METADATAEND should only be sent for successful requests (either those numerics or one error numeric must be the reply).


DanielOaks:
Looks fine to me. +1

PR #202, opened 2016-01-05
Add ctctags specification

nwf:
This spec introduces a new subspace of message-tags which are intended for client-to-client communication and are just relayed by servers.  It then defines one such tag to add subject labels to messages, a la Zephyr and Zulip.

### DIFF
--- a/extensions/ctctags-3.3.md
+++ b/extensions/ctctags-3.3.md
@@ -1,0 +1,52 @@
+---
+title: IRCv3.3 `ctctags` Extension
+layout: spec
+copyrights:
+  -
+    name: "Nathaniel Filardo"
+    period: "2015"
+    email: "nwfilardo@gmail.com"
+---
+
+This document defines a new sub-class of IRC message-tags, so-called
+"client-to-client" message-tags.  Syntactically, these are tag names whose
+key
+
+* is composed entirely of capital ASCII letters, numbers, and/or hyphens;
+  and
+* contains at least one capital letter
+
+The `ctctags` capability causes the server to forward any such
+client-to-client message-tags sent by other clients on `PRIVMSG` commands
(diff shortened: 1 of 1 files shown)


kaniini:
I would prefer a formal namespace for these tags ala the `data-` proposal instead of this.

nwf:
I was worried about colliding with existing tags and all the existing ones seemed lower case.  Would the use of a sigil character that's currently not in the spec, like $ or @ or something, be acceptable for namespacing client-to-client tags?  (i.e. `@subj` or `example.com/@subject`)

kaniini:
I really do not like the idea of using a sigil in that way.

gwillen:
From some chattering on IRC just now (sorry @nwf I hope I didn't misrepresent you there) -- it seems like the use of data- tags without the metadata portion is exactly client-to-client tags. So I'm not sure how I feel about the metadata part yet, but the data- part alone seems like a good fit for subject-tagging?

extensions/ctctags-3.3.md
```diff
@@ -0,0 +1,52 @@
...
+    email: "nwfilardo@gmail.com"
+---
+
+This document defines a new sub-class of IRC message-tags, so-called
+"client-to-client" message-tags.  Syntactically, these are tag names whose
+key
+
+* is composed entirely of capital ASCII letters, numbers, and/or hyphens;
+  and
+* contains at least one capital letter
```
SaberUK: Under IRC insensitive case mapping rules `TAG` is equivalent to `tag`.

jwheare:
Vote to close this issue in favour #165. Would be better to have a unified effort to address all concerns.

nwf:
Agreed; let's figure out what to do about #165.

PR #203, opened 2016-01-07
sasl-3.2: Weaken requirement on reauthentication from MUST to MAY.

### LINKED ISSUE #192: SASL reauthentication causes problems with implementations where accounts are linked to server access
The sasl-3.2 specification states:

>Servers MUST allow a client, authenticated or otherwise, to reauthenticate by sending a new AUTHENTICATE message at any time.

>Servers MAY disconnect ANY client at any time as a result of failed authentication, including both unregistered and registered clients, but MUST provide the reason for the authentication failure prior to disconnection.

This works for most uses but causes problems for implementations where your account defines your access to the server and it is undesirable to allow people to log out (e.g. in a corporate environment).

My proposed solution is to change the verb in the first section above from "MUST" to "SHOULD" and to restore the requirement to send ERR_SASLALREADY if reauthentication is not available.

cc: @rburchell

rburchell:
There are circumstances in which reauthentication are undesirable, and it is not
the job of a specification to dictate software functionality, only the means
through which it achieves that functionality.

Fixes #192.

### DIFF
--- a/extensions/sasl-3.2.md
+++ b/extensions/sasl-3.2.md
@@ -56,8 +56,10 @@ requested the `cap-notify` notification.
 
 ## SASL Reauthentication
 
-Servers MUST allow a client, authenticated or otherwise, to reauthenticate by
-sending a new `AUTHENTICATE` message at any time.
+Servers MAY allow a client, authenticated or otherwise, to reauthenticate by
+sending a new `AUTHENTICATE` message at any time. If reauthentication is not
+possible, an error appropriate to the circumstances MUST be sent (for instance,
+`ERR_SASLALREADY` is appropriate if changing accounts is not supported).
 
 Servers MAY disconnect ANY client at any time as a result of failed authentication,
 including both unregistered and registered clients, but MUST provide the reason
@@ -119,3 +121,13 @@ Example where the server disconnects a client for excessive reauthentications:
     Client: AUTHENTICATE ...
     Server: :irc.server.tld 904 modernclient :SASL authentication failed
     Server: ERROR :Too many SASL authentication failures
+
+## Errata
+
+A previous version of this specification had reauthentication listed as a MUST
+requirement. This was weakened to a MAY, to allow for server implementations to
+be compliant in circumstances where reauthentication is not a desired feature.
+
+See
+[issue #192](https://github.com/ircv3/ircv3-specifications/issues/192) for more
+information.


attilamolnar:
Please add an errata section to the spec, as seen in metadata for example listing what was changed

rburchell:
Ha, I was on it before I even saw your comment :)

attilamolnar:
LGTM

extensions/sasl-3.2.md
```diff
@@ -56,8 +56,10 @@ requested the `cap-notify` notification.
 
 ## SASL Reauthentication
 
-Servers MUST allow a client, authenticated or otherwise, to reauthenticate by
-sending a new `AUTHENTICATE` message at any time.
+Servers MAY allow a client, authenticated or otherwise, to reauthenticate by
+sending a new `AUTHENTICATE` message at any time. If reauthentication is not
+possible, an error appropriate to the circumstances MUST be sent (for instance,
+ERR_SASLALREADY is appropriate if changing accounts is not supported).
```
jwheare: Could be backticked: `ERR_SASLALREADY`

rburchell: Done

jwheare:
:+1:

Mikaela:
I think it should be changed to SHOULD as may makes it seem less important for servers to implement REAUTH.

attilamolnar:
I don't think we need to particularly encourage either behavior, so MAY seems fine:

    5. MAY   This word, or the adjective "OPTIONAL", mean that an item is
       truly optional.  One vendor may choose to include the item because a
       particular marketplace requires it or because the vendor feels that
       it enhances the product while another vendor may omit the same item.

Mikaela:
If it's not encouraged, certfp and /msg nickserv identify are going to be more popular ways to handle this if IRCds don't implement it.

Mikaela:
No matter which it's changed to, how is it tracked on ircv3.net who supports SASL REAUTH and who doesn't? Currently it's required, so SASL 3.2 support means REAUTH support. But what if it's changed?

Will IRCds join http://ircv3.net/docs/sasl-mechs.html ? Or would SASL REAUTH be stripped into it's own spec?

rburchell:
> No matter which it's changed to, how is it tracked on ircv3.net who supports SASL REAUTH and who doesn't?

I'm not sure if there's any need to track things down to that particular level, as it could well even be a configuration (rather than software) choice.

Mikaela:
> I'm not sure if there's any need to track things down to that particular level, as it could well even be a configuration (rather than software) choice.

And if I was going to setup IRC server, how would I know if it's configuration choice if I have no idea the IRCd supports it?

rburchell:
> And if I was going to setup IRC server, how would I know if it's configuration choice if I have no idea the IRCd supports it?

Presumably you'd read the configuration documentation for the software you were setting up long before you'd read documentation about the standards that that software implemented.

Mikaela:
> Presumably you'd read the configuration documentation for the software you were setting up long before you'd read documentation about the standards that that software implemented.

Have you ever seen any of these pages or thought that they could help with software choices?

* http://ircv3.net/software/clients.html
* http://ircv3.net/software/servers.html
* http://ircv3.net/software/libraries.html
* http://ircv3.net/docs/sasl-mechs.html

I think you are just derailing from the issue.

Mikaela:
I went to read [the RFC](https://www.ietf.org/rfc/rfc2119.txt) and there is:

```
3. SHOULD   This word, or the adjective "RECOMMENDED", mean that there
   may exist valid reasons in particular circumstances to ignore a
   particular item, but the full implications must be understood and
   carefully weighed before choosing a different course.
```

In my opinion it makes it clear that it's recommended that servers implement SASL REAUTH, but the minority cases where people aren't going to change roles are allowed to not implement it.

rburchell:
What you are talking about here is a policy question, not a technical question. I think you misunderstand the point of a technical specification document. A technical specification is written by technical people, for technical people, describing how different pieces of technology are supposed to interoperate together. It's not to mandate technical solutions ("you SHOULD implement reauthentication") to social problems ("I think that every IRC network SHOULD allow me to logout and in again").

Reauthentication, as common sense (and its absence from the original 3.1 spec) should dictate is an _optional_ feature. I hope we can agree on that. While I can agree with you that for certain network types, sure, it makes sense to recommend that they _use software that supports features users want_, like reauthentication. But it does not make sense to say that all software _should_ implement optional features that make no sense for their particular usecases.

Mikaela:
```
2016-01-07 20:26:24+0200 < w00t> I think I described fairly well why I chose what I chose in https://github.com/ircv3/ircv3-specifications/pull/203#issuecomment-169757134
```

I think you missed my answer

```
2016-01-07 20:01:41+0200 < Mikaela> can someone tell me have I accidentally said "network" in place of "IRCd" somewhere, as I don't know how else I should understand this comment. If I mean IRCd, I mean something we hopefully have in the future, if I say network, I mean a network which we currently have e.g. freenode and I am not expecting freenode to do anything modern in near future
```

And you possibly also missed the fact that I am requesting a vote as this is otherwise going nowhere as you were already told:

```
2016-01-07 20:20:25+0200 < w00t> I wrote the change, I don't want to change it any further, and you already signalled your acceptance; so why are we still discussing it? :)
2016-01-07 20:20:55+0200 < jwheare> because this is a consensus based organisation
2016-01-07 20:21:01+0200 < jwheare> and someone signalled an objection
2016-01-07 20:21:35+0200 < jwheare> and i like it when everyone can be happy
```

(assuming you are w00t, I have no idea)

+1 on SHOULD

rburchell:
I don't care anymore.

PR #205, opened 2016-01-22
cap-3.1: clarify that caps can be re-requested

DanielOaks:
If this is the desired behaviour from servers, I think it's is fairly non-obvious from reading the specification. This tries to make it more clear by specifying it directly in the `CAP REQ` section of the spec, for when people are implementing that subcommand.


TingPing:
:+1:

Mikaela:
:+1: this seems a lot more clear to me than getting NAKs on capabilities that should be supported.

attilamolnar:
LGTM

Aerdan:
:+1:

PR #206, opened 2016-01-28
Add compatibility with old INVITE notice

ChickenNuggers:
@jwheare Any comments?

### DIFF
--- a/extensions/invite-notify-3.2.md
+++ b/extensions/invite-notify-3.2.md
@@ -26,11 +26,13 @@ clients to learn when another client does an /INVITE.
 
 The format of the INVITE message is as follows:
 
-    :<inviter> INVITE <target> <channel>
+    :<inviter> INVITE <target> <channel> [status]
 
 The message source, `inviter` is the user doing the INVITE, target
 is the user being invited, `channel` is the channel where `target` is
-being invited.
+being invited. `status` is an optional status character, which MUST be
+present in the `STATUSMSG` token in the server's `RPL_ISUPPORT` (005)
+values if used.
 
 ## Example
 
@@ -38,10 +40,20 @@ being invited.
 
 This message translates to ChanServ has invited Attila to #channel.
 
+    :ChanServ!ChanServ@example.com INVITE Alissa #channel @
+
+This message translates to ChanServ has invited Alissa to #channel,
+but only shows the message to operators with status of @ or higher.
+
 ## Notes
 
 The server is not required to send the INVITE message described in
 this document to all clients supporting this capability on a channel.
 For example, a server may choose to only send the message to clients
 having channel op, or only to clients that have the privilege to do
-an /INVITE on the channel themselves.
+an /INVITE on the channel themselves. In this case, the server MAY 
+use an appropriate `STATUSMSG` status on the channel argument.
+
+## Errata
+
+Added a mention of the `STATUSMSG` argument.


SaberUK:
This is overly specific IMO. Just allow the prefix to be specified on the channel parameter with no details about how it is used.

Its also missing an errata.

extensions/invite-notify-3.2.md
```diff
@@ -34,7 +34,7 @@ being invited.
 
 ## Example
 
-    :ChanServ!ChanServ@example.com INVITE Attila #channel
+    :ChanServ!ChanServ@example.com INVITE Attila +#channel
```
jwheare: I would use the much more common @ in the example.

extensions/invite-notify-3.2.md
```diff
@@ -34,7 +34,7 @@ being invited.
```
jwheare: Add new sentence to this para:

```
`prefix` is an optional status prefix, which MUST be present in the `STATUSMSG` token in the
server's `RPL_ISUPPORT` (005) values.
```

extensions/invite-notify-3.2.md
```diff
@@ -44,4 +44,6 @@ The server is not required to send the INVITE message described in
 this document to all clients supporting this capability on a channel.
 For example, a server may choose to only send the message to clients
 having channel op, or only to clients that have the privilege to do
-an /INVITE on the channel themselves.
+an /INVITE on the channel themselves. The server should, if the
+message is sent based off a user's status in the channel, include the
+minimal prefix of who is allowed to view the message.
```
jwheare: Change this sentence to:

```
In this case, the server should use an appropriate `STATUSMSG` prefix on the
channel argument.
```

jwheare:
Info on the widely implemented `STATUSMSG` token for `RPL_ISUPPORT` is here: http://www.irc.org/tech_docs/005.html

extensions/invite-notify-3.2.md
jwheare:
```
## Errata

Added a mention of the `STATUSMSG` prefix.
```

dequis:
I probably missed some discussion elsewhere, but why is this needed? What problem does it address? What "old INVITE notice"?

jwheare:
This spec is specifically designed to replace the NOTICE sent to other channel members (usually just chanops) when someone is invited to the channel. That NOTICE used the STATUSMSG prefix to indicate that it was only being sent to a certain subset of users. The new spec currently makes no mention of it.

extensions/invite-notify-3.2.md
Zarthus: I personally think it should also have an example of without prefix, otherwise the text beneath reads oddly I feel.

jwheare:
See:
https://github.com/unrealircd/unrealircd/blob/unreal40/src/modules/m_invite.c#L285
https://github.com/inspircd/inspircd/blob/7343ac33fbc64fd6d4426da4203530c3aef95e29/src/coremods/core_channel/cmd_invite.cpp#L125

ChickenNuggers:
> I would use the much more common @ in the example.

Done. 

> I personally think it should also have an example of without prefix, otherwise the text beneath reads oddly I feel.

Done.

> Add new sentence to this para:
```
`prefix` is an optional status prefix, which MUST be present in the `STATUSMSG` token in the
server's `RPL_ISUPPORT` (005) values if used.
```

Done.

> Change this sentence to:
```
In this case, the server MAY use an appropriate `STATUSMSG` prefix on the
channel argument.
```

Done.

> Add Errata
```
## Errata
Added a mention of the `STATUSMSG` prefix.
```

Done.

jwheare:
:+1: LGTM, please squash.

attilamolnar:
NAK. This is a backwards incompatible change to a released spec and it is not justified. Parsing the NOTICE was never a standard method so there is nothing to be "compatible with" (except clients and servers who already implemented this spec as is)

jwheare:
STATUSMSG and before that WALLCHOPS enabled NOTICEs were standard according to the isupport spec, which is a valid IRC spec imo.

Without this prefix invite-notify loses the previous ability to know who the message is being targeted at, which for some channel moderators may be useful information.

There is one existing implementation of invite-notify in a client that this might affect, Colloquy. I asked @zadr to comment in IRC.

https://github.com/colloquy/colloquy/blob/main/Chat%20Core/MVIRCChatConnection.m#L4050

ChickenNuggers:
> There is one existing implementation of invite-notify in a client that this might affect

What about IRCCloud?

jwheare:
That's me :)

SaberUK:
I don't think it would break compatibility if instead of prepending it to the channel if it were to be appended as an optional parameter.

ChickenNuggers:
Guess that's handled then. I'll go slap it into Moon Moon real quick in case it gets implemented. @SaberUK that could also work.

ChickenNuggers:
@jwheare What's your opinion on @SaberUK's solution?

jwheare:
Sure, I'm not wedded to it being a prefix. It should probably be renamed to `status` instead of `prefix` anyway.

Adam-:
AFAIK the reason the old notice behavior does this is to prevent breaking the protocol (as in some people in a channel receive a channel notice but other people do not), and it solves this by abusing statusmsg. I also don't see why it is useful to know who the server intends to receive the message prefix-wise, and it is an odd behavior that nothing else really does.

Adam-:
Additionally, I think that who the server will end up sending the message to and not to should be implementation dependent and will likely involve more complex things than simply channel prefix.

ChickenNuggers:
So, some thoughts from IRC via @attilamolnar and @jwheare:

We should move this to the 3.3 release; I am in favor of having it added as `invite-notify` and having two different versions; one 3.3+ and one 3.2.  @jwheare discussed having a `statusmsg` capability.

The perks to having `invite-notify`:

 * Doesn't add an extra capability
 * Should be simple to implement

The perks of having `statusmsg`:

 * Tag-based format (i.e. `@status=+`)
 * Should also be easy to implement
 * Discusses as well that unknown tags should be ignored

jwheare:
I discussed this with @zadr in IRC and consensus that using standard STATUSMG-style prefix is the way to. There was a slight preference for moving it to a 3.3 upgrade, but the compatibility risk was acknowledged as minor and no strong feelings either way.

One issue with adding this as a version bump in 3.3 is how to version the CAP. There are 3 options I can think of, all with issues:

* A new cap name, e.g. `invite-notify-v2`. This has CAP proliferation implications, and thought should be given to this naming scheme to make sure it's applicable for other CAPs that might need this in future.
* A cap value, e.g. `invite-notify=v2`. Same issues as new name, plus removes the ability to use the value for something else in future.
* Version bump with `CAP LS 303`. This makes it hard to support just the invite-notify change without whatever else might be defined in 3.3. There should be no need to bump every CAP just for sake of one, LS bumps should be restricted to LS format changes, e.g. the multiline change in 302.

Perhaps we need to figure out how we're going to do versioning as a separate topic.

My votes on how to resolve this issue.

* Extra argument format: -1
* STATUSMSG-style prefix: +1
* @statusmsg CAP: -1
* Added as errata to 3.2: +1
* Added as new version in 3.3 without resolving versioning concerns: -1
* Added as new version in 3.3 after waiting for a formal policy on versioning: -0

attilamolnar:
This should be resolved by adding a new tag sent to clients requesting message-tags (#236) and invite-notify.

jwheare:
Closing due to lack of consensus/importance.

Issue #213, opened 2016-02-08
[meta] feature requests/suggestions should include functional mocks


kaniini:
I have noticed a lot of change requests / feature requests are based solely in theory, which ultimately are not feasible to include for whatever reason.

This request is a procedural request to require specifications to include mocks so we (as well as observers/vendors/etc) can see the change in action, which should help with both improved specs as well as improved discussion.

Thanks.

ChickenNuggers:
This could be implemented via a `CONTRIBUTING.md` file as users are actually notified before creating an issue.

kaniini:
Sure, I dont know what language to use for that, and would rather defer to the people presently running the show.  I just think a mock requirement is helpful to demonstrate the process that went into making the spec.

SaberUK:
:+1:

SoniEx2:
:-1: Disallows IRC **users** to contribute to IRCv3. I think we should cater to users, not developers.

kaniini:
@SoniEx2 with all due respect, that is the point.

ChickenNuggers:
> Disallows IRC users to contribute to IRCv3. I think we should cater to users, not developers.

... No. The point is that users should not be able to request things at random.

SoniEx2:
@ChickenNuggers They should be able to request what they want/need. If for some reason developers decide not to satisfy their request, the issue can be safely closed and optionally locked. That's how any properly-functioning issue/suggestion tracker should work.

SaberUK:
I'm not sure I can imagine a use case where a user would be capable of suggesting a feature but not be capable of implementing it.

ChickenNuggers:
If they want to request a feature, they can at least go on IRC and discuss it first before creating an issue.

SoniEx2:
@ChickenNuggers I'd love to do that but I'm banned.

ChickenNuggers:
@SoniEx2 And who's fault is that?

SaberUK:
Lets not bring drama into issues.

kaniini:
The IRCv3 specifications process is *not a free software project*.  While it has some characteristics that are similar, what we produce here are formal specifications meant for developers to use when implementing software.  Thusly, in my opinion, it is inappropriate for non-developers to participate here.

[...]

PR #215, opened 2016-02-09
[meta] Contributing document and process

DanielOaks:
This document lays out my thoughts for a process similar to the one specified in #213, including the proposed `IRCv3/ideas` repository.

This sort of a process would come with a tightening-up of the process for becoming a stakeholder. I think something along these lines should help us produce better specifications and help the members of the IRCv3 WG better spend their time and effort where they want to, without having to pay attention to every single issue or feature request brought up by everyone if they don't want to.

Additionally, this would come with the creation of the `IRCv3/ideas` repository, which is something we can spec up if we decide moving in this direction is a good idea.

Formatted version of this document can be found [here](https://github.com/DanielOaks/ircv3-specifications/blob/contributing/CONTRIBUTING.md).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,57 @@
+# How to Contribute
+
+There are a few different ways that you can contribute to the IRCv3 Working Group:
+
+* **Bug Fixes and Changes**: These fix bugs and issues with existing specifications, with errata updates where necessary.
+* **Ideas**: These outline new features that have not yet been worked out in full and do not have a reference implementation.
+* **Feature Proposals**: These outline new features and capabilities for the IRC protocol.
+
+## Bug Fixes and Changes
+
+A bug fix may be relatively minor, such as [#186](https://github.com/ircv3/ircv3-specifications/pull/186), or may fix issues found with the specification after it has been merged or ratified, such as [#204](https://github.com/ircv3/ircv3-specifications/pull/204).
+
+Minor bug fixes, such as fixing spelling mistakes or making formatting more consistent, may be submitted without much process. However, try to minimise the number of fixes you submit, as having 13 separate PRs to fix a few spelling and formatting issues can become grating to other participants and observers of the IRCv3 WG.
+
+Submitting bug fixes and changes of this type does not have any stakeholder requirements.
+
+Fixes and changes which address the actual behaviour or meaning of a specification *after they have been released*, such as [#204](https://github.com/ircv3/ircv3-specifications/pull/204) generally require 'errata updates'. Examples of errata updates can be found in various specifications, but they basically leave a log as to how a specification has been changed after it was released. Before submitting a fix of this type, you should talk with other members on the [#ircv3](http://ircv3.net/contact.html) channel and ensure there is support for your change.
+
+## Ideas
+
+Feature proposals contributed to the IRCv3-Specifications repository require a concrete specification and at least one reference or existing implementation.
+
+If you have an idea or a protocol proposal doesn't have these things or is not yet at this stage, you should make an issue on the [IRCv3/ideas](https://github.com/ircv3/ideas) repository instead. This repository is intended for working group members, as well as developers and users to be able to suggest ideas that are not yet full realised or without a reference implementation.
+
+Contributing to the [IRCv3/ideas](https://github.com/ircv3/ideas) repository does not have any stakeholder requirements. It was created so that contributors may decide to only review ideas that have already been specified and have support from other stakeholders of the IRCv3 WG, or review those that aren't yet at that stage.
+
+## Feature Proposals
+
+A new feature proposal includes things like [`account-tag`](http://ircv3.net/specs/extensions/account-tag-3.2.html), [`cap-notify`](http://ircv3.net/specs/extensions/cap-notify-3.2.html) and [`echo-message`](http://ircv3.net/specs/extensions/echo-message-3.2.html).
+
+Contributing a proposal to the IRCv3-Specifications repo requires a few things. These include:
+
+### Stakeholder Support
+
+Specifications submitted to this repository should be proposed or sponsored by a stakeholder of the IRCv3 Working Group.
+
+### Concrete Specification
+
+Your specification *must* describe all behaviour that comes with your proposal. You cannot simply say *"Yeah, and this part we'll work out later, it should be fine."*.
+
+The protocol syntax specified in your proposal may, of course, be changed after it is submitted. However, it should *not* be submitted with the attitude of *"Well, I didn't think about this too hard and I don't think it's great, what do other people think?"*. Proposals of that type should go through the [IRCv3/ideas](https://github.com/ircv3/ideas) process before being submitted to this repository.
+
+### Vendor Tag
+
+Any capabilities that your proposal introduces *must* be vendor-tagged. The vendor tag will be removed once it is made into a proper specification through the IRCv3 release process.
+
+For instance, if my spec introduces a new capability which will be named `cool-tag`, in the spec and in the reference implementation it should be specified as something like `example.com/cool-tag`.
+
+### Reference Implementation
+
+Your proposal must be implemented by at least one piece of IRC software. This is an IRC client, server, bouncer, or preferrably one of each. This implementation *must* be either downloadable, or be able to be accessible online (so it can be tested by other interested members).
+
+If your proposal is accepted, in the course of drafting an IRCv3 release it will require wider implementation.
+
+### Existing Implementations
+
+If a feature specifies capabilities that already exist in the IRC ecosystem, such as [`monitor`](http://ircv3.net/specs/core/monitor-3.2.html), they may be specified without a vendor prefix and this implementation may count as the 'Reference Implementation'. It should be ensured that any proposal of this type is backwards-compatible with existing implementations.


DanielOaks:
Basically, here's the process I imagine:

## Ideas
1. Someone posts their idea in `IRCv3-ideas`.
2. People decide it's a bad idea and work stops, or it's a good idea and work happens to further specify it.
3. At some point, a Stakeholder writes up and implements their version of that idea of it in their client/server behind their vendor capability.
4. Specification gets submitted to `ircv3-specification`, gets talked about and accepted or rejected.

## Non-Stakeholder Proposals
1. Specification is written up for the proposal.
2. Reference implementation is created for it.
3. Stakeholder submits it on behalf of the Non-Stakeholder, or Non-Stakeholder submits it with a Stakeholder 'sponsoring' it.
4. Gets discussed and accepted or rejected.

## Stakeholder Proposals
1. Specification is written up for the proposal.
2. Reference implementation is created for it.
3. Stakeholder submits it.
4. Gets discussed and accepted or rejected.

kaniini:
:+1:

jwheare:
My issue with this is that it would block proposals that are well-specified but have no implementation, or if the implementation is one sided because it's a server spec proposed by a client developer or vice versa. It also makes sense not to sink too much time into an implementation if you're expecting a proposal to be significantly modified through discussion with the working group.

That said, I emphasise the *well-specified* requirement, and would support guidelines that quantify that somehow, e.g. by requiring protocol-level examples of the specification in action, and presenting an understanding of possible implementation challenges.

DanielOaks:
Perhaps the two places we should have the implementation requirements are at the merging stage, and at the ratifying stage. Require at least one implementation before it's merged into the repo proper, and the more stringent implementation requirements before we actually ratify it as a full IRCv3 specification.

For stuff that gets put towards this specification, it MUST be well-specified, as stuff that isn't should be kept in the `ircv3/ideas` repo until such time as someone specifies it properly and to the point that we can have proper discussion on it. Of course, we specify what *"well-specified"* actually means.

jwheare:
Yup, before merging and ratifying is fine, but PRs shouldn't be dismissed outright *just* because they don't have an implementation.

DanielOaks:
I can see that, so long as it is tightened up in that they must be well-specified, and the stakeholders requirements. The main issue is stopping everyone from having to unnecessarily waste their time and energy debating ideas that aren't fully thought through, and already have almost no support from the actual working group members.

But merging and ratifying, I definitely think that forcing implementations to be written is definitely a good thing. Especially with the requirement that the vendor capability is left on the spec until it gets ratified as part of an IRCv3 release.

jwheare:
I'm a little bit conflicted about vendor prefixes. In some cases they useful without a doubt, e.g. a spec that is only applicable to a specific vendor, but I have mixed feelings about them in other circumstances, e.g. CSS and X-prefixed http headers. I don't feel like there is a correct rule on them either way. Sometimes they're useful, sometimes not, for a variety of reasons. It should really depend on the specification whether a prefix is required or recommended, and not a general rule.

DanielOaks:
I think that keeping them prefixed while we are working on them prevents us from getting into a situation where real clients and servers may respond to the unprefixed version of the capability with a WiP mode of operation. Once it's made into an official IRCv3 specification and we actually recommend implementing it for general use, the vendor prefix is stripped off.

Of course, we can tell software vendors not to put that code that responds to the capability with WiP modes of operation all we like, but this ensures that they don't.

DanielOaks:
One of the other ideas that came up is just doing something more like, for instance:  `wip.ircv3.org/sts-1`, `wip.ircv3.org/sts-2`, `wip.ircv3.org/sts-3`

This prevents the vendor's tag from causing any grief in terms of people wanting to implement it or not, and I think solves the issue of production software reacting to capabilities with codepaths that were written in the work-in-progress stage of the specification and don't actually reflect how the spec has changed before it was ratified.

I think that while we're doing a process change of this type, it is good to tackle this sort of issue at the same time, since I think it could become a bigger issue.

ChickenNuggers:
:+1: to non-stakeholder requiring stakeholders.

Do stakeholders have the ability to submit issues to ircv3-specifications or should it be discussed in -ideas first?

CONTRIBUTING.md
```diff
@@ -0,0 +1,57 @@
...
+
+### Vendor Tag
+
+Any capabilities that your proposal introduces *must* be vendor-tagged. The vendor tag will be removed once it is made into a proper specification through the IRCv3 release process.
+
+For instance, if my spec introduces a new capability which will be named `cool-tag`, in the spec and in the reference implementation it should be specified as something like `example.com/cool-tag`.
+
+### Reference Implementation
+
+Your proposal must be implemented by at least one piece of IRC software. This is an IRC client, server, bouncer, or preferrably one of each. This implementation *must* be either downloadable, or be able to be accessible online (so it can be tested by other interested members).
```
DarthGandalf: s/be able to//

CONTRIBUTING.md
```diff
@@ -0,0 +1,57 @@
...
+
+Your specification *must* describe all behaviour that comes with your proposal. You cannot simply say *"Yeah, and this part we'll work out later, it should be fine."*.
+
+The protocol syntax specified in your proposal may, of course, be changed after it is submitted. However, it should *not* be submitted with the attitude of *"Well, I didn't think about this too hard and I don't think it's great, what do other people think?"*. Proposals of that type should go through the [IRCv3/ideas](https://github.com/ircv3/ideas) process before being submitted to this repository.
+
+### Vendor Tag
+
+Any capabilities that your proposal introduces *must* be vendor-tagged. The vendor tag will be removed once it is made into a proper specification through the IRCv3 release process.
+
+For instance, if my spec introduces a new capability which will be named `cool-tag`, in the spec and in the reference implementation it should be specified as something like `example.com/cool-tag`.
```
DarthGandalf: Not necessary... There was a sad story of `znc.in/server-time` which was rewritten and became `znc.in/server-time-iso`, which became `server-time` as is. If spec is changed, it can't preserve the same name after `/`

CONTRIBUTING.md
```diff
@@ -0,0 +1,57 @@
...
+
+Contributing a proposal to the IRCv3-Specifications repo requires a few things. These include:
+
+### Stakeholder Support
+
+Specifications submitted to this repository should be proposed or sponsored by a stakeholder of the IRCv3 Working Group.
+
+### Concrete Specification
+
+Your specification *must* describe all behaviour that comes with your proposal. You cannot simply say *"Yeah, and this part we'll work out later, it should be fine."*.
```
DarthGandalf: Should STS be dropped until issues with it are resolved?

DanielOaks: Under this process change it would continue being workshopped in `IRCv3/ideas` until the whole base was specced out.

DanielOaks:
The vendor tag discussion has been dropped from this issue, and I will make a separate issue discussing my thoughts on it.

SaberUK:
I'd like to propose that we adopt something similar to what Rust does with its RFC system for ideas.

https://github.com/rust-lang/rfcs/blob/master/0000-template.md

attilamolnar:
Why is a separate, ideas repo required for idea issues? Issue labels are supported, we could use them: have a "triaged" label and filter for that if you want to only see those.

attilamolnar:
From a discussion on IRC: the benefit of a separate issue-only repo is to not have the ideas in the default view and the possibility to opt-out from notifications of comments on those issues. These make sense and with labels, these are not possible. This also means that issues on the ideas issue tracker can remain open forever and be ignored in case there is no interest instead from stakeholders.

CONTRIBUTING.md
attilamolnar: I'm willing to accept this, but I'd rather prefer this to be judged by stakeholders individually for each specification. If the specification works well as-is but could be improved somehow (STS falls into this category) then that should not block pulling the specification. What do you think?

DanielOaks:
Strengthened the language in the *Concrete Specification* section, which should hopefully be non-contentious.

Also added a section about actually ratifying feature proposals. I expect this section to be somewhat of a bikesheddy topic, but I think something along these lines should give us fairly decent coverage of the sort of issues that would pop up in specs we produce.

jwheare:
A line should be added to remind those with access to the main ircv3 repos to still open PRs from personal forks, to avoid branch proliferation and rebase risk in the primary repos and to not premature imply consensus. (I just did this in #239)

attilamolnar:
What's blocking this?

attilamolnar:
Bump

DanielOaks:
I'll take another pass over this on the weekend, make sure @jwheare's thing is solved and I can't see any further issues.

attilamolnar:
@DanielOaks have you had the time yet to finish the work you mentioned on this?

DanielOaks:
@attilamolnar I have some time so I'll do it now, will post once I'm happy with it.

DanielOaks:
Alright I'm happy with this

attilamolnar:
Thanks, I'll take a look in a few hours.

CONTRIBUTING.md
```diff
@@ -36,13 +38,13 @@ Specifications submitted to this repository should be proposed or sponsored by a
 
 ### Concrete Specification
 
-Your specification should describe all behaviour that is intended to be implemented by it before being submitted to this repo. If it is not, it may be sent it back to the [IRCv3/ideas](https://github.com/ircv3/ideas) repository to be workshopped until it is able to be fully specified.
+Your specification should describe all behaviour that is intended to be implemented by it before being submitted to this repo. If it is not, it may be sent it to the [IRCv3/ideas](https://github.com/ircv3/ideas) repository to be workshopped until it is able to be fully specified.
```
jwheare: Typos. Should be: "If it does not, it may be sent to the"

DanielOaks: Fixed, thanks

attilamolnar:
My only suggestion is that if the ideas repo is supposed to be forked for some reason then it should be called ircv3-ideas because otherwise when forked the fork will be at username/ideas which is too general and might already be taken. If there is no reason for forking currently (i.e. repo is an issue tracker only) this should still be taken into account as there can be a use for forking in a future workflow for ideas down the road.

DanielOaks:
Makes sense, I'll rename it to ircv3-ideas to match this repository

PR #229, opened 2016-02-14
Add a template for specifications to follow.

SaberUK:
This is just an initial draft. Feedback would be appreciated.

### DIFF
--- a/template.md
+++ b/template.md
@@ -1,0 +1,79 @@
+---
+title: Specification Name
+layout: spec
+work-in-progress: true
+copyrights:
+  -
+    name: "Your Name"
+    email: "you@yourdomain.com"
+    period: "Year Range (e.g. 2010-2015)"
+---
+
+<!--
+Edit this file and save it to either core/ or extensions/ and file a pull
+request against https://github.com/ircv3/ircv3-specifications.
+-->
+
+## Introduction
+
+<!--
+A quick summary of the feature which this specification describes.
+-->
+
+Insert text here.
+
+## Motivation
+
+<!--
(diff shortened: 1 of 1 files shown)


dequis:
Sections for numerics, capabilities, and isupport tokens?

And i think it'd be better to remove those sections if they are empty instead of having "This specification defines no client commands."

Also the template might get too verbose with all of those, i'd say just put them one per line, like

```
### Server Commands
### Client Commands
### Channel Modes
### User Modes
### Numerics
### Capabilities
### ISUPPORT tokens
```

And tell people to remove or fill as needed.

DanielOaks:
:+1:

I agree with @dequis as well, those sections seem like ones that can be omitted if they're not required.

SaberUK:
@dequis Done!

jwheare:
As of yesterday (https://github.com/blog/2111-issue-and-pull-request-templates) we can name this file `.github/PULL_REQUEST_TEMPLATE.md` for added magic.

SaberUK:
This is a template for specifications not for pull requests though.

jwheare:
Ah true, never mind.

SaberUK:
I don't think there are any more comments on this. Can it be merged now?

jwheare:
The following sections should be optional: Security implications, Alternatives, Errata. For new specifications, there won't ever be errata, so it's a bit redundant to include it at all in the template.

attilamolnar:
I don't think this should be merged like this. Empty/placeholder sections aren't needed in specifications. No IRCv3 specifications add user or channel modes. Why reserve a section for them?

Issue #209, opened 2016-02-04
Clarify STS implementation requirements


jwheare:
Opening this issue as a less inflammatory/more actionable version of #207 (which can be closed)

1. Clarify that server admins have a choice whether or not to enable this CAP.
2. Clarify that core compliance with 3.3 refers to:

  a) Clients following their side when instructed
  b) Servers supporting the CAP, if a server operator decides to enable it via configuration
3. Add a "when enabled" to this part (it means it must be present on plaintext and secure connections)
```The capability value, which MUST always be present```

thwg:
I see this is a hot topic, but I'll chime in with my thoughts.  Feel free to correct if I'm mistaken on any parts, as I'm not an expert w.r.t. X.509.

For each previous specification (except for the SNI one), the server extends its capabilities to match a feature that the client wants to improve the user experience.  From client authors' perspective, it allows us to create a better client, so it's easier to get them adopted.

This specification provides server rules to provide info that the client might want (the secure port and instructions to use TLS in the future), but the problem was never with the info the servers provided -- the problem is reliably verifying the identity of the server against a trusted root CA.  The specification mentions it, but only briefly:

> Clients MUST use sufficient certificate verification to ensure the connection is secure without any errors.

Given my rusty understanding, but based on what web browsers do, in the future we would do something to the effect of:
 - Collect a list of the Common trusted root CAs (e.g. Debian's ca-certificates)
 - Keep an up-to-date list of CRLs (maybe OCSP nowadays?)
 - Distribute them to mirrors (PGP signed)
 - Include an update mechanism in client, and update regularly from mirrors
 - Ship the public PGP key with the client distribution to verify the authenticity of each update
 - Keep an additional list of certificates that the user has personally trusted.

If this were easy then we would all be TLS-preferred in the default servers list a long time ago.  I think some good advice for what "use sufficient certificate verification" means would be helpful to most client authors.  Not included with the spec, but perhaps placed the documentation/ directory.

My other concern: I'm not a daemon author, but I do see @kaniini's point in the other thread.  This (transparent TLS preference-setting) does seem like a very specific network policy-oriented feature.  With the other IRCv3 features, it is easy to justify supporting them in the ircd software because there are likely many users on networks that benefit from them.  But if there are not any known networks being run with this feature and your software, then you've got an unused and under-tested security-related feature in your ircd to be considered IRCv3-compliant.

[...]

PR #230, opened 2016-02-14
Move STS specification to extensions.

Aerdan:
As it is currently written, the STS specification is not suitable to be a core specification. Among other things, it provides for a MITM attack to filter STS out, falsely present to the server a secure connection, or to prevent the client from connecting at all. Additionally, there are no provisions for clients to opt out of STS should they wish to insist on a plaintext connection. The specification also depends on TLS certificates to not be compromised, which can be done without the client's knowledge while still being verifiable.

Until such time as these issues are addressed by IRCv3 or this specification, this specification should not be mandatory for IRCv3.3 compliance even as a draft specification.

(For a more complete list of issues that need to be addressed, see #231.)


highflyer77:
I agree 100% percent. :+1:

SaberUK:
:-1:

ChickenNuggers:
:+1: until the issues with STS are resolved.

kaniini:
@SaberUK

First off please understand that I am not against STS being an optional component of the spec.  I have said that all along.  My concern is with the aggression that this is being pushed.

> This is not a defect in the specification. This is a client implementation detail which is outside the scope of the specification. There is nothing in the specification preventing a client from doing this.

I feel that it is important to mention that without proper guidance, your intentions may be misinterpreted.  A lack of the spec preventing it does not mean that clients will provide an opt-out, in fact I would argue that they will not.

An opt-out is important because some networks are intentionally configured as plaintext.  In practice, insecure networks are used by both state actors and activists as part of a multilateral information disclosure strategy (this is what "security clearances" are about).  The reason why is because the psychological knowledge of *knowing* that you're connected to an *insecure* network usually encourages participants to refrain from discussing sensitive information.  As part of an operation's MLS strategy, a secure "airgapped" network is used to discuss sensitive information.  If a user is tricked into thinking her connection is secure, she may volunteer classified information.  I fear that STS can be used as a method for tricking users who should otherwise know better.

Additionally, STS is an excellent way to cause a permanent denial of service to a target.  If you "upgrade" the user to an MITM (admittedly signed by a nation state, CNNIC ROOT CA looks like a good option here) with a valid certificate chain (again, you're probably trusting CNNIC ROOT CA right now) then the client will pin that network to continue using TLS.  Once the client shifts networks (for example, from the cellular network to wifi), the pinning ensures that the denial of service occurs as a result of the MITM no longer being present to provide a fake TLS relay, cutting off contact with their team.  While it is certainly true that there are other ways to deny service to a target, this method is particularly devious because it's client-side and will likely be difficult to debug, because STS is intentionally a black box.  This is another reason why a permanent opt-out for specific servers is essential, and why the spec needs to discuss this possibility.  A proper risk assessment would have shown this issue.

Further, IRC is not the same as HTTPS in several ways, notably that compromising only one user is all that is required to compromise every other user on a channel.  HTTPS is a 1-to-1 protocol where this is typically not the case, so IRC requires careful thinking about how adversaries may attack security attributes of the protocol since the use cases are different.  You can't just do the same thing as HTTP here, there are cases where an STS "upgrade" is not an upgrade at all, and forcing users to swallow that pill without giving them an opt-out is dangerous and may very well result in people in the real world being harmed.

IRC is fundamentally different than HTTPS because of the differences in use case and scope.  IRC has users outside of casual IRC chat networks and large networks like freenode.  It is used by everyone ranging from activists, to the military, to intelligence services to conduct operations.  **There is a strong potential this spec may intersect with those uses in ways that put actual people at risk**, which is something that needs to be *strongly considered* when considering security-related changes for IRC.

I'd like to also point out that there are people who compare STS as being similar to the SASL binding.  This is complete nonsense.  SASL does not have the ability to deny service, nor does it potentially put a user at risk because they may have been unwittingly compromised by an MITM (after which future service is denied to that user if they are not in the MITM network, so you can collect information and then disconnect them from their team).   STS should not be in core at this time until a complete adversarial risk analysis can be done on it, and all pitfalls documented so that clients can be aware of them.  Comparing STS to HSTS is also a poor comparison, IRC is not a website and therefore has potentially completely different security consequences as a result of getting the security wrong.  If your bank screws up their security, you lose money, but you're not likely to be *killed*.  If an activist communication server for a group in say, Iran for example, is suddenly "secure", and sensitive information is unwittingly discussed by the victim, people are likely to die.  Client authors need to realize that their users may be in the field, involved in an operation where their safety is at risk, not every user of your IRC clients are connecting to a typical IRC network.

jwheare:
Firstly, this issue as title can be closed. The STS draft is still in progress. The discussion of whether or not it should be in core is a derailing tactic. Let's rather address the concerns enough that it can be in core (using the existing issues #209, #231), rather than make the meaningless change of moving a by definition non-recommended draft spec to a different priority level. It's issue tracker noise.

The stated issues with the spec are discussable and resolvable, no one is trying to rush anything through.

----

To address the actual issues raised (which should be moved to one of the previously linked issues if valid/needed)

> It provides for a MITM attack to filter STS out, falsely present to the server a secure connection, or to prevent the client from connecting at all.

Any upgrade mechanism will have this property, this is not a criticism of the spec itself, but of an unavoidable reality. The fact is, people *are* on insecure, vulnerable connections, we are trying to move as many of them as possible to a more secure alternative, in an automated way, without expecting users to manually edit their connection details.

> Additionally, there are no provisions for clients to opt out of STS should they wish to insist on a plaintext connection.

Covered in #231

> The specification also depends on TLS certificates to not be compromised, which can be done without the client's knowledge while still being verifiable.

Any mechanism involving TLS certificates will have this property, not a criticism of the spec, but of reality. Alternatives to CA-backed certificates have been discussed, though no concrete proposal has been outlined. Fingerprint lists published by servers, or out of band, combined with a trust on first use model have been mentioned. They should be explored in detail and I would love to see a proposal.

---

@kaniini, if you can summarise that response to actionable bullet points (on one of the existing tickets) I would appreciate it. As it stands there are a lot assumptions bundled up in there and I don't really know where to begin.

Aerdan:
HSTS, which this specification appears to want to emulate, specifically instructs clients to ignore STS headers on plaintext connections. Redirection, therefore, does not belong in STS, particularly since there is an existing mechanism which some clients already support (`RPL_REDIR`/numeric `010`), notably mIRC. #231 points this out, too.

Regardless of #231, however, the lack of an opt-out mechanism (e.g., non-binding redirect) means that someone could plausibly supply a valid cert for pinning that will then prevent the client from connecting when they are on a different network. Such use of certs has been employed for TLS relays in network equipment in the recent past.

At any rate, I don't see how a specification being in development excuses it from being unsuitable as a core spec. Core specifications should generally not be ones for which philosophical objections can be mounted, since that threatens to undermine the importance of being compliant with IRCv3. @kaniini raised several significant points on IRC regarding this particular point.

Aerdan:
Owing to entirely too much argument for clarity on anyone's part, I have endeavoured to author an alternative proposal which, by necessity, means this PR is no longer required.

Issue #231, opened 2016-02-15
Issues with STS that need to be fixed


kaniini:
These fixes should be made to STS:

- [ ] The effect of encountering an `sts` cap should not cause a TLS upgrade, this should be decoupled from the spec and handled differently.
- [ ] The STS "pinning" should explicitly only be accepted if:
  - [ ] the client has accepted the TLS connection (i.e. accepted the certificate)
  - [ ] the certificate is verifiable to a root stored in the client's CA bundle (creating a small bundle for IRC-specific use may also be interesting, not covered here) OR has explicitly accepted the certificate (hopefully after examining it)
- [ ] The client MUST have the ability to override STS, for example if it were erroneously set either on purpose by an attacker (with verifiable certificate, such as one signed for example, by CNNIC ROOT CA) or by accident by the admin
- [ ] The client MUST have the ability to *ignore* STS if explicitly configured to do so (to prevent false upgrades)

It is probably desirable to include RPL_REDIR as an upgrade mechanism, which can be provided as a hint that TLS is available.  Such a specification should probably mention that RPL_REDIR which violate STS policy if pinned SHOULD be ignored.  Once a TLS connection is made, then the STS pinning can be presented and placed into effect if the TLS connection is valid.

jwheare:
Thanks.

>  The effect of encountering an sts cap should not cause a TLS upgrade, this should be decoupled from the spec and handled differently.

At the moment the spec uses just one CAP for both the upgrade redirect (the `port` key) and the pinning (the `duration` key). These are two separate mechanisms that trigger separate client behaviour. They are combined in one syntax, but they do not need to be, but I don't feel like it's a major point of contention as the distinction is clearly defined in the spec.

What are the pros and cons of combining them vs using a separate numeric e.g. RPL_REDIR? It's worth noting that the redirect should take place as soon as possible in the connection phase.

> The STS "pinning" should explicitly only be accepted if:
 the client has accepted the TLS connection (i.e. accepted the certificate)
 the certificate is verifiable to a root stored in the client's CA bundle (creating a small bundle for IRC-specific use may also be interesting, not covered here) OR has explicitly accepted the certificate (hopefully after examining it)

This is covered in the spec
* "Clients MUST use sufficient certificate verification to ensure the connection is secure without any errors."
* "If the client receives the duration key on a non-secure connection, it is invalid and MUST be ignored."

[...]

PR #232, opened 2016-02-16
Fix STS issues enumerated in #231 and add baseline TLS security practices

kaniini:
This pull request fixes STS issues as enumerated in #231, and adds non-normative baseline TLS security practices as guidance for client and server authors to follow.

### DIFF
--- a/core/sts-3.3.md
+++ b/core/sts-3.3.md
@@ -108,7 +108,19 @@ unable to do anything about except retrying at a later time.
 This is to ensure that it is not possible to fool users into allowing a
 man-in-the-middle attack.
 
-## Security
+## Security considerations
+
+### Acceptance of STS policy
+
+STS policies SHOULD ONLY be accepted for connections where the TLS certificate
+chain can be validated entirely, to either a user-specified root or the system
+root certificate bundle.
+
+### Visual feedback of STS upgraded connections
+
+Clients SHOULD NOT display visual feedback indicating that an STS-upgraded TLS
+connection has any enhanced security over a plaintext connection unless a user
+explicitly trusts the certificate.
 
 ### STS policy injection
 
@@ -116,6 +128,15 @@ It is possible for attackers to inject the STS policy into a plaintext
 connection. This causes the client to reconnect over TLS and either fail,
 or connect successfully over TLS but discover that the `sts` cap is not
 present.
+
+Additionally, it is possible for an attacker to inject an STS policy into
+a proxied TLS connection. This causes the client to continue following the
+STS policy once the proxy is removed, which may cause a denial of service.
+Clients SHOULD provide the ability to flush any relevant STS cache entries
+for the server to allow for re-learning the STS policy for that server, if
+any. Clients MAY additionally provide the ability to reject STS policies
+from servers the user knows to not provide an STS policy as an additional
+mitigation.
 
 ### Policy expiration
 
--- a/documentation/tls-best-practices.md
+++ b/documentation/tls-best-practices.md
@@ -1,0 +1,71 @@
+---
+title: IRCv3 TLS Best Practices
+layout: spec
+copyrights:
+  -
+    name: "William Pitcock"
+    period: "2016"
+    email: "nenolod@dereferenced.org"
+---
+
+This document specifies a series of baseline security practices for managing TLS on IRC networks.
+It should be considered non-normative.
+
+
+# Clients
+
+* Clients SHOULD upgrade the default port to 6697 and enable TLS per IETF [RFC 7194][rfc7194].
+
+  [rfc7194]: https://tools.ietf.org/html/rfc7194
+
+* Clients SHOULD allow a user to downgrade to plaintext if they explicitly configure the client
+  to do so.  A suggested `/server` command may be something like:
+
+    /server -insecure irc.example.com 6667
+
+* Clients SHOULD implement [STS][sts-3.3], with the following suggestions:
+
+  * Clients SHOULD implement the ability to explicitly disable STS for a connection if the user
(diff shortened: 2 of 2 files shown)


documentation/tls-best-practices.md
```diff
@@ -0,0 +1,71 @@
...
+# Clients
+
+* Clients SHOULD upgrade the default port to 6697 and enable TLS per IETF [RFC 7194][rfc7194].
+
+  [rfc7194]: https://tools.ietf.org/html/rfc7194
+
+* Clients SHOULD allow a user to downgrade to plaintext if they explicitly configure the client
+  to do so.  A suggested `/server` command may be something like:
+
+    /server -insecure irc.example.com 6667
```
SaberUK: I'm not so sure we should be recommending client interface implementation details.

kaniini: Why?  It's non-normative text, intended in part, to serve as guidance to client authors.  They can surely ignore it if it is inappropriate to them.

DanielOaks:
I think this makes sense. :+1:

documentation/tls-best-practices.md
Aerdan: If security is within the scope of this working group, as demonstrated by the desire to promote STS, then anything that touches on security is also within scope. Therefore, it's entirely reasonable to offer guidelines on user interfaces to client authors.

core/sts-3.3.md
```diff
@@ -108,7 +108,19 @@ unable to do anything about except retrying at a later time.
...
+
+STS policies SHOULD ONLY be accepted for connections where the TLS certificate
+chain can be validated entirely, to either a user-specified root or the system
+root certificate bundle.
+
+### Visual feedback of STS upgraded connections
+
+Clients SHOULD NOT display visual feedback indicating that an STS-upgraded TLS
+connection has any enhanced security over a plaintext connection unless a user
+explicitly trusts the certificate.
```
jwheare: If the certificate has a trusted root, then that is explicitly trusted by the user. And if not, then the connection will fail as an error as mentioned in the "No user recourse" section. So this paragraph on visual feedback is redundant.

documentation/tls-best-practices.md
```diff
@@ -0,0 +1,71 @@
...
+---
+
+This document specifies a series of baseline security practices for managing TLS on IRC networks.
+It should be considered non-normative.
+
+
+# Clients
+
+* Clients SHOULD upgrade the default port to 6697 and enable TLS per IETF [RFC 7194][rfc7194].
+
```
jwheare: This will result in a failure to connect for the second largest IRC network, and thus represent a support overhead for client authors. As such I wouldn't recommend it. We could say something like:

"Clients should prefer TLS connections on default port 6697 if available when populating their preloaded server lists."

jwheare:
Barring my inline comments, I'm mostly happy with the sts changes, but I think their inclusion will need to be followed up by some restructuring, with more clarity about normative vs non-normative sections. As such, I'd accept the changes with a TODO added.

However I think the TLS best practices should be split off from this PR.

While I appreciate the general intention of the TLS best practices, and am in favour of such a document in general, many of these suggestions would be difficult or infeasible to implement given the current landscape. If we are to publish such a document, I think we should focus on things that will not on the whole be discarded by authors.

That's not to say some of the more difficult/contentious recommendations shouldn't be discussed further, or even documented somewhere. But perhaps under a different title.

Also, if it's a non-normative document, it should avoid the use of SHOULD, etc, as suggested by https://tools.ietf.org/html/draft-hansen-nonkeywords-non2119-03.

core/sts-3.3.md
Aerdan: No, the cert being on a trusted root is an *implicit* trust that the cert is valid. *Only when the *\user verifies the cert themselves* is an explicit trust created.

documentation/tls-best-practices.md
Aerdan: I think the way it's currently phrased indicates that it can be overridden somewhere, presumably via the redirect announcement or by the user themselves.

core/sts-3.3.md
jwheare: OK, so can the client provide visual feedback that the TLS connection has been verified by the trusted root? But then, if you're a user who doesn't know how to manually verify a certificate, isn't the fact that it's verified and implicitly trusted by your system a security enhancement?

Are we allowed to show a lock icon but just not make it green? How are we allowed to indicate a verified trusted root connection vs plain text. They *are* different.

documentation/tls-best-practices.md
jwheare: The wording suggests to me that if a user simply types: "irc.quakenet.org" into their client, it will default to TLS on port 6697 and fail to connect.

core/sts-3.3.md
Aerdan: The precise details of how the client should represent untrusted vs implicit trust vs explicit trust is up to the clients, but I would look at best practices with browser UI regarding security for inspiration.

documentation/tls-best-practices.md
Aerdan: That would be the case if the client doesn't ship with server information preloaded, though. A smart client which did include such information should be able to figure it out, however.

core/sts-3.3.md
jwheare: Browsers use a green padlock for implicit root trusted connections. Does that indicate enhanced security? Open to interpretation I think. This paragraph remains a problematic recommendation.

jwheare:
Addressed the spec part in #249. Would need to be reworked/revised as a best practice document to be mergeable. Closing for now.

PR #235, opened 2016-02-18
core/metadata-3.3: Add a specification for IRCv3.3 metadata.

Aerdan:
This commit adds the initial work for IRCv3.3's extensions to metadata. It is a core specification because the changes being made are relevant to all implementers of the IRCv3.2 metadata specification, which is also core.

### DIFF
--- a/core/metadata-3.3.md
+++ b/core/metadata-3.3.md
@@ -9,7 +9,7 @@ copyrights:
 ---
 ## Introduction
 
-This specification incorporates the IRCv3.2 metadata specification and
+This specification extends the [IRCv3.2 metadata specification][metadata] and
 incorporates it by reference.
 
 ## METADATA
@@ -31,13 +31,10 @@ and on-change notifications, `change` permitting only on-change notifications,
 and `request` requiring that the client explicitly request metadata with the
 relevant keys.
 
-`<Keys>` may optionally be an asterisk (`*`), in which case the priority change
-applies to all keys.
-
 If any key is invalid, that key MUST generate an `ERR_KEYINVALID` event. If any
 key is not intended to be visible to the user due to permissions, that key MUST
-generate an `ERR_KEYNOPERMISSION` event. As usual, if the target is invalid,
-the whole command receives `ERR_TARGETINVALID`.
+generate an `ERR_KEYNOPERMISSION` or `ERR_NOMATCHINGKEY` event. As usual, if
+the target is invalid, the whole command receives `ERR_TARGETINVALID`.
 
 ## Numerics
 
@@ -54,3 +51,6 @@ silently filtered, numeric 763 is to be defined as follows:
 ## Examples
 
 ...
+
+[metadata]: http://ircv3.net/specs/core/metadata-3.2.html
+
--- a/core/metadata-3.3.md
+++ b/core/metadata-3.3.md
@@ -1,0 +1,56 @@
+---
+title: IRCv3.3 Metadata
+layout: spec
+copyrights:
+  -
+    name: "Kiyoshi Aman"
+    period: "2016"
+    email: "kiyoshi.aman@gmail.com"
+---
+## Introduction
+
+This specification incorporates the IRCv3.2 metadata specification and
+incorporates it by reference.
+
+## METADATA
+
+### METADATA PRIORITY
+
+This subcommand provides a mechanism for clients to inform the server what
+priorities to assign to particular keys. The default priority is 'on
+subscription' for all keys. The format of `METADATA PRIORITY` MUST be as
+follows:
+
+    METADATA <Scope> PRIORITY <Priority> :<Keys>
+
+`<Scope>` is one of `global`, `channel`, or `user`, reflecting all subscriptions,
+channel-related subscriptions, and subscriptions via `MONITOR`.
+
+`<Priority>` operates in descending order, with `sub` enabling on-subscription
+and on-change notifications, `change` permitting only on-change notifications,
+and `request` requiring that the client explicitly request metadata with the
+relevant keys.
+
+`<Keys>` may optionally be an asterisk (`*`), in which case the priority change
+applies to all keys.
+
+If any key is invalid, that key MUST generate an `ERR_KEYINVALID` event. If any
+key is not intended to be visible to the user due to permissions, that key MUST
+generate an `ERR_KEYNOPERMISSION` event. As usual, if the target is invalid,
+the whole command receives `ERR_TARGETINVALID`.
+
+## Numerics
+
+For the purposes of reporting that a `METADATA` change was blocked rather than
+silently filtered, numeric 763 is to be defined as follows:
+
+
+| No. | Label                 | Format                                   |
+| --- | --------------------- | ---------------------------------------- |
+| 763 | `ERR_VALUEINVALID`    | `<Target> <Key> :was blocked; <Reason>`  |
+
+`<Reason>` contains a localized reason for the value to have been rejected. 
+
+## Examples
+
+...


core/metadata-3.3.md
```diff
@@ -0,0 +1,56 @@
...
+
+    METADATA <Scope> PRIORITY <Priority> :<Keys>
+
+`<Scope>` is one of `global`, `channel`, or `user`, reflecting all subscriptions,
+channel-related subscriptions, and subscriptions via `MONITOR`.
+
+`<Priority>` operates in descending order, with `sub` enabling on-subscription
+and on-change notifications, `change` permitting only on-change notifications,
+and `request` requiring that the client explicitly request metadata with the
+relevant keys.
```
attilamolnar: What's the rationale for this feature? What keys should it be used with? Please provide some examples as this seems to be unnecessary complication.

core/metadata-3.3.md
```diff
@@ -0,0 +1,56 @@
...
+
+`<Keys>` may optionally be an asterisk (`*`), in which case the priority change
+applies to all keys.
```
attilamolnar: I think this should only be possible when subscribing to user or channel keys, not neighbors. I can't see a use case for subscribing to every key of every neighbor and it can be abused easily.

core/metadata-3.3.md
```diff
@@ -0,0 +1,56 @@
...
+layout: spec
+copyrights:
+  -
+    name: "Kiyoshi Aman"
+    period: "2016"
+    email: "kiyoshi.aman@gmail.com"
+---
+## Introduction
+
+This specification incorporates the IRCv3.2 metadata specification and
```
attilamolnar: This should link to the old specification

core/metadata-3.3.md
attilamolnar: My alternative:
`METADATA <Scope> SUB :<Keys>` where Scope and Keys are the same as in your version and subscribing means all notifications, not subscribing means no notifications but the key is still visible otherwise (whois, METADATA LIST or GET)

core/metadata-3.3.md
```diff
@@ -0,0 +1,56 @@
...
+
+## Numerics
+
+For the purposes of reporting that a `METADATA` change was blocked rather than
+silently filtered, numeric 763 is to be defined as follows:
+
+
+| No. | Label                 | Format                                   |
+| --- | --------------------- | ---------------------------------------- |
+| 763 | `ERR_VALUEINVALID`    | `<Target> <Key> :was blocked; <Reason>`  |
```
attilamolnar: I think the "was blocked; " text should not be there, rather, the reason should be the last parameter to make the reason easy to parse for clients. Clients should understand the numeric and present it to the user properly.

attilamolnar: METADATA should be in the name of the numeric to have a name which is more unique and is less likely to collide (we have no global registry of numerics...)

core/metadata-3.3.md
```diff
@@ -0,0 +1,56 @@
...
+### METADATA PRIORITY
+
+This subcommand provides a mechanism for clients to inform the server what
+priorities to assign to particular keys. The default priority is 'on
+subscription' for all keys. The format of `METADATA PRIORITY` MUST be as
+follows:
+
+    METADATA <Scope> PRIORITY <Priority> :<Keys>
+
+`<Scope>` is one of `global`, `channel`, or `user`, reflecting all subscriptions,
```
attilamolnar: `global` is a misnomer, it can be understood as subscribing to all clients on the network which is not the case. What do you think about `neighbor` or `visible`? I don't mind what we end up with as long as it's not misleading.

core/metadata-3.3.md
jwheare: I'm still not convinced `Scope` is needed even. Again, an example would help.

core/metadata-3.3.md
jwheare: I'm not even sure it's needed on user or channel keys. A client should only request notifications for keys it understands.

core/metadata-3.3.md
Aerdan: Given the nebulousness of what keys are acceptable, scope's sticking around.

core/metadata-3.3.md
Aerdan: ...Given the nebulousness regarding key names, this isn't going away.

core/metadata-3.3.md
Aerdan: That criticism would apply to IRCv3.2 metadata too, in which case a separate PR should be made to address it.

core/metadata-3.3.md
Aerdan: How is it a misnomer? `global` covers all channel subs and all user subs. By inference, `channel` and `user` cover all channel and user subs, respectively. The only alternative is having to specify `channel,user`, which is absurd.

dequis: `global` works for me, even if you interpret it as all clients. Just think of priority as a mask that is AND'ed rather than OR'd. `global` sets the mask of desired keys for all clients, `channel` sets it for channels, etc. And that mask is AND'ed to the actual subscriptions.

core/metadata-3.3.md
dequis: I don't think there are many reasons for or against scope. I'm leaning on keeping it, maybe reconsider once the metadata registry has more real use cases.

core/metadata-3.3.md
DanielOaks: So long as the client can track what keys it's subscribed to, I don't think this is necessary. If the client is able to understand another key, it can request that key. There's no use to subscribing to keys that the client doesn't understand? Allowing `*` complicates the checking server-side, and without a useful reason I don't think it should be allowed.

core/metadata-3.3.md
```diff
@@ -0,0 +1,56 @@
...
+and on-change notifications, `change` permitting only on-change notifications,
+and `request` requiring that the client explicitly request metadata with the
+relevant keys.
+
+`<Keys>` may optionally be an asterisk (`*`), in which case the priority change
+applies to all keys.
+
+If any key is invalid, that key MUST generate an `ERR_KEYINVALID` event. If any
+key is not intended to be visible to the user due to permissions, that key MUST
+generate an `ERR_KEYNOPERMISSION` event. As usual, if the target is invalid,
```
DanielOaks: This should be "generate an `ERR_NOMATCHINGKEY` or `ERR_KEYNOPERMISSION` event", to stay consistent with 3.2's handling of keys that aren't visible.

[3.2 `GET`](http://ircv3.net/specs/core/metadata-3.2.html#metadata-get)
> Servers MAY replace certain metadata, which is considered not visible for the requesting user, with `ERR_NOMATCHINGKEY` or with `ERR_KEYNOPERMISSION`.

core/metadata-3.3.md
DanielOaks: Can some example keys be provided where the different `PRIORITY` values would be used and useful?

I can't see the benefits of priorities. It seems something like Attila suggested above with just `SUB`, `UNSUB`, and `SUBS` (something to see current subscriptions) or similar would make more sense?

dequis: @DanielOaks I think we already agreed on irc that, regarding priority, just a "boolean" sub/unsub is fine, mapping to the change/request values of the current text.

That's the regarding the priority parameter, where AFAIK all that's missing is updating the text.

Not so much consensus on the scope parameter, which is what most comments in this subthread ended up being about.

core/metadata-3.3.md
attilamolnar: It is misleading because the word global is used to refer to all people on the network. Consider these examples:

- In reply to LUSERS "Current global users: X Max: Y" means number of all users on the entire network.
- A "global notice" means a notice sent to all users on the server.
- The "Global" service is a service designed to send global notices.
- Nicknames and channel names are globally unique (i.e. on the entire network).

Also in other areas "global" refers to concerning all parts of something, e.g.:
- A global variable in a program is a variable that is visible throughout the entire program.

But Scope being "global" does not subscribe to all users on the network, it only subscribes to other users the client sending the command can see.

core/metadata-3.3.md
attilamolnar: Okay but please address this one in this PR.

jwheare:
Closing in favour of #250

PR #236, opened 2016-02-18
3.3 tags

Aerdan:
This initial specification for IRCv3.3 message tags adds support for key/value pairs and lists (and the combination) for message tags and defines a capability to indicate general support for message tags.

### DIFF
--- a/core/message-tags-3.3.md
+++ b/core/message-tags-3.3.md
@@ -37,6 +37,9 @@ The updated BNF for keys is as follows:
 
     <key>           ::= [ '+' ] [ <vendor> '/' ] <sequence of letters, digits, hyphens (`-`)>
 
+Client-to-client tags MUST be forwarded on `PRIVMSG` and `NOTICE` events, and
+MAY be forwarded on other events.
+
 ### Examples
 
 `@+znc.in/network PRIVMSG user :lol`
--- a/core/message-tags-3.3.md
+++ b/core/message-tags-3.3.md
@@ -43,11 +43,4 @@ The updated BNF for keys is as follows:
 
 `@+displayname=user2 PRIVMSG #channel :"..."`
 
-## Security Considerations
-
-There are no security considerations that implementations should note when
-implementing this specification.
-
-## Errata
-
 [irctags]: http://ircv3.net/specs/core/message-tags-3.2.html
--- a/core/message-tags-3.3.md
+++ b/core/message-tags-3.3.md
@@ -1,0 +1,53 @@
+---
+title: IRCv3.3 Message Tags
+layout: spec
+copyrights:
+  -
+    name: "Kiyoshi Aman"
+    period: 2016
+    email: "kiyoshi.aman@gmail.com"
+---
+## Introduction
+
+This specification adds a new capability for general message tags support and
+a prefix for expressing client-to-client tags.
+
+## Motivation
+
+A new capability is required in order to indicate support for client-to-client
+tags. Additionally, some implementations felt constrained to apply filtering
+rules to prevent clients from receiving tags they didn't request.
+
+## Architecture
+
+### Capabilities
+
+This specification adds the `message-tags` capability. Clients which request
+this capability indicate that they are capable of parsing all tags, regardless
+of support for tags specified in other IRCv3.2 or later specifications.
+
+### Tags
+
+Client-to-client tags are tags which the server forwards from the client which
+sent them to all other clients which received the relevant event. A
+client-to-client tag starts with a plus sign (`+`) and otherwise obeys the
+format provided in [IRCv3.2 tags][irctags].
+
+The updated BNF for keys is as follows:
+
+    <key>           ::= [ '+' ] [ <vendor> '/' ] <sequence of letters, digits, hyphens (`-`)>
+
+### Examples
+
+`@+znc.in/network PRIVMSG user :lol`
+
+`@+displayname=user2 PRIVMSG #channel :"..."`
+
+## Security Considerations
+
+There are no security considerations that implementations should note when
+implementing this specification.
+
+## Errata
+
+[irctags]: http://ircv3.net/specs/core/message-tags-3.2.html


SaberUK:
What about client to server tags?

Aerdan:
I didn't think we had a consensus, and in any event this is a modification of a document that's been sitting on my gist for several months.

jwheare:
I have a python sample parser for the new value sequence stuff knocking about from back then too. Will check it still applies to this proposal and post it.

jwheare:
Is equals (`=`) escaped in pairwise values, or just reserved in pairwise keys?

i.e. how would you encode `{'=key': '=value'}`?

description | tag | parsed
--- | --- | ---
escaped in both | `@tag:=\=key=\=value` | `{'=key': '=value'}`
escaped in key only | `@tag:=\=key==value` | `{'=key': '=value'}`
forbidden in key | `@tag:=key==value` | `{'key': '=value'}`
forbidden in key, escaped in value | `@tag:=key=\=value` | `{'key': '=value'}`

---

Also, looking at my old parser, it looks like at one point we considered value and pairwise sequences to be used in the same tag. e.g.
```
@intent:=announce,noreply,redirect:#channel

parsed to JSON =>
[
  'announce',
  'noreply',
  {
   'redirect': '#channel'
  }
]
```
It also looks like we were using colon (`:`) as the separator for pairwise values, making it reserved in the pairwise key but forbidden in the value.

Do we still want to allow that?

jwheare:
In lieu of a diff for the squashed edit:
```diff
-<value pair>          ::= <pairwise value> '=' <pairwise value>
+<value pair>          ::= <pairwise value> ['=' <pairwise value>]
...
-In addition to IRCv3.2's escape sequences, `\.` is added, which evaluates to `,`.
+In addition to IRCv3.2's escape sequences, `\.` and `\+` are added, which evaluate to `,` and `=` respectively.
```

So `{'=key': '=value'}` would be encoded as `@tag:=\+key=\+value`

And 
```JSON
[
  "announce",
  "noreply",
  {
   "redirect": "#channel"
  }
]
```
is encoded as: `@intent:=announce,noreply,redirect=#channel`

jwheare:
As suggested on IRC, shall we allow merging the values of duplicate pairwise keys into a sequence:

`@foo:=a=b,a=c`

```json
[
  {
    "a": [
      "b",
      "c"
    ]
  }
]
```

jwheare:
After discussion in IRC, we are considering shelving value lists/dicts in tags for the moment. This PR will be revised accordingly, in addition to a draft of hyphen-prefixed client-2-client pass through tags.

core/message-tags-3.3.md
```diff
@@ -0,0 +1,41 @@
...
+## Capabilities
+
+This specification adds the `message-tags` capability. Clients which request
+this capability indicate that they are capable of parsing all tags, regardless
+of support for tags specified in other IRCv3.2 or later specifications.
+
+## Client-to-Client Tags
+
+Client-to-client tags are tags which carry a hyphen as a prefix, immediately
+after the `/` if the tag has a vendor. Such tags SHOULD be forwarded by the
```
DarthGandalf: An example would help

DarthGandalf:
The intention of previous spec was to send only known tags to client. This spec seems to assume that previous spec was sending all tags to client which knows some tag.

Aerdan:
No, the purpose of this specification is to *add* such a capability.

jwheare:
To summarize the previous change: sigil changed from a hyphen `-` to a plus `+` and moved to precede the vendor prefix. I'm fine with this. It modifies the grammar for tags, which warrants a new pseudo-BNF, I think.

jwheare:
My suggestion following an IRC discussion:

> Servers that enable this capability MUST forward client-to-client tags on `PRIVMSG` and `NOTICE`
> and MAY forward them on other events.

This clarification is for server implementations: a server that accepts tags on `JOIN` might want to forward them when the join is relayed, but not store them around in case a netjoin happens later.

This also allows clients to gauge their expectations, emphasising that client-to-client tags are designed to facilitate *message*-level metadata, rather than extra *server*-level metadata (channel membership, away status, topics, invites, etc). Tags *can* be forwarded for those events (to save server side logic), but clients shouldn't expect them to be. Subsequent client-side specifications for tags should also be proposed with this in mind.

core/message-tags-3.3.md
```diff
@@ -0,0 +1,49 @@
...
+
+This specification adds a new capability for general message tags support and
+a prefix for expressing client-to-client tags.
+
+## Motivation
+
+A new capability is required in order to indicate support for client-to-client
+tags. Additionally, some implementations felt constrained to apply filtering
+rules to prevent clients from receiving tags they didn't request.
+
```
jwheare: Suggested alternate wording:

> A new capability is required to allow negotiating support for client-to-client
> tags, as well as any well-formed, unrecognised tag. This removes the need to
> negotiate a capability for each new tag-only specification, and frees servers from
> having to filter tags individually when responding to clients.

core/message-tags-3.3.md
```diff
@@ -0,0 +1,49 @@
...
+client-to-client tag starts with a plus sign (`+`) and otherwise obeys the
+format provided in [IRCv3.2 tags][irctags].
+
+The updated BNF for keys is as follows:
+
+    <key>           ::= [ '+' ] [ <vendor> '/' ] <sequence of letters, digits, hyphens (`-`)>
+
+Client-to-client tags MUST be forwarded on `PRIVMSG` and `NOTICE` events, and
+MAY be forwarded on other events.
+
```
jwheare: Suggested wording change (and perhaps move PRIVMSG/NOTICE sentence above BNF):

```
Client-to-client tags are client-initiated tags that servers MUST attach as-is
to any relevant event relayed to other clients. A client-to-client tag is prefixed
with a plus sign (`+`) and otherwise conforms to the format specified in
[IRCv3.2 tags][irctags].

Client-to-client tags MUST be relayed on `PRIVMSG` and `NOTICE` events, and
MAY be forwarded on other events.

The expected client behaviour of individual client-to-client tags SHOULD be defined
in separate specifications, in the same way as server-initiated tags.
```

core/message-tags-3.3.md
```diff
@@ -0,0 +1,49 @@
...
+### Examples
+
+`@+znc.in/network PRIVMSG user :lol`
+
+`@+displayname=user2 PRIVMSG #channel :"..."`
+
```
jwheare: Improved examples:
```
This section is non-normative

`@+icon=https://example.com/favicon.png :URL_Titler!bot@example.com PRIVMSG #channel :A bot that provides titles and favicons for URLs`

`@+example.com/foo :irc.example.com NOTICE #channel :A vendor-prefixed client-to-client tagged message`
```

jwheare:
Suggest adding a security implications section:

```
```

jwheare:
```
### Security implications
```

jwheare:
Closing in favour of #251 which contains my suggestions.

PR #238, opened 2016-02-23
3.3 intents

Aerdan:
Something something talk about something productive.


kaniini:
I no longer find intents to be a good solution, they are an overengineered framework that essentially came about for replacing CTCP ACTION.  Feel free to take my name off this specification.

extensions/intents-3.3.md
```diff
@@ -0,0 +1,93 @@
...
+
+The `intent` tag carries the sender's intentions on how the message should be interpreted.
+The following intentions SHOULD be supported:
+
+| Intention   | CTCP? | Additional tag      | Meaning                                        |
+|-------------|-------|---------------------|------------------------------------------------|
+| `action`    | Yes.  | N/A                 | This message is an 'action'.                   |
+| `announce`  | No.   | N/A                 | Mark this message as an announcement.          |
+| `error`     | No.   | N/A                 | This message is an 'error' notification.       |
+| `log`       | No.   | `loglevel=<string>` | This message contains logging information.     |
```
SaberUK: As I have said before I don't think anything that requires an additional tag like log, name, and redirect should be part of intents and should instead go in their own specification under the proposed `message-tags` cap. In my view intents should be for basic semantic labels that require no extra information.

Aerdan: As I've said before, I don't think they're significant enough to deserve their own specification, particularly since they fit within the umbrella of client-side intended behaviour.

grawity:
Does `announce` mean a staff announcement (highlighted), or a news announcement (the opposite)?

Overall, this seems like an assorted pile of tags with unrelated purposes. Some intents change the message's presentation, others request some action (or inaction), yet others only attach metadata. For example, what's the point of having `response` an intent if it could very well be just a single end-to-end tag?

SaberUK:
It would also be good if you could adapt it to use the template in #229.

Aerdan:
@grawity: `response` is pending some specification which implements some sort of message ID tag, though honestly in that case it should be tied to that spec rather than an intent. `announce` is for generic announcements, both from server staff and from channel operators.

@SaberUK: I can do that, sure, but I'd prefer to wait until #229 is actually merged.

extensions/presentation-tags-3.3.md
```diff
@@ -0,0 +1,103 @@
...
+The `<Target>` must be a valid channel name.
+
+#### The `log` Tag
+
+The `log` tag MAY carry a string indicating a severity. It is intended
+primarily for automatic reporting of server state. The format is as follows:
+
+    log=<String>
+
+#### The `name` Tag
```
DarthGandalf: &lt;bikeshed>`as_nick`?&lt;/bikeshed>

extensions/presentation-tags-3.3.md
```diff
@@ -0,0 +1,103 @@
...
+
+`@intent=announce,noreply PRIVMSG #example :Please be advised that this is a test.`
+
+`@intent=noreply,announce;redirect=#example PRIVMSG user :Follow the rules at...`
+
+`@log=info NOTICE user :Foobar opered up.
+
+### Legacy translation
+
+The `intent=action`, `intent=noreply`, and `redirect` tags SHOULD be
```
DarthGandalf: What about the other direction?

extensions/presentation-tags-3.3.md
```diff
@@ -0,0 +1,103 @@
...
+translated as follows for legacy clients:
+
+* `intent=action` → `\x01ACTION ...`
+* `intent=noreply` becomes a `NOTICE`.
+* `redirect=<target>` → `NOTICE user :[<target>] ...`
+
+## Security Considerations
+
+There are no security considerations that implementations should note
```
DarthGandalf: "In order to prevent abuse" from `name` sounds like a security consideration for me.
Also, `@redirect=#a nick!u@h PRIVMSG #b :foo` when used to message on channel even when banned there.

Aerdan: Would only really work if users in `#b` are also in `#a`, but sure, the server should probably filter such cases,

highflyer77:
I think that it would be much easier if the name tag is only allowed to use names that are grouped to the users account.  

Then, there would be no "masquerading" language necessary if the server is ONLY allowed to accept name tags that the user has grouped to the account, thus being ones that they own.  

And therefore, there would be no security implications involved about names being "misused", and additionally therefore no need to "visually" associate the actual handle with a "masqueraded" one.  

Users SHOULD be allowed to use ANY nicks that are associated/grouped with their account at any time, or a method to  `REGAIN`  the additional ones that they own in order to use those.  

If this is agreed, then, I believe that "masquerading" should be stricken from the language.
Or, IF users are allowed to use nicks that are NOT grouped to their account THEN the language of "masquerading" would be included as such that the name being used would be the same as using any other unregistered nick.

#### To Summarize:
Only allowing grouped nicks associated with the account would relieve any security concerns and simplify administration.

Aerdan:
Grouping is an implementation-defined mechanism, so I don't feel that it's appropriate to incorporate it into the specification other than as one way to ameliorate the security implications that `name` offers.

jwheare:
Closing via #270

PR #239, opened 2016-03-01
cap-notify: Clarify behaviour for implicit support

### LINKED ISSUE #217: Clarify cap-notify behaviour in CAP LS/REQ handshake
Issue #205 refers to an issue where a server treated a REQ for `cap-notify` after initiating `CAP LS 302` as an error and sent a `NAK` for the entire REQ line. That issue refers to explicitly stating that re-requesting CAPs should be allowed, but there's a further issue that servers and clients don't really know what to do about the implicit nature of `cap-notify`.

A few options:

* With LS 302, `cap-notify` MUST appear in LS and MAY/MUST be REQed (redundant)
* With LS 302, `cap-notify` MUST NOT appear in LS and MUST not be REQed (error-prone)
* With LS 302, `cap-notify` MAY appear in LS and if REQed, MUST be treated as a no-op (safe)
* `cap-notify` MUST ONLY appear in LS for CAP 3.1 (i.e. no LS 302) (arguably pointless extra logic for the server)
* Get rid of the `cap-notify` CAP entirely (makes it impossible to request in 3.1, arguably not a big deal)

Given `cap-notify` MUST be enabled in a client that supports LS 302, its existence as a CAP is only to allow servers and clients to support it in 3.1 mode. As it's a 3.2 spec, this seems like an unlikely combination.

Looking at client implementations:

* Colloquy will send a REQ for `cap-notify` immediately after requesting `LS` and without waiting for a response [[1]](https://github.com/colloquy/colloquy/blob/a094cfd37a5a53c8401074dddbb8f7e598b49724/Chat%20Core/MVIRCChatConnection.m#L1065)
* Hexchat will send a REQ for `cap-notify` if it's seen in LS [[2]](https://github.com/hexchat/hexchat/blob/4362085847f359ed13df6f8f488a06eb52ecd767/src/common/inbound.c#L1822)
* IRCCloud will never REQ `cap-notify`, (due to the error with mammon referenced in #205), relying on it being implicit with LS 302. Prior to that, it followed the same process as Hexchat.

All these clients support CAP NEW and DEL, regardless of how the handshake goes down.

Since we probably can't remove the CAP entirely (which would probably be the cleanest solution), the spec should be clarified to say that the CAP is entirely optional to show in response to LS 302 and if REQed, it should not be an error, but simply a no-op.

This line from the spec:
> The `cap-notify` capability MUST be implicitly enabled if the client requests `CAP LS` with a version of 302 or newer (`CAP LS 302`), as described in the capability negotiation 3.2 specification. Further, the `cap-notify` capability MAY NOT be disabled if the client requests `CAP LS` with a version of 302 or newer.

could be followed up with:

> In this situation, the `cap-notify` capability MAY appear in the `LS` response from the server. Additionally a client MAY request the capability with `REQ`, which MUST be accepted by the server without side effects.

If we were writing this before any implementations, I would change the MAYs to MUST NOTs, but it's probably too late for that. I'd personally favour SHOULD NOT, but I expect that will be met with resistance from people who don't want to feel like they've implemented something incorrectly.

jwheare:
Closes #217

### DIFF
--- a/extensions/cap-notify-3.2.md
+++ b/extensions/cap-notify-3.2.md
@@ -19,14 +19,21 @@ When enabled, the server MUST notify clients about all new capabilities
 and about existing capabilities that are no longer available via the messages
 specified in this document.
 
+## Implicit vs explicit support
+
 The `cap-notify` capability MUST be implicitly enabled if the client requests
 `CAP LS` with a version of 302 or newer (`CAP LS 302`), as described in the
 [capability negotiation 3.2 specification](../core/capability-negotiation-3.2.html).
 Further, the `cap-notify` capability MAY NOT be disabled if the client requests
 `CAP LS` with a version of 302 or newer.
 
-Legacy clients which request the `cap-notify` capability MAY disable the
-capability at any time.  This to ease the adaptation of new features.
+When implicitly enabled via this mechanism, servers MAY list the `cap-notify` capability
+in `CAP LS` responses. Additionally client MAY request the capability with `CAP REQ`,
+and capable servers MUST accept and `CAP ACK` the request without side effects.
+
+Clients that do not support `CAP LS` version 302 MAY request the `cap-notify` capability
+explicitly. Such clients MAY disable the capability at any time.  This to ease the
+adaptation of new features.
 
 ## Subcommands
 
@@ -96,3 +103,6 @@ If capability negotiation 3.2 was used, extensions listed MAY contain values.
   `NEW`/`DEL` subcommands, but had it in the examples anyway.
 
 * Earlier version of this spec did not mention whether `NEW` and `DEL` can have values or not.
+
+* Earlier versions of this spec were missing clarification of client and server behaviour
+  when the capability is implicitly enabled with `CAP LS` version 302 or newer.


extensions/cap-notify-3.2.md
```diff
@@ -19,14 +19,21 @@ When enabled, the server MUST notify clients about all new capabilities
...
 The `cap-notify` capability MUST be implicitly enabled if the client requests
 `CAP LS` with a version of 302 or newer (`CAP LS 302`), as described in the
 [capability negotiation 3.2 specification](../core/capability-negotiation-3.2.html).
 Further, the `cap-notify` capability MAY NOT be disabled if the client requests
 `CAP LS` with a version of 302 or newer.
 
-Legacy clients which request the `cap-notify` capability MAY disable the
-capability at any time.  This to ease the adaptation of new features.
+When implicitly enabled via this mechanism, servers MAY list the `cap-notify` capability
+in `CAP LS` responses. Additionally client MAY request the capability with `CAP REQ`,
```
SaberUK: What about `CAP LIST`?

jwheare:
Reopened this from my personal fork, and will be removing this branch from the main repo #240

PR #245, opened 2016-03-08
Rework metadata notifications

attilamolnar:
Placeholder PR. This will not be included as errata. Needs syncing up with #250 in a new version later.

### DIFF
--- a/core/metadata-3.2.md
+++ b/core/metadata-3.2.md
@@ -6,6 +6,10 @@ copyrights:
     name: "Kiyoshi Aman"
     period: "2012"
     email: "kiyoshi.aman@gmail.com"
+  -
+    name: "Attila Molnar"
+    period: "2015-2016"
+    email: "attilamolnar@hush.com"
 ---
 ## Introduction
 
@@ -94,11 +98,19 @@ It is an error for users to set keys on targets for which they lack
 authorization from the server, and the server MUST respond with
 `ERR_KEYNOPERMISSION` and fail the request.
 
+Servers MAY respond with only an `ERR_METADATARATELIMIT` event and fail the
+request. When a client receives an `ERR_METADATARATELIMIT` event, it SHOULD
+retry the `METADATA SET` request at a later time. If the
+`ERR_METADATARATELIMIT` event contains the OPTIONAL `<RetryAfter>` parameter,
+the parameter value MUST be a positive integer indicating the minimum number
+of seconds the client SHOULD wait before retrying to request.
+
 If the request is successful the server MUST carry out the requested change and
 the response MUST be one `RPL_KEYVALUE` event, representing what was actually
 stored by the server, and one `RPL_METADATAEND` event.
 
-*Errors*: `ERR_METADATALIMIT`, `ERR_KEYINVALID`, `ERR_KEYNOTSET`, `ERR_KEYNOPERMISSION`
+*Errors*: `ERR_METADATALIMIT`, `ERR_KEYINVALID`, `ERR_KEYNOTSET`, `ERR_KEYNOPERMISSION`,
+`ERR_METADATARATELIMIT`
 
 ### METADATA CLEAR
 
@@ -187,7 +199,7 @@ keys per-user; the format in that case MUST be `METADATA=<integer>`, where
 
 ## Numerics
 
-The numerics 760 through 769 MUST be reserved for metadata, carrying the
+The numerics 760 through 770 MUST be reserved for metadata, carrying the
 following labels and formats:
 
 | No. | Label                 | Format                                   |
@@ -201,6 +213,7 @@ following labels and formats:
 | 767 | `ERR_KEYINVALID`      | `<Key> :invalid metadata key`            |
 | 768 | `ERR_KEYNOTSET`       | `<Target> <Key> :key not set`            |
 | 769 | `ERR_KEYNOPERMISSION` | `<Target> <Key> :permission denied`      |
+| 770 | `ERR_METADATARATELIMIT` | `<Target> <Key> [<RetryAfter>]`        |
 
 The `<Visibility>` field for numerics follows the same rules and requirements
 as specified for notifications' visibility.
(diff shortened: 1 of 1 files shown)


attilamolnar:
It has been discussed that this should become `metadata-notify-2` and server implementations can stop advertising the original `metadata-notify` if they wish to make use of the new protocol.

Aerdan:
:-1: on modifying 3.2; this needs to be its own document. I can go ahead and open a PR for 3.3 and we can work this into that, however.

dequis:
Yeah this change is way too big to be just errata. You can't just start giving old clients a rate limit numeric they don't know about.

attilamolnar:
@jwheare asked me to open a PR with it anyway, it's pending a rebase after some of the metadata 3.3 stuff lands

core/metadata-3.2.md
```diff
@@ -120,6 +132,23 @@ using `ERR_KEYNOPERMISSION` with an asterisk (`*`) in the `<Key>` field.
 
 *Errors*: `ERR_KEYNOPERMISSION`
 
+### METADATA SYNC
+
+This subcommand requests the full synchronization of metadata associated with
```
dequis: I think "requests the full synchronization" doesn't describe this command very well. It looks like it actually means "request the next page of results after getting `ERR_METADATASYNCLATER`".

Or is this meant for when `PRIORITY` is set to `request` and clients want to poll all the metadata of a target? Like a bulk get. A "full" synchronization of all keys regardless of priority. I think that wouldn't work well with pagination - the on-subscription notifications would be more limited than the ones requested by this command.

attilamolnar: > I think "requests the full synchronization" doesn't describe this command very well. It looks like it actually means "request the next page of results after getting ERR_METADATASYNCLATER".

It does not mean that, it means what it says.

Or is this meant for when PRIORITY is set to request and clients want to poll all the metadata of a target? Like a bulk get. A "full" synchronization of all keys regardless of priority.

It's a full synchronization of every key that the client is subscribed to.

dequis: Nevermind, I misread the examples and reached the wrong conclusions. Also there's no guarantee that PRIORITY as described in the other ticket will be what is merged with this one. Maybe just delete these three comments.

core/metadata-3.2.md
```diff
@@ -94,11 +98,19 @@ It is an error for users to set keys on targets for which they lack
 authorization from the server, and the server MUST respond with
 `ERR_KEYNOPERMISSION` and fail the request.
 
+Servers MAY respond with only an `ERR_METADATARATELIMIT` event and fail the
+request. When a client receives an `ERR_METADATARATELIMIT` event, it SHOULD
+retry the `METADATA SET` request at a later time. If the
+`ERR_METADATARATELIMIT` event contains the OPTIONAL `<RetryAfter>` parameter,
+the parameter value MUST be a positive integer indicating the minimum number
+of seconds the client SHOULD wait before retrying to request.
+
```
jwheare: typo "before retrying *the* request"

core/metadata-3.2.md
```diff
@@ -120,6 +132,23 @@ using `ERR_KEYNOPERMISSION` with an asterisk (`*`) in the `<Key>` field.
 
 *Errors*: `ERR_KEYNOPERMISSION`
 
+### METADATA SYNC
+
+This subcommand requests the full synchronization of metadata associated with
+the given target. It is only to be used when `metadata-notify` is enabled
+by the client.
+
```
jwheare: I think this can be allowed even without `metadata-notify` to allow users to manually check metadata, even if their client doesn't enable notifications, just as they can with WHO.

Aerdan:
Honestly, this whole thing is a complete departure from the original design, and I think that, like my `METADATA PRIORITY` proposal this has a real danger of being overengineered. I don't know if there's a way to design a mechanism that flows well with the original design, but I don't think this is it.

dequis:
@Aerdan What's described in this diff is just rate limiting and postponed synchronization, I have no idea what danger of overengineering you see.

And what's wrong with a complete departure from the original design of metadata notifications? Flooding clients with all metadata keys when they join a channel clearly wasn't a great idea.

If anything, this PR is a bit too conservative with the changes, allowing the old behavior of servers sending clients metadata notifications on join and making ERR_METADATASYNCLATER optional  instead of requiring clients to do bulk requests if they care.

attilamolnar:
Experimental implementation here: https://github.com/attilamolnar/inspircd/tree/master%2Bmetadata

core/metadata-3.2.md
jwheare: Disregard, that's what `METADATA LIST` is for.

jwheare: That said, LIST doesn't let you get the metadata for an entire channel member list in one go.

core/metadata-3.2.md
```diff
@@ -201,6 +253,8 @@ following labels and formats:
 | 767 | `ERR_KEYINVALID`      | `<Key> :invalid metadata key`            |
 | 768 | `ERR_KEYNOTSET`       | `<Target> <Key> :key not set`            |
 | 769 | `ERR_KEYNOPERMISSION` | `<Target> <Key> :permission denied`      |
+| 770 | `ERR_METADATARATELIMIT` | `<Target> <Key> [<RetryAfter>]`        |
+| 771 | `ERR_METADATASYNCLATER` | `<Target> [<RetryAfter>]`               |
```
jwheare: These clash with the numerics in #250

The test implementation in inspircrd uses:

`ERR_METADATASYNCLATER = 774`

https://github.com/inspircd/inspircd/commit/3d52a33d3c650292b33825ce3458456790149caa#diff-410482bb2c797687c7731b0efe472abeR38

We should also set

`ERR_METADATARATELIMIT = 775`

jwheare: I'd also propose that `ERR_METADATARATELIMIT` should include the value, to make it easier for clients to schedule a retry. Since the value would need to be in the last position, an optional `RetryAfter` would need to be encoded as *.

e.g.

```
# Retry in 5 seconds
:server 775 client_nick target key 5 :value
# No retry value
:server 775 client_nick target key * :value
```

jwheare:
Closing in favour of #339

PR #249, opened 2016-03-10
STS: Improved sections for Client/Server/Security implementation considerations

jwheare:
This can replace the STS spec portions of #232.

### DIFF
--- a/core/sts-3.3.md
+++ b/core/sts-3.3.md
@@ -154,6 +154,8 @@ user recourse" section.
 
 ## Server implementation considerations
 
+This section is non-normative.
+
 ### Policy expiry time
 
 IRCd configurations and network administrators should consider whether expiry times
--- a/core/sts-3.3.md
+++ b/core/sts-3.3.md
@@ -23,11 +23,15 @@ MUST only connect to the server securely (using TLS, aka. SSL), and the port
 number to use for it.
 
 Clients MUST use sufficient certificate verification to ensure the connection
-is secure without any errors.
+is secure without any errors. This may involve validation against a system root
+certificate bundle or a user-specified trust root.
+
+In compliant server software, this capability MUST be made available as an optional
+configuration setting. Server administrators may have valid reasons not to enable it.
 
 ## Details
 
-The capability value, which MUST always be present, is a comma (`,`) separated
+When enabled, the capability has a REQUIRED value: a comma (`,`) separated
 list of tokens. Each token is a key which may have a value attached. If there
 is a value attached, the value is separated from the key by a `=`. That is,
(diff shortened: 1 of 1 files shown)


core/sts-3.3.md
```diff
@@ -23,11 +23,15 @@ MUST only connect to the server securely (using TLS, aka. SSL), and the port
...
+is secure without any errors. This may involve validation against a system root
+certificate bundle or a user-specified trust root.
+
+In compliant server software, this capability MUST be made available as an optional
+configuration setting. Server administrators may have valid reasons not to enable it.
 
 ## Details
 
-The capability value, which MUST always be present, is a comma (`,`) separated
+The capability value, which MUST always be present when enabled, is a comma (`,`) separated
```
attilamolnar: I think you mean the capability should be present when enabled, not only its value.

attilamolnar:
LGTM

jwheare:
TODO clarify how this is reconciled with the "no user recourse" section.

core/sts-3.3.md
```diff
@@ -96,28 +100,94 @@ Servers MUST send this key to non-securely connected clients.
...
+Fixed expiry times will involve a dynamic `duration` value being calculated on each
+connection attempt.
+
+Which approach to take will depend on a number of considerations. For example, a server
+may wish their STS Policy to expire at the same time as their domain certificate.
+Alternatively, a server may wish their STS policy to last for as long as possible.
+
+Server implementations should consider employing a value of `duration=0` in their default
+configurations. This will require server administrators to deliberately set an expiry
+according to their specific needs rather than an arbitrary generic value.
```
attilamolnar: Why advertise a policy if not explicitly configured by default?

attilamolnar: Why advertise a policy if not explicitly configured by default?

jwheare: This is meant in e.g. the default example config, and would probably also be commented out. It's just to avoid example configs having arbitrary values that people will blindly choose instead of deciding what suits them.

core/sts-3.3.md
```diff
@@ -125,6 +195,27 @@ client to the server until the policy expires. At that time, the client will
...
+
+### Denial of Service
+
+STS could be used to inflict Denial of Service (DoS) on IRC servers in a number of ways.
+Some non-exhaustive examples include:
+
+* An attacker can inject an STS policy into an insecure connection. This causes the
+client to reconnect on a secure port, which may fail persistently if the server isn't
+listening for secure connections. This can be mitigated in clients by allowing for
+STS policy rejection as described in the "Client implementation considerations" section.
```
attilamolnar: Note that this can only happen persistently if the attacker is present at all times and is actively stripping the redirect, it will not be persistent otherwise.

jwheare:
Last chance for comments before merge.

PR #250, opened 2016-03-16
Add rewritten metadata specification

attilamolnar:
TODO:

* [ ] Update numerics to match [inspircd implementation](https://github.com/inspircd/inspircd/blob/3d52a33d3c650292b33825ce3458456790149caa/src/modules/m_ircv3_metadata.cpp#L35-L39)
* [ ] Rewrite to include the parts of 3.2 that still apply, clarifying if necessary and mention that 3.2 is deprecated (see #279)
* [ ] Add a requirement that clients should expect to receive notifications even if they haven't subscribed to any key, e.g. if a server needs to update a client's own metadata, e.g. on oper, or auth, or if an oper sets a key on another user.
* [ ] Specify a cap value that advertises a limit of key and/or value names
* [ ] Drop the 3.3 versioning. (see #265)
* [ ] #278 Additional non-normative sections
* [ ] Remove RPL_METADATAEND replies, clients should use labeled-response
* [ ] Allow async processing of requests
* [ ] Change ERR_KEYINVALID to drop the message reason and include the invalid key as the last arg since it could contain spaces
* [ ] Clarify ERR_KEYNOPERMISSION to expand "not settable by the requesting user" with "or otherwise disallowed by the server"
* [ ] Allow hyphens in key names

Discuss:

* Rename CAP to just `metadata` (no need for version number, values are about more than notifications)

### DIFF
--- a/extensions/metadata-3.3.md
+++ b/extensions/metadata-3.3.md
@@ -1,0 +1,423 @@
+---
+title: IRCv3.3 Metadata
+layout: spec
+work-in-progress: true
+copyrights:
+  -
+    name: "Attila Molnar"
+    period: "2015-2016"
+    email: "attilamolnar@hush.com"
+---
+
+## Notes for implementing work-in-progress version
+
+This is a work-in-progress specification.
+
+Software implementing this work-in-progress specification MUST NOT use the
+unprefixed `metadata-notify-2` capability name. Instead, implementations SHOULD
+use the `draft/metadata-notify-2` capability name to be interoperable with other
+software implementing a compatible work-in-progress version.
+
+The final version of the specification will use an unprefixed capability name.
+
+## Description
+
+This specification extends the existing [metadata v3.2 specification](http://ircv3.net/specs/core/metadata-3.2.html).
+
+## Metadata notify v2
+
+This document describes the `draft/metadata-notify-2` capability.
+In contrast to `metadata-notify`, where clients get METADATA events for all
+keys, clients using the new capability need to explicitly subscribe to metadata
+keys they are interested in.
+
+### Capability value
+
+The value of the metadata notify capability introduced in this document MUST
+have the following format:
+
+    [<anything>,]maxsub=<N>[,<anything>]`
+
+`<N>` is an integer indicating the maximum number of keys a client is allowed
+to be subscribed to.
+
+## Mechanics
+
+A client can either be subscribed to a key, or not subscribed to it.
+
+The server MUST allow a client to subscribe to any valid keys, even to
+privileged keys when the client has no privilege to access that key at
+the time of subscription.
+
+If a client is subscribed to a metadata key and has adequate privileges to get
+notifications about that key then it gets METADATA events about the key as
+described [here](http://ircv3.net/specs/core/metadata-3.2.html#metadata-notifications).
+
+If a client is not subscribed to a metadata key then it will not get METADATA
+events about it, however the client can use `METADATA GET`, `METADATA LIST` or
+other means available to obtain the value of the key.
+
+By default, the client is not subscribed to any keys.
+
+Managing subscriptions are possible with the protocol described below.
+
+## Subcommands
+
+Clients who negotiated the extension described in this document can use the
+following new subcommands.
+
+### METADATA SUB
+
+This subcommand is used to subscribe to metadata keys.
+
+Syntax:
+
+    METADATA * SUB <key1> [<key2> ...]
+
+The server MUST reply with zero or more numerics of the following types in any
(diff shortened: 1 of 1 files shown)


extensions/metadata-3.3.md
```diff
@@ -0,0 +1,389 @@
...
+It is pointless to use `metadata-notify-2` and `metadata-notify` (as described
+in the [metadata v3.2 specification](http://ircv3.net/specs/core/metadata-3.2.html))
+together.
+
+This is because `metadata-notify` implicitly subscribes the client to all keys,
+while `metadata-notify-2` requires the client to tell the keys it wants to
+subscribe to.
+
+Servers and clients MAY support both of the mentioned extensions, but MUST NOT
+negotiate both of them at the same time in the same connection.
```
Aerdan: Depending on the order, I'd allow the client to either request both (`metadata-notify` and then `...-2`) or allow `...-2` and reject the other.

Aerdan:
The key subscription needs to be disclosed to the client somewhere, e.g. via a `METADATA` key on `RPL_ISUPPORT`.

attilamolnar:
> The key subscription needs to be disclosed to the client somewhere, e.g. via a METADATA key on RPL_ISUPPORT.

I don't exactly understand what you mean. Can you elaborate? If this is about advertising the availability of the new subcommands, the presence of `metadata-notify-2` implies support for them.

Aerdan:
Sorry, I missed a word there; the key subscription *limit* needs to be disclosed.

attilamolnar:
I agree that should be advertised, I think it should be in the cap's value because we can't change the format of the value of the METADATA ISUPPORT token after release and adding a new ISUPPORT token is not necessary if there is a cap already.

Aerdan:
I disagree; the `METADATA` token could be changed to use the same key/value format other `RPL_ISUPPORT` tokens use (`TARGMAX`, for instance), and clients should be able to figure out which version is which trivially.

extensions/metadata-3.3.md
attilamolnar: Any good reason to do it that way? I've deliberately tried to avoid adding special cases like "if both are present, accept only one" or "if the order is X Y accept if Y X then reject". Servers must accept all caps in a CAP REQ or reject them all per the cap spec.

jwheare:
Changing the ISUPPORT format only if the CAP is negotiated seems reasonable.

extensions/metadata-3.3.md
Aerdan: By 'order' here I mean specifically each capability is in a `REQ` of its own; it doesn't make sense to reject `...-2` if `metadata-notify` was negotiated previously. 

>Clients MAY request `metadata-notify-2` after a successful `metadata-notify`.

...may be the best way to phrase it.

attilamolnar:
> I disagree; the METADATA token could be changed to use the same key/value format other RPL_ISUPPORT tokens use (TARGMAX, for instance), and clients should be able to figure out which version is which trivially.

Clients released before this specification can't figure out "which version is which" because they expect a single integer as the value of the ISUPPORT token.

> Changing the ISUPPORT format only if the CAP is negotiated seems reasonable.

Clients can request metadata-notify-2 after ISUPPORT (say after they receive a CAP NEW) and they won't get the value that way. Also existing servers have a cached ISUPPORT and send that to all clients. Having per client ISUPPORT is something they'd have to add support for separately for this case and it has no benefit over advertising as a cap value. (In fact, the cap value has benefits as it allows changing the limit on the fly for example).

Aerdan:
>Clients released before this specification can't figure out "which version is which" because they expect a single integer as the value of the ISUPPORT token.

This is a valid criticism.

>Changing the ISUPPORT format only if the CAP is negotiated seems reasonable.

This is a *terrible* idea.

extensions/metadata-3.3.md
attilamolnar: Any good reason to deliberately allow that?

Aerdan: ...Well, if a client author's developing support for IRCv3.3 metadata, they might have already negotiated `metadata-notify` by default. It's not a particularly good reason, though, so I won't insist on it.

attilamolnar:
What's the status of this? Do we need implementations for this to move forward?

jwheare:
Yes, and I think it should be merged with #245 so we can point to a single PR from the tracker issues.

attilamolnar:
Somewhat experimental implementation here: https://github.com/attilamolnar/inspircd/tree/master%2Bmetadata

attilamolnar:
More ideas: Remove the `End of metadata` numeric sent in reply to every METADATA command, clients that want to track when the reply ends should use labeled-response. Also allow the server to asynchronously process any METADATA request (but still in order).

jwheare:
Closing in favour of #339

Issue #256, opened 2016-04-27
STS suggestions / improvements


kaniini:
Two things:

1. The spec should use "non-TLS" in places where it uses "non-secure" right now, as the spec is specific to TLS upgrades.

2. Say you have two IRCds on a single IP + DNS name, production IRCd uses STS while development IRCd doesn't.  Now you cannot connect to development IRCd via plaintext.  It should probably be documented that STS imposes policy for the entire DNS name.

jwheare:
1) While much of the implementation details will be TLS specific, I think the spec should be as agnostic about the specifics of the underlying transport as possible. 2) is addressed in #257

PR #257, opened 2016-04-28
STS: Add considerations for offering multiple IRC servers on the same domain

attilamolnar:


attilamolnar:
Partially addresses #256.

core/sts-3.3.md
```diff
@@ -179,6 +179,21 @@ Server implementations should consider using a value of `duration=0` in their ex
 configurations. This will require server administrators to deliberately set an expiry
 according to their specific needs rather than an arbitrary generic value.
 
+### Offering multiple IRC servers at alternate ports on the same domain
+
+The STS policy is imposed for an entire domain name. This means that mixing
```
kaniini: Maybe make this section explicitly non-normative?

jwheare: The whole parent section states it's explicitly non-normative.

PR #258, opened 2016-04-29
Clarify where cap values may be shown

digitalcircuit:
For ```CAP LS 302```, clarify that ```<name>[=<value>]``` only apply to ```CAP LS``` and ```CAP NEW``` replies.

For example, this is valid:
```
Server: CAP * LS :sasl=EXTERNAL,FOO,PLAIN batch cap-notify
Client: CAP REQ :sasl
Server: CAP modernclient ACK :sasl
```

While this is **not** valid:
```
Server: CAP * LS :sasl=EXTERNAL,FOO,PLAIN batch cap-notify
Client: CAP REQ :sasl
Server: CAP modernclient ACK :sasl=EXTERNAL,FOO,PLAIN
```

This simplifies logic for clients by ensuring new values don't need parsed and tracked in ```CAP ACK``` and ```CAP NAK``` replies.

### DIFF
--- a/core/capability-negotiation-3.2.md
+++ b/core/capability-negotiation-3.2.md
@@ -42,7 +42,7 @@ If no other arguments are present, version 3.1 is assumed.
 ### Capabilities with values
 
 Servers MAY specify additional data for each capability using the
-`<name>[=<value>]` format.
+`<name>[=<value>]` format for `CAP LS` and `CAP NEW`.
 
 The meaning of the value (if present) depends on the capability in question.
 
@@ -91,3 +91,9 @@ reply:
     Client: CAP LIST
     Server: CAP modernclient LIST * :example.org/example-cap example.org/second-example-cap account-notify
     Server: CAP modernclient LIST :invite-notify batch example.org/third-example-cap
+
+## Errata
+
+A previous version of this specification did not specify when to use
+the `<name>[=<value>]` format.  This was clarified to limit capability
+values to `CAP LS` and `CAP NEW` replies.


DanielOaks:
Add an errata section mentioning the change. I think "only the capability names" should be removed since it makes the sentence a bit complex. I'd just say "`CAP ACK` and `CAP NAK` MUST NOT contain any values".

digitalcircuit:
I agree with simplifying the sentence, and I've added the errata.  Thanks for the guidance!

DanielOaks:
Awesome, looks good to me!

grawity:
Looks good except for "Replies to CAP ACK and CAP NAK…" – there _aren't_ any replies to those.

digitalcircuit:
@grawity Good catch!  I meant the replies the server sends, rather than something the client must reply to.  I've corrected that with the most recent update.

core/capability-negotiation-3.2.md
```diff
@@ -42,7 +42,8 @@ If no other arguments are present, version 3.1 is assumed.
 ### Capabilities with values
 
 Servers MAY specify additional data for each capability using the
-`<name>[=<value>]` format.
+`<name>[=<value>]` format for `CAP LS` and `CAP NEW`.  `CAP ACK` and
+`CAP NAK` replies MUST NOT contain any values.
```
attilamolnar: The last sentence incomplete and unnecessary so please remove it.

ChickenNuggers: How is it incomplete? It states that the `CAP ACK` and `CAP NAK` shouldn't contain values, which is the entire reason for this pull request.

attilamolnar: It is incomplete because it does not contain all cases when the reply must not contain values. The cases where values are allowed are already listed, so in all other cases values are not allowed.

digitalcircuit: Good catch, as this could become a maintenance burden should other non-value-accepting replies be added.  In order to reinforce that *only* `CAP LS` and `CAP NEW` may contain values without explicitly listing, would this be acceptable?
```md
Servers MAY specify additional data for each capability using the
`<name>[=<value>]` format for `CAP LS` and `CAP NEW`.  Other replies
MUST NOT contain any values.
```

If it's good, I'll update the Errata, too.

attilamolnar: That's not ok because later we may wish to add values to some other place.

digitalcircuit: Alright, that makes sense.  I was concerned some servers might simply do `<name>=<value>` in all responses, but that may be an unfounded worry.

```md
 Servers MAY specify additional data for each capability using the
`<name>[=<value>]` format for `CAP LS` and `CAP NEW`.
```

Then, for the Errata, keep a note about the situation as it is right now.  This should more clearly inform to anyone whom originally implemented values in all replies:
```md
## Errata

A previous version of this specification did not specify when to use
the `<name>[=<value>]` format.  This was clarified to limit capability
values to `CAP LS` and `CAP NEW` replies.

Servers currently including values in `CAP ACK` and `CAP NAK` replies
MUST modify these replies to not include values.
```

Alternatively, should the Errata avoid mentioning them as well?

dequis: Just not mention those at all, unless you happen to know about servers that go out of their way to have a `CAP ACK` payload that doesn't match what was `REQ`'d...

attilamolnar: > Alternatively, should the Errata avoid mentioning them as well?

Yes. The second paragraph from the errata section you show should be dropped.

digitalcircuit: At the moment, I do not know of any servers that do this.  My concern was more for client-side parsing - if any servers offer a `CAP ACK` payload regardless of whether or not it differs, values would need to be discarded, e.g. ```if reply.contains('=') { reply.split('=')[0] […] }```.

That said, given the low likelihood of this being an issue (the examples encourage implementing the right way), it's probably less overhead to reach out to the authors of any IRC server who happen to have implemented it incorrectly.

I'll update the commit to not mention `CAP ACK` or `CAP NAK` at all.

attilamolnar:
LGTM

attilamolnar:
Thanks!

PR #261, opened 2016-05-06
STS: Document relationship with STARTTLS and cap tls

attilamolnar:


attilamolnar:
Last chance for comments before merge.

jwheare:
As mentioned on the tracker, I thought it was worth mentioning some downsides of the `tls` cap specifically: it doesn't help for people with bookmarked insecure ports and its recommendations around making the upgrade sticky aren't well fleshed out.

attilamolnar:
> it doesn't help for people with bookmarked insecure ports

Covered implicitly by the "This specification mandates that clients must only attempt secure connections to a server which has a policy..." bullet point.

> and its recommendations around making the upgrade sticky aren't well fleshed out.

Covered by the "This specification is not merely a hint but requires comforming clients to switch to a secure connection." bullet point.

attilamolnar:
I've added more text to cover the bookmarked server case as it is a good example worth pointing out.

attilamolnar:
Typos fixed

PR #263, opened 2016-07-10
Add a `certificate-transparency` key to sts for requiring CT validation

kylef:
#### Why

> "TLS is strong, but it's not enough. You also need to have confidence that the certificate that you're using to validate that TLS connection represents the right server."

It's possible for an attacker to get a certificate with a hostname matching an IRC server, and then the attacker could use that certificate. It's hard for client to distinguish the attackers certificate from our real certificates. This is possible if a certificate authority makes a mistake, or if their keys get compromised and someone else uses their private keys to sign a certificate.

Certificate transparency attempt to make is harder for an attacker to do this.

#### What

This key allows a server to instruct clients that they SHOULD perform CT validation when connecting to the server. I've worded this as SHOULD so that clients that don't support CT validation do not have to, but it is highly recommended.

#### Resources

- [What is CT?](https://www.certificate-transparency.org/what-is-ct)
- [How it works](https://www.certificate-transparency.org/how-ct-works)
- [Comparison with Other Technologies](https://www.certificate-transparency.org/comparison)
- [RFC 6962](https://tools.ietf.org/html/rfc6962)

### DIFF
--- a/core/sts-3.3.md
+++ b/core/sts-3.3.md
@@ -141,6 +141,12 @@ Servers MUST send this key to non-securely connected clients.
 Servers MAY send this key to securely connected clients, but it will be
 ignored.
 
+### The `certificate-transparency` key
+
+This key indicates that the server supports certificate transparency and
+clients SHOULD require connections to the server to validate the certificate
+logs as per [RFC6962](https://tools.ietf.org/html/rfc6962).
+
 ### Handling disconnection
 
 IRC connections may be long-lived. Connections lasting for more than a month


jwheare:
I don't think this belongs in STS. While I support the CT concept, it's still in its infancy, and building in client support is very complicated.

I don't want to discourage people from implementing STS by making it more complicated than it needs to be. A small paragraph that doesn't explain CT, or how to implement it, and requires a considerable investment in reading and understanding extra documentation, background, and theory before you can act on it seems hostile to client authors. Even as an optional component. I think responsibility for implementing CT should be squarely with clients for now.

It's definitely worth keeping an eye on CT, but I don't think we need to jump ahead of what's happening in the browser landscape here. There's already a *huge* amount of catching up to do in terms of how IRC clients handle invalid SSL. Let's take it a step at a time.

attilamolnar:
I agree with the previous comment. I think the reason it was put into the STS spec is to make the enforce CT flag permanent like the STS policy, however, this is not really explained by the text.

The text being added by is vague in other areas as well: it does not tell when the `certificate-transparency` key should be considered valid (it should not be accepted by clients unless received on a secure connection), when the CT check should be performed (do it only on subsequent connections? pause registration and perform it now, in the middle of the CAP negotiation?) and lacks examples explaining these.

kylef:
Makes sense, thanks for the feedback.

Issue #266, opened 2016-07-15
STS: Support setting policy for multiple hostnames


attilamolnar:
A single server should be able to set the STS policy for multiple hostnames. This is useful when all servers on a network are advertising STS but the servers are on different subdomains. The desirable behavior in this case is that if a client has seen one STS policy from any server on the network then it would not connect non-securely to any other server belonging to the network, not just that single server it received the policy from.

For example server1.example.net should be able to set an STS policy for server2.example.net by sending a `certhosts` key and presenting a certificate which is valid for server2.example.net (either a wildcard cert or one which contains the appropriate hostname).

PR #267, opened 2016-07-15
STS: Add DoS example for untrusted 3rd party DNS aliases

jwheare:
This addresses the irc.kde.org -> irc.freenode.net CNAME case.

Also, remove TODOs:

* This commit covers the motivation for advertising hosts the server has a valid certificate for.
* #266 covers the motivation the the `sub` key
* SNI is now a formal recommendation: https://github.com/ircv3/ircv3-specifications/blob/7d6f6a4/core/sni-3.3.md

### DIFF
--- a/core/sts-3.3.md
+++ b/core/sts-3.3.md
@@ -280,7 +280,7 @@ the "Server implementation considerations" section to avoid inadvertent expiry i
 
 ### Denial of Service
 
-STS could be used to inflict Denial of Service (DoS) on IRC servers in a number of ways.
+STS could result in a Denial of Service (DoS) on IRC servers in a number of ways.
 Some non-exhaustive examples include:
 
 * An attacker could inject an STS policy into an insecure connection that causes clients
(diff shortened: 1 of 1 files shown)


attilamolnar:
I am not sure if mentioning CNAMEs at all is a good idea. The same thing can happen with an A record, etc. It is not specific to CNAMEs.

jwheare:
Removed reference to CNAME

attilamolnar:
Still wrong I think: it says "via DNS alias" but an A (or AAAA) record is not an alias.

attilamolnar:
Can't the "via..." part be dropped?

core/sts-3.3.md
```diff
@@ -290,6 +290,10 @@ an administrator has regained control of their server. This can be mitigated in
 by allowing for STS policy rejection as described in the "Client implementation
 considerations" section.
 
+* A 3rd party host that points to an STS-enabled host via DNS alias, where the 3rd party
+isn't listed on the server's certificate. This configuration would fail certificate validation
+even without STS, but users may be relying on it for unsecure access.
```
attilamolnar: The spec is using the "insecure" and "non-secure" nomenclature for plaintext connections, I suggest we pick one and stick to it instead of adding unsecure as a third one.

Issue #259, opened 2016-05-03
Allow Unicode in nicknames


ttepasse:
RFC 1459 only allows ASCII letters, numerals and some special characters in Nicknames, leaving people from non-anglophone countries at a disadvantage. Using the wealth of human writing is possible in the body of messages, it should be possible in the nicknames too.

SaberUK:
There are existing implementations of this (e.g. InspIRCd's m_nationalchars) but nothing standard. I believe that @DanielOaks was looking into trialling RFC 3454 in @mammon-ircd though.

clokep:
There's also cases of servers improperly implementing rfc1459 vs. strict-rfc1459 (see inspircd/inspircd#1017).

Ideally, wouldn't we want this to match how it is done for channel names?

TingPing:
For what its worth I made a test branch for hexchat supporting rfc3454 though no network implements it afaik to try it.

clokep:
For what it's worth, we just experimented a bit on moznet and things like the zero-width space character is accepted as a valid room name...which shows up as an empty in whois:

(Additionally, there's also a channel which is *just* the prefix, `#`, which is a bit funky.)

![screen shot 2016-05-03 at 2 06 50 pm](https://cloud.githubusercontent.com/assets/517124/14993387/902f2468-1138-11e6-8a60-aabb5efd7f75.png)

dequis:
Relevant reading:

[UTR 36: Unicode Security Considerations](http://unicode.org/reports/tr36/)

[UTS 39: Unicode Security Mechanisms](http://unicode.org/reports/tr39/)

Bitlbee has a 'utf8_nicks' setting, disabled by default and with a small warning about potential breakage in the help text. It doesn't perform any cleanup, deferring that to the IM server (XMPP for example cleans them with the nodeprep/resourceprep profiles of stringprep), but i'd really like to change this.

I haven't heard of clients with big issues when enabling this, just minor visual issues like miscalculating the width when displaying the nicks in a terminal.

MicroDroid:
How are we going to maintain the backward compatibility?
I'd upvote this otherwise

grawity:
In practice, it's probably already compatible because many clients don't care.

MicroDroid:
Hmm, well then this should really be in IRCv3.2, it's awesome

DanielOaks:
> masquerading with characters that look similar (e.g. character 97 "a" looks very similar to character 1072 "а").

[...]

PR #272, opened 2016-09-15
Add document for Unicode casemapping

DanielOaks:
Unicode names have been wanted for a while, and used in experimental implementations as well in certain bouncers integrating with other messaging systems.

This document outlines a method based on [RFC 7700](https://tools.ietf.org/html/rfc7700) which should represent a reasonable, modern solution for those projects that wish to allow unicode characters and casemap them appropriately.

There's previous discussion around this in #259.

This casemapping does not specify any sort of backwards-compatibility measures. Being compatible with clients and servers that cannot correctly handle unicode has been brought up many times during discussions about unicode casemappings. Below outlines some of the most reasonable suggestions, and why I haven't included them in this specification:
### Encoding names so non-rfc7700 servers can accept them

This suggestion revolves around the client encoding nicknames and channel names into currently IRC-friendly characters before it sends them to the server (allowing them to be used on every server out there today). When receiving these encoded names, other unicode-aware clients will decode them to their proper unicode counterpart before displaying them.
#### Pros
- Unicode nicknames and channel names can be used on servers that don't natively support unicode.
- Non-unicode-aware clients can connect to servers that are unicode-aware.
#### Cons
- Non-unicode-aware servers will allow nicknames nicknames that look like duplicates, due to the encoding required and the server not being able to enforce the name preperation described above.
- Possible duplication of names by encoding names that contain only irc-friendly characters (or otherwise, strict client-side checking that is likely to be misinterpreted or misimplemented).

Because of the security implications this would bring up, I think this is an extremely bad idea.
### Encoding names so that non-unicode-aware clients can accept them

This suggestion revolves around the server encoding nicknames and channel names into currently IRC-friendly characters before it sends them to the client (allowing them to be accepted by every client out there). When receiving these encoded names, unicode-aware clients will decode them into their proper unicode counterpart before displaying them.
#### Pros
- We can be assured that any client, including ones that can't do unicode, will be able to accept the names.
#### Cons
- Due to the encoding, these encoded nicknames are not going to be easily readable by non-unicode-aware clients, and are going to appear as a blob of unreadable text.
- Even if we only encode names that contain special characters, that complicates message sending in ways that's likely going to irritate server authors into not implementing this.
- The decoding/encoding required by this (particularly if only certain names are encoded) complicates client programming in ways that is likely to be misimplemented.

I don't think this is required because getting this casemapping widely implemented will take time. By the time this casemapping gets into large enough use to warrant worrying about legacy clients, I think a large majority of the clients currently in use will support unicode names without issues. As well, a number of clients already successfully accept unicode names.

Because of the complexity this process adds and how much I see it as a non-issue at this point this is implemented, I don't think this should happen and believe it's more effort than it's worth I think this measure would just cause more problems than it would solve.


documentation/rfc7700.md
```diff
@@ -0,0 +1,151 @@
...
+* `('\n', 0x0A)` - Used in message framing.
+* `('\r', 0x0D)` - Used in message framing.
+
+
+### Nicknames
+
+Nicknames cannot contain the following characters:
+
+* `(' ', 0x20)` - Separates parameters.
+* `(':', 0x3A)` - Separates trailing parameter.
```
grawity: This & identical entries below seem unnecessary; `:` only has special meaning as the _first_ character of a parameter.

documentation/rfc7700.md
```diff
@@ -0,0 +1,151 @@
...
+* Channel membership prefixes as defined in the `PREFIX` parameter of `RPL_ISUPPORT` - Protocol conflicts.
+* Channel prefixes as defined in the `CHANTYPES` parameter of `RPL_ISUPPORT` - Protocol conflicts.
+
+
+### Usernames
+
+Usernames cannot contain the following characters:
+
+* `(' ', 0x20)` - Separates parameters.
+* `(':', 0x3A)` - Separates trailing parameter.
```
grawity: Same.

documentation/rfc7700.md
```diff
@@ -0,0 +1,151 @@
...
+* `('!', 0x21)` - Separates username from hostname.
+* `('@', 0x40)` - Separates username from hostname.
+
+
+### Hostnames
+
+Hostnames cannot contain the following charactes:
+
+* `(' ', 0x20)` - Separates parameters.
+* `(':', 0x3A)` - Separates trailing parameter.
```
grawity: Same.

documentation/rfc7700.md
```diff
@@ -0,0 +1,151 @@
...
+### Usernames
+
+Usernames cannot contain the following characters:
+
+* `(' ', 0x20)` - Separates parameters.
+* `(':', 0x3A)` - Separates trailing parameter.
+* `(',', 0x2C)` - Used as a separator.
+* `('*', 0x2A)` - Used in mask matching.
+* `('?', 0x3F)` - Used in mask matching.
+* `('!', 0x21)` - Separates username from hostname.
```
attilamolnar: Fix: separates nickname from username

documentation/rfc7700.md
```diff
@@ -0,0 +1,151 @@
...
+### Order of Operations
+
+Names being prepared MUST apply the following rules in the order shown:
+
+1. Preperation using the PRECIS [Nickname profile](https://tools.ietf.org/html/rfc7700#section-2).
+2. Check for restricted characters.
+
+These steps MUST happen in the order shown, or else the restricted characters check may miss characters that should be legitimately restricted.
+
+If a name does contain a restricted character (whether disallowed by the [Nickname profile](https://tools.ietf.org/html/rfc7700#section-2.2) or this document), it MUST be rejected by the server and MUST NOT be propogated to other clients. This is done through the appropriate numeric for the command which tried to set or use the invalid name such as `ERR_ERRONEUSNICKNAME`, `ERR_NOSUCHCHANNEL`, or whichever numeric is most appropriate.
```
grawity: At this point better use the named link syntax:

```markdown
The `rfc7700` casemapping uses the PRECIS [Nickname profile][precis] as defined in [Section 2 of RFC 7700][precis].

[precis]: https://tools.ietf.org/html/rfc7700#section-2
```

documentation/rfc7700.md
DanielOaks: It does, but since it's already disallowed and I could see it causing possible confusion with libraries that split parameters strangely, figured it was better to disallow it. If we figure it's not required I can definitely remove it though.

grawity: Might as well forbid `:` in privmsgs, topics, etc. A library blindly splitting on `:` is not "strange", it's outright buggy.

DanielOaks: Hmm, that's fair. In that case, I can just note that the first letter of one can't be `:`? (since if i.e. a nickname started with `:` then you wouldn't be able to use it in normal messages)

documentation/rfc7700.md
DanielOaks: Cool, I changed the other links above to use the named link syntax since they were all referring to the same URL. The link here and the link to rfc7700 up the top differ from the others since they're linking to different sections (and the url's only replicated once throughout the doc).

documentation/rfc7700.md
grawity: Yeah, it should be fine under the "first character" list. (For channels it's already implied by CHANTYPE.)

documentation/rfc7700.md
```diff
@@ -0,0 +1,153 @@
...
+
+Nicknames cannot contain the following characters:
+
+* `(' ', 0x20)` - Separates parameters.
+* `(':', 0x3A)` - Separates trailing parameter.
+* `(',', 0x2C)` - Used as a separator.
+* `('*', 0x2A)` - Used in mask matching.
+* `('?', 0x3F)` - Used in mask matching.
+* `('.', 0x2E)` - Denotes a server name.
+* `('!', 0x21)` - Separates username from hostname.
```
attilamolnar: Same as previously

documentation/rfc7700.md
```diff
@@ -0,0 +1,164 @@
...
+
+* `(':', 0x3A)` - Separates trailing parameter.
+
+
+### Hostnames
+
+Hostnames cannot contain the following charactes:
+
+* `(' ', 0x20)` - Separates parameters.
+* `(':', 0x3A)` - Separates trailing parameter.
```
attilamolnar: IPv6 IPs need `:` in the hostname

attilamolnar:
@DanielOaks Could you add some examples, particularly ones that illustrate how comparisons work?

documentation/rfc7700.md
jwheare: They even technically can have it as the first character. Which is a bit problematic for 352 RPL_WHOREPLYand 311 RPL_WHOISUSER.

attilamolnar: Servers add a `0` prefix to IPv6 IPs beginning with `:` so that's not a problem.

documentation/rfc7700.md
```diff
@@ -0,0 +1,164 @@
...
+### Channel Names
+
+Channel names must start with a channel prefix as defined in the `CHANTYPES` parameter of `RPL_ISUPPORT` (005).
+
+Channel names cannot contain the following characters:
+
+* `(' ', 0x20)` - Separates parameters.
+* `(',', 0x2C)` - Used as a separator.
+* `('*', 0x2A)` - Used in mask matching.
+* `('?', 0x3F)` - Used in mask matching.
```
jwheare: Is mask matching in channel names a thing? `*` and `?` are valid channel characters at the moment, this seems overly restrictive.

grawity:
For clarification: Does this affect only comparisons or display as well? If it affects display, does the PRECIS case-folding rule mean that it's impossible to use mixed-case nicknames (since they get mapped to lowercase)?

attilamolnar:
@grawity As I understand it only affects comparisons, if adopting it meant losing upper case characters in nicks then it would be a step backwards.

documentation/rfc7700.md
```diff
@@ -0,0 +1,164 @@
...
+* `('0', 0x30)` - Disallowed.
+* `('1', 0x31)` - Disallowed.
+* `('2', 0x32)` - Disallowed.
+* `('3', 0x33)` - Disallowed.
+* `('4', 0x34)` - Disallowed.
+* `('5', 0x35)` - Disallowed.
+* `('6', 0x36)` - Disallowed.
+* `('7', 0x37)` - Disallowed.
+* `('8', 0x38)` - Disallowed.
+* `('9', 0x39)` - Disallowed.
```
attilamolnar: There's nothing stopping servers from accepting a subset of nicks allowed by this spec (they can send an invalid nick numeric for any nick they don't like) so servers can still disallow digits if they want but they cannot allow more nicks than what this spec allows. Also clients must be prepared to see nicks starting with digits.

documentation/rfc7700.md
DanielOaks: Cool, removed those

DanielOaks:
@grawity @attilamolnar Correct, PRECIS (and casemapping) only affects display, similarly to how casemapping works currently.

documentation/rfc7700.md
DanielOaks: My mistake, meant to remove those with another change. This has been removed.

DanielOaks:
Made the Disallowed Characters section recommended instead of required, as suggested by @attilamolnar, threw `*` and `?` back into the usernames section, various other minor edits of the copy.

M2Ys4U:
PRECIS (RFC 7564) defines two classes, `IdentifierClass` and `FreeformClass`, the Nickname profile (RFC 7700) uses the latter.

To quote from 7564 (with emphasis added by me):

>IdentifierClass:  a sequence of letters, numbers, and some symbols that is **used to identify or address a network entity such as** a user account, **a venue (e.g., a chatroom)**, an information source (e.g., a data feed), or a collection of data (e.g., a file); the intent is that this class will minimize user confusion in a wide variety of application protocols, with the result that **safety has been prioritized over expressiveness for this class.**

> FreeformClass:  a sequence of letters, numbers, symbols, spaces, and other characters that is **used for free-form strings, including** passwords as well as **display elements such as human-friendly nicknames** for devices or **for participants in a chatroom**; the intent is that this class will allow nearly any Unicode character, with the result that expressiveness has been prioritized over safety for this class.  Note well that protocol designers, application developers, service providers, and end users might not understand or be able to enter all of the characters that can be included in the FreeformClass -- see Section 12.3 for details.

With that context out of the way, here's my question:

Should we be re-using the Nickname profile (RFC 7700) for *channel names* as well as nicks and usernames?

It would make more sense to me to restrict channel names to the `IdentifierClass`

DanielOaks:
That's a good point... Using multiple algorithms (one for chans, one for nicks), imo is just begging for trouble but I'll certainly have a closer look into and read of that, thanks for pointing it out.

documentation/rfc7700.md
```diff
@@ -0,0 +1,149 @@
...
+* `(',', 0x2C)` - Used as a separator.
+
+
+## Security Considerations
+
+New unicode-capable casemappings have a considerable security impact.
+
+With the large numbers of new characters allowed comes the risk of introducing confusion for users. The PRECIS framework (much like the earlier framework [stringprep](https://tools.ietf.org/html/rfc3454)) aims to avoid this through mapping confusable characters to a single base character, and by allowing specific known-good characters.
+
+The PRECIS framework represents the most modern standardized solution today for doing this sort of mapping and handling of internationalized names, and should mitigate most of the issues around this.
```
M2Ys4U: I think this is a highly misleading statement.

Reading [Section 12.5 (Security Considerations - Visually Similar Characters) of RFC 7564](https://tools.ietf.org/html/rfc7564#section-12.5) it says:

>Because PRECIS-compliant strings can contain almost any properly encoded Unicode code point, it can be relatively easy to fake or mimic some strings in systems that use the PRECIS framework.  The fact that some strings are easily confused introduces security vulnerabilities of the kind that have also plagued the World Wide Web, specifically the phenomenon known as phishing.

>[...]

>Because it is impossible to map visually similar characters without a great deal of context (such as knowing the font families used), the PRECIS framework does nothing to map similar-looking characters together, nor does it prohibit some characters because they look like others.

>[...]

>The challenges inherent in supporting the full range of Unicode code points have in the past led some to hope for a way to programmatically negotiate more restrictive ranges based on locale, script, or other relevant factors; to tag the locale associated with a particular string; etc.  As a general-purpose internationalization technology, the PRECIS framework does not include such mechanisms.

DanielOaks: Unless I'm mistaken, I believe this would be covered by the rules of the Nickname profile itself [here](https://tools.ietf.org/html/rfc7700#section-2.1) (specifically, 3+4+5). Regardless, I'll have another read over both those documents and probably adjust the text here to make it more clear exactly what I'm referring to, thanks for pointing this out.

documentation/rfc7700.md
```diff
@@ -0,0 +1,149 @@
...
+The `rfc7700` casemapping uses the PRECIS [Nickname profile][precis] as defined in [Section 2 of RFC 7700][precis].
+
+[precis]: https://tools.ietf.org/html/rfc7700#section-2
+
+
+### Order of Operations
+
+Names being prepared MUST apply the following rules in the order shown:
+
+1. Preperation using the PRECIS [Nickname profile][precis].
```
SaberUK: s/Preperation/Preparation/

documentation/rfc7700.md
```diff
@@ -0,0 +1,149 @@
+---
+title: rfc7700 IRC Casemapping
+layout: spec
+copyrights:
+  -
+    name: "Daniel Oaks"
+    period: "2016"
+    email: "daniel@danieloaks.net"
+---
+This document describes a unicode-aware casemapping for IRC, based on the recommendations in [RFC 7700](https://tools.ietf.org/html/rfc7700).
```
SaberUK: Unicode is a proper noun so it should be capitalised.

documentation/rfc7700.md
SaberUK: We (InspIRCd) use glob matching on channel names in various places.

jwheare: @SaberUK Example? Do you also forbid those characters in channel names or is there just no way to specify them without accidentally over-globbing?

Actually, I just tested and was able to create a channel on Insp with both `*` and `?`. I think the recommendation should probably not go against existing valid characters.

SaberUK: @jwheare We don't presently forbid them although they are used in various places like e.g.

https://github.com/inspircd/inspircd/blob/master/docs/conf/modules.conf.example#L714

This does unfortunately result in some problems like what you mentioned though.

attilamolnar: It is possible to block them, as documented: https://github.com/inspircd/inspircd/blob/insp20/docs/conf/modules.conf.example#L417-L419

CarrotCodes:
On IRC we discussed the use case of emoji in channel names (#🥕 for example) - irccloud and others use this in production right now.

@DanielOaks is investigating the difficulty of a custom precis profile that permits such modifications to other profiles.

syzop:
Hmm. I can't find any C library that has PRECIS and those profiles. But that could also be my current lack of knowledge with regards to unicode (and utf8). In any case, the availability of a library or drop-in code that various IRCd's could use for checking "is this nick permitted?" and "are these nicks the same?" would make implementing this much more doable, possibly even crucial for success. And of course, not just for IRC servers but also for services and (I suppose) clients.

Also, I read that as of October 2017 RFC8265 obsoletes RFC7613 and RFC8266 obsoletes RFC7700.

DanielOaks:
Yeah, there's some trouble with this approach around confusable characters, so I've got this specification 'on hold' until I work out those issues. Once I've got those issues worked out I'll change this spec from 7613 to one of the newer RFC numbers.

syzop:
I only saw your edit just now:
> To be specific, PRECIS doesn't in any way attempt to map confusable characters to a single codepoint. Well, it does, really, but only *certain* confusable characters, and not others. Which means you can actually get two nicknames that look exactly the same following this method.

That is disappointing. I guess I misunderstood what PRECIS does then (could be because I didn't read it :D). I must say that the *Security Considerations* in your draft gave me a bit of a false sense of security.. it starts with saying it has considerable security impact but then outlines the avoiding of confusing characters etc. etc... it sounded quite reassuring. So you may want to reword that or, better, see if a solution is possible (see next).

I think for something workable on IRC you would have to "solve" that problem as well. Or give suggestions about what should be done in the IRCd. Don't you agree?
As an UTF8 noob I'm not really in the position to do this but perhaps suggesting only to allow certain scripts or only specific certain combinations...

DanielOaks:
For sure, yeah. I assumed PRECIS protected against that as well (because it does map a fair few of those characters together, just not the identical-looking ones). I wrote up the spec, then someone demonstrated certain pairs of characters that look identical, but the PRECIS UsernameCaseMapped profile keeps separate, so yeah.

Don't you love Unicode?

I'm planning on something along those lines as well, similar to the PRECIS suggestions around possibly only allowing one script or similar (as much as that feels like a copout).

syzop:
@DanielOaks: I tried to contact you a while back via email (27 Nov 11:53 UTC) from syzop@vunscan.org. I could put part of that here in the open:

I've added experimental UTF8 support in [set::allowed-nickchars](https://www.unrealircd.org/docs/Nick_Character_Sets) in UnrealIRCd which allows the admin to allow certain utf8 characters in nick names. In the [release notes](https://github.com/unrealircd/unrealircd/blob/unreal40/doc/RELEASE-NOTES) I mention that, like the original set::allowed-nickchars, it does not do any special CASEMAPPING or "similar looking character detection", and summing up the known problems with the lack of such support. I also noticed that for example anope does not seem to allow such characters which further limits the current use.
So, I'm not happy with the present state. In practice for serious networks, it's not so much usable. It's more of an experimental thing so users can play around, hoping to get that UTF8 ball rolling a bit. It's sad to see that UTF8 nick name support is still lacking in IRC in 2017.

In my opinion the goal of IRCv3, or in any case IRC in general, should be to add a new CASEMAPPING in some standard way/library/tables so the same casemapping (and other stuff PRECIS does) is applied the same way to irc servers and services (and clients). If every software implementation is going to choose it's own casemapping it's rather annoying and confusing. This is especially notable in the servers vs services case where f.e. account names are compared. Anyway, just saying, the spec is just as important as having common code/lib/implementations.

In my email to you I also ask for some technical suggestions with regards to that.
Just checking you received it. If you did and don't think you have anything useful to reply, don't want to or don't have time, that's fine too of course. Just checking..

DanielOaks:
Heyo @syzop! Sorry for not responding, emails have fallen behind a little with lots going on at work and home. I'm thinking of changing this proposal slightly to better integrate with existing servers that use something like `CASEMAPPING=rfc1459`, as well as clarifying more precisely in the spec the issues with this folding method and how to avoid those issues.

Totally agree with the standard way to do such a thing, that's been mostly the intention of this since it started coming up and since I threw this proposal in.

I'll respond to your email and either later this week or over the weekend throw those changes into this specification to clear things up and make it easier for existing servers to implement. Thanks for the push with this and I'm excited to see what comes of Unreal's new experimental char support :)

syzop:
Great. And no problem at all! Glad to see your continued interest and look forward to working with you.

jwheare:
s/preperation/preparation/g

Should there be a way to specify an allowed list of characters/sets as described in the visually similar section? Another ISUPPORT token?

DanielOaks:
Thanks, will fix that.

Nah, no way to do something like that. It'd end up being a multi-KB (maybe even MB) blob sent to the client on connection/registration, and we'd be inventing the format from scratch. Since the server can just refuse that channel/nick name they can just refuse it with an appropriate error message in the numeric if they want to.

jwheare:
Fair enough. Just wondering if something like Unreal's allowed-nickchars labels might work instead of a huge list of chracters https://www.unrealircd.org/docs/Nick_Character_Sets

hebrew-utf8, etc aren't really standard labels but maybe something similar exists?

DanielOaks:
You could try, but they wouldn't be able to be used as anything more than a rough _"Hey this is what we sorta allow"_.

In addition to standard character sets, since the issue is around confusable characters you'd also need to have the ability to list explicit characters as well, which combinations of characters specifically aren't allowed and which combinations aren't allowed together.

For instance, in my (admittedly fairly lax) server I'm looking at doing an interesting approach around the confusables lists distributed by Unicode and somewhat-heuristically along with character sets determining whether specific names are allowed. Those are the sorta details you can't really codify.

Given the above, I think just plain leaving it to the server and them explicitly telling the user why that nick/channel name isn't allowed would be the best option.

DanielOaks:
As a note, there's a running implementation of this on [`testnet.oragono.io`](ircs://testnet.oragono.io:6697/) for anyone interested in giving it a shot and seeing how it works.

It should be noted that aside from the banned characters in the spec, it doesn't implement any sort of additional protections recommended (it's planned, just a lot of work to either build up whitelists or build up some sort of blacklist system based around Unicode's confusables list). This means it's pretty similar to get similar-looking nicks using this [homograph attack generator](https://www.irongeek.com/homoglyph-attack-generator.php), but that's just what you get when you don't implement proper protections. I don't consider this a spec issue because the spec leaves those specific protection mechanisms up for debate. We can't legislate a single good one because there isn't any defined good one. Even browsers change how they do it regularly, and they've had to deal with it a lot longer than us.

DanielOaks:
Yo @slingamn if you could take a look over the two commits just added that'd be ace. You're the one that's looked deeper into the skeletonisation requirements so if I've got anything wrong just let me know ;)

documentation/rfc8265.md
```diff
@@ -73,7 +73,9 @@ As noted in the [Visually Similar Characters section](https://tools.ietf.org/htm
...
+Unicode skeletonisation is the method we recommend to combat this. For each identifier (nick/channel name) on the server, a 'skeleton' is generated by taking the **casefolded** name, and then applying to it the transformations described in the [Unicode Security Mechanisms document](http://unicode.org/reports/tr39/#Confusable_Detection). These skeletons, if used, MUST ONLY be used for comparison, and not as any user-visible identifier (as they intentionally contain complicated mixes of scripts and characters). When users change nicknames or create new channels, the casefolded names should be compared and the skeletons should also be compared to ensure that both are globally-unique (with any non-unique names rejected outright). This seems to be the most reliable method as of right now, but does require storing the skeletons of all in-use names for comparison purposes.
```
slingamn: This isn't what we implemented in oragono --- we skeletonize the *unfolded*, original identifier (the one that is displayed to the users), and only then apply a round of width and case normalization. The rationale is that an initial round of casefolding may lose information about visual confusability. Hypothetically, you could have a non-Latin character with both uppercase and lowercase forms, such that its uppercase form is visually confusable with a Latin character but its lowercase form is visually distinct. Casefolding first would allow an impersonation attack using the uppercase character.

To be honest, I think it would help to see how this plays out in the wild --- try to get real-world user stories from people using non-Latin scripts, also see if we can get some Unicode experts to play with the implementation and try to break it.

DanielOaks:
I think the only thing to fix with this spec is that we described our _skeletonisation_ a bit incorrectly, but otherwise this should describe our implementation pretty well.

As far as a general i18n name solution, I like the prospect of display names more than this spec because of the simpler implementation (once you've got Metadata at least) and the flexibility. It takes a lot to keep Unicode identifiers unique, evident from our skeletonisation description above.

We're gonna keep this implementation in Oragono, but I might publish this as a vendor thing on our site instead of keeping it as a PR here. Particularly as PRECIS libraries that IRC servers written in C (and the like) don't really seem to be available.

PR #274, opened 2016-10-11
cap/tags: Specify that idents are to be parsed as opaque identifiers

DanielOaks:
Mentioned in IRC with projects implementing draft support for specs with names like `draft/client-tags` and such, and subsequently some clients having issues with names like that since they were looking at parsing out and validating vendor prefixes.

It makes much sense to just make sure clients treat them as opaque identifiers, makes clients more resilient and leaves it open for further future clarification/expansion.

### DIFF
--- a/core/capability-negotiation-3.1.md
+++ b/core/capability-negotiation-3.1.md
@@ -166,6 +166,9 @@ continue as normal.
 
 ## Rules for naming capabilities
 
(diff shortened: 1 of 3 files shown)


DanielOaks:
@jwheare sounds good, given servers parse client reqs for caps and parse client tags as well, should be fine to have it there in general.

PR #275, opened 2016-10-11
Message tags 3.3 clarifications

jwheare:
* Clarify that servers advertising the CAP understand all client-initiated tags, prefixed or otherwise.
* Server-initiated tags take precedence over client-only tags
* pseudo-BNF whitespace
* Security note about unprefixed client-initiated tags
* More/improved examples

### DIFF
--- a/core/message-tags-3.3.md
+++ b/core/message-tags-3.3.md
@@ -50,6 +50,9 @@ with a plus sign (`+`) and otherwise conforms to the format specified in
 Client-only tags MUST be relayed on `PRIVMSG` and `NOTICE` events, and
 MAY be relayed on other events.
 
+Any server-initiated tags attached to messages MUST be included before client-only
+tags to prevent them from being pushed outside of the 512 byte tag limit.
+
 The expected client behaviour of individual client-only tags SHOULD be defined
 in separate specifications, in the same way as server-initiated tags.
 
--- a/core/message-tags-3.3.md
+++ b/core/message-tags-3.3.md
@@ -37,7 +37,7 @@ This specification adds the `message-tags` capability. Clients requesting
 this capability indicate that they are capable of parsing all well-formed tags,
 even if they don't handle them specifically.
 
-Servers advertising this capability indicate that they are capable of handling
+Servers advertising this capability indicate that they are capable of parsing
 any tag received from a client, with or without the client-only prefix.
 
 ### Tags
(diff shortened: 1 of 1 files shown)


core/message-tags-3.3.md
```diff
@@ -38,7 +38,7 @@ this capability indicate that they are capable of parsing all well-formed tags,
 even if they don't handle them specifically.
 
 Servers advertising this capability indicate that they are capable of handling
```
DanielOaks: handling -> parsing ?

Mostly because 'handling' sounds like it refers to actually implementing the functionality that every tag out there is supposed to represent, where parsing just means that it can parse it on incoming messages.

core/message-tags-3.3.md
```diff
@@ -87,18 +92,53 @@ that act on tag content as well.
...
 ```
+
+---
+
+A client-only tag `+example` with a value containing valid raw and escaped characters:
+
+* Escaped value: `raw-values+:=,escaped-values\:\s\\`
+* Decoded value: `raw-values+:=,escaped-values; \`
+
+Plus signs, colons, equals signs and commas can appear unescaped in tag values, but semicolons, spaces and backslashes must be escaped.
```
Zarthus: A semicolon is `;`, a colon is `:`.

Zarthus: Also "MUST", probably.

jwheare: `;` is escaped as `\:` which part are you referring to?

MUST et al are not to be used in non normative sections, the escaping rules are specified in full in message-tags-3.2

core/message-tags-3.3.md
jwheare: Good call.

core/message-tags-3.3.md
```diff
@@ -87,18 +92,53 @@ that act on tag content as well.
...
 ```
+
+---
+
+A client-only tag `+example` with a value containing valid raw and escaped characters:
+
+* Raw value: `raw+:=,escaped; \`
+* Escaped value: `raw+:=,escaped\:\s\\`
+
+In this example, plus signs, colons, equals signs and commas are trasnmitted raw in tag values, while semicolons, spaces and backslashes are escaped. [Escaping rules](./message-tags-3.2.html#escaping-values) are unchanged from IRCv3.2 tags.
```
Zarthus: transmitted*

PR #276, opened 2016-11-07
Account Management Framework

DanielOaks:
The `ACC` command framework provides a standardized way for clients to create and manage accounts. In particular, the `ACC REGISTER` and `ACC VERIFY` subcommands should allow a network to provide a consistent interface to register new accounts, no matter what services software is used on the backend. They also let clients build nice interfaces for account registration. This framework is extensible, letting us provide more commands and interfaces in future.

This is the same spec as in #152 but cleaned up and with some bugs fixed. Asked kaniini if I could submit some fixes a while back and he said I should just throw a new PR in.

*edit:* I think the last thing missing here is the ability to know whether the net uses your nickname as the account name or not. I'll take a look at that and get something worked out there

### DIFF
--- a/extensions/acc-core.md
+++ b/extensions/acc-core.md
@@ -11,146 +11,171 @@ copyrights:
     period: "2016-2019"
     email: "daniel@danieloaks.net"
 ---
+## Notes for implementing work-in-progress version
+This is a work-in-progress specification.
 
-The `ACC` command framework provides a unified, standardized interface for account management.
+Software implementing this work-in-progress specification MUST NOT use the unprefixed `acc` capability name. Instead, implementations SHOULD use the `draft/acc` capability name to be interoperable with other software implementing a compatible work-in-progress version.
 
-This specification defines the `ACC REGISTER` and `VERIFY` subcommands, which standardise account creation and validation to the network's authentication layer.  This allows for a network to provide a common interface regardless of what authentication layer they have chosen for their network to operate, i.e. traditional services, a different central authority, or a decentralized model similar to [IRCX][https://en.wikipedia.org/wiki/IRCX].
+The final version of the specification will use an unprefixed capability name.
+
+
+## Introduction
+The `ACC` command framework provides a unified, standardized interface for account management. Typically, managing this account has been done through `NickServ` or other services bots, which has made it difficult for clients to present clean, consistent interfaces to users.
+
+This specification defines the `ACC REGISTER` and `VERIFY` subcommands, which standardise account creation and validation to the network's authentication layer.  This allows for a network to provide a common interface regardless of what authentication layer they have chosen for their network to operate, i.e. traditional services, a different central authority, or a decentralized model similar to [IRCX](https://en.wikipedia.org/wiki/IRCX).
    
-We believe that the benefit of adding a common interface for account management will result in more user-focused interfaces for common account tasks including registration. Further, standardised account registration allows client authors to provide guided account creation and validation to their users, regardless of how the network performs these tasks.
+We believe that the benefit of adding a common interface for account management will result in more user-focused interfaces for common account tasks including registration (just as SASL has allowed clients to create clean account authentication interfaces). Further, standardised account registration allows client authors to provide guided account creation and validation to their users, regardless of how the network performs these tasks.
+
+
+## The `draft/acc` capability
+This capability indicates that the server supports the `ACC` command as described by this specification, and can use `ACC LS` to get a list of supported subcommands. Clients MAY request the `draft/acc` capability, but it will not affect whether or not they can use the command.
+
 
 ## The `ACC` command
+The `ACC` command exposes a number of subcommands as described below. Servers MAY send `ACC` messages to clients in response to client commands, as described by the specifications describing those commands.
 
-The `ACC` command is the main command used by the account registration subsystem.  It provides several subcommands listed in the `ACCCOMMANDS` ISUPPORT type.  The core subcommands are listed below.
+Servers MUST use the [standard replies extension](https://github.com/ircv3/ircv3-specifications/pull/357) to notify the client of failed `ACC` subcommands. The specific codes and format to use are given with each command's description.
 
-## The `ACC REGISTER` sub-command
 
-The `ACC REGISTER` command signals the intent of a client to register an account with the network's authentication layer.  It is similar to current methods of signalling that intent.
+## The `ACC LS` subcommand
+The `ACC LS` subcommand signals that a client wishes to receive a list of supported `ACC` subcommands. The response is one or more `ACC LS` messages sent to the client with this format:
 
-An `ACC REGISTER` command has the following format:
+    :<server> ACC LS :subcommand names go here
+
+If the reply contains multiple lines (due to IRC line length limitations), all but the last reply MUST have a parameter containing only an asterisk (`*`) preceding the subcommand list. This lets clients know that more `ACC LS` lines are incoming. For example:
+
+    :<server> ACC LS * :LS REGISTER VERIFY
+    :<server> ACC LS * :subcommand names go here
+    :<server> ACC LS * :subcommand names go here
+    :<server> ACC LS :final names go here
+
+If this subcommand is given with any additional parameters, servers MUST ignore these and process the subcommand as normal. This allows for future extension of this subcommand.
+
+
+## The `ACC REGISTER` subcommand
+The `ACC REGISTER` subcommand signals the intent of a client to register an account with the network's authentication layer.  It is similar to current methods of signalling that intent.
+
+An `ACC REGISTER` subcommand has the following format:
 
     ACC REGISTER <accountname> [callback_namespace:]<callback> [cred_type] :<credential>
 
-The `accountname` field specifies the account name the client wishes to register.  If the account name is `*`, it indicates that the client wishes to register with their current nickname as this field.
+The `accountname` parameter specifies the account name the client wishes to register.  If the account name is `*`, it indicates that the client wishes to register with their current nickname as the account name.  If the `REGNICK` ISUPPORT token is advertised, the client MUST send `*` for this parameter.
 
-The `credential` field specifies a credential of `cred_type`.  If there is no specified credential, then the `credential` field MUST be `*`.  A passphrase type `credential` MAY have spaces in it. New credential types may or may not allow spaces inside `credential`.
+The `callback` parameter indicates where the verification code should be sent. This parameter MAY contain an asterisk `*`, indicating that no verification code is required and that account registration shall complete immediately. Otherwise, this parameter contains a standard URI, the namespace being contained in the IANA URI Schemes registry (as per [RFC 7595](https://tools.ietf.org/html/rfc7595)) and listed in the `REGCALLBACKS` ISUPPORT token. If a callback namespace is not explicitly provided, the IRC server MUST use `mailto` as a default.
 
-The `callback` field designates an opaque value which indicates where a verification code, if any, will be sent.  The callback field SHOULD contain either a standard URI, the namespace being contained in the IANA URI Schemes registry (as per [RFC 7595][rfc7595]) and listed in the `REGCALLBACKS` ISUPPORT token, or a `*` as decribed below. An invalid callback parameter should result in the `ERR_REG_INVALID_CALLBACK` error.
+The `cred_type` parameter indicates the type of authentication that's required to login to the account once it has been registered. The supported credential types are:
 
-In the event that a callback is not provided, the client SHOULD send `*` to indicate that they are not providing a callback resource.  If a callback namespace is not explicitly provided, the IRC server MAY choose to use `mailto` as a default.
+- `passphrase`: `<credential>` is a plain-text passphrase that the client will use to authenticate with SASL once the account is registered. Passphrases MAY contain spaces.
+- `certfp`: The client's TLS certificate is used as the as the authentication mechanism for future connections. In short, if the client connects and presents the same TLS certificate, they can use the `SASL EXTERNAL` method to authenticate. The client SHOULD provide an empty credential (`*`), and the server MUST ignore the credential value provided by the client. If no certificate is presented by the client, the server MUST respond with the `REG_INVALID_CRED_TYPE` `FAIL` code.
 
-The `cred_type` field designates a value which indicates the type of supplied credential.  The accepted values of the `cred_type` field is implementation-specific, with credential types being listed using the `REGCREDTYPES` ISUPPORT token.  An invalid credential type parameter should result in the `ERR_REG_INVALID_CRED_TYPE` error.  The client SHOULD supply a `cred_type` value, however if one is not provided the IRC server SHOULD use `passphrase` as a default credential type.
+If the client does not provide the `cred_type` parameter, the server MUST use the `passphrase` credential type. If the client provides a credential type that's now supported, the server MUST respond with the `REG_INVALID_CRED_TYPE` `FAIL` code.
 
-The IRC server MAY forward the `ACC REGISTER` command to a central authority, or process it locally. Clients SHOULD NOT be sent unnecessary legacy (e.g. from NickServ) private messages or notices in response to `ACC` commands, where native numerics such as the ones defined in this document can replace them.
+Clients SHOULD NOT be sent legacy (e.g. from NickServ) privmsgs or notices in response to `ACC` commands where numerics and messages defined in this document can replace them.
 
 After this, if verification is required, the IRC server MUST send the `RPL_REG_VERIFICATION_REQUIRED` numeric. If verification is not required, the IRC server MUST instead send the `RPL_REG_SUCCESS` numeric.
 
-| No. | Label                           | Format                                                                                                       |
-| --- | ------------------------------- | ------------------------------------------------------------------------------------------------------------ |
-| 920 | `RPL_REG_SUCCESS`               | `:<server> 920 <user_nickname> <accountname> :Account created`                                               |
+| No. | Label | Format |
+| --- | ----- | ------ |
+| 920 | `RPL_REG_SUCCESS` | `:<server> 920 <user_nickname> <accountname> :Account created` |
 | 927 | `RPL_REG_VERIFICATION_REQUIRED` | `:<server> 927 <user_nickname> <accountname> <callback_namespace:><callback> :A verification token was sent` |
 
-Upon error, the IRC server MUST send an error code that is relevant.  We suggest these numerics:
+Upon error, the IRC server MUST send a [`FAIL` message](https://github.com/ircv3/ircv3-specifications/pull/357) with one of the codes below using the given format, and including an appropriate description of the error:
 
-| No. | Label                        | Format                                                                                          |
-| --- | ---------------------------- | ----------------------------------------------------------------------------------------------- |
-| 921 | `ERR_ACCOUNT_ALREADY_EXISTS` | `:<server> 921 <user_nickname> <accountname> :Account already exists`                           |
-| 922 | `ERR_REG_UNSPECIFIED_ERROR`  | `:<server> 922 <user_nickname> <accountname> :<descriptive_text>`                               |
-| 928 | `ERR_REG_INVALID_CRED_TYPE`  | `:<server> 928 <user_nickname> <accountname> <cred_type> :Credential type is invalid`           |
-| 929 | `ERR_REG_INVALID_CALLBACK`   | `:<server> 929 <user_nickname> <accountname> <callback> :Callback token is invalid`             |
-| 440 | `ERR_REG_UNAVAILABLE`        | `:<server> 440 <user_nickname> <subcommand> :Authentication services are currently unavailable` |
+| Code | Format |
+| ---- | ------ |
+| `ACCOUNT_ALREADY_EXISTS` | `:<server> FAIL ACC ACCOUNT_ALREADY_EXISTS <accountname> :Account already exists` |
+| `REG_INVALID_CALLBACK` | `:<server> FAIL ACC REG_INVALID_CALLBACK <accountname> <callback> :Callback is invalid` |
+| `REG_INVALID_CRED_TYPE` | `:<server> FAIL ACC REG_INVALID_CRED_TYPE <accountname> <cred_type> :Credential type is invalid` |
+| `REG_MUST_USE_REGNICK` | `:<server> FAIL ACC REG_MUST_USE_REGNICK <accountname> :Must register with current nickname instead of separate account name` |
+| `REG_UNAVAILABLE` | `:<server> FAIL ACC REG_UNAVAILABLE :Account registration is currently unavailable` |
+| `REG_UNSPECIFIED_ERROR` | `:<server> FAIL ACC REG_UNSPECIFIED_ERROR <accountname> [<contexts>...] <description>` |
 
-The server MAY send additional informative text upon registration success or failure using `RPL_TEXT` or `NOTICE`.
 
-## The `ACC VERIFY` sub-command
+## The `ACC VERIFY` subcommand
+The `ACC VERIFY` subcommand lets a client submit a verification token to complete their account registration.  This token is one that the server has sent to the user out-of-band with the given callback namespace and method.
 
-The `ACC VERIFY` command signals the intent of a client to submit a verification token for their account registration.  If a valid callback is specified by the client during the `ACC REGISTER` command, the server MUST send the token to the specified callback.
-
-A `ACC VERIFY` command consists of the following format:
+A `ACC VERIFY` subcommand consists of the following format:
 
     ACC VERIFY <accountname> <auth_code>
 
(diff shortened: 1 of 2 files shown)


extensions/reg-core-3.3.md
```diff
@@ -0,0 +1,224 @@
...
+comma-separated list of supported commands.  Other extensions MAY include additional credential
+types than those listed here.
+
+The following credential types are defined:
+
+  * `passphrase`: an unencrypted passphrase which is sent to the server.
+  * `certfp`: the client certificate fingerprint as determined by the server if available.
+    The client MAY provide an empty credential (`*`) in this case.  The server SHOULD calculate
+    the credential using the client certificate fingerprint, if available.  If no client certificate
+    is presented, the server SHOULD NOT advertise the availability of this credential type.
```
ProgVal: What hash algorithm should be used to give a fingerprint other than `*`? And what format?

extensions/reg-core-3.3.md
```diff
@@ -0,0 +1,224 @@
...
+| 928 | `ERR_REG_INVALID_CRED_TYPE`  | `:<server> 928 <user_nickname> <accountname> <cred_type> :Credential type is invalid`           |
+| 440 | `ERR_REG_UNAVAILABLE`        | `:<server> 440 <user_nickname> <subcommand> :Authentication services are currently unavailable` |
+
+The server MAY send additional informative text upon registration success or failure using `RPL_TEXT` or `NOTICE`.
+
+## The `REG VERIFY` sub-command
+
+The `REG VERIFY` command signals the intent of a client to submit a verification token for their
+account to the authentication layer.  The verification token MUST be sent to a valid callback
+resource as specified by the client in the `REG CREATE` command.
```
ProgVal: “**If a verification token is required by the server,** it MUST be sent **by the IRC network** to a valid callback resource [...]”

extensions/reg-core-3.3.md
```diff
@@ -0,0 +1,224 @@
...
+
+| No. | Label               | Format                                                                         |
+| --- | ------------------- | ------------------------------------------------------------------------------ |
+| 923 | `RPL_VERIFYSUCCESS` | `:<server> 923 <user_nickname> <accountname> :Account verification successful` |
+
+The IRC server SHOULD also send an `RPL_LOGGEDIN` (900) numeric and consider the client to be logged in to the
+account that has been successfully verified.
+
+Upon error, the IRC server MUST send an error code that is relevant.  We suggest these
+numerics:
```
ProgVal: Could you add the requirement for the error code to be in a given list?
Otherwise, servers may send something a bot does not expect to be a reply to this, and the bot will not know what to do with it.

extensions/reg-core-3.3.md
```diff
@@ -0,0 +1,224 @@
...
+    The client MAY provide an empty credential (`*`) in this case.  The server SHOULD calculate
+    the credential using the client certificate fingerprint, if available.  If no client certificate
+    is presented, the server SHOULD NOT advertise the availability of this credential type.
+
+A sample 005 reply indicating support for credential types is:
+
+    :irc.example.com 005 kaniini REGCREDTYPES=passphrase,certfp :are supported by this server
+
+The first credential type SHOULD be the implementation-defined default if default credential types
+are supported.
```
ProgVal: I did not understand this line at first.
Do you mean this? “The **order** credential type SHOULD be the implementation-defined default if **both** credential types are supported.

extensions/reg-core-3.3.md
```diff
@@ -0,0 +1,224 @@
...
+| 920 | `RPL_REGISTRATION_SUCCESS` | `:<server> 920 <user_nickname> <accountname> :Account created` |
+
+If the server requires a verification token, it MUST also reply with the `RPL_REG_VERIFICATION_REQUIRED`
+numeric, which looks like:
+
+| No. | Label                           | Format                                                                                                       |
+| --- | ------------------------------- | ------------------------------------------------------------------------------------------------------------ |
+| 927 | `RPL_REG_VERIFICATION_REQUIRED` | `:<server> 927 <user_nickname> <accountname> <callback_namespace:><callback> :A verification token was sent` |
+
+Upon error, the IRC server MUST send an error code that is relevant.  We suggest these numerics:
```
ProgVal: Could you add the requirement for the error code to be in a given list?
Otherwise, servers may send something a bot does not expect to be a reply to this, and the bot will not know what to do with it.

extensions/reg-core-3.3.md
DanielOaks: Yeah the language here is a bit iffy, thanks for pointing it out. I'll clean this up a bit.

DanielOaks:
@ProgVal I'll look at making sure the error codes are in the given list, but how do those changes look to you? Should clean up the language in those sections a fair bit

ProgVal:
Why do you want to prevent the client from setting a fingerprint that is not the one of its current certificate?
We will have to allow that at some point if we want to allow users to change certificates or use more than one.

DanielOaks:
Different networks can use whatever fingerprint methods they want for certs. In addition, allowing clients to register with fingerprints other than one that they explicitly control sounds a bit dodgy in terms of spoofing other users. Say that you know the fingerprint of my cert – you register an account with the fingerprint in the certfp's cred field and now I can't register an account linked to my cert because one already exists for it.

Because of that and the difficulty of conveying which algorithm is being used. Does certificate fingerprinting use standardised algorithms like SHA1/SHA256 **always** or are there cases where people can do whatever special stuff they want, say using both SHA1 and SHA256 to better protect against dodgy collisions. Is there a 100% solid way that we know to convey exactly which method is being used to generate the fingerprint, should we also define a numeric that dumps what your certfp is on connection, so that clients can then use that later in the `REG CREATE` command?

Just using the client's current cert makes the most sense to me, though if there are answers for all these then I'm open to investigating it further.

grawity:
Existing services already allow "claiming" any fingerprint anyway. But I like the restriction because it can reduce accidents – if you're only allowed to use your current cert, then registering implies you can _successfully connect_ with it and won't end up with an unusable account.

DanielOaks:
Updating your cert would be handled separately, if appropriate could be through another `REG` command we describe later (similar to updating your email address).

ProgVal:
But to update the cert, you have to be able to give a fingerprint at some point. Unless you find a way of connecting with the old and the new cert at the same time (IRC in TLS in TLS?), but that may be hard to implement in some/most clients.

DanielOaks:
I think restricting registration to the certificate you currently have makes sense, for reasons @grawity put above. For updating the cert on an existing account that could definitely make sense, but I think looking at that is the job of another spec (which may also look at updating passphrase/etc for the account as well).

ProgVal:
Ok

extensions/reg-core-3.3.md
```diff
@@ -0,0 +1,261 @@
...
+
+The `accountname` field specifies the account name the client wishes to register.  If the account name
+is `*`, it indicates that the client wishes to register with their current nickname as this field.
+
+The `credential` field specifies a credential of `cred_type`.  If there is no specified credential, then
+the `credential` field MUST be `*`.  A passphrase type `credential` MAY have spaces in it. New credential
+types may or may not allow spaces inside `credential`.
+
+The `callback` field designates an opaque value which indicates where a verification code, if any,
+will be sent.  The callback field is implementation-specific, with namespaces being listed using the
```
M2Ys4U: If we're using IANA-registered scheme names for the callback namespace, does it make sense that the callback field is implementation-specific? Why not just use a proper URI here?

DanielOaks: Makes sense, thanks for the suggestion. I'll look at this.

extensions/reg-core-3.3.md
```diff
@@ -0,0 +1,261 @@
...
+
+## The `REG VERIFY` sub-command
+
+The `REG VERIFY` command signals the intent of a client to submit a verification token for their
+account registration.  If a valid callback is specified by the client during the `REG CREATE`
+command, the server MUST send the token to the specified callback.
+
+A `REG VERIFY` command consists of the following format:
+
+    REG VERIFY <accountname> <auth_code>
```
Zarthus: I think this should include the (optional?) password if we're going to immediately log the user in [L111]. Or indicate that this command is only available while already logged in (how do you log in?) (in which case the sentence below "the irc server should also send an rpl_loggedin ..." would no longer be valid)

There are many users who accidentally paste their authentication code in chat by accident.

Zarthus:
Seems to be missing a way to drop the existing nickname.  What if I misspell my email?

DanielOaks:
That's handled by a separate spec. This is just about registering your account

extensions/reg-core-3.3.md
```diff
@@ -0,0 +1,261 @@
...
+will be sent.  The callback field is implementation-specific, with namespaces being listed using the
+`REGCALLBACKS` ISUPPORT token.  An invalid callback parameter should result in the `ERR_REG_INVALID_CALLBACK`
+error.  In the event that a callback is not provided, the client SHOULD send `*` to indicate that
+they are not providing a callback resource.  If a callback namespace is not explicitly provided, the IRC
+server MAY choose to use `mailto` as a default.
+
+The `cred_type` field designates a value which indicates the type of supplied credential.  The accepted
+values of the `cred_type` field is implementation-specific, with credential types being listed using the
+`REGCREDTYPES` ISUPPORT token.  An invalid credential type parameter should result in the `ERR_REG_INVALID_CRED_TYPE`
+error.  The client SHOULD supply a `cred_type` value, however if one is not provided the IRC server SHOULD
```
Zarthus: This sounds better defined as a MUST.

prawnsalad:
> C: REG CREATE * mailto:dan@example.com passphrase :testpassphrase123
> S: 920 dan- dan- :Account registered
> S: 927 dan- dan- dan@example.com :A verification code was sent

From the client dev perspective, receiving the 920 when verification is still required can cause issues in the UI.
* User attempts to register, UI shows a loading sign
* Client receives the 920, UI shows "congrats" screen
* Client must wait for an arbitrary length of time incase it receives a 927, UI must now change to "Sorry, you're not actually finished. verify now".

Two possible ways to improve this would be to either send 927 before 920, or to send one or the other but never both. The latter would be cleaner IMO.

DanielOaks:
920's sent when it's successful, so if they do mailto: then it'd send that after they send the REG VERIFY command, not after REG CREATE. If they REG CREATE with skipping verification, they wouldn't get 927 since it requires no verification code to be sent/etc.

extensions/reg-core-3.3.md
```diff
@@ -0,0 +1,261 @@
...
+The IRC server MAY forward the `REG VERIFY` command to a central authority, or process it
+locally.
+
+Upon success, the IRC server MUST send the `RPL_VERIFYSUCCESS` numeric, which looks like:
+
+| No. | Label               | Format                                                                         |
+| --- | ------------------- | ------------------------------------------------------------------------------ |
+| 923 | `RPL_VERIFYSUCCESS` | `:<server> 923 <user_nickname> <accountname> :Account verification successful` |
+
+The IRC server SHOULD also send an `RPL_LOGGEDIN` (900) numeric and consider the client to be logged in to the
```
Zarthus: Sounds better defined as a MUST.

prawnsalad:
@DanielOaks In that case then the examples need updating to reflect that. They currently show verification mailto: being sent and then a 920 being received before 927.

DanielOaks:
@prawnsalad oh, weird. yeah I'll def go over that and make sure it's made clearer and all, thanks for pointing that out!

MuffinMedic:
Rather than have a specification solely for registration purposes, a general `ACCOUNT` spec would be more extensible and allow for easier implementation of adding on new features later. Account management includes actions other than registration and verification, such as updating details, deleting, etc...

With this current spec, each of these additional features would need to be written as its own spec from the ground up, which will fragment similar features across multiple. Using an `ACCOUNT` spec and implementing `REG` as one of potentially many extensions enables things such as `DELETE` and `UPDATE` to be wrapped into a single spec as extensions, keeps related features together, and makes improvements easier down the line.

Having a single `ACCOUNT` spec also enables standardization of certain elements across various account features, such as the same recommended security checks for all account features and consistency with the command format. At the same time, this allows for individual extensions to add the changes they need (new sub-command) while maintaining this consistency.

Zarthus:
I personally like how it is extensible, it allows networks to implement some components, make their own, and omit some of the others.

DanielOaks:
Sure, makes sense. I'll push an update that puts these changes into action.

DanielOaks:
Should address most of the concerns here.

Named it `ACC` rather than `ACCOUNT` to avoid confusion with the `ACCOUNT` verb added in [`account-notify`](http://ircv3.net/specs/extensions/account-notify-3.1.html) and the related account-info-tracking specs. Let's not bikeshed this for ages, `ACC` or `ACCT` works fine and avoids confusion while allowing for nice extension of the command later (particularly if different commands want to respond in turn with the same verb).

@M2Ys4U How's that language regarding the IANA URI Schemes registry? We'll leave it as 'SHOULD' as it also allows `*` as a special char and to leave possibilities open.

@prawnsalad Made it so that after `ACC REGISTER` the server replies with EITHER `RPL_REG_SUCCESS` or `RPL_REG_VERIFICATION_REQUIRED`, and not both.

extensions/acc-core.md
```diff
@@ -8,11 +8,11 @@ copyrights:
     email: "nenolod@dereferenced.org"
   -
     name: "Daniel Oaks"
-    period: "2016"
+    period: "2016-2017"
     email: "daniel@danieloaks.net"
 ---
 
-The `REG` command framework provides a unified, standardized interface for account creation and
+The `ACC` command framework provides a unified, standardized interface for account creation and
```
MuffinMedic: s/creation/management/

extensions/acc-core.md
```diff
@@ -25,20 +25,20 @@ authority layer of the network.  Further, we believe it is helpful to client aut
 they may provide guided account creation and validation for their users, regardless of network
```
MuffinMedic: s/creation/management/

extensions/acc-core.md
```diff
@@ -83,23 +81,23 @@ Upon error, the IRC server MUST send an error code that is relevant.  We suggest
...
-account registration.  If a valid callback is specified by the client during the `REG CREATE`
+The `ACC VERIFY` command signals the intent of a client to submit a verification token for their
+account registration.  If a valid callback is specified by the client during the `ACC REGISTER`
 command, the server MUST send the token to the specified callback.
 
-A `REG VERIFY` command consists of the following format:
+A `ACC VERIFY` command consists of the following format:
 
-    REG VERIFY <accountname> <auth_code>
+    ACC VERIFY <accountname> <auth_code>
```
MuffinMedic: Server should allow overrides - i.e. oper verification of accounts without token.

extensions/acc-core.md
```diff
@@ -197,65 +195,64 @@ description.
...
     S: 923 kaniini rabbit :Account verification successful
     S: 903 kaniini :Authentication successful
 
 ### Registering the account "rabbit" with an unsupported verification method:
 
-    C: REG CREATE rabbit 1vBjNBdhjWFFbbbbVBHJEWBHJWcfbbvjkhbea passphrase :testpassphrase123
+    C: ACC REGISTER rabbit 1vBjNBdhjWFFbbbbVBHJEWBHJWcfbbvjkhbea passphrase :testpassphrase123
     S: 929 kaniini rabbit 1vBjNBdhjWFFbbbbVBHJEWBHJWcfbbvjkhbea :Callback token is not valid
 
 ### Registering the account "rabbit", but providing an invalid verification token:
```
MuffinMedic: Add token for oper overrides (account verified and user notified account was verified for them).

extensions/acc-core.md
```diff
@@ -83,23 +81,23 @@ Upon error, the IRC server MUST send an error code that is relevant.  We suggest
...
 | 922 | `ERR_REG_UNSPECIFIED_ERROR`  | `:<server> 922 <user_nickname> <accountname> :<descriptive_text>`                               |
-| 929 | `ERR_REG_INVALID_CALLBACK`   | `:<server> 929 <user_nickname> <accountname> <callback> :Callback token is invalid`             |
 | 928 | `ERR_REG_INVALID_CRED_TYPE`  | `:<server> 928 <user_nickname> <accountname> <cred_type> :Credential type is invalid`           |
+| 929 | `ERR_REG_INVALID_CALLBACK`   | `:<server> 929 <user_nickname> <accountname> <callback> :Callback token is invalid`             |
 | 440 | `ERR_REG_UNAVAILABLE`        | `:<server> 440 <user_nickname> <subcommand> :Authentication services are currently unavailable` |
 
 The server MAY send additional informative text upon registration success or failure using `RPL_TEXT` or `NOTICE`.
 
-## The `REG VERIFY` sub-command
+## The `ACC VERIFY` sub-command
```
MuffinMedic: Each subcommand should be it's own extension rather than in the framework spec (extensions/acc/verify).

This allows for a more modular system in which networks can choose which features to support.

extensions/acc-core.md
```diff
@@ -25,20 +25,20 @@ authority layer of the network.  Further, we believe it is helpful to client aut
...
+## The `ACC` command
 
-The `REG` command is the main command used by the account registration subsystem.  It provides
-several subcommands listed in the `REGCOMMANDS` ISUPPORT type.  The core sub-commands are listed
+The `ACC` command is the main command used by the account registration subsystem.  It provides
+several subcommands listed in the `ACCCOMMANDS` ISUPPORT type.  The core sub-commands are listed
 below:
 
-## The `REG CREATE` sub-command
+## The `ACC REGISTER` sub-command
```
MuffinMedic: Each subcommand should be it's own extension rather than in the framework spec (extensions/acc/register).

This allows for a more modular system in which networks can choose which features to support.

extensions/acc-core.md
```diff
@@ -25,20 +25,20 @@ authority layer of the network.  Further, we believe it is helpful to client aut
 they may provide guided account creation and validation for their users, regardless of network
 configuration.
 
-## The `REG` command
+## The `ACC` command
 
-The `REG` command is the main command used by the account registration subsystem.  It provides
-several subcommands listed in the `REGCOMMANDS` ISUPPORT type.  The core sub-commands are listed
+The `ACC` command is the main command used by the account registration subsystem.  It provides
```
MuffinMedic: s/registration/management/

DanielOaks:
Makes sense, thanks. I'll rewrite a bunch of stuff around how we're structuring it now and throw in an update

DanielOaks:
`ACC_REQUIRED` numeric added to facilitate future use where networks required clients to register an account before connecting. This will come further into play with this particular spec later if/when pre-reg `RPL_ISUPPORT` makes its way onto the stage.

ProgVal:
What is the status of implementations?

DanielOaks:
Oragono has an up-and-running implementation (including verification). That's about it right now.

geajack:
Should other core account related verbs be implemented as ACC subcommands? For example ACC LOGIN or ACC LOGOUT.

SaberUK:
@geajack Logging in is already supported with SASL's `AUTHENTICATE` command.

geajack:
@SaberUK Is that intended to be used in concert with this spec? If so it should surely be a mandatory part of the standard that the account registered here is the same accepted by the AUTHENTICATE command.

Also, should the "account name" used here be (or at least MAY be) totally independent of previous IRC notions of user identity (namely nick, username and real name)? What do we do with nickname collisions on a server that supports accounts?

For example, if I make an account with username Jack, I might be hoping to use the nickname Jack. If I make my account and then one day log in while somebody else is using the nickname Jack, what should happen? How will a non-technical user react to being told they can't have their usual nickname because somebody else is connected with it, after they went to the trouble of making an account?

KoraggKnightWolf:
This could perhaps be solved via implementing nickname ownership as well, at least optionally. The majorly used services packages, Atheme, Anope and forks thereof, often have an option similar  in naming to no_nickname_ownership which can be set to yes for accounts like quakenet/undernet handle them or set to no so that they operate like on most networks that use services, owning also a nickname.

The only thing to decide on would be the default though if something like NO_NICK_OWNING is for example used, I would set it to no as imo one registers a nickname so noone else can use it. Also most newbies (which IRC strongly needs more of) would probably register for the same/similar reasons.

Hope this can clear some things up and maybe even spark up new ideas/specs!

Regards,

Koragg

RyanSquared:
>Also, should the "account name" used here be (or at least MAY be) totally independent of previous IRC notions of user identity (namely nick, username and real name)? What do we do with nickname collisions on a server that supports accounts?

As of right now, IRC service support typically includes "recover", "ghost", and identifying as the current nick. While the spec doesn't outright say it - which it should IMO - it is reasonable to assume that an account name is associated with a nick used in registration.

As was brought up, multiple services also include either holding a nick (preventing use of it without "recovery") or giving a client a certain amount of time to identify, like 60 seconds.

glguy:
If the supported callback and credential types were advertised in CAP LS a client could know what options to present to the user in the case that account registration was to be done before connection-registration. 005 is too late for this.

Examples with and without `CAP 302`

```
CAP * LS :acc
CAP * LS :acc=REGCREDTYPES=passphrase,certfp;REGCALLBACKS=mailto,sms
```

Making this extension opt-in via CAP negotiation ensures that the new, more descriptive error messages using FAIL aren't sent to clients that don't understand them and expect the current error numerics. Many clients I've used, particularly mobile ones, completely ignore messages they don't expect or understand. Changing error messages in cases like joining a channel that requires registration could cause that failure to turn into a silent failure.

It would be slightly more work for server implementations to know whether they should send a standard numeric error or the new FAIL error, but it should just be a simple case of checking negotiated capabilities.

It will be no more work for clients to have a negotiated capability because clients will need to continue to support servers that don't implement this extension. The numeric errors sent when joining a registered-users-only channel will continue to exist in the wild. A client negotiating this capability would be declaring the ability to handle the new error message format.

extensions/acc-core.md
```diff
@@ -0,0 +1,270 @@
...
+In the event that verification is not required by the implementation, the `REGCALLBACKS` token should
+be an empty value.
+
+A sample 005 reply indicating support for verification callbacks is:
+
+    :irc.example.com 005 kaniini REGCALLBACKS=mailto,sms :are supported by this server
+
+A sample 005 reply indicating that no verification callback methods are supported is:
+
+    :irc.example.com 005 kaniini REGCALLBACKS= :are supported by this server
```
jesopo: Surely there should be a way to specify that the server offers registration without callback but also offers e.g. mailto? That way a server can allow accounts to be e.g. tied to an email address or not, depending on what the user chooses?

e.g. `:irc.example.com 005 kaniini REGCALLBACKS=*,mailto :are supported by this server`

DanielOaks: Eh, sure. I'd be fine with this. Haven't dine any work on ACC for ages because of the lack of implementation interest, but when I do update I'll include this.

extensions/reg-core-3.3.md
DanielOaks: Resolved now

extensions/reg-core-3.3.md
DanielOaks: Resolved now

extensions/reg-core-3.3.md
DanielOaks: This line has been reworded, should make more sense now

DanielOaks:
So, this is essentially a complete rewrite of the ACC Command Framework spec with the (lots of) feedback kept in mind. No longer relies on `ISUPPORT`, no longer defines a ton of error numerics and makes use of the `FAIL` command instead, etc. I think this is much easier to implement, I rewrote the confusing / arbitrarily complex text, and it should support the use cases a lot better in general now.

Let us know what you think.

xPaw:
Is there a reason why this is called `acc` and not just `account` or something like that?

Even looking at the title `The ACC command framework` gives me no meaningful information at to what this framework is supposed to do.

DanielOaks:
The [account-notify spec](https://ircv3.net/specs/extensions/account-notify-3.1.html) already defines the `ACCOUNT` verb. I can change the title to something better regardless of the actual command name, though.

DanielOaks:
Bit of a better name, if we wanna brandy alternative suggestions about we can though.

DanielOaks:
Oragono now implements the updated spec as of the latest git

DanielOaks:
Mention from the channel on something this could use: "it has no provision afaict for serated verification measures (think "go to this URL and solve a captcha" or "generate this proof of work")"

DanielOaks:
This tries to do a lot and I think misses the mark in some ways and would probably be better getting rewritten. Especially this long with no real implementations sorta shows that it's probably better to get a more well-defined look at what specifically we should solve in a single spec (e.g. lump registration as well as authentication changing, maybe ^ and whatever other registration/cred-specific things come up), rather than trying to add on later with patch specs like this one's tried to.

I dunno, I'll submit something to -ideas along the lines of a problem statement so we can better nail down what we want out of this and see if there's any actual interest from the not-client side.

PR #277, opened 2016-11-15
Update message-tags and sts with WIP disclaimer, prefix cap names

attilamolnar:


lol768:
Discussed on IRC, seems like a good modification to me.

PR #282, opened 2016-11-20
Add example specification to assist new spec authors

DanielOaks:
This was originally proposed by @SaberUK in #229 but it didn't get in. We should have something like this to help guide authors to writing well-formed and somewhat similarly-formatted specs.


jwheare:
Better, but is a bit conflicted on whether it's an example or guidelines. And i don't think there should be commented sections, they're invisible in rendered markdown.

DanielOaks:
Updated the language. I think that now it's fine having just the _"Feel free to use this as a rough guideline while writing up your specification. However, keep in mind that the sections here are not set in stone."_ text there.

If we used more strong language than 'feel free' then I could see it being an issue, but as-is I think it conveys well-enough that it's just a suggestion.

attilamolnar:
Thoughts about adding the draft disclaimer recently added to some of the newer specs?

DanielOaks:
@attilamolnar Thanks, I added the draft section, added a non-normative Implementation Considerations section and an examples syntax that works better for multi-client examples.

It should be "messages" yeah, that's the language widely-used to refer to those by other specifications. I've clarified what messages are though, to make it more clear to people using this as a base.

I considered replacing the numerics in the examples, but I think the benefit of displaying the real, raw traffic outweighs the benefits gained by using the names instead. For instance, when I run into issues / while implementing I tend to go compare my raw traffic to the raw traffic in the example section, to make sure that the format and numerics are correct.

jwheare:
:+1: LGTM

jwheare:
Is this good to merge?

example-spec.md
```diff
@@ -0,0 +1,134 @@
...
+## Security Considerations
+
+New features may have security-specific considerations that authors should take note of
+while implementing your specification. This section lists those considerations in one place,
+so implementors can look over them.
+
+If no security considerations exist, you should include this text instead:
+
+There are no security considerations that implementations should note when implementing this
+specification.
```
attilamolnar: Do we really want this boilerplate in all specs unrelated to security?

example-spec.md
```diff
@@ -0,0 +1,134 @@
...
+New numerics that this spec defines are listed here. For example:
+
+| No. | Label          | Format                                                                  |
+| --- | ---------------| ----------------------------------------------------------------------- |
+| 001 | `RPL_WELCOME`  | `<client> :Welcome to the Internet Relay Network <nick>!<user>@<host> ` |
+| 002 | `RPL_YOURHOST` | `<client> :Your host is <servername>, running version <version> `       |
+
+### RPL_ISUPPORT Tokens
+
+`RPL_ISUPPORT` tokens that this spec defines are listed here.
```
attilamolnar: If there isn't any (like most specs) this section should be omitted. This is similar to the alternatives section which says omit if empty. Thoughts?

example-spec.md
jwheare: I think this can have the "can be omitted" text to.

DanielOaks: Sure, so long as people carefully consider it I'm happy

example-spec.md
DanielOaks: Yep, same can go for this and the Caps/Messages/Numerics sections closeby.

DanielOaks:
Cool, I'm happy for this to be merged in if there aren't any further comments?

jwheare:
@DanielOaks do you want to have a go at adding a note in the motivation section to address  ircv3/ircv3-ideas#7 before merge?

DanielOaks:
Sure, won't be a 100% defined thing but a 'you should consider X' thing. I'll throw in an update later today probably.

PR #260, opened 2016-05-05
sasl-3.1: Clarify when RPL_SASLMECHS is sent

attilamolnar:
Fixes issue #254.

PR #283, opened 2016-11-23
echo-message: Don't echo client messages sent to themselves

lp0:
Servers MUST NOT send `PRIVMSG` and `NOTICE` messages back to the
client when they send a message to themselves.

### DIFF
--- a/extensions/echo-message-3.2.md
+++ b/extensions/echo-message-3.2.md
@@ -18,7 +18,8 @@ This client capability MUST be named `echo-message`.
 If enabled, servers MUST send `PRIVMSG` and `NOTICE` messages back to
 the client that sent them. If servers apply any modifications to these
 messages, they MUST send the final version of the message back to the
-originating client.
+originating client. Servers MUST NOT send `PRIVMSG` and `NOTICE`
+messages back to the client when they send a message to themself.
 
 For clients, receiving a message with themselves as the sender acts as
 an acknowledgement that the message has been delivered to the server.
@@ -57,3 +58,9 @@ formatting and sends the final version back:
 
     --> PRIVMSG #ircv3 :back from \02lunch\0F
     :example!ex@example.com PRIVMSG #ircv3 :back from lunch
+
+
+# Errata
+
+* Previous versions of this specification did not describe what should
+happen when a client sends a message to themself.


ChickenNuggers:
```
<SaberUK> inspircd.org/replay-message (what echo-message is based on) doesnt send an echo message if they privmsg themselves
```

jwheare:
Since echo-message is a published spec, changes like this need to be checked against all existing implementations, and if it's deemed safe enough, an errata needs to be added. If not, we need to solve this with other means, for example the proposed labelled-replies spec #162

jwheare:
From a quick check of the support tables though, this looks pretty safe. Add an errata heading as in e.g. #260 and this LGTM.

lp0:
Done. Also `s/themselves/themself/g` in the added text as it's singular.

attilamolnar:
I do not see any reasoning behind this change. Please explain why is a backwards incompatible change to a published specification required and how is it justified.

DarthGandalf:
+1 to Attila.
How messages to themselves are any different from any other messages?

jwheare:
Without echo-message, clients can simply not print their own local version of a message when PMing to themselves, and only display the server sent version, to avoid duplicates.

With echo-message, you end up receiving two identical PRIVMSGs, one as the echo of the sent message, and one as the received message, so you'd end up having to perform string comparisons, or just skip every other message, neither of which are very robust.

As a client developer, I'd personally wait for this problem to be solved with #260, but it's still an oversight of the spec.

DanielOaks:
this change requires special-case code on both the server and the client to correctly interpret. why is this a special case where the client doesn't need a notification that their message was correctly sent? it makes sense to do the opposite and specify that the response must be sent back to the user, and particularly after labeled replies come in it'll mean everything can use one code path.

jwheare:
This is the one case where the client already gets notification that their message was sent without echo-message. Clients already have (straightforward) special case code paths to prevent dupes.

DarthGandalf:
> Clients already have (straightforward) special case code paths to prevent dupes.

Some clients maybe have such case. It's not a reason for all other clients to add a special case.

jwheare:
The clients that don't care can continue not to and would be unaffected by this change.

DanielOaks:
if clients already have code in-place to prevent duplicate messages from echo-message being displayed then this case is perfectly covered by that. in this case they should only show the message to the server, and them receiving one back from the server (particularly once echo-message becomes more robust with #162).

jwheare:
Without echo-message:

```
-> PRIVMSG me :hi
<- me!user@host PRIVMSG me :hi
```

With echo-message:

```
-> PRIVMSG me :hi
<- me!user@host PRIVMSG me :hi
<- me!user@host PRIVMSG me :hi
```

The special casing clients currently have is to *remove* a code path on their end (fake printing their own message). With echo-message they must add an extra messy code to determine which server sent messages to ignore.

DanielOaks:
With echo-message sending to someone else:
```
[sent msg] -> PRIVMSG dan :hi
[sent msg]<-  me!user@host PRIVMSG dan :hi
```

With echo-message sending to self:
```
[sent msg] -> PRIVMSG dan :hi
[sent msg]<-  dan!user@host PRIVMSG dan :hi
[received msg]<-  dan!user@host PRIVMSG dan :hi
```

this seems like it's the expected result from such a privmsg/notice?

jwheare:
If echo-message didn't send an extra message back to yourself then you could have the one holy grail code path: print all received PRIVMSGs and don't print any fake messages for yourself.

jwheare:
What does `[sent msg]<-` mean?

DanielOaks:
And if you print all received PRIVMSGs without printing any fake messages for yourself, right now it'll look like you sent a message and received one. with this change, it'll look like you sent one and got no response from the server.

The `[sent msg]` and `[received msg]` are the logic blocks that the messages are a part of. first sent + received msg is the logical block of "sent message" and the next received message is logical block of "received message".

jwheare:
After sleeping on this, and some discussion in IRC, I believe the current implied behaviour that even private messages to yourself should be echo-ed back is the correct one. Any difficulties clients might have correlating deduplicating this case should be solved with the #162 labelled replies spec (a client can ignore any labelled replies for self pms).

That spec is already needed to solve the similar issue that clients can't easily correlate and replace a temporary local message used to reduced perceived input lag, with the server echo'ed message.

If anything, the spec should clarify the opposite of what this PR says, but arguably that's already strongly implied.

Changing my vote to -1 and reverting to my first commend on this thread.

lp0:
> The [sent msg] and [received msg] are the logic blocks that the messages are a part of.

You're assuming the client has some magic to identify which two pairs of messages it's dealing with when it receives two of them. The specification states a second use case of "multiple clients". These additional clients won't be able to tell what is going on when they receive two messages that they've apparently sent to themselves.

If the server doesn't echo clients own messages to themselves then the client will still receive the message it sent and be able to handle it appropriately, whether that involves displaying it once or twice.

jwheare:
The multiple clients point is interesting. The labelled-replies spec should probably instruct bouncers to forward the the label tag to all attached clients. We should consider how that interacts with e.g. znc's route_replies. However, discussion for that should continue in the labelled replies spec, which is how this issue should be resolved. I'll raise it over there.

lp0:
There's also no guarantee that an ircd implementation couldn't interleave responses:
```
-> PRIVMSG myself :hello
-> PRIVMSG myself :world
<- :myself PRIVMSG myself :hello
<- :myself PRIVMSG myself :world
<- :myself PRIVMSG myself :hello
<- :myself PRIVMSG myself :world
```

There is no requirement to send responses back immediately. You'd also confuse a "multiple clients" scenario when they're both sending messages to themselves and get an expected message order.

jwheare:
Even if servers interleave (which I think the assumptions is they don't) this isn't a problem.

With labelled replies, your example would be (non-finalised tag)

```
-> @label=1 PRIVMSG myself :hello
-> @label=2 PRIVMSG myself :world
<- @label=1 :myself PRIVMSG myself :hello
<- @label=2 :myself PRIVMSG myself :world
<- :myself PRIVMSG myself :hello
<- :myself PRIVMSG myself :world
```

A multiple clients `Client1` and `Client2` talk to `Bouncer1` and `Bouncer2` and both bouncers talk to `Server` (via single bouncer connection to server) example:

```
Client1  -> @label=1/1 PRIVMSG myself :hello
Client2  -> @label=2/1 PRIVMSG myself :world
Server   <- @label=1/1 :myself PRIVMSG myself :hello
Bouncer1 <- @label=1/1 :myself PRIVMSG myself :hello
Bouncer2 <- @label=1/1 :myself PRIVMSG myself :hello
Server   <- @label=2/1 :myself PRIVMSG myself :world
Bouncer1 <- @label=2/1 :myself PRIVMSG myself :world
Bouncer2 <- @label=2/1 :myself PRIVMSG myself :world
Server   <- :myself PRIVMSG myself :hello
Bouncer1 <- :myself PRIVMSG myself :hello
Bouncer2 <- :myself PRIVMSG myself :hello
Server   <- :myself PRIVMSG myself :world
Bouncer1 <- :myself PRIVMSG myself :world
Bouncer2 <- :myself PRIVMSG myself :world
```

The client behaviour could simply ignore all @label tagged messages if nick == theirnick. This does highlight the need for bouncer clients to use non-overlapping labels. Perhaps another point of discussion for that spec.

This does mean that clients wishing to deduplicate self pms need to discard all temporary local messages in that case, as the labels are being ignored. Since those clients are already special casing self pms that shouldn't be an issue, but should be mentioned.

If a client doesn't wish to deduplicate self pms, then it only needs to ignore messages with labels that it doesn't recognise (again assuming each bouncer client uses a unique namespaced labelled).

jwheare:
Closing this issue in favour of further discussion in #162.

Issue #227, opened 2016-02-12
Server sent ids


highflyer77:
This would open up all kinds of possibilities of being able to reference a message by id #. 

It was discussed yesterday by: sagerdearia that: 

>  <sagerdearia>  Hi all. I have been noticing that LinkedIn has been gradually becoming similar to
>                 Facebook, including ability to like posts, and I thought I'd suggest that IRC v3 
>                 protocol should incorporate the ability to like privmsgs sent to channels and 
>                 users
- **a globally sequential id # , could make this a possibility.**

And just now a message by: @dequis on issue #226 

> I'd rather leave that up to the implementation. It just needs to be unique-ish. It could be a reference to a globally sequential id in a server side log database, for example
- **a globally sequential id # , could make this a possibility.**  

Finally I'd like to say that I can think of many uses for this and doing this could be useful for a lot of different kinds of implementations to make them a reality.

dequis:
When I mentioned globally sequential ids that was just an example, and one that isn't very representative of what servers will want to do. It was just meant to contrast against the idea of requiring "a hash of the message+username+timestamp". The only requirement should be uniqueness.

Globally sequential ids are terrible for many reasons. Ensuring that a sequence is kept in order across a distributed system is painful and/or very annoying. Keeping server side logs in a centralized database is rare and often undesirable for privacy reasons among others. That's something I can only imagine non-traditional servers will attempt to do, and even then they'd probably prefer databases that don't provide sequential IDs

>sagerdearia

Uhm I thought we had agreed that sagerdearia was a fairly obvious troll. Well, "troll" isn't the best way to describe it as that was light hearted. Still, terrible way to start a ticket.

Aerdan:
As long as the server can map it to an id, global uniqueness is not required; s2s is for servers to figure out, and they already have.

grawity:
> Ensuring that a sequence is kept in order across a distributed system is painful and/or very annoying.

What if the requirement was just "unique"? Allocate the 'counter' the same way as connection UIDs then, with server prefix + per-server counter.

> Keeping server side logs in a centralized database is rare and often undesirable for privacy reasons among others

You don't need logs, just a counter. Stash it in `ircd.tune`.

highflyer77:
@dequis 
>Keeping server side logs in a centralized database is rare
I didn't mean that servers should do that, but to only serialize the messages. The implementer could to the logging for what they were trying to accomplish.
@grawity 
[...]

PR #285, opened 2016-11-28
[message-ids] draft msg-id tag spec for server supplied message IDs.

jwheare:
Addresses #227


extensions/message-ids.md
```diff
@@ -0,0 +1,100 @@
...
+    S: @draft/msgid=63E1033A-051D-4B41-B1AB-1FA3CF4B243E :nick!user@host NOTICE #channel :Hello!
+
+A private `NOTICE` sent by the server:
+
+    S: @draft/msgid=D444852B-77EF-4829-AB6B-4D01C5036C01 :nick!user@host NOTICE me :Hello!
+
+A channel `PRIVMSG` sent by the server, and a possible client response. The `reply` tag is a non-standard example:
+
+    S: @draft/msgid=062F321D-425E-4951-A3E8-D644513E2ECD :nick!user@host PRIVMSG #channel :Hello!
+    C: @reply=062F321D-425E-4951-A3E8-D644513E2ECD :nick2!user2@host2 PRIVMSG #channel :Hello to you!
```
dequis: Prefix non-standard example tags with `example/`, like `@example/reply=...`. Same thing below.

jwheare: Not a bad plan.

extensions/message-ids.md
```diff
@@ -0,0 +1,100 @@
...
+
+In order to guarantee uniqueness, message IDs can't be implemented as simple numeric counters that risk clashing with other servers on the network, or being reset if the server restarts.
+
+Some examples of appropriate IDs are:
+
+* An [RFC 4122 UUID](https://tools.ietf.org/html/rfc4122)
+* An ID generated by a shared external provider, used by all servers on the network
+* A numeric counter, prefixed with a server ID, with the latest value saved externally to the IRCd on every issue, to allow for power failures.
+* A numeric counter, prefixed with a server ID as well as the server startup timestamp at a suitable precision.
+
```
DarthGandalf: * MD5 hash of one of the above.

extensions/message-ids.md
```diff
@@ -0,0 +1,100 @@
...
+Communication on IRC has historically been limited to a flat structure of sequential messages without unique identifiers shared across clients. As a result, the protocol lacks an ability to establish relationships between messages and other entities. This places limits on client-side enhancements such as reply tracking, message editing, rating or sending other forms of message annotation.
+
+The message ID tag is a way for servers to enable these enhancements.
+
+## Architecture
+
+### Dependencies
+
+The `draft/message-tags` capability MUST be negotiated for servers wishing to use this tag.
```
DarthGandalf: Typo in the cap name. It should be something like `msgid`, not `message-tags`

dequis: Pretty sure that's intended. The new message-tags-3.3 spec says clients should accept unknown tags without an associated cap, which allows servers to reuse messages. There's no cap for this spec.

extensions/message-ids.md
jwheare: It's not meant to be an exhaustive list. But any reason why you'd want to use a hash?

extensions/message-ids.md
DarthGandalf: Ah, got it.
Let's add an example with CAP negotiation step.

extensions/message-ids.md
jwheare: Hmm. My issue with md5 is that it's no shorter than a UUID, has no specific guarantee of uniqueness (just a low collision probability), and requires an already unique input so it doesn't really help an implementor. I can mix up the examples though.

grawity:
Considering that _every_ message would get this tag, suggesting an UUID in hex might be overkill... bad for both mobile bandwidth use, and for the 512-byte tag limit. Email can get away with ridiculous Message-IDs, but here the examples should IMHO be shorter.

("An ID generated by a shared external provider" probably refers to twitter/snowflake?)

extensions/message-ids.md
dequis: FWIW, Bitlbee uses MD5 for XMPP IDs. At startup it seeds the digest object with some bytes from /dev/urandom and each time it needs to generate an ID it increases a counter and appends its value to the digest object. Getting the hash out of it is essentially serializing its internal state, and the unfinalized digest is kept for the next time.

Seems to be very efficient, or at least, clever. But don't take this as a recommendation of that particular method, I haven't done any benchmarks.

But you got 128 bits so it's more or less the same guarantee of uniqueness as UUIDs

extensions/message-ids.md
```diff
@@ -0,0 +1,112 @@
...
+
+    S: @draft/msgid=8FD37885-C4B0-4977-A698-76786BFDDEB2;example/split :nick!user@host PRIVMSG #channel :Hello
+    S: @draft/msgid=DACE5EEB-9BCC-4597-85F4-AB13562992E8;example/concat=8FD37885-C4B0-4977-A698-76786BFDDEB2 :nick!user@host PRIVMSG #channel : World
+
+A client negotiating the `draft/message-tags` capability to enable and disable messages tagged with IDs.
+
+    S: :nick!user@host PRIVMSG #channel :Hello
+    C: CAP REQ draft/message-tags
+    S: :irc.example.com CAP me ACK :draft/message-tags
+    C: CAP END
```
dequis: Isn't CAP END pointless after connection registration is over?

extensions/message-ids.md
```diff
@@ -0,0 +1,112 @@
...
+
+### Tags
+
+This specification adds the `draft/msgid` tag, which has one required value.
+
+This tag MAY be sent by a server on `PRIVMSG` and `NOTICE` messages.
+
+#### Tag value
+
+The tag value is chosen by the server and MUST be unique, meaning any other message transmitted on the entire network at any time MUST NOT share the same value.
```
dequis: How about re-delivered messages, such as the ones in a chathistory batch? Getting repeated IDs for those makes sense, as they represent the exact same message that was sent before. Might be worth rewording to allow cases like this.

extensions/message-ids.md
jwheare: Yeah, but our logic is just send it after ACK every time (or start SASL). Extraneous in the examples though so I'll drop it.

extensions/message-ids.md
jwheare: That does result in a situation where you can get duplicate messages in a buffer that share an ID (by cycling more than once on a chathistory channel). So I guess clients need to be able to handle that and treat the most recent version as the target.

jwheare:
@grawity I was thinking of http://code.flickr.net/2010/02/08/ticket-servers-distributed-unique-primary-keys-on-the-cheap/ but yeah, similar concept to snowflake.

UUIDs aren't *that* long. `msgid=A4B9DA138BBC4F328386F46C2C6A562E` is 38 bytes, or 7.4% of the 512 allowance. The bandwidth/tag limit tradeoff is at the discretion of individual servers. UUIDs are probably the easiest standardised way to achieve uniqueness, with no additional state tracking required. I'd much rather recommend something that people are less likely to screw up.

If length becomes a problem, servers can always switch to a different scheme when it does, and I added some shorter examples already. We could maybe add a discussion of length/complexity tradeoffs to the server implementation considerations.

grawity:
> UUIDs aren't that long. msgid=A4B9DA138BBC4F328386F46C2C6A562E is 38 bytes

That's already different from your original examples. I'm not opposed to UUIDs in general, but there are more space-efficient ways of representing 128 bits than hex...

Though I guess if it ever becomes a problem we'll already have bigger limits in general. ¯\\\_(ツ)\_/¯

jwheare:
Well, the examples weren't explicitly concerned with shaving bytes. All I did was remove the draft/ and optional hyphens to simulate real world usage of a more space efficient uuid tag. Could update the examples to ditch hyphens too I suppose.

jwheare:
Any more issues with this or are we good to merge as a draft? I've done client and server implementations of this in IRCCloud now.

extensions/message-ids.md
```diff
@@ -0,0 +1,128 @@
...
+## Server implementation considerations
+
+This section is non-normative.
+
+In order to guarantee sufficient uniqueness, message IDs can't be implemented as simple numeric counters that risk clashing with other servers on the network, or being reset if the server restarts.
+
+Some examples of appropriate IDs are:
+
+* An [RFC 4122 UUID](https://tools.ietf.org/html/rfc4122)
+* An ID generated by a shared external provider, used by all servers on the network
```
DarthGandalf: For consistency, either add dot in the end of sentence, or remove it in following elements

extensions/message-ids.md
```diff
@@ -0,0 +1,128 @@
...
+* A numeric counter, prefixed with a server ID, with the latest value saved externally to the IRCd on every issue, to allow for power failures.
+* A numeric counter, prefixed with a server ID as well as the server startup timestamp at a suitable precision.
+
+When using timestamps, make sure they're correctly synchronised using NTP or similar.
+
+Although clients are required to treat IDs as case sensitive opaque values, servers might still choose a case insensitive ID format internally.
```
DarthGandalf: clients are REQUIRED, servers MAY

extensions/message-ids.md
```diff
@@ -0,0 +1,128 @@
...
+
+Certain message ID formats could present a length issue if other tags are being used. The trade-offs between length and complexity of implementation should be carefully considered when choosing an ID format.
+
+For instance, a UUID in hex form takes up 32 bytes (or 22 bytes in unpadded base64), and has widely available standardised implementations in many languages and no external dependencies. Conversely, a custom format can be much shorter, and might depend on an external service, disk I/O, additional state being tracked by the server, or untested proprietary code. 
+
+If the 512-byte tag limit is only being sparsely used overall, the simpler and longer format might be chosen. But if the server's capabilities put messages at a greater risk of reaching the limit, a shorter custom format might be more suitable.
+
+Another consideration is the chance of collisions. For example, some versions and implementations of the UUID specification rely on conditions that may not always produce collision-free outputs. However, if the probability is sufficiently low, a "practical" (if not guaranteed) uniqueness might be acceptable, given the requirement on clients to accept shared IDs.
```
DarthGandalf: This paragraph allows server to assign the same ID to every message and is in direct contradiction with what was stated before.

extensions/message-ids.md
dequis: But it's a non-normative section

extensions/message-ids.md
jwheare: Yeah how do we resolve this?

As I stated on IRC:

> [2016-11-29 19:18:13] <jwheare> there is a *bit* of tension in saying ids MUST be unique and then saying "well, there may be like a 0.00000000000000000000001 chance of a collision but don't worry about it"
> [2016-11-29 19:18:46] <jwheare> i don't really want to downgrade the uniqueness to a SHOULD, but i do want to allow people to use "practically" unique values even if they aren't 100% guaranteed

extensions/message-ids.md
jwheare: Yeah that's deliberate, the client requirements are specified using the correct shouting earlier on.

jwheare:
I added echo-message as a recommended requirement (SHOULD). As a client author, I also needed labeled-response to implement this, but since that hasn't been merged yet I left it out, and maybe should be covered in the echo-message spec directly.

Also fixed full stops.

DarthGandalf:
There will be a strange interaction between labelled responses and message IDs in case of echo-message.
```
-> @label=1 PRIVMSG you :hi
<- @label=1,msgid=A :nick PRIVMSG you :hi
```
Labels are unique for one connection only, IDs are unique for the whole network, so server has to assign the message an ID independently of its label.

extensions/message-ids.md
DarthGandalf: Or if the server doesn't satisfy the requirements of the specific UUID version they use. In that case they can either fix the code to satisfy the requirements (e.g. store the number on disk to reinitialize the UUID library with a large number than it got the last time), or change version.

jwheare:
Yeah I wasn't suggesting the label should match the ID, just that in order to deal with echo-message properly in user facing clients, you need labels as well.

extensions/message-ids.md
DanielOaks: I think explicitly bringing this up just muddies the spec and isn't required. It bring up questions around exactly what a 'practical level of uniqueness' is, and weakens the "unique" language that's used through it.

imo we can just leave it as-is without this paragraph, say make it unique, and if a server really wants to do this and have a fantastically low chance of a collision happening then they'll decide whether it's an acceptable tradeoff.

jwheare:
Alright, dropped that paragraph.

DanielOaks:
👍, LGTM

DarthGandalf:
LGTM

jwheare:
Barring any further responses, I'll merge this at the end of the coming week.

extensions/message-ids.md
```diff
@@ -0,0 +1,130 @@
...
+Software implementing this work-in-progress specification MUST NOT use the
+unprefixed `msgid` tag name. Instead, implementations SHOULD use
+the `draft/msgid` tag name to be interoperable with other software
+implementing a compatible work-in-progress version.
+
+The final version of the specification will use an unprefixed tag name.
+
+## Introduction
+
+This specification describes a mesage tag indicating a server supplied unique ID for `PRIVMSG` and `NOTICE` messages.
```
attilamolnar: Any reason this doesn't apply to other events as well such as all channel events, NICK and QUIT? Having a unique id on those events as well allows clients to deduplicate all messages in case of e.g. history playback.

jwheare: No real reason, since it's a MAY and there's no guarantee it'll appear at all (there's no CAP) we can just say all message events.

DarthGandalf:
I've been thinking how to implement this in ZNC, and it conflicts with echo-message.
```
CLI → ZNC: PRIVMSG foo :hi
ZNC → CLI: @draft/msgid=znc-11 cli!u@h PRIVMSG foo :hi
ZNC → SRV: PRIVMSG foo :hi
SRV → ZNC: @draft/msgid=srv-22 cli!u@h PRIVMSG foo :hi
ZNC → CLI: @draft/msgid=znc-11 cli!u@h PRIVMSG foo :hi
```
ZNC would need to remap all the IDs. Or it needs some way in the protocol to tell client that message `znc-11` is the same as `srv-22`.
What is more reasonable?

dequis:
@DarthGandalf I'd start by not sending an echo to the client if the server has echo-message enabled (you don't want to send echoes twice anyway). Not sure if variations of the issue are still present after solving that.

jwheare:
@DarthGandalf is there a reason ZNC needs to generate its own msgids? They're meant to be a shared reference for everyone on a network.

jwheare:
Dropped the section on length vs complexity. It's a bit of a moot point given the proposal in #287 to increase the tag space size limit.

extensions/message-ids.md
```diff
@@ -0,0 +1,128 @@
...
+Software implementing this work-in-progress specification MUST NOT use the
+unprefixed `msgid` tag name. Instead, implementations SHOULD use
+the `draft/msgid` tag name to be interoperable with other software
+implementing a compatible work-in-progress version.
+
+The final version of the specification will use an unprefixed tag name.
+
+## Introduction
+
+This specification describes a mesage tag indicating a server supplied unique ID for events.
```
MuffinMedic: Why is this limited to a server-supplied ID? The ability for clients to send a unique ID along with their message / event would be useful - it allows for the client to know the ID of the message it sent, which is required for proper processing of #288 

Rather than limit the ID supply to the server, it may be more useful to have the original sender generate the ID.

jwheare: Client-supplied IDs will have no guaranteed of uniqueness. Knowing the ID of your own messages is provided by echo-message.

DarthGandalf: s/mesage/message/

extensions/message-ids.md
```diff
@@ -0,0 +1,128 @@
...
+
+### Dependencies
+
+This specification doesn't define any capabilities of its own, but the `draft/message-tags` capability MUST be negotiated for servers wishing to use this tag.
+
+### Tags
+
+This specification adds the `draft/msgid` tag, which has a required value.
+
+Servers MAY attach this tag on any event. If this tag is being used, it SHOULD be attached to all `PRIVMSG` and `NOTICE` events.
```
DarthGandalf: What are events?
For example, should nicks in response to `NAMES` have own IDs? What about `MONITOR`? Response to remote `WHOIS`? `WALLOP`?

extensions/message-ids.md
```diff
@@ -0,0 +1,128 @@
...
+
+#### Tag value
+
+The tag value is chosen by the originating server and MUST be unique, meaning any other message transmitted on the entire network at any time MUST NOT share the same value.
+
+However, if a message is re-transmitted as-is, for example with the [`chathistory`](./batch/chathistory-3.3.html) batch type, the ID SHOULD be reused. As a result, clients MUST accept shared IDs.
```
DarthGandalf: If a bouncer transparently translates messages between two languages (English to Russian), should ID stay the same?

extensions/message-ids.md
```diff
@@ -0,0 +1,128 @@
...
+
+When using timestamps, make sure they're correctly synchronised using NTP or similar.
+
+Although clients are required to treat IDs as case sensitive opaque values, servers might still choose a case insensitive ID format internally.
+
+## Client implementation considerations
+
+This section is non-normative.
+
+Message IDs have no guarantee of being universally unique across different IRC networks, nor will they necessarily share the same format. There is also no requirement that numeric IDs increase monotonically. Don't attempt to correlate them beyond their scope and don't use them for message ordering.
```
DarthGandalf: FWIW, it can prevent linking two smaller networks into one bigger network, if there is a chance that the same ID was generated by both networks in the past.

Related: during temporary netsplit, I assume the parts of network are still "single network".

extensions/message-ids.md
jwheare: It's up to the server, no reason why not to. The client needs to handle messages with ids on a case by case basis, there's no guarantee that they will be sent on any message as there's no CAP to enable this beyond the generic `message-tags` CAP.

extensions/message-ids.md
jwheare: If all clients are expected to treat the message as the same (e.g. for the sake of consistent replies) then yes.

extensions/message-ids.md
jwheare: If two networks join, they could add unique server prefixes to their ID space to avoid collisions.

DarthGandalf: They can't add such prefixes to all previously sent messages.

1 Jan: irc.example.com: `@msgid=AXM5 :a JOIN #channel`
2 Jan: irc.example.net: `@msgid=AXM5 :b PRIVMSG #chan :Hi`
3 Jan: these 2 servers join. They both have some kind of published logs, where one could lookup the message by ID. Someone tries to lookup message AXM5, and sees ...

DarthGandalf:
To clarify my position: I'm not opposed to the spec, I'm brainstorming various possible future issues. Some of them are non-issues.

extensions/message-ids.md
jwheare: Ah I see what you mean. There are probably some id schemes that mitigate that better than others. I suppose it would require forward thinking on behalf of the servers to use those.

This is also mitigated somewhat by the recommendation to clients.

> Handling duplicates gracefully is also useful in the case of servers accidentally reusing an ID

This sort of thing is why I wanted to keep a section on the probability of collisions. Mentioning the possibility of your example occurring might help servers to choose a more appropriate id scheme.

DarthGandalf: Yes, I meant that IRC client is not the only entity which needs to handle possible collisions, but also such server-side viewer of logs, which is (probably) outside of IRC protocol.

jwheare: Do you think it's worth bringing back the paragraph dropped in 7183c2839cafe344684f324b5bddaaa299f8ad71, with a specific warning about this?

DarthGandalf: With a different focus. Probability of collisions in UUID implementation is not really a concern IMHO. But if servers do use UUID as message ID, they are protected from this issue too.

extensions/message-ids.md
DarthGandalf: May this be mentioned in some example?

jwheare: Which part of the spec suggests that this wouldn't be the case? Or what would be the best wording or placement of such a clarification?

extensions/message-ids.md
jwheare: How about this? 002283378399fcd5e67ee14a4f9372c59093ca71

PR #286, opened 2016-12-06
Make SASL-3.1 spec more understandable.

DarthGandalf:


DarthGandalf:
@grawity It doesn't. It rewords that. As I understand the existing spec _now_, `AUTHENTICATE +` is used not only for empty responses, but also for responses of length 400, 800, 1200, etc.
This commit only encodes this my assumption to an actual text in a readable form.

extensions/sasl-3.1.md
```diff
@@ -42,16 +42,21 @@ is received, the IRCD will attempt to establish an association with a SASL
...
 
 A set of regular AUTHENTICATE messages transmits a response from client to
 server or vice versa. The server MAY intersperse other IRC protocol messages
-between the AUTHENTICATE messages of a set. The "+" form is used for an empty
-response. The server MAY place a limit on the total length of a response.
+between the AUTHENTICATE messages of a set.
+The response is encoded in Base64
+([RFC 4648](https://tools.ietf.org/html/rfc4648)), then split to 400-byte
+chunks. If length of response divides by 400 (e.g. if the response is empty), an
+additional `AUTHENTICATE +` message is transmitted in the end.
```
grawity: Every time I re-read this, it seems _more_ confusing than before. Yes, `0 % 400 == 0`, but it doesn't seem to translate well.

Regardless of the above, I think it should also be clarified that the length is of the _encoded_ response...

How about:

> ... then split to 400-byte chunks, and each chunk is sent as a separate `AUTHENTICATE...` command. Empty (zero-length) responses are sent as `AUTHENTICATE +`. If the last chunk was exactly 400 bytes long, it must also be followed by `AUTHENTICATE +` to signal end of response.

ChickenNuggers: :+1:

PR #292, opened 2017-01-02
Add the SCROLLBACK batch type specification.

MuffinMedic:


jwheare:
Interesting spec. Can you elaborate why `msgid` is required?

prawnsalad:
As mentioned on IRC I've been getting this idea into the next version of Kiwi IRC. Something that stands out during implementation:
> No acknowledgement by the server of the command is required other then returning the requested content. 

It would be tidier if this was required to send an empty batch if there is no content to be returned. Just so the client doesn't sit there waiting for nothing.

MuffinMedic:
@jwheare The `draft/msgid` is to allow for usage of the `draft/reply` tag in servers where multiple users may have the same scrollback content and a user wishes to reply to an older message. The `draft/msgid` would be that which the server originally sent to clients. If a user replies to a scrollback message, other users with the same history can scrollback as well and see the original message replied to.

@prawnsalad Agreed & updated.

dequis:
Thanks for taking the time to write a spec for this!

This should probably be together with the other extensions, not just a batch type. In fact, I don't see a strong reason for this to use a *new* batch type, if it's just chathistory plus event replay. 

Please read the previous discussion on why event replay wasn't added as part of the chathistory batch: #156 - the short version is that the way th

>to allow for usage of the `draft/reply` tag

While I agree that sounds desirable and that all messages from history *should* have an ID (also helps with deduplication), there's no need to make that into a requirement.

MuffinMedic:
@dequis Do you suggest the extension be implemented as an RPL_ISUPPORT 005 command or as the `scrollback` capability? The original implementation I had was as a CAP but initial feedback suggested to simply send the 005 numeric instead.

> While I agree that sounds desirable and that all messages from history should have an ID (also helps with deduplication), there's no need to make that into a requirement.

Agreed. Updated to make `draft/msgid` recommended but optional.

dequis:
Yeah just an ISUPPORT token is fine for this.

prawnsalad:
The changes to use `chathistory` makes it more confusing now. Sending the `SCROLLBACK` command to get a `chathistory` batch type now uses 2 different terms to refer to the same thing - scrollback.

Would there be anything stopping us from using a `CHATHISTORY` command instead to keep things consistent?

extensions/scrollback-3.3.md
```diff
@@ -0,0 +1,70 @@
...
+The examples below are written with `draft/msgid` tags included. This tag is optional but recommended.
+
+### Begin
+    :irc.host BATCH +ID chathistory target
+    :irc.host BATCH +XNyDSitp9MvcX scrollback #channel
+### PRIVMSG
+    @batch=ID;draft/msgid=UUID;time=YYYY-MM-DDThh:mm:ss.sssZ :nick!ident@host PRIVMSG target :message
+    @batch=XNyDSitp9MvcX;draft/msgid=eb703092-0782-4c28-bf7c-f9e5c45963ca;time=2016-11-19T18:02:01.001Z :foo!bar@example.com PRIVMSG #channel :The SCROLLBACK specification is going to be fantastic!
+### NOTICE
+    @batch=ID;draft/msgid=UUID;time=YYYY-MM-DDThh:mm:ss.sssZ :nick!ident@host NTOICE target :message
```
SaberUK: Typo.

extensions/scrollback-3.3.md
```diff
@@ -0,0 +1,70 @@
...
+
+The `server-time` should be the time at which the message was originally sent, the `batch id` a randomly generated string unique to the entire batch, and the optional `draft/msgid`, if included, the `draft/msgid` originally sent with the message.
+
+### `SCROLLBACK` Command
+Scrollback content can be requested by the client to the server by sending the `SCROLLBACK` command to the server. A `batch` must be returned by the server. If no content exists to return, an empty batch should be returned to avoid the client waiting for a reply. Command support is sent to the client as the RPL_ISUPPORT 005 numeric `:irc.host 005 nick SCROLLBACK :are supported by this server`
+
+#### Format
+The `scrollback` content can requested using timestamps:
+
+    SCROLLBACK target timestamp=YYYY-MM-DDThh:mm:ss.sssZ message_count
```
SaberUK: Wouldn't it be better to make it two parameters rather than using one separated with `=`?

extensions/scrollback-3.3.md
```diff
@@ -0,0 +1,70 @@
...
+
+A method for securely identifying the requesting user must exist to ensure content is sent only to the appropriate users and clients. See below for more information.
+
+## Security considerations
+Secure identification of users and clients must exist in order to ensure that users cannot obtain history that does not belong to them. Use of account names, internal account identifiers, or certificate fingerprints should be strongly considered when matching content to users. The server must verify the current user's identity matches that of the desired content. This information is not sent as part of the `SCROLLBACK` command and must be validated via other means, such as those stated above.
+
+## Current Implementations
+A ZNC-module implementation exists as [znc-scrollback](https://github.com/MuffinMedic/znc-scrollback). Client-side support for the ZNC module is in progress. Interest in supporting the batch type and command without ZNC has also been obtained, although specifics are not available at this time.
+
+[batch]: batch-3.2.md
```
SaberUK: This should link to the website pages for those specs rather than the markdown documents.

extensions/scrollback-3.3.md
SaberUK: Also:

- Can `message_count` be negative to go backwards?
- Should the max `message_count` be in the ISUPPORT token?
- What numerics are sent if `message_count` is invalid?

extensions/scrollback-3.3.md
MuffinMedic: Fixed in d261046

extensions/scrollback-3.3.md
MuffinMedic: Fixed in d261046

MuffinMedic:
@prawnsalad Fixed in d261046 - Use `chathistory` as both extension and batch type for consistency.

SaberUK:
I'm wondering if for the purpose of this specification we might be better off introducing a new 'HISTORY' batch type which works similar to `CHATHISTORY` but for all messages instead of just `PRIVMSG`/`NOTICE`.

extensions/scrollback-3.3.md
MuffinMedic: > Wouldn't it be better to make it two parameters rather than using one separated with `=`?

No preference either way. Any reason two parameters are better?

> Can `message_count` be negative to go backwards?

Backwards in what sense? Send messages in reverse order or else? What is the use case?

 > Should the max message_count be in the ISUPPORT token?

`:irc.host 005 nick CHATHISTORY=message_count :are supported by this server`?

> What numerics are sent if message_count is invalid?

Oversight and not considered. Suggestions?

dequis:
>a new 'HISTORY' batch type which works similar to chathistory but for all messages instead of just PRIVMSG/NOTICE

A new batch type will have the same problems the chathistory batch had. Batch types by themselves can't introduce the sort of behavior changes we need to stop (for example) a QUIT from the past making someone quit in the present.

We haven't talked much about how to do the event replay thing, maybe it's worth doing it in a new issue. But the ideas i have in mind could reasonably be added retroactively to chathistory.

Zarthus:
I have some serious concerns about how this is going to be properly handled security wise.

I think for bouncers this'll be great, but for an ircd I can think of many ways in which this can be abused, and i'm not sure to which point it would be a violation of privacy and where the line'll be drawn. Logs would be very unlikely to have the state of a channel knowledge.

Some of the below points don't necessarily need to be mentioned in the spec, but I'm curious on how some things would be handled nevertheless.

----

I would consider getting some numerics for errors (to which point it is relevant to IRCv3): What should happen if I do a `/CHATHISTORY #channel-i'm-not-in` or `/CHATHISTORY #secret-channel`? Personally, I think access should be rejected, but the specification gives no guidance on how to do this.

----

- User1 sends STATUSMSG(@) to `#coolchannel`
- User1 ops User2 in `#coolchannel`
- User2 does a `/CHATHISTORY #coolchannel` 

Should they receive the opnotice they were not opped for?
I think it'll be hard to keep track of every single edge case.

----

- User1 accidentally pastes their password, email, or address. While they change it, do we want to permit a way to delete chathistory? Should the user be able to do it for their own messages?
  - This doesn't have to be the user themselves, plenty of doxxing attempts happen on IRC.

---- 

- Some channels have confidentiality settings (think +mz, or +qz on charybdis)
- opped User1 talks with deopped User{2,3,4}
- Spam stops. -mz gets set.
- Should User2 be able to request chathistory for the logs of when +mz was enabled?

----

- coolserver.coolnework.net splits from hub.coolnetwork.net
- User{1,2,3} chat on coolserver.* whilst split, User{4,5,6} chat on hub.*
- coolserver.* rejoins hub.*
- User1 wants to view the chat on hub.*'s chat, do they have to reconnect to hub to see it?

With that, an optional `server` parameter might be worth considering. Although with different specifications (like `server-time`) some sort of resync between messages for split servers could also exist in the future.

----

- Services are down. My channel is SET RESTRICTED.
- User2 joins my channel, and requests a CHATHISTORY.

----

With the above points, I really think that for a bouncer oriented specification, it'll be great. And while I'd like to keep the option for ircds to implement it, I have a hard time seeing how they would cover some of the problems. I imagine some sort of opt-in mechanic could work, but this comment is already getting quite long and it's a bit too much outside of the IRCv3 realm of influence.

MuffinMedic:
ircv3/ircv3-ideas#4

MuffinMedic:
#162

extensions/chathistory-3.3.md
```diff
@@ -0,0 +1,79 @@
...
+
+When a client with the above mentioned capabilities requests `chathistory` content from the server (using the `CHATHISTORY` command outlined below), the server should return to the client a single batch containing a number of desired raw IRC lines equal to the `message_count` parameter specified, ending directly before the given timestamp or with the message directly proceeding the one with the specified `draft/msgid`. The raw IRC lines should be formatted and returned to the client as they were originally, with the addition of the above capability tags.
+
+The returned `batch` must not be blank. If no content exists to return, the appopriate error message should be sent instead.
+
+Both the `message_count` and `max_message_count` must be integers greater than 0. The client should not request a `message_count` greater than the `max_message_count` parameter sent in the command. If the `message_count` exceeds the `max_message_count`, server should return a number of lines equal to the `max_message_count`.
```
DarthGandalf: Move this below the `CHATHISTORY` Command header.
Before reading syntax of the command `message_count` and `max_message_count` don't make sense for the reader.

extensions/chathistory-3.3.md
```diff
@@ -0,0 +1,79 @@
...
+#### Format
+The `chathistory` content can requested using timestamps:
+
+    @draft/label=ID CHATHISTORY target timestamp=YYYY-MM-DDThh:mm:ss.sssZ message_count
+
+Alternatively, content can be requested using a `draft/msgid`:
+
+    @draft/label=ID CHATHISTORY target draft/msgid=ID message_count
+
+If no message_count is known, `*` should be used to specify the default value.
```
DarthGandalf: What example use case for this do you have in mind?

extensions/chathistory-3.3.md
```diff
@@ -0,0 +1,79 @@
...
+
+#### Errors
+If the server receives an improperly formatted `CHATHISTORY` command, the `CMD_INVALID` error code should be returned.
+
+If no `chathistory` exists to return, the server should return the appropriate error code. `ACCESS_DENIED` should be sent if the user requests content they do not have permission to view. `NOT_FOUND` should be sent if no content matching the request is found and no `ACCESS_DENIED` error exists.
```
DarthGandalf: What if client requested 40 lines, but server allows access only to 20 (which happened after client JOINed the channel), should it silently decrease size of batch, or report `ACCESS_DENIED` somewhere?

extensions/chathistory-3.3.md
```diff
@@ -0,0 +1,79 @@
...
+
+### Examples
+The examples below are written with `draft/msgid` tags included. This tag is optional but recommended.
+
+#### Begin
```
DarthGandalf: Is it all one example? Why does it have the same msgid for two messages?

extensions/chathistory-3.3.md
MuffinMedic: Fixed in 78b39f1

extensions/chathistory-3.3.md
MuffinMedic: It was a leftover from an earlier version of the draft spec - removed in 78b39f1

extensions/chathistory-3.3.md
MuffinMedic: One example with generic values - the IDs are a simple placeholder, not specific values.

extensions/chathistory-3.3.md
MuffinMedic: Addressed in 	78b39f1 - the `max_message_count` should be returned an `WARN MAX_MESSAGE_COUNT_EXCEEDED` returned to let the client know the value was too high.

prawnsalad:
The latest changes now forbids an empty BATCH response if there is no content to return, and instead forces an error to be returned. However if I request `CHATHISTORY` for a channel that has no more content then this isn't an error - there's just no more content. An empty BATCH is more descriptive and easier to handle here.

It now also states that `draft/labeled-response` is required when this could be easily be optional but recommended such as `draft/msgid`. Relying on any `draft/*` specs could hold back any clients wanting to implement `CHATHISTORY` for some time.

LFP6:
@Zarthus While a little belated...
> I'm not sure to which point it would be a violation of privacy and where the line'll be drawn

I think this is something best determined by the server. What's appropriate on an internal server is potentially much different than on an open network. I would find this to be true for history as a whole as well (personally, I'd find a channel mode for history to be warranted).

Zarthus:
@LFP6, indeed, we covered it further in https://github.com/ircv3/ircv3-ideas/issues/7 if I recall correctly.

extensions/chathistory-3.3.md
```diff
@@ -8,6 +8,13 @@ copyrights:
     period: "2017"
     email: "evan@muffinmedic.net"
 ---
+## Notes for implementing work-in-progress version
+This is a work-in-progress specification.
+
+Software implementing this work-in-progress specification MUST NOT use the unprefixed `CHATHISTORY` command. Instead, implementations SHOULD use the `draft/CHATHISTORY` command to be interoperable with other software implementing a compatible work-in-progress version.
```
Zarthus: in all caps? None other capabilities are capitalized are they?

MuffinMedic: The extension itself is lowercase. Only thee command itself is capitalized for consistency (CHATHISTORY, PRIVMSG, NOTICE, etc...).

LFP6:
Could this be extended so that you can request a certain number of messages to either side of a given timestamp/id? I'm imagining ircv3/ircv3-ideas#9 being implemented in a client so that you can jump back to that point in history (as Slack and Discord do).

Zarthus:
Extensions to the spec would probably make more sense as a separate specification, rather than to have one do-it-all spec full of "this is all an optional part of the spec".

MuffinMedic:
Adding "x lines before and after" would introduce an unneeded level of complexity as it requires an additional parameter for specify the "direction" of the history. I think a better solution is allowing the `message_count` to accept both positive and negative integers, where a positive value means "x lines after the timestamp" and a negative number represents "x lines before the timestamp". 

A search-type spec could simply request content both before and after. This would give a simpler yet more versatile approach I feel, although I'd like to see what others say before implementing it into the spec.

LFP6:
That seems good to me. While less concise, it also means that the server doesn't need to parse through the history nor does the client need to parse through the response to determine what comes before and what comes after.

jwheare:
Can we get an example of a full client/server exchange for this?

extensions/chathistory.md
```diff
@@ -0,0 +1,88 @@
...
+
+When a client with the above mentioned capabilities requests `chathistory` content from the server (using the `CHATHISTORY` command outlined below), the server should return to the client a single batch containing a number of desired raw IRC lines equal to the `message_count` parameter specified, ending directly before the given timestamp for a negative `message_count`, after the number of messages specified for a positive `message_count`, or with the message directly proceeding the one with the specified `draft/msgid`. The raw IRC lines should be formatted and returned to the client as they were originally, with the addition of the above capability tags.
+
+The `server-time` should be the time at which the message was originally sent and the `batch id` a unique ID to the entire batch. `draft/label` SHOULD be inclued and MUST be a unique ID used to identify the `chathistory` request and any replies. `draft/msgid` SHOULD be the `draft/msgid` originally sent with the message.
```
Zarthus: inclued

kylef:
I spotted https://github.com/ircv3/ircv3-specifications/issues/293 which states:

> The chathistory batch is currently limited to PRIVMSG and NOTICE and has no way to some events because they may be processed by the client in a way that affects the present channel state.

I don't see this actually mentioned anywhere so I think it should be mentioned that only `PRIVMSG` and `NOTICE` messages can be apart of the chat history.

extensions/chathistory-3.3.md
```diff
@@ -0,0 +1,79 @@
...
+
+The `server-time` should be the time at which the message was originally sent and the `batch id` a unique ID to the entire batch. `draft/label` is recommended and should be a unique ID used to identify the `chathistory` request and any replies. Also recommended, `draft/msgid` should be the `draft/msgid` originally sent with the message.
+
+### `CHATHISTORY` Command
+Chathistory content can be requested by the client to the server by sending the `CHATHISTORY` command to the server. A `batch` must be returned by the server. If no content exists to return, an empty batch should be returned to avoid the client waiting for a reply. Command support is sent to the client as the RPL_ISUPPORT 005 numeric `:irc.host 005 nick CHATHISTORY=max_message_count :are supported by this server`
+
+Both the `message_count` and `max_message_count` must be integers greater than 0. The client should not request a `message_count` greater than the `max_message_count` parameter sent in the command. If the `message_count` exceeds the `max_message_count`, server should return a number of lines equal to the `max_message_count` and the appropriate warning as described below.
```
MuffinMedic: Agreed - Will update accordingly to set 0 = no limit.

extensions/chathistory.md
```diff
@@ -0,0 +1,88 @@
...
+Software implementing this work-in-progress specification MUST NOT use the unprefixed `CHATHISTORY` command. Instead, implementations SHOULD use the `draft/CHATHISTORY` command to be interoperable with other software implementing a compatible work-in-progress version.
+
+The final version of the specification will use an unprefixed command.
+
+## Description
+This document describes the format of the `chathistory` extension. The `chathistory` extension uses the [chathistory][batch/chathistory] batch type, adding a client-side command for requesting `chathistory` content from the server.
+
+The `chathistory` extension adds an optional `draft/msgid` to the `chathistory` batch reply.
+
+Clients MUST support the [`batch`][batch] and [`server-time`][server-time] capabilities. Clients SHOULD support the [`draft/labeled-response`][draft/labeled-response] and [`draft/msgid`][draft/msgid] capabilities.
```
MuffinMedic: My understanding is that CAPS are negotiated before the ISUPPORT is sent - if the CAPs are not supported (as required by `MUST`), then `CHATHISTORY` should not appear in the 005. This is already standard behavior to my best of my knowledge.

Will update capability URL.

extensions/chathistory.md
```diff
@@ -0,0 +1,88 @@
...
+
+The `server-time` should be the time at which the message was received from the IRC server and the `batch id` a unique ID for the entire batch. `draft/label` SHOULD be included and MUST be a unique ID used to identify the `chathistory` request and any replies. A `draft/msgid` to identify each individual message MUST be the `draft/msgid` included when each message was first received from the IRC server.
+
+### `CHATHISTORY` Command
+Chathistory content can be requested by the client to the server by sending the `CHATHISTORY` command to the server. A `batch` must be returned by the server. If no content exists to return, an empty batch should be returned to avoid the client waiting for a reply. Command support is sent to the client as the RPL_ISUPPORT 005 numeric `:irc.host 005 nick CHATHISTORY=max_message_count :are supported by this server`
```
MuffinMedic: Yes, it's required. As per https://github.com/ircv3/ircv3-specifications/pull/292#discussion_r96109865, it will be updating such that 0 can be used to indicate no limit exists.

extensions/chathistory.md
```diff
@@ -0,0 +1,88 @@
...
+
+Both the `message_count` and `max_message_count` MUST be non-zero integers and `max_message_count` MUST be positive. The client should not request a `message_count` with an absolute value greater than the `max_message_count` parameter sent in the command. If the `message_count` absolute value exceeds the `max_message_count`, server should return a number of lines equal to the `max_message_count` and the appropriate warning as described below.
+
+#### Format
+The `chathistory` content can requested using timestamps:
```
MuffinMedic: This is required to account for certain use cases in which `draft/msgid` tags are not available, such as in ZNC 1.6+ logs, in the format of `YYYY-MM-DD.log [HH:mm:ss]`.

extensions/chathistory.md
```diff
@@ -0,0 +1,88 @@
...
+
+    @draft/label=ID CHATHISTORY target timestamp=YYYY-MM-DDThh:mm:ss.sssZ message_count
```
MuffinMedic: The format given is required for consistency with the  [server-time](http://ircv3.net/specs/extensions/server-time-3.2.html) extension.

extensions/chathistory.md
```diff
@@ -0,0 +1,88 @@
...
+
+Alternatively, content can be requested using a `draft/msgid`:
+
+    @draft/label=ID CHATHISTORY target draft/msgid=ID message_count
```
MuffinMedic: What would the standard format be for the `timestamp=YYYY-MM-DDThh:mm:ss.sssZ` option? The use of `server-time` for either seems inappropriate.

extensions/chathistory.md
```diff
@@ -0,0 +1,88 @@
...
+    @batch=ID;draft/msgid=ID;time=YYYY-MM-DDThh:mm:ss.sssZ :nick!ident@host NOTICE target :message
+#### End
+    :irc.host BATCH -ID
+#### Error
+    @draft/label=ID :nick!ident@host CHATHISTORY ERR ERROR_CODE
+#### Warning
+    @draft/label=ID :nick!ident@host CHATHISTORY WARN WARN_CODE
+
+## Use Cases
+The batch type and supporting command are useful for allowing an "infinite scroll" type capability within the client. A client will, upon scrolling to the top of the active window or a manual trigger, may request `chathistory` from the server and, after receiving returned content, append it to the top of the window. Users can repeat this historic scrolling to retrieve prior history until limitations are met (see below).
```
MuffinMedic: > If `message_count` is positive, content MUST be retrieved from after and not including the specified timestamp or `draft/msgid`. If the `message_count` is negative, content MUST be retrieved from before the specified timestamp or `draft/msgid`.

The client is sending a known message as a starting point. If the oldest message a client has is at a particular time, this timestamp will be provided to `CHATHISTORY` and anything before it will be returned. The same goes for `draft/msgid`.

This works in the forward direction as well, in which the client knows the last message it has and wants to retrieve content from directly after it.

Theoretically, this works for any known message as a reference, and could be used to add context if a client wants to "jump" to a message, and then request X lines before and Y lines after.

extensions/chathistory.md
```diff
@@ -0,0 +1,88 @@
...
+
+## Limitations
+Logging of messages and other actions must be enabled server-side and can be stored in any format desired, given appropriate software exists to retrieve and format such stored contnet. Scrollback can only be retrieved as far back as logs exist for the requesting user in the specified channel or query.
+
+A method for securely identifying the requesting user must exist to ensure content is sent only to the appropriate users and clients. See below for more information.
+
+## Security Considerations
+Secure identification of users and clients must exist in order to ensure that users cannot obtain history that does not belong to them. Use of account names, internal account identifiers, or certificate fingerprints should be strongly considered when matching content to users. The server must verify the current user's identity matches that of the desired content. This information is not sent as part of the `CHATHISTORY` command and must be validated via other means, such as those stated above. Access must be checked first and return an `ACCESS_DENIED` error as appropriate. If no `ACCESS_DENIED` error exists, the server can check for returnable content.
```
MuffinMedic: > Secure identification of users and clients must exist in order to ensure that users cannot obtain history that was not previously available to them at the time the messages were originally sent.

Using the reworded version:

> Could I not request the chat history for a public channel on a server?

While you can certainly request `chathistory` from a public channel, you should not be able to request history if you weren't in the channel at the time the messages were originally sent. Allowing users to obtain history for channels they never used or are no longer joined to would essentially allow people to "spy" on channels.

MuffinMedic:
`PRIVMSG` and `NOTICES` aren't limitations with this specification - it derives from limitations with the batch type, which this extension requires be supported. This is a dependency limitation.

If #293 addresses these issues and `batch/chathistory` no longer has these limitations, then `CHATHISTORY` would automatically support the additional event types.

extensions/chathistory.md
```diff
@@ -0,0 +1,88 @@
...
+    @draft/label=ID CHATHISTORY target draft/msgid=ID message_count
+
+If `message_count` is positive, content MUST be retrieved from after and not including the specified timestamp or `draft/msgid`. If the `message_count` is negative, content MUST be retrieved from before the specified timestamp or `draft/msgid`.
+
+#### Errors and Warnings
+If the server receives an improperly formatted `CHATHISTORY` command, the `CMD_INVALID` error code should be returned.
+
+If the absolute value of `message_count` exceeds the `max_message_count`, warn code `MAX_MSG_COUNT_EXCEEDED` should be returned. The command should continue to be processed as described above.
+
+If no `chathistory` exists to return, the server should return the appropriate error code. `ACCESS_DENIED` should be sent if the user requests content they do not have permission to view.
```
MuffinMedic: I wil get rid of it due to other privacy concerns.

extensions/chathistory.md
kylef: > While you can certainly request chathistory from a public channel, you should not be able to request history if you weren't in the channel at the time the messages were originally sent.

What would the point of requesting history if I have already seen the history because I was in the channel?

I think this should be down to networks/implementors, as a network might decide that channels without `+p` (private) and `+s` (secret) have publicly available logs. Or some other reasons behind it.

MuffinMedic: > you should not be able to request history if you weren't in the channel at the time the messages were originally sent.

>> What would the point of requesting history if I have already seen the history because I was in the channel?

This limitation exists to prevent retrieval of content you *weren't* in the channel for. If you were in there, then that qualifies as content "previously available to them."

extensions/chathistory.md
kylef: > I will specify the target as a single channel or query.

Okay if there is no wildcard, I am not sure how a client would know about the unread queries they have opened for individual nicks. Is there anyway a client would be able to load the available chat history targets?

[znc-playback](http://wiki.znc.in/Playback) currently allows the wildcard so that clients can ask for all queries during launch so that they can load the nick queries.

extensions/chathistory.md
kylef: > This limitation exists to prevent retrieval of content you weren't in the channel for.

Shouldn't this be a consideration on an IRC server and not rules enforced by the specification? An implementation of `CHATHISTORY` might want to provide public history for some channels. For example some private IRC server at a company where all employees could join later and view existing history.

MuffinMedic: Good point. Will update to:

> Secure identification of users and clients MUST exist in order to ensure that users cannot obtain history they are not authorized to view.

extensions/chathistory.md
MuffinMedic: `CHATHISTORY` is not intended to be a replacement for a playback buffer. Unless the spec is reworked, it currently has no concept of "read messages".

extensions/chathistory.md
```diff
@@ -0,0 +1,97 @@
...
+
+    @draft/label=ID CHATHISTORY <target> <start> <end>
+
+The `chathistory` content can requested using timestamps:
+
+    @draft/label=ID CHATHISTORY <target> timestamp=YYYY-MM-DDThh:mm:ss.sssZ message_count=<message_count>
+
+Alternatively, content can be requested using a `draft/msgid`:
+
+    @draft/label=ID draftCHATHISTORY <target> draft/msgid=<message_id> timestamp=<timestamp>
```
SaberUK: draftCHATHISTORY?

extensions/chathistory.md
```diff
@@ -0,0 +1,97 @@
...
+If the server receives a`CHATHISTORY` command with missing parameters, the `ERR_NEEDMOREPARAMS` error code SHOULD be returned.
+
+If the number of lines between the `start` and `end` parameters exceeds the `max_message_count`, warn code `MAX_MSG_COUNT_EXCEEDED` SHOULD be returned. The command SHOULD continue to be processed as described above.
+
+If the target has no `chathistory` content to return or the user does not have permission to view the requested content, `ERR_NOSUCHNICK` or `ERR_NOSUCHCHANNEL` SHOULD be returned accordingly.
+
+### Examples
+The examples below are written with `draft/msgid` and `draft/label` tags included. These tags are recommended.
+
+#### Begin
```
SaberUK: Does each line need a title?

extensions/chathistory.md
```diff
@@ -0,0 +1,97 @@
...
+    @draft/label=ID :irc.host BATCH +ID chathistory target
+#### PRIVMSG
+    @batch=ID;draft/msgid=ID;time=YYYY-MM-DDThh:mm:ss.sssZ :nick!ident@host PRIVMSG target :message
+#### NOTICE
+    @batch=ID;draft/msgid=ID;time=YYYY-MM-DDThh:mm:ss.sssZ :nick!ident@host NOTICE target :message
+#### ACTION
+    @batch=ID;draft/msgid=ID;time=YYYY-MM-DDThh:mm:ss.sssZ :nick!ident@host PRIVMSG target :ACTION message
+#### End
+    :irc.host BATCH -ID
+#### Error
```
SaberUK: You should use a separate example for errors/warnings imo.

extensions/chathistory.md
```diff
@@ -0,0 +1,97 @@
...
+Alternatively, content can be requested using a `draft/msgid`:
+
+    @draft/label=ID draftCHATHISTORY <target> draft/msgid=<message_id> timestamp=<timestamp>
+
+Content can also be requested up to a specified timestamp or `draft/msgid` in place of the `message_count`. The start and end parameter types do not have to match:
+
+    @draft/label=ID draftCHATHISTORY <target> timestamp=<timestamp> +draft/msgid=<message_id>   
+
+#### Errors and Warnings
+If the server receives a`CHATHISTORY` command with missing parameters, the `ERR_NEEDMOREPARAMS` error code SHOULD be returned.
```
Zarthus: a\`CHATHISTORY

extensions/chathistory.md
```diff
@@ -0,0 +1,97 @@
...
+
+If the number of lines between the `start` and `end` parameters exceeds the `max_message_count`, warn code `MAX_MSG_COUNT_EXCEEDED` SHOULD be returned. The command SHOULD continue to be processed as described above.
+
+If the target has no `chathistory` content to return or the user does not have permission to view the requested content, `ERR_NOSUCHNICK` or `ERR_NOSUCHCHANNEL` SHOULD be returned accordingly.
```
Zarthus: Why are we mixing numerics and "return verbs" (L77) like WARN and ERROR. If one format is superior to the other we should strive to include it for all cases, not some of them, and some of them not.

MuffinMedic: Can you explain what you mean? I'm not clear on the distinction. Are you suggesting we stick with `ERROR` and `WARN` and provide plaintext reasons rather than the codes?

extensions/chathistory.md
```diff
@@ -0,0 +1,96 @@
...
+
+A method for securely identifying the requesting user MUST exist to ensure content is sent only to the appropriate users and clients. See below for more information.
+
+## Security Considerations
+Secure identification of users and clients MUST exist in order to ensure that users cannot obtain history they are not authorized to view. Use of account names, internal account identifiers, or certificate fingerprints SHOULD be strongly considered when matching content to users. The server MUST verify the current user's identity matches that of the desired content. This information is not sent as part of the `CHATHISTORY` command and MUST be validated via other means, such as those stated above. Access MUST be checked first and return an `NO_SUCH_NICK` or `NO_SUCH_CHANNEL` error as appropriate. If no authorization error exists, the server can check for returnable content. If no returntable content is found, the server MUST send an `NO_TEXT_TO_SEND` error. The server MUST NOT expose the existence of valid targets to unauthorized users.
```
Zarthus:
```
(18:53:52) <Zarthus> before i make an ass out of myself
(18:53:58) <Zarthus> return an `NO_SUCH_NICK` or `NO_SUCH_CHANNEL` error
(18:54:04) <Zarthus> typo "an" -> "a" yes/no
(18:54:14) <Sadie> y
```

Zarthus: MUST send an `NO_TEXT_TO_SEND` error

extensions/chathistory.md
```diff
@@ -0,0 +1,96 @@
...
+    @batch=ID;draft/msgid=ID;time=YYYY-MM-DDThh:mm:ss.sssZ :nick!ident@host PRIVMSG target :message
+
+    @batch=ID;draft/msgid=ID;time=YYYY-MM-DDThh:mm:ss.sssZ :nick!ident@host NOTICE target :message
+
+    @batch=ID;draft/msgid=ID;time=YYYY-MM-DDThh:mm:ss.sssZ :nick!ident@host PRIVMSG target :ACTION message
+
+    :irc.host BATCH -ID
+
+    @draft/label=ID :nick!ident@host CHATHISTORY ERR ERR_CODE
+
```
Zarthus: Without explicitly mentioning ERR as a verb (and WARN) implementors may think that this is not a constant string and get confused. A reference to some sort of document explaining subcommands may be useful

extensions/chathistory.md
```diff
@@ -0,0 +1,96 @@
...
+    @draft/label=ID CHATHISTORY <target> timestamp=YYYY-MM-DDThh:mm:ss.sssZ message_count=<message_count>
+
+Alternatively, content can be requested using a `draft/msgid`:
+
+    @draft/label=ID CHATHISTORY <target> draft/msgid=<message_id> timestamp=<timestamp>
+
+Content can also be requested up to a specified timestamp or `draft/msgid` in place of the `message_count`. The start and end parameter types do not have to match:
+
+    @draft/label=ID CHATHISTORY <target> timestamp=<timestamp> +draft/msgid=<message_id>
```
Zarthus: timestamp without an `@`  symbol?

extensions/chathistory.md
```diff
@@ -0,0 +1,96 @@
...
+
+The `server-time` SHOULD be the time at which the message was received from the IRC server and the `batch id` a unique ID for the entire batch. `draft/label` SHOULD be included and MUST be a unique ID used to identify the `chathistory` request and any replies. A `draft/msgid` to identify each individual message MUST be the `draft/msgid` included when each message was first received from the IRC server.
+
+### `CHATHISTORY` Command
+`CHATHISTORY` content can be requested by the client to the server by sending the `CHATHISTORY` command to the server. A `batch` MUST be returned by the server. If no content exists to return, an empty batch SHOULD be returned to avoid the client waiting for a reply. Command support is sent to the client as the RPL_ISUPPORT 005 numeric `:irc.host 005 nick CHATHISTORY=max_message_count :are supported by this server`
```
Zarthus: Perhaps instead of "an empty batch ..." we could do "CHATHISTORY SUCCESS NOOP" / NO_DATA whatever can be returned

extensions/chathistory.md
MuffinMedic: `timestamp` in this case is a command parameter, not a client tag.

jwheare:
Just to record the issues with this proposal as written so far:

It doesn't explicitly/obviously support a client requesting "give me x number of messages that arrived after timestamp/msgid y"

Questions over whether the ERR subcommand would work better with a top level generic error numeric/command reply.

MuffinMedic:
Moved to #349

PR #294, opened 2017-01-03
Add the NICKLIST extension specification.

MuffinMedic:
[Rendered view available here.](https://github.com/MuffinMedic/ircv3-specifications/blob/nicklist/extensions/nicklist-3.3.md)

The discussion came up regarding multiple nick parameters vs. a comma-separated list. Initial spec uses multiple parameters to ensure compatibility on networks allowing commas in nicks.

### DIFF
--- a/extensions/nicklist-3.3.md
+++ b/extensions/nicklist-3.3.md
@@ -1,0 +1,26 @@
+---
+title: IRCv3 nicklist extension
+layout: spec
+work-in-progress: true
+copyrights:
+  -
+    name: "Evan Magaliff"
+    period: "2017"
+    email: "evan@muffinmedic.net"
+---
+## Description
+This document describes the format of the `nicklist` capability extension. 
+
+This specification allows for sending a list of nicks the client would like to use to the server. The server then attempts to change the user's nick to each in the list until a successful nick change occurs or the list is exhausted.
+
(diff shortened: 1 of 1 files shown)


extensions/nicklist-3.3.md
```diff
@@ -0,0 +1,26 @@
...
+    name: "Evan Magaliff"
+    period: "2017"
+    email: "evan@muffinmedic.net"
+---
+## Description
+This document describes the format of the `nicklist` capability extension. 
+
+This specification allows for sending a list of nicks the client would like to use to the server. The server then attempts to change the user's nick to each in the list until a successful nick change occurs or the list is exhausted.
+
+The given nicklist takes the format: `:foo!bar@example.com NICK :nick1 nick2 nick3`
```
SaberUK: Theres no need to make it a <trailing> parameter. Just make it a series of <middle> parameters.

attilamolnar:
Can you describe the practical problem solved by this? It cannot be used for reducing roundtrips when sending the initial NICK when registering because at that time there are no caps negotiated and NICK is sent first by mainstream clients to support servers having no CAP (such as quakenet). It can be used after registration but the optimization achieved is minimal at best and nothing in cases when the primary nick is available.

Clients may use ISON, WATCH or MONITOR to determine which nicks are in use from a bunch of nicks then send a NICK, which is not the same but works good enough in practice.

PR #295, opened 2017-01-05
sts-3.3: Add the preload key

attilamolnar:
This is one way to check whether a server wishes to be included in preload lists. Another method, proposed by `e`  on IRC is to request a signature from a valid certificate when submitting servers to a preload list.

### DIFF
--- a/core/sts-3.3.md
+++ b/core/sts-3.3.md
@@ -169,6 +169,17 @@ Servers SHOULD verify whether the hostname provided to them via SNI is a
 hostname that is whitelisted for preloading by administrators to determine
 whether or not to advertise this key.
 If no hostname is available via SNI, this key SHOULD NOT be sent.
+This is to ensure that server administrators have control on which hostname
+may be preloaded in case the server has multiple. It is possible that a server
+has a wildcard certificate or a certificate with Subject Alternative Names but
+its administrators only wish to get a subset of those preloaded.
+
+For example a server presents a wildcard certificate for `*.example.net`.
+`irc.example.net`, `example.net`, `www.example.net` and `test.example.net` all
+point to the IP address of the server. The administrators may not wish to get
+`test.example.net` or `www.example.net` preloaded. Verifying the hostname sent
+via SNI before advertising the `preload` key ensures that only
+administrator-approved hostnames will get preloaded.
 
 See the [Pre-loaded STS policies section](#pre-loaded-sts-policies) for more
 information on preload lists.
--- a/core/sts-3.3.md
+++ b/core/sts-3.3.md
@@ -154,6 +154,22 @@ Servers MUST send this key to non-securely connected clients.
 Servers MAY send this key to securely connected clients, but it will be
 ignored.
 
+### The `preload` key
+
+This OPTIONAL key, if present, indicates that the server requests to be
+included in STS preload lists. If it has a value, the value MUST be ignored.
+
+Clients not looking to confirm whether the server requests to be included
+in STS preload lists SHOULD ignore the presence of this key.
+
+Servers SHOULD verify whether the hostname provided to them via SNI is a
+hostname that is whitelisted for preloading by administrators to determine
+whether or not to advertise this key.
+If no hostname is available via SNI, this key SHOULD NOT be sent.
+
+See the [Pre-loaded STS policies section](#pre-loaded-sts-policies) for more
+information on preload lists.
+
 ### Handling disconnection
 
 IRC connections may be long-lived. Connections lasting for more than a month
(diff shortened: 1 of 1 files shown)


attilamolnar:
Added links to sections the text is talking about.

attilamolnar:
Added paragraph telling clients to ignore the preload key unless they are looking to confirm that the server requests to be included in STS preload lists. We could also mention that this applies to the vast majority of clients.

core/sts-3.3.md
```diff
@@ -154,6 +154,25 @@ Servers MUST send this key to non-securely connected clients.
...
+Clients not looking to confirm whether the server requests to be included
+in STS preload lists SHOULD ignore the presence of this key.
+
+If the client receives the `preload` key on a non-secure connection, it is
+invalid and MUST be ignored.
+
+Servers SHOULD verify whether the hostname provided to them via SNI is a
+hostname that is whitelisted for preloading by administrators to determine
+whether or not to advertise this key.
+If no hostname is available via SNI, this key SHOULD NOT be sent.
```
DarthGandalf: How does HTTP do it?

attilamolnar: HTTP has the Host: header

attilamolnar:
I'll merge this at the end of the week if there are no objections.

jwheare:
Sorry for the late suggestion, but could you flesh out this part some more?

> If no hostname is available via SNI, this key SHOULD NOT be sent

It should be explained to a reader why this recommendation is being made, I don't think it's completely obvious as written. An example of the risk to server operators of not checking the hostname would be good.

lol768:
>An example of the risk to server operators of not checking the hostname would be good.

I guess the main risk here is that an IRC network has other A record entries (or indeed a wildcard entry that points all subdomains at a certain IP address) that they may not want an STS policy stored persistently for - blindly sending the `preload` key regardless of the hostname sent in the SNI extension would cause these to be stored, if the client was able to open a secure conneciton. They'd need to be using a wildcard (or multi-SAN cert) for this situation to be applicable, so it's something of an edge case - but probably worth outlining nonetheless.

attilamolnar:
@jwheare which section do you suggest putting the explanation in?

jwheare:
Just right after it's mentioned for now. We can move it around and finesse it if needed once the main gist is written down.

jwheare:
Thanks a lot, I'll take a closer look at this when I have a chance and suggest some improved wording.

jwheare:
I split out the SNI recommendation to cover the whole policy, rather than be limited to preload.

attilamolnar:
@jwheare Makes sense, LGTM

jwheare:
Thanks, any final comments from anyone else before this is merged?

core/sts-3.3.md
```diff
@@ -205,7 +241,8 @@ discussed in General security considerations.
 As further protection against bootstrap MITM vulnerabilities, clients may choose to
 include a pre-loaded list of known hosts with STS policies. Such lists should be
 compiled on an opt-in basis, by request of IRC network administrators. Hosts should be
-verified to be correctly advertising an STS policy before inclusion.
+verified to be correctly advertising an STS policy, as well as a 
+[preload policy](#the-preload-key) before inclusion.
```
ProgVal: Could you make it explicit that `preload` implies “by request of IRC network administrators”?

jwheare: How should this be worded to make it more clear to you?

ProgVal: “by request of IRC network administrators *via the `preload` key*”

jwheare: How's this? 4730fb6c875eff5290d40c3c8529d6aa48dda4a6

ProgVal: perfect

core/sts-3.3.md
```diff
@@ -274,7 +311,7 @@ server may be offered at a different domain name, for example a subdomain.
 
 It's possible for an attacker to strip the STS `port` value from an initial connection
 established via an insecure connection, before the policy has been cached by the client.
-This represents a Bootstrap MITM (man-in-the-middle) vulnerability.
+This represents a bootstrap MITM (man-in-the-middle) vulnerability.
```
SISheogorath: "MITM" is explained in this part, but not in line 236. Since 236 is earlier it would make sense to move the `(man-in-the-middle)` up there or remove it completely.

jwheare: Yeah this bothered me a tiny bit too, but well, this is the part that actually talks about it, line 236 is a reference to this part of the spec.

We could alternatively move this section further up, changing the order to:

* General Security considerations
* Client implementation considerations
* Server implementation considerations

jwheare: I'll punt that over to #301 though.

lol768: Nit: Most of the literature I've seen doesn't capitalise the minor words in the initialism, so I'd expect to see `MitM`(e.g. see http://ieeexplore.ieee.org/document/7546529/?reload=true)

lol768:
LGTM, will be nice to see the preload stuff in the spec

PR #298, opened 2017-01-09
Fix SASL 3.1, aborted is 906 not 904

kylef:
The specification mentions:

> The client can abort an authentication by sending an asterisk as the data. The server will send a 904 numeric.

However later we say this returns 906 numeric:

> `906` aka `ERR_SASLABORTED` is sent when the SASL authentication is aborted because the client sent an `AUTHENTICATE` command with `*` as the parameter.
 >
>     :server 906 <nick> :SASL authentication aborted


attilamolnar:
Please add this to the errata section

grawity:
Last time I remember looking into this, client-initiated abort wasn't actually _implemented_ as such &ndash; services merely treated `*` as invalid data and triggered a failure that way... I think that's where the 904 comes from.

I'm fine either way; 906 would certainly make more sense.

PR #301, opened 2017-01-16
STS edits

jwheare:
This is a WIP PR

* [x] Copy edits
* [x] Drops 3.3
* [x] Moves from core to extensions
* [x] Adds an example of clients ignoring CAP DEL
* [x] Add redirect-from yaml to avoid breaking old links
* [x] Make sure all non-normative sections are marked appropriately
* [x] The spec mostly talks about servers, maybe it should refer to hostnames, since that's what policies are tied to.
* [x] Sort out what to do on policy expiry/removal. The spec currently implies that clients need to revert to the previously insecure settings that were chosen by the user prior to upgrade, but I think it makes more sense to keep the upgraded port but just no longer enforce upgrades for new connections. This saves having to remember the original connection details, and just treat port=6697 upgrades as stateless.
* [x] Note that an sts port but no duration can be used as a soft temporary upgrade, if a server doesn't want to commit to a full policy.
* [x] Add a stronger recommendation for SNI. Done in #295
* [x] Try not to use "must", "should", "may" in non-normative sections
* [x] Merge in changes from #295
* [x] Reorder "General Security considerations" above "Client implementation considerations"
* [x] Use CAP NEW to update/disable an STS policy for active connections.

Leaving this out of scope for now:

* [x] Possibly out of scope, but maybe discuss what clients should do when a cert error is seen on a secure connection *without* an sts policy. At least initially, I'll be using the presence of sts to enforce cert validation, but still allow self-signed, unrecognised CAs on non-sts connections (though possibly with a more visible error). Given the number of non-secure certs still out there, I think that's reasonable. I'll also allow treat sts=duration=0 upgraded connections the same way, i.e. not hard fail, so it may be in scope after all.

### DIFF
--- a/extensions/sts.md
+++ b/extensions/sts.md
@@ -165,11 +165,9 @@ Clients might consider allowing users to explicitly define an STS policy for a g
 
 ### STS policy deletion or rejection
 
-Clients should consider allowing users or administrators to delete cached STS policies on a per-host basis, in case a server's policy is accidentally or maliciously injected on a secure connection.
+Clients might consider allowing users or administrators to reject or delete cached STS policies on a per-host basis, in case a server's policy is accidentally or maliciously injected on a secure connection.
 
-Clients might additionally provide the ability to reject STS policies on a per-host basis as an additional mitigation.
-
-These features should be made available very carefully in both user interface and security senses. Deleting or rejecting a cache entry for a known STS host should be a very deliberate and well-considered act -- it shouldn't be something that users get used to doing as a matter of course: e.g., just "clicking through" in order to get work done. In other words, these features should not violate the [no immediate user recourse](#no-immediate-user-recourse) section.
+Such a feature should be made available very carefully from both user interface and security standpoints. Deleting or rejecting a cache entry for a known STS host should be a very deliberate and well-considered act -- it shouldn't be something that users get used to doing as a matter of course: e.g., just "clicking through" in order to get work done. In other words, these features should not violate the [no immediate user recourse](#no-immediate-user-recourse) section.
 
 ## Server implementation considerations
 
--- a/extensions/sts.md
+++ b/extensions/sts.md
@@ -58,76 +58,50 @@ this document MUST only occur at most once.
 
 Clients MUST ignore every token with a key that they don't understand.
 
-An STS capability having at least a `duration` key expresses an STS policy when advertised on a secure connection.
+An STS capability advertised on a secure connection with at least a `duration` key present expresses an **STS persistence policy** for the server at the requested hostname.
+
+An STS capability advertised on an insecure connection with at least a `port` key present expresses an **STS upgrade policy** for the server at the requested hostname.
 
 See [capability negotiation 3.2](capability-negotiation-3.2.html) for more
 information about capabilities with values.
 
 ### Mechanism
 
-When the client sees an STS capability advertised over a non-secure
-connection, it MUST first establish a secure connection and confirm that the
-STS capability is still present.
+When a client sees an STS upgrade policy over a non-secure connection, it MUST first establish a secure connection (see the `port` key) and confirm that the STS persistence policy is present.
 
-Once the client has confirmed that the STS capability is indeed offered over
-a secure connection with a valid policy, it then MUST only use a secure transport to
-connect to the server in future, until the policy expires (see the `duration` key).
-It MUST refuse to connect if a secure connection cannot be established with the
-server for any reason during the lifetime of the policy.
+Once a client has connected securely, and it has verified that an STS persistence policy is in place, it then MUST only use a secure transport to connect to the server at the requested hostname in future, until the policy expires (see the `duration` key). Once an STS persistence policy has been verified, clients MUST refuse to connect if a secure connection cannot be established to the server for any reason during the lifetime of the policy.
 
-However, if the client fails to connect securely for any reason, the connection
-attempt SHOULD be considered a failure, similar to a network error.
-The client SHOULD retry the secure connection next time it receives the STS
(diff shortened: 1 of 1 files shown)


extensions/sts.md
```diff
@@ -283,14 +278,13 @@ section.
 
 ### Policy expiry
 
-It is possible that a client successfully receives the STS policy from a server
+It is possible that a client successfully receives an STS policy from a server
 but later on attackers begin to block all secure connection attempts from the
-client to the server until the policy expires. At that time, the client will
+client to the server until the policy expires. At that time, the client might
```
lol768: Why is this a might now?

jwheare: Because it's not guaranteed. The way I've implemented it, this would only happen if the user manually edited the connection to be non-secure. If left alone it would remain on the failing secure connection. See the TODO about making this more explicit and having the redirect be stateless.

jwheare:
Ticked off a few TODOs (and split some apart)

1. Rewrote most of the normative sections, hopefully for the better.
2. Introduced the concept of "upgrade" policies (port) and "persistence" policies (duration) to help clarify the distinction. Open to other labels though.
3. "Requested hostnames" are now mentioned, no SNI yet.
4. It's now explicitly mentioned that clients are allowed to continue using a secure connections after policy expiry.
5. Mentioned servers sending a port but no duration for a soft upgrade.

jwheare:
Latest batch of edits https://github.com/ircv3/ircv3-specifications/compare/03d7a4a...2ee66d6?expand=1

syzop:
Hi James, you asked for some feedback. I have gone through the entire specification (as in this pull request) and here are my comments:

The section *Server Name Indication* says this:
```Before advertising an STS policy, servers SHOULD verify whether the hostname provided by clients, for example, via TLS Server Name Indication (SNI), has been whitelisted by administrators in the server configuration.```
However, draft/tls is sent when the client is connected without SSL/TLS, the client then re-connects to the SSL port and has to verify the certificate.
In other words, the "to send draft/sts or not, depending on hostname" happens before SNI.
Now, I realize some clients send the servername in USER but this is not always the case and you may not have received USER before CAP LS so...
My advice would be to simply remove this section.

The non-normative *Client implementation considerations* has a section *STS policy deletion or rejection*. I agree there should be some (preferably obscure) way to delete cached STS objects but I really don't see a good reason to have some kind of "ignore STS" list as this section suggests. As you rightfully point out this would be dangerous. I would suggest to remove that part. We shouldn't encourage developers or users to do something that dangerous.

As for the non-normative *Server implementation considerations*: IMO it contains a lot of unnecessary information, such all the text regarding fixed-time and relative-time policy duration. The suggestion (or possibility) that you could have sts & certificate expiry at the same time, for instance. I don't see why you would ever want to do that, they are not related to each other. And, as the section rightly points out, it won't work anyway. Expiry at a fixed-point-in-time (eg: date XYZ) is only an illusion due to the "rescheduling" requirement on disconnection.
Anyway, it's in a non-normative section and it doesn't hinder me. I just wanted to give you my full feedback :)

Finally, I believe I saw a typo or markdown error, in the text it shows as ```(STS policy stripping)[#sts-policy-stripping]``` - I presume it was meant to be a link or reference.

Other than the few comments from above it's a really good specification IMO.

syzop:
I forgot one thing:
If you compare https://hstspreload.org/#removal with the section called **The ```preload``` key**, there's quite a difference between the text regarding updates/removal. HSTS has a big warning about the effect of including your domain in preload and telling you that it may take months to undo the process and giving no guarantees with regards to clients. The draft/sts on the other hand has a lot of SHOULD's that suggest client providers to have reasonable frequent updates. I tend to be on HSTS's side on this, if you submit your server to the preload list then you really shouldn't be whining about the time involved to undo (remove) your domain.
Anyway, there's something to say for both arguments and they are all SHOULD's so client devs can just ignore them.

syzop:
I guess what I meant with the last comment is not so much that the client SHOULD's are a bad idea. It's more that they may give the wrong impression to IRC administrators that this will be done, that is: that they may count on timely removal. With that in mind, you *could* keep the text, but still add some stronger wording regarding IRC administrators considering *preload* (that it may take a long time etc. etc.)

jwheare:
@syzop thanks for your comments!

1) SNI hostname checks: This isn't aimed at insecure connections, I'll clarify the wording. This would only apply to the duration and preload keys.

2) STS policy deletion or rejection: I feel it's better to pre-empt clients that might want to provide this sort of advanced option and explain the risks, rather than leave it up to chance whether it's implemented in a safe and sensible way. Rejecting an STS advertisement could be the only way to prevent denial of service for a misconfigured server.

3) Expiry guidelines: I feel there needs to be some discussion here on how to choose an expiry time. Server admins run networks in all sorts of unpredictable ways, and we are explicitly recommending IRCd configurations not to suggest default values, so some guidance here is probably useful. One reason you might want to sync (at least approximately) the sts duration with a cert expiry is to avoid a situation where the cert expires before you can renew it and then connections are no longer possible due to STS.

4) Typo: thanks, will fix!

5) Preload warning: agree, will add some stronger wording for servers to not expect timely removal.

jwheare:
Made some edits to address those comments. Let me know if they seem sensible.

syzop:
Much better :)

jwheare:
Will merge this next week if there aren't any further comments. This is the last remaining blocker for ratifying sts and removing the draft prefix.

jwheare:
Merged in 285aa38f376678f6a86094c542846928ae71683f

jwheare:
Site updated: http://ircv3.net/specs/extensions/sts.html

PR #302, opened 2017-01-25
Icons client tag spec

jwheare:
This allows clients (most notably bots) to associate icons to individual messages. This is distinct from persistent user avatars which will be better implemented as metadata keys.

Rather than just a URL, I've specced this to allow alternate sources, dimensions and type hints, to allow clients to select an appropriate image for a specific context and pixel density.

The format is inspired by the web app manifest spec: https://w3c.github.io/manifest/#icons-member

Simple:

`@+draft/icons=[{"src":"https://example.com/favicon16.png"}]`

Extended:

`@+draft/icons=[{"src":"https://example.com/favicon16.png","sizes":"16x16","type":"image/png"},{"src":"https://example.com/favicon32.png","sizes":"32x32","type":"image/png"},{"src":"https://example.com/favicon64.png","sizes":"64x64","type":"image/png"}]`

### DIFF
--- a/client-tags/icon.md
+++ b/client-tags/icon.md
@@ -68,6 +68,7 @@ Client tags are untrusted data and icon URIs might point to any arbitrary resour
 * Unexpected dimensions: Scale or crop images to fit within allocated user interface space.
 * Unknown formats: Validate content types and reject unaccepted image formats.
 * Malicious content: Allow users to block specific icon URIs or icons from specific users.
+* Revealing IP addresses to 3rd parties: Add an option to only download icons from trusted sources.
 
 ## Examples
 
--- a/client-tags/icon.md
+++ b/client-tags/icon.md
@@ -1,5 +1,5 @@
 ---
-title: `+icons` client tag
+title: `+icon` client tag
 layout: spec
 work-in-progress: true
 copyrights:
@@ -14,15 +14,15 @@ copyrights:
 This is a work-in-progress specification.
 
 Software implementing this work-in-progress specification MUST NOT use the
-unprefixed `+icons` tag name. Instead, implementations SHOULD use the
-`+draft/icons` tag name to be interoperable with other software
+unprefixed `+icon` tag name. Instead, implementations SHOULD use the
+`+draft/icon` tag name to be interoperable with other software
 implementing a compatible work-in-progress version.
 
 The final version of the specification will use an unprefixed tag name.
 
 ## Introduction
 
-This specification defines a client-only message tag to indicate message icons
+This specification defines a client-only message tag to indicate message icons.
 
 ## Motivation
 
@@ -40,9 +40,9 @@ Clients wishing to use this tag MUST negotiate the [`draft/message-tags`](../cor
 
 ### Format
 
-The icons tag is sent by a client with the client-only prefix `+` and its value is a JSON encoded array of image objects that reference icon resources for the message.
+The icon tag is sent by a client with the client-only prefix `+` and its value is a JSON encoded array of image objects that reference icon resources for the message.
 
-    +draft/icons=<json>
+    +draft/icon=<json>
 
 Image objects MAY contain any of the following members:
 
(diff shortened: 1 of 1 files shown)


client-tags/icon.md
```diff
@@ -0,0 +1,85 @@
...
+## Client implementation considerations
+
+This section is non-normative
+
+Client tags are untrusted data and icon URIs might point to any arbitrary resource. Some examples of unexpected icon conditions and potential mitigations:
+
+* Large file sizes: Validate file sizes and abort if appropriate limits are reached.
+* Unexpected dimensions: Scale or crop images to fit within allocated user interface space.
+* Unknown formats: Validate content types and reject unaccepted image formats.
+* Malicious content: Allow users to block specific icon URIs or icons from specific users.
```
DarthGandalf: * Automatically detect IPs of all users on the channel, regardless of any cloaks/bouncers used.

client-tags/icon.md
```diff
@@ -0,0 +1,85 @@
...
+
+## Examples
+
+In this example, a bot `C` responds to a link sent to the channel by the server `S` by fetching the title and favicon of the link. It then sends a message back to the channel with the link title in the message body and favicon included as a message tag:
+
+```
+S :nick!user@example.com PRIVMSG #channel :https://example.com/a-news-story
+C @+draft/icons=[{"src":"https://example.com/favicon16.png","sizes":"16x16","type":"image/png"},{"src":"https://example.com/favicon32.png","sizes":"32x32","type":"image/png"}],{"src":"https://example.com/favicon64.png","sizes":"64x64","type":"image/png"}] PRIVMSG #channel :Example.com: A News Story
```
DarthGandalf: Too many `]`

client-tags/icon.md
jwheare: Can add a security consideration about that, but worth noting that clients that embed image links have the same issue.

grawity:
I'd say strip `"type":"image/png"` and require PNG support at minimum. Clients can determine the actual type from HTTP headers.

As for the address leak... can a 32x32 icon fit into a `data:` URL?

jwheare:
An example of the type field being useful is clients that want to allow animated icons can prefer a gif src before falling back to png. Also allows clients to use scalable sources like svg if they want.

Yes it's available in the header and should be validated once it's downloaded, but allowing prefiltering is useful.

And you can't necessarily fit a 32x32px png in a 1024 byte data uri, no. It depends on the complexity of the image.

jwheare:
TODO

* [ ] add the `any` size for scalable sources.

RyanSquared:
I give a :+1: to including the type in the data instead of using HTTP headers.

client-tags/icon.md
```diff
@@ -0,0 +1,86 @@
...
+
+| Member           | Description |
+| :--------------- | :---------- |
+| `src` (string)   | The Uniform Resource Indicator (URI) of an icon resource |
+| `sizes` (string) | A space separate list of pixel image dimensions in `wxh` format supported by the resource. Clients MAY use this to select an appropriate icon resource for the context and pixel density of the display |
+| `type` (string)  | A media type hint for the icon resource. Clients MAY use this to filter out unsupported icon types without downloading the resource |
+
+When an icon `src` is an HTTP resource, it SHOULD be referenced with an HTTPS protocol where available.
+
+When the icon is encoded as an [RFC2397 Data URI](https://tools.ietf.org/html/rfc2397) the length of the URI SHOULD be no more than 1024.
```
Zarthus: SHOULD be no more than? Is perhaps SHOULD NOT exceed 1024 characters a better phrasing?

LFP6:
Could you describe some more specific use cases? The biggest benefit I could see for such a feature would be for link previews (including file upload previews), which I would think would be delegated either to the client once a message is received (theoretically requiring nothing over the wire), or the server (not a client tag) before sending the message. I like the idea, I'm just wondering if it's handled in the wrong place.

jwheare:
Originally I was thinking this would be a per-message avatar overlay/replacement, i.e. associated with the user, not a link/image preview. But I'm now considering dropping this and instead focusing on a broader "attachments" format, e.g. https://api.slack.com/docs/message-attachments with a set of fields that could be extended in future.

I haven't had a chance to revise it yet, but it'll depend on the use cases of any reference implementation I end up working on.

Zarthus:
If the idea was avatar support, why involve bots? 

I thought this sounded more for embedded urls than avatars, as it sounds like something you can repeatedly send rather than you only need send once, and if the idea is to send it with every message (so that an user always has an avatar) it just sounds like a lot of wasted bandwidth and a metadata-like solution would surely be better?

jwheare:
Well it's utterly moot to discuss now, but it was so that a bot avatar could be changed per message, not persistently.

Zarthus:
I agree it's moot, but I would imagine a combination of "`avatar-notify`" and some sort of metadata (like WHOX, which is polled on channel entry) would be a more bandwidth efficient solution in the case a similar implementation wants to come along for avatar support.

If we get a lot of specs that have data that has to be "sent with every message", you quickly exceed the 4k character limit you've set forth in a different issue.

LFP6:
At any rate, I'm all for an attachment spec. When we get there though, it will be important to consider who has the responsibility of doing things like file previews, url previews, etc., possibly using Slack as a model. I like their concept of having the ability to define it manually for bots, but their [unfurling](https://medium.com/slack-developer-blog/everything-you-ever-wanted-to-know-about-unfurling-but-were-afraid-to-ask-or-how-to-make-your-e64b4bb9254#.6s40uh3xf) concept seems neat and something I need to read up on.

client-tags/icon.md
```diff
@@ -0,0 +1,122 @@
...
+
+    +draft/icon=<json>
+
+Image objects MAY contain any of the following members:
+
+| Member           | Description |
+| :--------------- | :---------- |
+| `src` (string)   | The Uniform Resource Indicator (URI) of an icon resource |
+| `sizes` (string) | A space separate list of pixel image dimensions in `wxh` format supported by the resource, or a value of `any` for scalable icons. Clients MAY use this to select an appropriate icon resource for the context and pixel density of the display |
+| `type` (string)  | A media type hint for the icon resource. Clients MAY use this to filter out unsupported icon types without downloading the resource |
```
kylef: I think if a client has some strict requirements on the image format, they don't necessarily need to download the image. When the src is HTTP, the client can sent a HTTP request with an `Accept` header requesting the format(s) that the client supports. If the server cannot offer the src resource in any media types that the user asked for it SHOULD return an error.

kylef: Note also, the server could offer the resource as multiple types depending on what the user asked for.

jwheare:
Closing this issue as a dead proposal. Future work on this theme should probably follow the attachments idea in https://github.com/ircv3/ircv3-ideas/issues/22

PR #303, opened 2017-02-19
Add WEBIRC extension

MuffinMedic:
[Rendered view available here.](https://github.com/MuffinMedic/ircv3-specifications/blob/webirc/extensions/webirc.md)


extensions/webirc-3.3.md
```diff
@@ -0,0 +1,65 @@
...
+
+## Limitations
+WebIRC requires that the both WebIRC gateway and IRC server be configured to accept the connection in advance. Specific reccomendations are outlined below in the [Security Considerations](#security-considerations) section.
+
+## Security Considerations
+WebIRC allows anyone with a valid password to successfully connect to the IRC server and spoof any desired hostname. IRC servers SHOULD use secure passwords, SHOULD whitelist the IP addresses for the allowed WebIRC gateway servers, and SHOULD close any connections with an ERROR command that attempt to use the `WEBIRC` command not originating from a whitelisted IP address or using an incorrect password. Because the possibility for hostname spoofing exists, IRC servers MAY attempt to further validate or resolve hostnames or and match them to the IP addess. It is the responsibility of the IRC server to verify the authenticity of connecting users and perform additional security checks as they see fit. To assist network operators and prevent abuse, IRC servers SHOULD show when a WebIRC connection is in use and SHOULD provide the original host when possible.
```
Zarthus: This block contains a few sentences that would make sense to separate on newlines and would also have you running out of breath saying them well before you could finish saying them.

extensions/webirc-3.3.md
```diff
@@ -0,0 +1,65 @@
...
+
+## Current Implementations
+Widespread adoption of WebIRC already exists for both clients and servers. Below is a non-exhaustive list of current implementions to demonstrate in-use instances.
+ - Clients
+  - [KiwiIRC](https://kiwiirc.com/docs/webirc)
+  - [Mibbit](https://wiki.mibbit.com/index.php/Webirc)
+  - [qwebirc](http://qwebirc.org/features)
+ - Servers
+  - [charybdis](https://github.com/charybdis-ircd/charybdis/blob/release/4/extensions/m_webirc.c)
```
Zarthus: github urls pointing to a branch are unlikely to be permanent. Worth considering using a git hash?

dequis: I suggested removing the whole "Current Implementations" section instead, since it doesn't belong inside the spec.

SaberUK: Current implementations is only really useful for draft stuff like STS.

Zarthus:
Perhaps it's just me not understanding the protocol, but how am I supposed to know the users IP address? Based on your example, kiwiirc.com is being used, does that mean I send the IP from my server (kiwiirc.com) or from the user (whatever their IP is)?

extensions/webirc-3.3.md
```diff
@@ -0,0 +1,65 @@
...
+
+## Description
+When a user connects through an indirect connection to the IRC server, the user's actual IP address and hostname are not visible. The `WEBIRC` command allows the WebIRC gateway to pass on the user's actual IP address and hostname to the IRC server. This information can be utilized by the network as if the user was connecting directly.
+
+## Format
+The `WEBIRC` command takes four parameters: `pass `, `user`, `hostname`, and `ip`. `pass` is the password authenticating the WebIRC gateway to the IRC server which must be agreed upon ahead of time, `user` is the WebIRC gateway service name, `hostname` is the resolved hostname of the user's IP address, and `ip` is the IPv4 or IPv6 address of the connecting user (IPv4-in-IPv6 MUST NOT be sent). If the IP address cannot be resolved to a DNS name, the IP address MUST be sent as both the `hostname` and `ip`.
```
SaberUK: >`user` is the WebIRC gateway service name

Is this new? In at least InspIRCd this parameter is ignored and I don't remember seeing this specific usage ever being implemented. It doesn't seem that useful to me imo.

prawnsalad:
It should also mention that `WEBIRC` **MUST** be the first command sent to the IRC server from the webirc gateways

extensions/webirc-3.3.md
SaberUK: Also, is there any reason for specifying this:

>IPv4-in-IPv6 MUST NOT be sent

Seems like its something that should be left up to the implementations to decide on how they want to handle that.

extensions/webirc-3.3.md
```diff
@@ -0,0 +1,65 @@
...
+
+If forward and reverse DNS do not match, the IP addesss SHOULD be sent as the `hostname` and the unverified `hostname` SHOULD NOT be sent (see [Security Considerations](#security-considerations)). If the connection fails to the IRCD, the WebIRC gateway MUST return an `ERROR` with an appropriate message and terminate the connection.
```
SaberUK: >addesss

Typo/

extensions/webirc-3.3.md
prawnsalad: It's never been used, but every major implementation has always used the service name in the `user` param so it makes sense to keep that consistent. This is mainly clarified here as some IRC devs have taken it to mean the connecting users ident is to be given here and overrides the `USER` params which no existing IRCD does that I've seen.

extensions/webirc.md
```diff
@@ -0,0 +1,56 @@
...
+    name: "Evan Magaliff"
+    period: "2017"
+    email: "muffinmedic@kiwiirc.com"
+  -
+    name: "Darren Whitlen"
+    period: "2017"
+    email: "prawnsalad@kiwiirc.com"
+---
+## Purpose
+This specification is intended to document the existing `WEBIRC` command that is already in use by multiple IRCDs and WebIRC gateways to provide a standard specification for implementation. This specification does not add any new features or changes, although new features, such as use of public-key cryptography over password authentication, can be submitted as extensions.
```
DarthGandalf: s/over/instead of/ ?

extensions/webirc.md
```diff
@@ -0,0 +1,56 @@
...
+
+## Description
+When a user connects through an indirect connection to the IRC server, the user's actual IP address and hostname are not visible. The `WEBIRC` command allows the WebIRC gateway to pass on the user's actual IP address and hostname to the IRC server. This information can be utilized by the network as if the user was connecting directly.
+
+## Format
+The `WEBIRC` command MUST be the first command sent from the WebIRC gateway to the IRC server.
```
DarthGandalf: , even before capability negotiation.

extensions/webirc.md
```diff
@@ -0,0 +1,56 @@
...
+If forward and reverse DNS do not match, the IP address SHOULD be sent as the `hostname` and the unverified `hostname` SHOULD NOT be sent (see [Security Considerations](#security-considerations)). If the connection fails to the IRCD, the WebIRC gateway MUST return an `ERROR` with an appropriate message and terminate the connection.
+
+### Examples
+Generic format.
+
+    WEBIRC password user hostname ip
+
+IP address resolves to hostname.
+
+    WEBIRC hunter2 gatewayname kiwiirc.com 104.25.32.27
```
DarthGandalf: Use one of IPs reserved for the documentation
https://en.wikipedia.org/wiki/Reserved_IP_addresses

DarthGandalf: kiwiirc.com sounds like hostname of gateway, not hostname of user

extensions/webirc.md
```diff
@@ -0,0 +1,56 @@
...
+
+## Limitations
+WebIRC requires that the both WebIRC gateway and IRC server be configured to accept the connection in advance. Specific reccomendations are outlined below in the [Security Considerations](#security-considerations) section.
+
+## Security Considerations
+WebIRC allows anyone with a valid password to successfully connect to the IRC server and spoof any desired hostname. IRC servers SHOULD use secure passwords and SHOULD whitelist the IP addresses for the allowed WebIRC gateway servers. If the `WEBIRC` command does not originate from a whitelisted IP address or uses an incorrect password, the gateway SHOULD close any connections with an `ERROR` command.
+
+Because the possibility for hostname spoofing exists, IRC servers MAY attempt to further validate or resolve hostnames and match them to the IP addess. It is the responsibility of the IRC server to verify the authenticity of connecting users and perform additional security checks as they see fit. To assist network operators and prevent abuse, IRC servers SHOULD show when a WebIRC connection is in use and SHOULD provide the original host when possible.
```
DarthGandalf: typo

extensions/webirc.md
```diff
@@ -0,0 +1,56 @@
...
+
+    ERROR :Invalid WebIRC password
+
+## Use Cases
+Webchat applications that make the connection to the IRCD via their servers are the primary intended use case. WebIRC allows network and channel operators to utilize the full extent of moderation and security tools available by exposing the user IP addresses and hosts.
+
+Other use cases can include bouncer services that wish to pass user information to the IRC server.
+
+## Limitations
+WebIRC requires that the both WebIRC gateway and IRC server be configured to accept the connection in advance. Specific reccomendations are outlined below in the [Security Considerations](#security-considerations) section.
```
DarthGandalf: typo

extensions/webirc.md
MuffinMedic: What ambiguity does this clear up that isn't addressed by "first"?

ProgVal:
Since it is not specific to a web IRC clients, shouldn't it be renamed using proxy terminology (eg. “real IP”, “forwarded for”/“forwarding for”, ...)?

MuffinMedic:
@ProgVal Which parts of the language are specific to web IRC clients? The generic use case for the spec is the same: a user connects to IRC through an intermediary WebIRC gateway.

ProgVal:
> Which parts of the language are specific to web IRC clients?

The name of the command

> The generic use case for the spec is the same: a user connects to IRC through an intermediary WebIRC gateway.

The command was created for that use case; but commands should be named according to their role/purpose, not their most common use case.

prawnsalad:
@ProgVal This spec is to document the existing command as it is currently used. This isn't to change any implementations or introduce anything new.

MuffinMedic:
I'm open to discussions for changing this from `WEBIRC` to `CGIIRC`, although only if a large amount of developers who currently implement it agree to the change and we retain backwards compatibility with the spec as it currently is.

dequis:
No, don't rename it. That's bikesheddy and missing the point of documenting the existing implementation.

prawnsalad:
Since this is only documenting the existing spec, this should be ready to merge in now I think?

extensions/webirc.md
```diff
@@ -0,0 +1,56 @@
...
+IP address does not resolve to hostname.
+
+    WEBIRC hunter2 gatewayname 198.51.100.3 198.51.100.3
+
+Error from invalid password.
+
+    ERROR :Invalid WebIRC password
+
+## Use Cases
+Webchat applications that make the connection to the IRCD via their servers are the primary intended use case. WebIRC allows network and channel operators to utilize the full extent of moderation and security tools available by exposing the user IP addresses and hosts.
```
SaberUK: s/IRCD/server/

extensions/webirc.md
```diff
@@ -0,0 +1,56 @@
...
+  -
+    name: "Darren Whitlen"
+    period: "2017"
+    email: "prawnsalad@kiwiirc.com"
+---
+## Purpose
+This specification is intended to document the existing `WEBIRC` command that is already in use by multiple IRCDs and WebIRC gateways to provide a standard specification for implementation. This specification does not add any new features or changes, although new features, such as use of public-key cryptography instead of password authentication, can be submitted as extensions.
+
+## Description
+When a user connects through an indirect connection to the IRC server, the user's actual IP address and hostname are not visible. The `WEBIRC` command allows the WebIRC gateway to pass on the user's actual IP address and hostname to the IRC server. This information can be utilized by the network as if the user was connecting directly.
```
jwheare: s/utilized/used.

extensions/webirc.md
```diff
@@ -0,0 +1,56 @@
...
+
+## Format
+The `WEBIRC` command MUST be the first command sent from the WebIRC gateway to the IRC server and MUST be sent before capability negotiation.
+
+The `WEBIRC` command takes four parameters: `pass `, `user`, `hostname`, and `ip`. `pass` is the password authenticating the WebIRC gateway to the IRC server which must be agreed upon ahead of time, `user` is the WebIRC gateway service name, `hostname` is the resolved hostname of the user's IP address, and `ip` is the IPv4 or IPv6 address of the connecting user. If the IP address cannot be resolved to a DNS name, the IP address MUST be sent as both the `hostname` and `ip`.
```
jwheare: s/`pass`/`password`

jwheare: Perhaps split out each parameter definition like this:

* `password` The password authenticating the WebIRC gateway to the IRC server which must be agreed upon ahead of time
* `user` The WebIRC gateway service name.
* ... etc

extensions/webirc.md
```diff
@@ -0,0 +1,56 @@
...
+
+If forward and reverse DNS do not match, the IP address SHOULD be sent as the `hostname` and the unverified `hostname` SHOULD NOT be sent (see [Security Considerations](#security-considerations)). If the connection fails to the IRCD, the WebIRC gateway MUST return an `ERROR` with an appropriate message and terminate the connection.
```
jwheare: s/IRCD/server

extensions/webirc.md
```diff
@@ -0,0 +1,56 @@
...
+    name: "Evan Magaliff"
+    period: "2017"
+    email: "muffinmedic@kiwiirc.com"
+  -
+    name: "Darren Whitlen"
+    period: "2017"
+    email: "prawnsalad@kiwiirc.com"
+---
+## Purpose
+This specification is intended to document the existing `WEBIRC` command that is already in use by multiple IRCDs and WebIRC gateways to provide a standard specification for implementation. This specification does not add any new features or changes, although new features, such as use of public-key cryptography instead of password authentication, can be submitted as extensions.
```
jwheare: s/IRCDs/IRC servers

extensions/webirc.md
```diff
@@ -0,0 +1,56 @@
...
+IP address does not resolve to hostname.
+
+    WEBIRC hunter2 gatewayname 198.51.100.3 198.51.100.3
+
+Error from invalid password.
+
+    ERROR :Invalid WebIRC password
+
+## Use Cases
+Webchat applications that proxy the connection to the IRC server are the primary intended use case. WebIRC allows network and channel operators to utilize the full extent of moderation and security tools available by exposing the user IP addresses and hosts.
```
jwheare: Rephrase suggestion:

> WebIRC allows network and channel operators to use available moderation and security tools to their full extent by exposing the user IP addresses and hosts.

extensions/webirc.md
```diff
@@ -0,0 +1,56 @@
...
+
+Other use cases can include bouncer services that wish to pass user information to the IRC server.
```
jwheare: > Other use case ~can~ include…

extensions/webirc.md
```diff
@@ -0,0 +1,56 @@
...
+
+## Limitations
+WebIRC requires that the both WebIRC gateway and IRC server be configured to accept the connection in advance. Specific recommendations are outlined below in the [Security Considerations](#security-considerations) section.
+
+## Security Considerations
+WebIRC allows anyone with a valid password to successfully connect to the IRC server and spoof any desired hostname. IRC servers SHOULD use secure passwords and SHOULD whitelist the IP addresses for the allowed WebIRC gateway servers. If the `WEBIRC` command does not originate from a whitelisted IP address or uses an incorrect password, the gateway SHOULD close any connections with an `ERROR` command.
```
jwheare: Invalid password is a MUST close.

extensions/webirc.md
```diff
@@ -0,0 +1,56 @@
...
+
+Because the possibility for hostname spoofing exists, IRC servers MAY attempt to further validate or resolve hostnames and match them to the IP address. It is the responsibility of the IRC server to verify the authenticity of connecting users and perform additional security checks as they see fit. To assist network operators and prevent abuse, IRC servers SHOULD show when a WebIRC connection is in use and SHOULD provide the original host when possible.
```
jwheare: > match them to **an** IP address

jwheare: s/the IRC server/IRC servers

jwheare: How do IRC servers show when a WebIRC connection is being used?

extensions/webirc.md
prawnsalad: Not all IRCds currently close the connection but rather ignore the WEBIRC line if the pass is not matching. Changing this to MUST would render those existing implementations as breaking. (Remember - this is to document *existing* usage)

jwheare: Ah, fair enough. Aware this is documenting existing usage, just thought this was an oversight.

extensions/webirc.md
MuffinMedic: It's up to each network how they decide to show WebIRC is being used. Some add a cloak `network/cgi-irc/gateway/ip` and show real IP and hostname information to opers on WHOIS. Networks are free to implement other options as well, such as setting the ident (as is common with Mibbit) or adding a line into the user's WHOIS displaying gateway information.

jwheare: Might be worth clarifying that, say it's non normative and implementation defined.

jwheare:
👍  LGTM

extensions/webirc.md
```diff
@@ -0,0 +1,60 @@
...
+- `user` WebIRC gateway service name
+- `hostname` Resolved hostname of the user's IP address
+- `ip` IPv4 or IPv6 address of the connecting user. If the IP address cannot be resolved to a DNS name, the IP address MUST be sent as both the `hostname` and `ip`.
+
+If forward and reverse DNS do not match, the IP address SHOULD be sent as the `hostname` and the unverified `hostname` SHOULD NOT be sent (see [Security Considerations](#security-considerations)). If the connection fails to the IRC server, the WebIRC gateway MUST return an `ERROR` with an appropriate message and terminate the connection.
+
+### Examples
+Generic format.
+
+    WEBIRC password user hostname ip
```
Zarthus: I think overall (if this is meant to help new implementers out as well) it would help to clarify which component is being asked (that of the server or of the user), as it's rather confusing and unclear to me.

Take the following for instance:
```
- `user` WebIRC gateway service name
   ^^^^
- `hostname` Resolved hostname of the user's IP address
                                      ^^^^
```

It doesn't help that `user` is also an irc verb for "username", or "ident". it's also called gatewayname a bit lower. It's confusing when it's mixed together like this, and I would clarify by prefixing something like `server_${param}` and `user_${param}`.

prawnsalad: You're right, and it has tripped people up in the past but we now have text to specifically state what it's used for. Every implementation I've seen calls this param `user` and may cause confusion when they read through this doc. Technically it makes no difference either way though..

MuffinMedic: @prawnsalad Can we change the `user` field to another name, such as `gateway`? The implementation would be the same, it would just be more clear as to what is being referred to.

Zarthus: I think you can be both future-proof and backwards-compatible by just adding a non-normative section explaining what some implementations might use as "denominator". 

Sure, the text explains it, but I don't think it hurts to change some namings around, as prawnsalad said it makes no technical difference, so being clear for the future is more beneficial than satisfying the past.

prawnsalad: Renaming it to `gateway` and Including a quick note that sometimes it's refereed to as `user`  should be fine IMO

extensions/webirc.md
```diff
@@ -0,0 +1,60 @@
...
+## Description
+When a user connects through an indirect connection to the IRC server, the user's actual IP address and hostname are not visible. The `WEBIRC` command allows the WebIRC gateway to pass on the user's actual IP address and hostname to the IRC server. This information can be used by the network as if the user was connecting directly.
+
+## Format
+The `WEBIRC` command MUST be the first command sent from the WebIRC gateway to the IRC server and MUST be sent before capability negotiation.
+
+The `WEBIRC` command takes four parameters: `password `, `user`, `hostname`, and `ip`.
+- `password` Password authenticating the WebIRC gateway to the IRC server which must be agreed upon ahead of time
+- `user` WebIRC gateway service name
+- `hostname` Resolved hostname of the user's IP address
```
Zarthus: To my (admittedly lacking) networking knowledge this sentence is a no-op. The resolving hostname of the user's IP address is already their IP address, no? I would just say "The hostname".

MuffinMedic: > of the user's IP address

The only thing that can be resolved in this case is an IP address, correct. This clarification is to make it clear that it's the user's IP address being resolved and not the IP of the gateway.

extensions/webirc.md
```diff
@@ -21,26 +21,28 @@ When a user connects through an indirect connection to the IRC server, the user'
...
 - `ip` IPv4 or IPv6 address of the connecting user. If the IP address cannot be resolved to a DNS name, the IP address MUST be sent as both the `hostname` and `ip`.
 
 If forward and reverse DNS do not match, the IP address SHOULD be sent as the `hostname` and the unverified `hostname` SHOULD NOT be sent (see [Security Considerations](#security-considerations)). If the connection fails to the IRC server, the WebIRC gateway MUST return an `ERROR` with an appropriate message and terminate the connection.
 
+Note: Previous implementations referred to the `gateway` field as `user`. This change is for documentation clarity only and maintains compatibility with all existing implementations.
+
 ### Examples
 Generic format.
 
     WEBIRC password user hostname ip
```
jwheare: s/user/gateway

dequis:
Merge?

PR #305, opened 2017-04-10
Add meta status keys to header yaml

DanielOaks:
This lets us, in the website, link between specifications where appropriate and makes the discoverability of things like new versions and extensions to existing capabilities much easier to see.

### DIFF
--- a/core/capability-negotiation-3.1.md
+++ b/core/capability-negotiation-3.1.md
@@ -1,6 +1,9 @@
 ---
 title: IRCv3.1 Client Capability Negotiation
 layout: spec
+updated-by:
+  "CAP Negotiation 3.2": core/capability-negotiation-3.2.html
+  "cap-notify Extension": extensions/cap-notify-3.2.html
 copyrights:
   -
     name: "Kevin L. Mitchell"
--- a/core/capability-negotiation-3.1.md
+++ b/core/capability-negotiation-3.1.md
@@ -2,8 +2,8 @@
 title: IRCv3.1 Client Capability Negotiation
 layout: spec
 updated-by:
-  "CAP Negotiation 3.2": core/capability-negotiation-3.2.html
-  "cap-notify Extension": extensions/cap-notify-3.2.html
+  - cap-3.2
+  - cap-notify
 copyrights:
   -
     name: "Kevin L. Mitchell"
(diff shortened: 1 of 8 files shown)


DarthGandalf:
Need a doc with list of keys with explanation. What is the difference between updates and extends?

dequis:
>What is the difference between updates and extends?

Extends is a jekyll thing, updates is a dan thing. It's just custom metadata for whatever thing he's doing with the template code of the website.

(Also, I think he has write access to the website repo but not this one)

DanielOaks:
Heh, I didn't realize extends was a jekyll thing. We can definitely do up some docs, prolly to put in the contributing document.

My thoughts are:

- `updates`: To do this spec, read the linked doc and then this one.
- `updated-by`: After doing the above, read these documents to get the latest version.
- `extends`: This is an optional extension of the linked documents.
- `extended-by`: These are optional extensions to this document.
- `obsoletes`: Ignore the linked doc, this one replaces it entirely.
- `obsoleted-by`: Ignore this doc, the linked one(s) replace it entirely.

Here specifically, `capability-negotiation-3.1.md` is a bit strange since we really recommend implementing 3.2 but it's not exactly deprecated (since you need to depend on the original doc), and `cap-notify` in particular is... somewhere in the middle of an extension and just a plain update.

DanielOaks:
I don't think `extends` should matter here too much, seems like an actual tag rather than part of the YAML metadata, so we should be fine to use it however we want.

dequis:
Oh, sorry, mixed things up. Yeah it's just a tag.

grawity:
Seems fine, although I think it would be enough to have:

```
updated-by:
    - core/capability-negotiation-3.2
    - extensions/cap-notify-3.2
```

DanielOaks:
If there's a way for Jekyll to extract info directly from the front-matter of files given just the filename then I'd be fine to add the shortnames to the files directly and reference them that way. Would also reduce duplication (since the shorter name gets duped for files like CAP 3.2 which is referenced by multiple other files).

grawity:
Well, I wouldn't see it as a problem even if it directly showed the names – they're already clear and concise.

DanielOaks:
Included an explanation of the current keys and how to use them

DanielOaks:
I've got an implementation of this based on the new registry files in the site, so we don't need to include native names or URLs! Should be good to merge now as far as I can see?

jwheare:
Go for it.

PR #306, opened 2017-04-13
Add `resume` extension

DanielOaks:
Occasionally, clients disconnect from IRC. What happens these days is that the client connects with a different nick, joins all their old channels again, waits for the old connection to time out (or manually kills it using services), and then changes back to their original nickname.

This feature intends to vastly simplify this form of reconnection, and reduce the amount of nick-switching, `JOIN` and `QUIT` notices, and general disruption that other clients see when this happens.

Note that this is not related to detaching and re-attaching an active connection (such as if a client is forced to change servers), but only to streamline the reconnection process for unintended connection drops.

[Rendered specification](https://github.com/DanielOaks/ircv3-specifications/blob/master+resume/extensions/resume.md)


extensions/resume.md
```diff
@@ -0,0 +1,137 @@
...
+Implementations should keep in mind that the only concrete indication that a `RESUME` attempt failed is that it does not receive a `RESUMED` message during registration. Clients MUST NOT rely on `ERR_CANNOT_RESUME` or any other numeric being received to trigger their 'resume has failed' state and process.
+
+`ERR_HISTORY_TRUNCATED` is used to indicate that chat history may be truncated for a given client/channel. It may be sent during a `chathistory` batch or otherwise during history playback. Bouncers that reconnect may also find this useful to send to their connected clients, to indicate that there may be missed history.
+
+
+## Connection Registration
+
+Upon connection, clients request the `draft/resume` capability. If they receive both this capability and the SASL capability, they can continue attempting to resume their old connection.
+
+After sending the `NICK` and `USER` commands, clients wishing to resume an old connection send an appropriate `RESUME` command indicating the nickname they wish to take over and how long they've been disconnected for.
```
grawity: Does requesting `resume` hold registration until CAP END, or does it rely on the registration being held by `sasl`?

extensions/resume.md
```diff
@@ -0,0 +1,152 @@
...
+    C2 - S: :irc.example.com 332 dan #test :Example topic
+    C2 - S: :irc.example.com 333 dan #test george 1442060874
+    C2 - S: :irc.example.com 353 dan @ #test :@dan @george +violet roger
+    C2 - S: :irc.example.com MODE #test +o dan
+
+Here is this reconnection seen by `george`, a client that does not have the `draft/resume` capability enabled:
+
+    C: :dan!~old@192.168.0.5 QUIT :Client reconnected
+    S: :dan!~d@127.0.0.1 JOIN #test
+    S: :irc.example.com MODE #test +o dan
```
grawity: Maybe the existing `CHGHOST` event would work here?

extensions/resume.md
```diff
@@ -0,0 +1,152 @@
...
+
+And here is this reconnection seen by `violet`, a client that has the `draft/resume` capability:
+
+    C: :dan!~old@192.168.0.5 RESUMED dan ~d 127.0.0.1 2017-04-13T15:12:51.620Z
```
dequis: Isn't this RESUMED essentially the same as CHGHOST with an additional timestamp? Not sure how that timestamp could be useful for other clients.

extensions/resume.md
DanielOaks: It's good to make it clear that some history may have been lost for the new client (so people talking to them will understand that their connection died), which is why I'd prefer sending an explicit QUIT+JOIN to clients that don't support this.

extensions/resume.md
```diff
@@ -0,0 +1,152 @@
...
+### Numerics
+
+Here are the new numerics that this extension defines:
+
+| No. | Label                   | Format                                                          |
+| --- | ----------------------- | --------------------------------------------------------------- |
+| ??? | `ERR_CANNOT_RESUME`     | `:<server> ??? <oldnick> :Cannot resume connection, <details>`  |
+| ??? | `ERR_HISTORY_TRUNCATED` | `:<server> ??? <client/channel> :Chat history may be truncated` |
+
+`ERR_CANNOT_RESUME` is used to indicate a general failure to resume a connection for a specific reason. The purpose of this numeric is to pass along human-readable information informing the user why their connection wasn't automatically resumed such as a lack of SASL authentication and login. If a more appropriate numeric exists for conveying the issue already exists (for example, `ERR_NOSUCHNICK` may indicate that the old nickname is no longer being present on the network), then it should be used instead.
```
dequis: >If a more appropriate numeric exists for conveying the issue already exists (for example, `ERR_NOSUCHNICK` may indicate that the old nickname is no longer being present on the network), then it should be used instead.

Can we require ERR_CANNOT_RESUME for all failures? I wouldn't mind additional errors like ERR_NOSUCHNICK afterwards. But associating arbitrary error numerics with outgoing commands has always been fiddly and this is a good opportunity to have an concrete mapping of RESUME -> ERR_CANNOT_RESUME.

DanielOaks:
The idea around using `CHGHOST` rather than an explicit separate verb for noting resumed connections.

One of the main things to convey when a reconnection happens is that message history may have been lost, which is why we send an explicit `QUIT`+`JOIN` for clients that don't support this capability. If we used `CHGHOST` with a `resume` tag, clients with message tags and `resume` would receive the `CHGHOST`, clients with `resume` but no tags would have to receive the `QUIT`+`JOIN` so the user knows that the connection was explicitly disconnected and reconnected. As well, not requiring message tags means that this can be implemented on clients and servers that don't currently support them.

Using an explicit new message reduces the dependencies on other extensions, which in this case I think is a good thing.

extensions/resume.md
```diff
@@ -0,0 +1,163 @@
...
+### Numerics
+
+Here are the new numerics that this extension defines:
+
+| No. | Label                   | Format                                                           |
+| --- | ----------------------- | ---------------------------------------------------------------- |
+| ??? | `ERR_CANNOT_RESUME`     | `:<server> ??? <oldnick> :Cannot resume connection[, <details>]` |
+| ??? | `ERR_HISTORY_TRUNCATED` | `:<server> ??? <client/channel> :Chat history may be truncated`  |
+
+`ERR_CANNOT_RESUME` is used to indicate a failure to resume a connection for a specific reason. The purpose of this numeric is to indicate that resumption was not successful and optionally provide a human-readable explanation. If a more appropriate numeric for conveying the issue already exists (for example, `ERR_NOSUCHNICK` may indicate that the old nickname is no longer being present on the network), then it should be sent after this numeric is sent (i.e. both numerics will be sent).
```
MuffinMedic: If the `ERR_CANNOT_RESUME` error is sent, should the client automatically attempt a new connection instead of a resume?

extensions/resume.md
```diff
@@ -0,0 +1,163 @@
...
+
+#### `RESUMED` Message
+
+This message is sent when a client successfully resumes a connection. For clients that have not received numerics indicating that their connection registration is completed, this numeric means that their `RESUME` attempt was successful. For registered clients, this indicates that the given client disconnected and reconnected to the server.
+
+This message has the following format:
+
+    :nick!olduser@oldhost RESUMED nick user host [timestamp]
+
+`nick` is the nickname of the client whose connection has been resumed. `user` and `host` are the username and hostname that the new connecting client has. The client receiving this message MUST update the username and hostname they have stored for the given client, in the same way that the `CHGHOST` message is parsed. `timestamp` if given, shows when the reconnecting client received their last message from the server. The timestamp may be used as a rough indication of how much message history has been lost.
```
MuffinMedic: To better support `chathistory`, allowing either timestamps or `draft/msgid` to be sent could be useful, as `chathistory` allows clients to request content by either parameter.

    :nick!olduser@oldhost RESUMED nick user host [timestamp=2017-04-13T15:12:51.620Z]
    :nick!olduser@oldhost RESUMED nick user host [draft/msgid=123456789]

extensions/resume.md
DanielOaks: Yeah it falls back to just continuing with a new connection. It's noted with this text:
> Connection resumption is done on a best-effort basis. If the resumption fails clients SHOULD continue with a regular reconnection.

I'll add an example to show how a failed RESUME attempt should look though

MuffinMedic: I see it now, thanks!

extensions/resume.md
```diff
@@ -0,0 +1,163 @@
...
+
+After sending the `NICK` and `USER` commands, clients wishing to resume an old connection send an appropriate `RESUME` command indicating the nickname they wish to take over and how long they've been disconnected for.
+
+After this, the client performs SASL authentication to prove that the indicated old connection does belong to them. If SASL authentication completes successfully, clients are sent the `RESUMED` message and then the regular connection burst (`001`-`005` and all).
+
+Once the connection burst has been received, the server sends the client channel `JOIN` messages as appropriate (and the usual numerics that are sent on joining a channel), and dispatches the relevant messages and numerics to other clients in related channels. The new client is also given the user modes that were active on the old client, as appropriate.
+
+Other clients that have the reconnecting user `MONITOR`'d MUST be sent one `RPL_MONOFFLINE` numeric and one `RPL_MONONLINE` numeric indicating that the user reconnected.
```
ProgVal: With `server-time` enabled, what should the timestamp of the `RPL_MONOFFLINE` message be?

extensions/resume.md
ProgVal: Could you make it mandatory for clients to display the disconnection one way or an other?
I may become an horrible mess otherwise, with other users loosing messages without any indication to the sender.

extensions/resume.md
```diff
@@ -40,11 +40,11 @@ This command is sent before starting SASL authentication, and indicates that the
...
+If the client's able to successfully resume their connection, the server sends them a `RESUMED` message (and their nickname is set to `<oldnick>`). Otherwise, the server sends the `ERR_CANNOT_RESUME` numeric (and optional additional error numerics).
 
-If successful, the `RESUMED` message is sent to the connecting user just after connection registration (including ident lookups and similar) completes. When this message is sent out, the connecting client's nickname is updated to be the old client's nick. From this point forward, the combination of old nickname and new connecting client's username+hostname is used to identify the client from that point forward.
+If successful, the `RESUMED` message is sent to the connecting user after ident lookups and similar have been completed completes. When this message is sent out, the connecting client's nickname is updated to be the old client's nick. From this point forward, the combination of old nickname and new connecting client's username+hostname is used to identify the client from that point forward.
```
ProgVal: typo: “have been completed completes.”

extensions/resume.md
```diff
@@ -0,0 +1,162 @@
...
+
+This command is sent after SASL authentication has successfully completed. After sending this command, the client continues connection registration as usual.
+
+If the client's able to successfully resume their connection, the server sends them a `RESUMED` message (and their nickname is set to `<oldnick>`). Otherwise, the server sends the `ERR_CANNOT_RESUME` numeric (and optional additional error numerics).
+
+If successful, the `RESUMED` message is sent to the connecting user after ident lookups and similar have been completed completes. When this message is sent out, the connecting client's nickname is updated to be the old client's nick. From this point forward, the combination of old nickname and new connecting client's username+hostname is used to identify the client from that point forward.
```
jwheare: > have been completed completes

typo

extensions/resume.md
```diff
@@ -0,0 +1,162 @@
...
+There are two new messages defined by this specification:
+
+- `RESUME`: Indicates that the client wishes to resume a previous connection.
+- `RESUMED`: Indicates that a client has disconnected+reconnected to the server.
+
+Below, we go through both of these.
+
+#### `RESUME` Command
+
+This command is sent before starting SASL authentication, and indicates that the client wishes to resume a previous connection to the server which may still be active. The command is sent with the following format:
```
jwheare: > This command is sent before starting SASL authentication

Should be "after SASL authentication has completed", or just drop this clause, it's mentioned further down.

extensions/resume.md
```diff
@@ -0,0 +1,162 @@
...
+### TLS
+
+Clients may initially connect with TLS, only to later attempt to `RESUME` from a connection without TLS. In this situation, servers SHOULD deny the attempt with a message such as _"Cannot resume connection, client is no longer using TLS"_.
+
+This is intended to ensure that TLS-only channel modes are not bypassed, and that clients prefer continuing to use TLS.
+
+
+## Examples
+
+A client with the nick `dan` reconnecting. The old connection used the username `~old` and the host `192.168.0.5`, and the new connection uses the username `~d` and the host `127.0.0.1`:
```
jwheare: Mention new nick is `dan-` and ideally choose a more obviously different name.

extensions/resume.md
```diff
@@ -0,0 +1,162 @@
...
+    S: :dan!~old@192.168.0.5 RESUMED dan ~d 127.0.0.1 2017-04-13T15:12:51.620Z
+
+
+## Implementation Considerations
+
+This section notes considerations software authors will need to take into account while implementing this specification. This section is non-normative.
+
+Right now, when clients detect that their connection to the server may have dropped they tend to send a `QUIT` command, close their current connection and then create a new connection to the server. In cases where the server supports resuming connections and SASL is configured for this server, clients may find it more useful to attempt to establish a new link to the server and resume the connection before closing their old one. If this is done, clients should be able to better take advantage of connection resumption.
+
+In addition, users sometimes manually reconnect when they see that their is lag on their connection. In these cases, clients may also wish to do the above rather than closing the connection and then reconnecting. However, it could be good to make this configurable.
```
jwheare: > see that their is lag 

"there"

extensions/resume.md
```diff
@@ -0,0 +1,162 @@
...
+
+
+## Connection Registration
+
+Upon connection, clients request the `draft/resume` capability. If they receive both this capability and the SASL capability, they can continue attempting to resume their old connection.
+
+After successfully completing SASL authentication, clients wishing to resume an old connection send an appropriate `RESUME` command indicating the nickname they wish to take over and how long they've been disconnected for. If resuming is successful, clients are sent the `RESUMED` message and then connection registration continues, eventually ending in the registration burst (`001`-`005` and all).
+
+Once the registration burst has been received, the server sends the client channel `JOIN` messages as appropriate (and the usual numerics that are sent on joining a channel), and dispatches the relevant messages and numerics to other clients in related channels. The new client is also given the user modes that were active on the old client, as appropriate.
```
jwheare: >registration burst has been received

This is for the server, so should be "has been sent"

prawnsalad:
There's no mention of the length of time a server may allow a connection to be resumed. Is the intended use to be able to resume after 5mins of disconnection? 1 hour? Or x2 ping timeout length?

And I'm assuming that once the RESUME cap has been negotiated that's when the server will not simply quit the user after a ping timeout. Which leads me to think the cap should also advertise the length of time a connection can be resumed within?

prawnsalad:
As mentioned on other specs, scoping any errors to the command will help tidy things up and remove the need for extra numerics. Eg. `resume ERR_CANNOT_RESUME`.

DanielOaks:
Honestly I haven't given thought to the possibility of letting/having the server extend the ping timeout length when the cap's negotiated, or similar. I suppose servers could definitely do so if they want to, that'd be interesting. Maybe the cap's value could be a series of k/v's similar to sts' params.

Right now, it's intended to just allow reconnection up to the point where the server disconnects the client due to ping timeout, but allowing servers to make it more clear (and extend how long the client connection stays open if they've negotiated the right cap) definitely makes sense, thanks for the yell with that!

extensions/resume.md
DanielOaks: It should be before SASL authentication has completed. Once SASL auth is finished, registration completes and the client is connected normally. This command replaces the usual registration process with playing the reg info, current modes, connection to any channels the client's currently joined to and then message history if the server has it.

Sending the command after SASL auth has been completed wouldn't really work since the client would already be connected and the server would need to rename+merge the client connections in a much weirder way, which just seems sketchy imo.

jwheare: Either way, the example transcript, and line 43 and contradicts it:

> This command is sent after SASL authentication has successfully completed

DanielOaks: Fair enough, thanks for the yell, I'll fix that up!

prawnsalad:
So if I've understood this correctly, the only way to make use of this is to guess the servers ping timeout length, and ping the server from the client at an interval shorter than that to try catch it before the server does. This sounds messy and ripe for a connection race condition.

Having an increased timeout once the cap has been negotiated will make this more useful.

extensions/resume.md
```diff
@@ -0,0 +1,162 @@
...
+
+Connection resumption is done on a best-effort basis. If the resumption fails clients SHOULD continue with a regular reconnection.
+
+#### `RESUMED` Message
+
+This message is sent when a client successfully resumes a connection. For clients that have not received numerics indicating that their connection registration is completed, this numeric means that their `RESUME` attempt was successful. For registered clients, this indicates that the given client disconnected and reconnected to the server.
+
+This message has the following format:
+
+    :nick!olduser@oldhost RESUMED nick user host [timestamp]
```
prawnsalad: This shows `nick!olduser@oldhost` yet further below says that they are the new info, `'user' and 'host' are the username and hostname that the new connecting client has`. Shouldn't this be displayed as `nick!newuser@newhost`?

DanielOaks: Nah, so the source is the old nickname, old username and old hostname, and the new nick/user/host are included as params on the message so clients receiving this message can update it.

prawnsalad:
Thoughts on making use of the "resume/d [subcommand]" format instead of adding new numerics? We keep on saying this is better going forward whenever it comes up, so we should start doing so.

extensions/resume.md
prawnsalad: Ah yes, my bad. It's referring to the params.

DanielOaks:
Eh, I'm happy to go either way if decent syntax comes up, but having the same message represent either both a request and a failure (if we use `RESUME` and then `RESUME ERROR`), or a success and a failure with different lengths of params (i.e. `RESUMED <blah>` and `RESUME FAIL`) feels iffy to me. I think it feels iffy just because neither of those currently have any sort of subcommand format, so adding one in feels dodgy.

extensions/resume.md
DanielOaks: Clarified the language there with 6451d1f5ef79303708c7d707a0b8f6693106908a, thanks for the note.

DanielOaks:
I've got a rough proof-of-concept implementation of this [here](https://github.com/oragono/oragono/releases/tag/v0.11.0-alpha), just using the fake numeric `999` for resume errors.

DanielOaks:
As a note, there's a running implementation of this on [`testnet.oragono.io`](ircs://testnet.oragono.io:6697/) for anyone interested in giving it a shot and seeing how it works.

jesopo:
Reading over this, I realised it didn't really talk enough about how authentication factors in to this to prevent people `RESUME`ing sessions they shouldn't be allowed to and that got me thinking that it might be a good idea to somehow share a session resumption token with a client when they successfully connect so that it can be saved and used when reconnecting to prove a client has permission to `RESUME` the session (this would lead to the syntax changing to something like `RESUME <nickname> <token> [timestamp]`.)

This would avoid the complexities of factoring SASL in to the process of `RESUME` and thus (I think) make adoption of this spec easier.

DanielOaks:
Well, even if a token did exist I'd like to gate it on some sort of account or similar authentication. You shouldn't just be able to guess a token and takeover someone's session – but having that token in addition to the SASL auth would prevent some other issues around different clients of yours being able to accidentally take over each-other's sessions.

Given the adoption of SASL out there, both in servers and clients, I don't think it'd be as much of a contributing factor to this getting adopted. The complexity, maybe, but honestly I'd prefer to be relatively certain about the security of this by piggybacking rather than creating a custom security method that may be nowhere near as secure as SASL depending on how the user's authenticating (e.g. SASL EXTERNAL would be much more secure than any sorta token system we could make, particularly because the resume token would need to be issued on/near connection to cover the use case this spec's intended to address).

**side note:** Was asked in-channel whether a PR to disconnect but ask the server to leave the session open would make sense. It... might be used by software, but I don't see it as significantly different to this proposal, apart from the relatively smaller application in that it could presumably only be used by clients that know they're going to intentionally disconnect. If anything, might make sense to add a similar sort of command to this since that would let both cases be covered nicely by the same mechanism?

It's also worth keeping #330 in mind – while the purpose is different, there is a biiit of overlap with the above proposed case and what's done by the linked `migrate` extension.

jesopo:
Fairly certain we can get pretty solid security with e.g. a 128char token with good entropy that is strong enough that we needn't worry plus if we have to integrate this with SASL, it means clients not set up to use SASL (even if they have SASL support) can't use it. I'd recommend we instead go for "when SASL has been used to create the session, it is also required from the connection `RESUME`ing session"

RyanSquared:
The main issue I see here is that if SASL is the only thing required to regain a session, people can then regain access to a session where someone using the nick of someone else - perhaps like a joke, which happens often - who could have access that the person regaining the session of, shouldn't have access to. A random token combined with SASL _matching the original user_ would be a good solution IMO, and something even as simple as 256 bits should be more than unguessable.

DanielOaks:
I'd wanna hear from other implementers. IRC Servers ultimately would need to be the ones to ensure how secure the token is, the sort of entropy used in creating it, etc. If we can do it while not making it easier to take over someone's connection / let someone else access their user account or info, cool.

DanielOaks:
@RyanSquared _"SASL matching the original user"_ - Yeah that's what's intended (and done in Ora's implementation). Reading through, I'm nowhere near explicit enough in the spec as it's written right now so will make changes so that's crystal clear. Sounds good on the combined stuff, that's the execution I have in mind.

jesopo:
@DanielOaks by "combined stuff" do you mean that SASL be optional for this unless the original connection used it?

DanielOaks:
@jesopo I'd much rather SASL be enforced and non-optional. Happy to hear opinions from other implementers on this point though, and specifically on whether we could do a token-only solution for the connections not originally using SASL without affecting security too much.

DanielOaks:
I think that:

1. The possibility of breaching the security of those clients isn't worth weakening the SASL requirement. But that's why I'm opening it up to other implementers to weigh in, esp on the security aspect (since we both know the opinions you and I hold on it).
2. Ehh, I think that's a reasonable tradeoff. The number of nets without services aren't too large these days, and the large ones that lack services are ones that're generally not going to implement (m)any new stuff from the v3 WG, regardless.

But def happy to hear other opinions

prawnsalad:
> This would not work at all ever for networks without nickserv

> Ehh, I think that's a reasonable tradeoff.

This tradeoff would be a huge annoyance for casual users which I see a lot of. If we put something in place, it should cover as many cases as possible.

A decently sized token based session should be ok, it's how websites+cookies work to keep user sessions around and has proven to be just fine. That is, as long as TLS is used which we may want to require to increase security there  in the same way. Mostly session hijacking via traffic sniffing in some cases.

RyanSquared:
That's definitely a good point about TLS that I hadn't considered - mostly because the network I manage is TLS only.

jwheare:
I'd personally be comfortable with a strong token only approach, not coupling this with nickserv accounts or otherwise. And yes, TLS only makes sense (fwiw browsers are gating new features on TLS too)

extensions/resume.md
DanielOaks: Added language clarifying that the timestamp on the `RPL_MONOFFLINE` message should be the current time. From the server's perspective the client never really disconnected, so I think it makes sense for it to just send them both as current. The (apparent) time'd out time from the client is given with the `RESUMED` message, so that should suffice.

grawity:
As I understand it, resuming with a token will kill (ghost) the old session so that the new client can take over. I would say that definitely makes TLS required, because otherwise it makes it much easier for an observer to take over your session – they no longer need fancy TCP injection or spoofing packets from your IP address, they can just use the token.

DanielOaks:
I'm rewriting a lot of this to be based on the token rather than SASL, force TLS, and take into account some of the other comments made here. Thanks very much for the feedback everyone :)

DanielOaks:
Sorry, this rewrite's reeeal rough, but should illustrate the direction I'm going with it. I'll re-evaluate this, edit it up some, and continue with it.

extensions/resume.md
DanielOaks: Added language to this affect

jesopo:
BitBot's got a completely untested implementation of this as of https://github.com/jesopo/bitbot/blob/2b106adc050703be3055f2408ce21020c56e53e1/modules/resume.py, it'd be nice to have a server implementation of this to test against.

DanielOaks:
I'll update Ora's one to match the -0.2 impl sometime soon, and yell in here once it's shipped

Ascrod:
Just to clarify, is this intended to work across port changes for STS upgrades? I seem to recall someone saying so at one point, but I can't find where. If so, is it worth documenting this relationship?

jesopo:
Just want to check because I'm not sure it's been clarified: should clients be taking note of the specific IP they're connected to so that they definitely end up doing the `RESUME` against the right server? I assume `RESUME` isn't cross-server and a round-robin DNS is likely to push you on to a different server.

DanielOaks:
@Ascrod it's worth the client asking the server to resume, sure. @jesopo I haven't added any particular notes, but I want to add a line like _Clients attempting to resume SHOULD try reconnecting to the same IP address_, but I assume servers that implement `resume` would attempt to make it cross-server for simplicity on the client end though.

jesopo:
Awesome.

Are there any server/testnet implementations of `-0.2` yet?

DanielOaks:
I believe https://irc.darwin.network implements it. Ora's proper testnet will once I update it

extensions/resume.md
```diff
@@ -0,0 +1,233 @@
...
+    C: CAP REQ :draft/resume-0.2
+    S: CAP * ACK :draft/resume-0.2
+    S: RESUME TOKEN JKbAypzFiovffzuD8VEfcs6bOLrXsSenxsyZNt87Bb0YaQlyOzuQciz6cB2R
+
+
+## Resuming A Connection
+
+When a client detects that it has become disconnected from a server, it SHOULD try to resume before it breaks the existing connection.
+
+Upon establishing the new connection, the client begins capability negotiation, negotiates supported capabilities, and MUST confirm that the `draft/resume-0.2` capability exists. If this capability does not exist, the client continues connection registration without attempting to resume. If this capability does exist, the client sends the `RESUME` command and MUST wait for either a `RESUME SUCCESS` or a `RESUME ERR` message before continuing registration. It should be noted that the client MUST NOT perform SASL authentication if the `draft/resume-0.2` capability exists, as this will end connection registration and abort the resumption attempt.
```
slingamn: When implementing the server side of this (oragono#304), I got confused about whether the server needs to wait for `CAP END` before processing the resume token. The correct behavior is that the token should be processed immediately, i.e., a successful resume ends capability negotiation the same way `CAP END` does. This could probably use some clarifying language in the specification.

extensions/resume.md
```diff
@@ -0,0 +1,233 @@
...
+    S: CAP * ACK :draft/resume-0.2
+    S: RESUME TOKEN JKbAypzFiovffzuD8VEfcs6bOLrXsSenxsyZNt87Bb0YaQlyOzuQciz6cB2R
+
+### Commands
+
+#### `RESUME` Command
+
+This command MAY ONLY be sent by a client during connection registration, and indicates that the client wishes to resume their session.
+
+    RESUME <oldnick> <token> [timestamp]
```
jesopo: why do we need `<oldnick>`? the server already knows what old nick the token is tied to. seems pointless to make the client keep this extra piece of information and there's scenarios in which the nick of the old connection could have changed while the client was pinging out which would cause the new connection's `RESUME` to fail

DanielOaks:
So, `<oldnick>`. While we could allow resumption based purely on the token, I like requiring the specific nickname as it increases the burden of brute-forcing. An attacker can't just send random tokens and hope that they stumble on an active session, they explicitly need to attack particular users, which makes detecting and preventing this sort of brute-forcing on the server side more effective.

Mentioned by ilbelkyr in-channel, requiring the nick also makes auditing takeover attempts possible. For example, servers may wish to send the particular user a note saying that someone is trying to take-over their session, and/or prevent resumption on their session for some amount of time if an invalid one/few is detected.

DanielOaks:
In addition, which I'm okay with making the `<token>` the new bar for authentication (instead of needing SASL credentials, just having the token lets you take over an account, I'm really uncomfortable with the idea of doing it globally rather than scoped per connected client. To convince me that global is fine, I'd need to see the opinion of some people well-versed in security.

Alexendoo:
Even with a moderately sized token the number of nicks in use on any given server do not compare to distinct values of said token.

The examples use a short (45bits of base64 by the looks of it?) token, a 45bit token has `35184372088832` distinct possible values, billions of nicks would need to be in use to yield any marginal increase in security.

The difference for larger tokens (64/128 bit) is greater still.

DanielOaks:
comment from sling:

> with a global namespace of valid tokens, there is a potential concern about timing attacks
> basically uh
> if you do something like put all the valid resume tokens in a hashmap, then information leaks out via timing about which tokens exist

Alexendoo:
Adding a nick doesn't increase security for attacking a single user, since you already know their nick. It does does against attacks against multiple nicks, by `log2(number of nicks)` bits. Increasing the token size directly achieves the same benefit (ie adding 32 bits onto the token length would be the same as a server with just over 4 billion nicks).

Since the token must be long enough to make brute-forcing a single user impractical already, removing the nick argument is fine

DanielOaks:
had a big discussion about the security up/downsides of various approaches in-channel. making it user-scoped rather than global doesn't help much on the entropy side at all.

main things we want to prevent:

- being able to attack specific (high-value) users. the current implementation/spec lets attackers pick a specific nick of someone they like and specifically brute-force them, which isn't good.
- being able to brute-force any particular user, or all the users at once.
- being able to use a timing-attack to discover the token's value. just putting the token into a hashmap isn't great, because most hashmaps aren't designed to lookup in constant-time. we need to make sure that server devs can as easily as possible use a constant-time comparison function when checking the token.
- the resume failing if the server changes the client's nickname in the meantime (e.g. NickServ enforcement renaming someone to Guest12345)

---------

rough proposal which I'll write up in detail a tough later:

each client on connecting and requesting the `resume` cap gets a `resume-id` and a `resume-token`.

the `resume-id` is just a random, session-specific identifier. on a resume attempt, servers can use this to look up the client connection using a hashmap.

the `resume-token` is the primary cryptographically-random value that actually prevents session takeover. on a resume attempt, servers compare the provided resume-token with the session's resume-token using a constant time comparison function.

both the `resume-id` and `resume-token` are private, with the `-id` just being something the server can use to lookup which token to compare. the `-id` isn't meant to provide any additional entropy, and honestly if a server wanted to just use the UID here that'd work fine (though being random would be cool). the `-token` provides the entropy, and it's not the primary lookup mechanism to prevent things like timing attacks on the hashmap lookup.

this system should prevent timing attacks, since the token comparison is done using a constant time function. for some extra effort, if an attacker does manage to find a valid `resume-id`, the server can detect implement specific protections for that particular account, maybe send the user a message saying that someone's trying to resume their connection, etc.

I don't think this increases the difficulty too much for clients – instead of just storing just the token they now store the ID and token, and provide both. and no more relying on the old nickname and hoping it hasn't changed in the meantime.

example:

    -- getting the resume id and token
    C1 - C: CAP LS
    C1 - S: CAP * LS :multi-prefix draft/resume-0.3 sasl
    C1 - C: CAP REQ :draft/resume-0.3
    C1 - S: CAP * ACK :draft/resume-0.3
    C1 - S: RESUME CREDENTIALS btTBR3rtyMTIYbe JKbAypzFiovffzuD8VEfcs6bO

    --- and resuming --
    C2 - C: CAP LS
    C2 - C: NICK dan-
    C2 - C: USER d * 0 :An example user!
    C2 - S: :irc.example.com CAP * LS :multi-prefix draft/resume-0.3 sasl
    C2 - C: CAP REQ :multi-prefix draft/resume-0.3
    C2 - S: :irc.example.com CAP dan- ACK :multi-prefix draft/resume-0.3
    C2 - S: :irc.example.com RESUME CREDENTIALS m6neEMTE3gtRdfg jyeA2e2VRBeergvw3DF25eHBa
    C2 - C: RESUME dan btTBR3rtyMTIYbe JKbAypzFiovffzuD8VEfcs6bO 2017-04-13T15:12:51.620Z
    C2 - S: :irc.example.com RESUME SUCCESS dan

thanks to sling, emily, ilbelkyr, Herringway and Alexendoo for some really good input during the discussion. feeling a bit more confident now about the risks, things that aren't actually risks, and the entropy side of things.

slingamn:
>being able to attack specific (high-value) users. the current implementation/spec lets attackers pick a specific nick of someone they like and specifically brute-force them, which isn't good.

I don't think this is a concern. The current spec states that the token MUST be cryptographically strong, in which case it cannot be brute-forced even for a specific nick. (128 bits of entropy is enough to ensure this. If we can't trust the CSPRNG to produce unguessable tokens, we can't trust TLS to protect them either.)

In my view, this new proposal does not improve security, but it doesn't harm it either. It fixes an edge case (nick changes the client is unaware of) at the cost of adding implementation complexity on the server side. This is probably a good trade.

ilbelkyr suggested a variant of this where the token remains a single, opaque value, which can then be processed by the server in a manner immune to timing attacks (e.g., the token is just the concatenation of `resume-id` and `resume-token`); @jesopo was in favor of this variant. That seems fine to me as long as we add guidance to implementers on how to handle this correctly.

DanielOaks:
The single opaque value method would be fine too, so long as we could ensure (or at least do our best to ensure) that servers do implement it correctly. If we have the single opaque value method, I would like to very explicitly refer to the ID and the token/authentication part in the spec when talking about server implementations, and make sure it's not just hidden away in a considerations or some other section where server/services implementers could simply skip over it.

DanielOaks:
Gonna try to hit an appropriate middle-ground in terms of the spec text. I'll update it to use the single-opaque-value method and we'll see how it ends up.

Alexendoo:
Perhaps it's easier to store tokens in a manner that doesn't require constant time considerations, as these can be difficult to implement correctly

E.g. The server generates a secret that is kept locally

The server generates a random token as normal and sends it to the client

Instead of entering the token into a map, use HMAC(secret, token) as the key, the token can then be discarded and the HMAC recalculated when received from a RESUME

slingamn:
Secure HMAC verification also requires constant-time string comparison: https://codahale.com/a-lesson-in-timing-attacks/

Alexendoo:
Yeah the value of the HMAC could be worked out, however in this case that is fine, since knowing the HMAC doesn't let an attacker spoof a resume

Alexendoo:
To clarify: instead of trying to prevent a value from leaking it's replacing the value with something that doesn't matter if it gets leaked

slingamn:
That sounds correct. So this would be an alternative server-side implementation strategy for a proposal with a single opaque token.

extensions/resume.md
DanielOaks: Resolved in next update

extensions/resume.md
DanielOaks: SASL now isn't required, so this issue is resolved!

extensions/resume.md
DanielOaks: Now clarified in the spec, thanks for the yell here.

extensions/resume.md
DanielOaks: `<oldnick>` no longer required after some good security discussion back-and-forth. Thanks for the yell!

DanielOaks:
Updated to a method which only uses a single opaque token value. Allows implementers to build the tokens however they like (for example, they could use that HMAC system described above), with a secure standard method described that fulfills our security requirements.

Give us a yell for how this looks! Hopefully a lot easier for clients, while still being good and secure.

extensions/resume.md
DanielOaks: The `[timestamp]` value only allows other clients to know how much message history may have been lost. The reconnecting client will attempt to retrieve their missing history themselves (and/or get it replayed), they don't use or depend on this value in any way so allowing them to send a message ID in this space would only complicate things for those other clients.

extensions/resume.md
```diff
@@ -0,0 +1,255 @@
...
+Servers may wish to check the new hostmask of resuming clients, to ensure that it does not fall under their list of banned hosts or hostmasks.
+
+
+## Security Considerations
+
+This section notes security-specific considerations software authors will need to take into account while implementing this specification. This section is non-normative.
+
+When servers apply the old client's session information to the new client, ensure that the new client retains their own unique resume token. Clients shouldn't share resume tokens under any circumstances.
+
+Servers must construct resume tokens in a secure way. The [Resume Token](#resume-token) section above lays out a method that should prevent timing attacks. Any implementers creating their own method must protect against timing attacks and other possible attacks against their authentication method.
```
slingamn: I'd add something at the end here like:

>It is recommended that the token provide 128 bits of security strength. In the typical method of generating tokens, this can be achieved by ensuring that the secret "resume key" portion of the token contains 128 bits of entropy from a cryptographically secure pseudorandom source.

jesopo: I think that should be an `"at least 128 bits of strength"` with explanation of why that number was chosen

jesopo:
`draft/resume-0.3` implemented in bitbot as of https://github.com/jesopo/bitbot/blob/14418286/modules/resume.py

extensions/resume.md
```diff
@@ -0,0 +1,255 @@
...
+- Metadata.
+- MONITOR list.
+- User modes.
+- Logged-in user account.
+
+Session information that MUST NOT be applied from the old client and replayed includes:
+
+- Enabled client capabilities.
+- The old client's resume token (after completing the resume request, this token is thrown away).
+    - Including both the old client's resume ID and resume key, if using the token values described above.
```
Alexendoo: This either no longer applies, or is referring to the values described below

DanielOaks: Oh, yeah I moved things around. wilfix, thanks

slingamn:
oragono's master branch (anything containing https://github.com/oragono/oragono/commit/5bd2133946ce519609fd7d23b09e8632793e245b) now implements the server side of `draft/resume-0.3`.

extensions/resume.md
slingamn: +1 for "at least 128 bits of security strength". Revised comment:

>It is recommended that the token provide at least 128 bits of security strength. This reflects recommendations such as that of NIST [1] that modern systems target a 128-bit security level. In the relevant threat model of online attack by a classical adversary, this can be achieved through the use of a secret with 128 bits of entropy from a cryptographically secure pseudorandom source. With the 'typical method' of generating tokens, it is sufficient to ensure that the 'resume key' portion of the token has this property.
>
>[1] https://csrc.nist.gov/publications/detail/sp/800-57-part-1/rev-4/final

(Not for inclusion in the spec, a more colorful version of the rationale: https://security.stackexchange.com/a/6149)

slingamn:
It might help to explicitly address how `RESUME` interacts with the `PASS` command, e.g., by specifying that a successful RESUME will complete the registration without requiring `PASS`.

Either way, it would help to add some language like this to the security considerations section:

>A successful resume immediately concludes the registration process. Implementations should therefore verify that `RESUME` cannot be used to evade any checks that are normally enforced during registration, such as the `PASS` command, IP bans, or nickmask bans: each such check must be imposed on at least one of the old connection and the new connection.

extensions/resume.md
DanielOaks: Language is now included, reads well to me and provides some good advice/rationale. Thanks!

extensions/resume.md
DanielOaks: Fixed this whole line up to be a lot clearer, thanks!

extensions/resume.md
```diff
@@ -0,0 +1,258 @@
...
+    C2 - C: AUTHENTICATE PLAIN
+    C2 - S: :irc.example.com AUTHENTICATE +
+    C2 - C: AUTHENTICATE YnVubnkAYnVubnkAYnVubnk=
+    C2 - S: :irc.example.com 900 dan-backup-nick * bunny :You are now logged in as bunny
+    C2 - S: :irc.example.com 903 dan-backup-nick :SASL authentication successful
+    C2 - S: :irc.example.com 001 dan-backup-nick :Welcome to the Internet Relay Network dan-backup-nick
+    ... regular connection ...
+
+
+## Implementation Considerations
```
DarthGandalf: If client reconnects to a different server, the new server should be allowed to resume the connection, so the connection would be transferred to the new server. And they need to handle corner case when several clients try to resume the same connection simultaneously.

jesopo: I'd honestly prefer `resume` to require you connect to the same server again - just because i think server jumping is a bit too much like #330 and will be complex for ircd devs to implement

RyanSquared: I think it would be useful to have optional support for it in case for some reason you can't contact the old server.

SaberUK: As a server maintainer I can't see any massive technical issues with implementing server jumping.

slingamn: Would it make sense to have a soft recommendation that clients attempt to reconnect to the same server initially (for compatibility with servers that don't synchronize resume tokens across nodes)?

DanielOaks: I thought that recommendation was already in the spec. If it's not, I'll add a note like that.

DanielOaks: Added guidance along both of these lines, for both server and client authors!

Alexendoo:
Brought up in channel:

There is a gap left between `resume` and `brb`, where the client loses connection to the server but the TCP connection is closed for whatever reason.

The flow of `resume` could be altered slightly to accommodate it, by considering the client active for the timeout duration even if the TCP connection closes (without a `QUIT`). 

`brb` would continue to be a way for clients to manually enter into the active but disconnected state, inheriting the same timeout logic.

jesopo:
A clean TCP socket disconnect without a `brb` should always fire off `QUIT`s imo - whether or not the client actually sent a `QUIT` or not. This circumstance is common and we don't want to just introduce a higher rate of ping timeout `QUIT` messages.

SaberUK:
>It's very common for clients to disconnect without properly `QUIT`ing

I don't see why we can't have sending `QUIT` when disconnecting as a mandatory part of the resume spec tbh.

jesopo:
> > It's very common for clients to disconnect without properly `QUIT`ing
> 
> I don't see why we can't have sending `QUIT` when disconnecting as a mandatory part of the resume spec tbh.

2 things.

1. it's not always the irc client's fault that the tcp socket was closed
2. do we really want to make a tcp connection close without a `QUIT` be an implicit "please hold this session open"?

SaberUK:
>* it's not always the irc client's fault that the tcp socket was closed

Yes, and they'd probably want to resume rather than disconnecting if possible in those cases.

>2\. do we really want to make a tcp connection close without a `QUIT` be an implicit "please hold this session open"?

I don't see any significant reasons why not?

Alexendoo:
Plus it requires opting in via a cap, so it's not entirely implicit

jesopo:
I just don't think the majority of these scenarios are ones in which people will want to `RESUME`

dequis:
Quoting from upthread:

**[prawnsalad](/prawnsalad)** commented [on Aug 21, 2017](#issuecomment-323659897)

> There's no mention of the length of time a server may allow a connection to be resumed. Is the intended use to be able to resume after 5mins of disconnection? 1 hour? Or x2 ping timeout length?
> 
> And I'm assuming that once the RESUME cap has been negotiated that's when the server will not simply quit the user after a ping timeout. Which leads me to think the cap should also advertise the length of time a connection can be resumed within?

**[DanielOaks](/DanielOaks)** commented [on Aug 21, 2017](#issuecomment-323696508)

> Honestly I haven't given thought to the possibility of letting/having the server extend the ping timeout length when the cap's negotiated, or similar. I suppose servers could definitely do so if they want to, that'd be interesting. Maybe the cap's value could be a series of k/v's similar to sts' params.
> 
> Right now, it's intended to just allow reconnection up to the point where the server disconnects the client due to ping timeout, but allowing servers to make it more clear (and extend how long the client connection stays open if they've negotiated the right cap) definitely makes sense, thanks for the yell with that!

I think we never addressed these, and I was left with the idea that any disconnect that isn't an explicit QUIT would mean the server keeps the client alive as if it were a longer ping timeout. Apparently that wasn't actually part of the spec.

FWIW, Oragono seems to have slightly larger timeouts when the resume cap is enabled ([IdleTimeoutWithResumeCap](https://github.com/oragono/oragono/blob/928caba92212fb5ae2fb96a2c037e8df776827f7/irc/idletimer.go#L20))

It would be a bad idea to restrict this feature to pure ping timeouts only, there are many things that can kill a connection and some of them look like clean disconnections.

BitlBee's session takeover feature is very similar to this spec, but relies on the old connection being still up, which makes it very unreliable. I dislike that feature because it fails most of the time, and it would be very easy for me to implement this spec based on that code, but that'd be just as unpleasant.

So IMO we need to:

* Have a timeout after non-QUIT disconnections 
* Make QUIT mandatory when actually quitting
* Mention that servers might want to have a longer ping timeout time when resume is enabled.
* Optional: roll up BRB in this spec? It can be separate but it's very closely related, and complements this spec nicely.

slingamn:
tentative +1 for including BRB in this spec

extensions/resume.md
```diff
@@ -0,0 +1,262 @@
...
+
+The `ERR` message indicates that the resume attempt failed. `<info>` is an information text string indicating the reson for the failure (e.g.: `"Cannot resume connection, token is not valid"`).
+
+#### `RESUMED` Message
+
+    :nick!olduser@oldhost RESUMED <user> <host> [timestamp]
+
+This message is sent to indicate that a client has reconnected. `<nick>`, `<olduser>`, and `<oldhost>` indicate the client that has reconnected, and are the details of the old client. `<user>` and `<host>` indicate the reconnecting client's new username and hostname, and the client MUST process this information as they would a regular [`CHGHOST`](https://ircv3.net/specs/extensions/chghost-3.2.html) message. `<timestamp>`, if given, is a timestamp of the form described above which indicates the last message received by the reconnecting clent.
+
+Upon receiving a `RESUMED` message, clients MUST display in some way that the given user has reconnected. If `<timestamp>` is given, clients SHOULD use this to display how much message history seems to have been lost.
```
Alexendoo: I think this is better left as a decision for clients to make, clients may wish not to display reconnects. They may not be displaying JOINs and QUITs at all

DanielOaks: Makes sense. I'll make the text here more exact as to why clients **should** make sure their users know that a reconnection happened and history might've been lost, and less demanding as it is right now.

DanielOaks:
For what it's worth, the spec already mentions extending the ping timeout length explicitly: _"In addition, servers MAY extend the "ping timeout" length for that client (with the expectation that if the client disconnects, it will try to resume its connection as described here)."_ Referring to clients that have negotiated the cap. I'll make this more explicit, though the specific duration will remain up to implementer's choice.

What I'm going to do, with jes's blessing:

- Pull `BRB` wholesale into this spec, and present it as a command that can be used if the resume cap is advertised/negotiated (since it already depends on having a resume token, etc).
- Make the server hold the connection open for some server-decided length of time if a client that's negotiated `resume` disconnects uncleanly in any way, in addition to just something like a ping timeout happens and their connection stays open. For example, if the socket closes without a `QUIT` or a `BRB`, and the client has negotiated resume capabilities, the server SHOULD leave the session open for the same length of time that they'd normally hold it open for a `BRB`.

Thanks for the comments on this, much appreciated. Should end up with something more clearly

jesopo:
I'd like the wording around missed-chat-playback to get a little tighter on this; there was confusion on IRC about whether this spec mandates chat playback or not.

I propose this spec mandates that servers automatically send chat playback on `RESUME` *unless* the client also supports `CHATHISTORY` (#349, would need to become a `CAP` though) - this would allow clients to lazy-load (when needed) channel chat playback as well as actually having control of what is played back.

DanielOaks:
Makes sense to go with lazy-loading in that way, for clients that do support actually grabbing chathistory on their own. If chathistory is made a cap it's a simple check on when the resume command is sent after reconnecting. If chathistory isn't a cap it's an optional argument between RESUME and the token which indicates flags to resume with, the one specified one being `'nohistory'`.

jwheare:
CAP is probably the best bet, even though it might seem weird that it's mainly needed to suppress the auto-playback feature behaviour that isn't chathistory.

extensions/resume.md
DanielOaks: This is now done

extensions/resume.md
DanielOaks: This is now done

DanielOaks:
So the cap stuff and chat history playback is weird. The BOUNCER proposal also requires a way to say "Yo don't replay all chat history to me I'll request it especially", which suggeeeests that the right way to tackle this is for `chathistory` to have a requestable cap which indicates that chat history should or should not be replayed. So I'm going to leave it alone in this spec for now and assume that the chathistory spec will be extended to say something explicit on this in future.

DanielOaks:
note: we can't just say they MUST treat disconnections as `BRB` because, hey, the server might not have brb available, and depending on spam/abuse the server may need to not treat all disconnections as brb. I think the current text lets servers do it if it makes sense (as we have in Ora)

Herringway:
what error does a client get when RESUME is used after a successfully RESUMEd connection?

slingamn:
According to my reading, a successful resume immediately concludes registration, so you should get `REGISTRATION_IS_COMPLETED`.

DanielOaks:
small case: client sends resume cmd, connection breaks in the meantime (maybe server needs a couple secs to validate the token and in the meantime connection diez), client needs to try again. a small para with this text might be appropriate but I will rewrite it to be less shoddy and bad and dodgy: "if the client does not receive any response from the RESUME command and the connection times out, the client MAY reconnect and attempt resumption with that same token" or whatevr

slingamn:
Spelling out the Byzantine fault scenario we were discussing:

1. Client negotiates resume, receives a token `1rSSfXrBW52mejcxBbRm-eTWWhdMLFSVNgtqTIEmN4Y`, completes registration
1. Client's connection drops
1. Client reconnects, negotiates resume, receives a token `2jjwBWK9mRA-9huWPZtas0er7tuS0Dmv0RQtQZkcMMA`
1. Client sends `RESUME 1rSSfXrBW52mejcxBbRm-eTWWhdMLFSVNgtqTIEmN4Y 2019-05-17T14:56:43.733Z`, i.e., attempts a resumption with the first token
1. Connection drops again before a response is received
1. Client is uncertain whether the resume was processed (making it necessary to do further resumptions with the second token), or whether to retry with the first token
1. It should be unproblematic to write language that allows the client to retry with either (possibly on a new connection?)

slingamn:
Per discussion in #ircv3, we're no longer going to propagate the new session's `USER` field to the old client, and we're removing the corresponding `user` field from the `RESUMED` message.

Rationale: it was not clear on the basis of the existing specification that the new session had to send a `USER` line before `RESUME` could be processed. This change removes that dependency

extensions/resume.md
```diff
@@ -0,0 +1,306 @@
...
+
+On a successful request, the server:
+
+1. Terminates the old client's connection.
+2. Updates the client's username, hostname, and resume token (or lack of one) to match the new client's information.
+3. Plays the 'joining server' burst to the new client.
+4. Replays the client's session to the new client.
+5. Sends `RESUMED` or `QUIT`+`JOIN` messages to other clients as appropriate.
+6. Sends clients that have the reconnecting user `MONITOR`'d one `RPL_MONOFFLINE` numeric and one `RPL_MONONLINE` numeric indicating that the user has reconnected. The timestamps on both these messages, if sent, SHOULD be the current time.
+7. If message history could not be replayed (because it is not stored, or for any other reason), sends the new client a `RESUME WARN` message making them aware of this.
```
slingamn: `WARN RESUME`?

slingamn: Also, should we standardize a `WARN RESUME` code for this? I propose `WARN RESUME HISTORY_LOST`.

extensions/resume.md
```diff
@@ -0,0 +1,306 @@
...
+| `INVALID_TOKEN` | `:<server> FAIL RESUME INVALID_TOKEN :Cannot resume connection, token is not valid` |
+| `REGISTRATION_IS_COMPLETED` | `:<server> FAIL RESUME REGISTRATION_IS_COMPLETED :Cannot resume connection, connection registration has completed` |
+| `CANNOT_RESUME` | `:<server> FAIL RESUME CANNOT_RESUME :Cannot resume connection, for a different reason described here` |
+
+If a client receives a `FAIL RESUME` message, regardless of the code, then they MUST abort the resume attempt and connect to the server normally instead.
+
+#### `RESUME` Message
+This message, sent from a server to a client, indicates that a `RESUME` request has been successful.
+
+    RESUME <oldnick>
```
jesopo: When compared to the server giving us a new RESUME token during CAP negotiation...

> RESUME TOKEN JKbAypzFiovffzuD8VEfcs6bOLrXsSenxsyZNt87Bb0YaQlyOzuQciz6cB2R

What if a user's nickname was `TOKEN`? Seems clumsy to use use `RESUME <nickname>` because of this.

extensions/resume.md
```diff
@@ -0,0 +1,306 @@
...
+
+If the token provided by the new client is validated by the server, the old client completed connection registration with the server, and both the old and new clients use TLS, then the attempt SHOULD be successful. If the attempt is successful, the server MUST send the client a `RESUME` message and complete connection registration immediately (at which time the state will begin to replay as described below). If the attempt is unsuccessful (for example, if either session is not using a secure connection), the server MUST send a `FAIL RESUME` message, and then allow the client to continue connection registration.
+
+If the client receives a `RESUME SUCCESS` message, connection registration will complete immediately and state will begin to replay as described below. If the client receives a `FAIL RESUME`, the client MUST continue registration as though they are joining the server normally.
```
jesopo: This disagrees with the description of s2c commands. `RESUME <nick>` is described above as being a success.

slingamn: I would be in favor of bringing back `RESUME SUCCESS <nick>` (that's what I had in my `draft/resume-0.3` implementation.

So the server-to-client message would be:

1. `RESUME TOKEN`
1. `RESUME SUCCESS`
1. `RESUMED`
1. `FAIL RESUME`, `WARN RESUME`, and friends (according to the standard replies extension)

slingamn:
The discussion of history in this specification is a little confusing:

1. "Upon receiving a RESUMED message, clients SHOULD display in some way that the given user has reconnected (as message history may have been lost and the users' chat may have been interrupted). If <timestamp> is given, clients SHOULD use this to display how much message history seems to have been lost."
1. "If message history could not be replayed (because it is not stored, or for any other reason), sends the new client a RESUME WARN message making them aware of this."
1. "When clients see a RESUMED message for another client which contains a timestamp, they can calculate how much time has passed since the timestamp and the current time and then display this next to the reconnect notice. Displaying this can assist users in knowing how much message history has been lost in private queries and channels."
1. "A client that disconnects and reconnects to the server should explicitly display the reconnection, even if they're able to resume successfully. This is so that the user knows why they may be missing message history and similar issues."

Basically, the spec is extremely risk-averse about history loss: all the clients are being told to assume the maximum amount of loss. But what about server implementations that are confident (e.g., because they have persistent message history) that no history was lost? Might they want to suppress the timestamp when sending the `RESUMED` message to friends?

slingamn:
Very rough proposal:

1. If it's possible that history was lost, the server MUST send the `WARN RESUME HISTORY_LOST` message. A `RESUME SUCCESS` message with no warning therefore indicates the server's belief that no history was lost.
1. When sent to friends, `RESUMED` with no timestamp indicates the server's belief that no history was lost. If it's possible that history was lost, the server MUST send a timestamp (if the client did not supply a timestamp, it can use the time of the last message sent from the client, or even an extreme lower bound like the server creation time).

jesopo:
Felt this might be good to jot some notes up about:

BitBot has fully working `draft/resume-0.4` capability and something I came across when putting BitBot against Oragono is that "millisecond rounding" caused a line BitBot saw prior to reconnection to appear in the  `chathistory` batch (had exact same timestamp as the timestamp requested in `c->s` `RESUME`). Overcame this by keeping track of the last-seen `msgid` per-channel and popping things before and including that `msgid` from the `chathistory` batch.

extensions/resume.md
```diff
@@ -0,0 +1,323 @@
...
+
+The `[status]` parameter is used to indicate the server's belief about how much history was lost. If it is the string `ok`, this means the server believes no history was lost, i.e., any messages that were missed will be replayed to the client. If it is a timestamp, then the client may have lost history messages that were sent between that time and the resumption. If it is omitted, then the client may have lost an unknown amount of history.
+
+Upon receiving a `RESUMED` message, clients SHOULD display in some way that the given user has reconnected (as message history may have been lost and the users' chat may have been interrupted). If `<timestamp>` is given, clients SHOULD use this to display how much message history seems to have been lost.
+
+The `RESUMED` message can only be sent to clients that have negotiated the `draft/resume-0.5` capability. Clients taht have not negotiated this MUST be either sent consecutive `QUIT` and `JOIN` messages that describe the reconnection (and how much history may have been lost), or no notification at all if no history has been lost.
```
jesopo: `taht` -> `that`

jesopo:
BitBot is up to `draft/resume-0.5` as of https://github.com/jesopo/bitbot/commit/6c64a2296ccbcee0e186448f2b9076c6b21648ee

slingamn:
How much interest is there in this from other vendors?

I'm considering removing this from oragono, because it doesn't seem to add much value relative to proposals like CHATHISTORY (#393) and [oragono.io/bnc](https://oragono.io/bnc).

edk0:
I'm interested in it. I agree about the direction you're going in with the bnc capability, but also think this is much closer to being compatible with the architecture I'm stuck with for now.

SadieCat:
I'm interested and have most of an implementation but it has a lot of things that need to be tested and its been deprioritised in favour of other things recently.

slingamn:
>It should be noted that the client MUST NOT perform SASL authentication if the draft/resume-0.5 capability exists and they wish to resume their session, as completing SASL auth will end connection registration and abort the resumption attempt.

Is this language still accurate? Successful SASL does not terminate registration on oragono, at least. It seems like the only things that can terminate registration are `NICK`, `USER`, `CAP END`, or `RESUME` itself.

DanielOaks:
suggestion from e, `WRONG_SERVER` `FAIL` code, used when the token tries to be redeemed on a different server and cannot be

DanielOaks:
another suggestion, explicitly tell clients how to act if they receive a WARN/FAIL code that they don't understand (if a new one is added in future how to act?)

edk0:
another suggestion: amend

> `<reason>` is used as an `AWAY` reason for the client until either the client resumes their session or the window of time for resumption closes - in which case it will be used as the displayed `QUIT` reason.

to

> `<reason>` is used as an `AWAY` reason for the client until either the client resumes their session or the window of time for resumption closes - in which case it will be used as the displayed `QUIT` reason. Upon successful resumption, the server SHOULD restore the connection's `AWAY` status to its previous value.

DanielOaks:
note to self: in the examples, could probably remove the source from most of the std-replies responses and possibly other things

DanielOaks:
Idea mentioned in the channel: make the server remember caps between sessions that are resumed? can probably cut down a lot on traffic needed to resume, client and server both already know what they've negotiated anyways.

dequis:
> Idea mentioned in the channel: make the server remember caps between sessions that are resumed? can probably cut down a lot on traffic needed to resume, client and server both already know what they've negotiated anyways.

Following that line of thought: how about assuming that in servers for which we have a resume token, we can just assume the cap is supported, and send the command before the CAP LS response?

So this example from the current version:

    C2 - C: CAP LS
    C2 - C: NICK dan-backup-nick
    C2 - C: USER d * 0 :An example user!
    C2 - S: :irc.example.com CAP * LS :multi-prefix draft/resume-0.5 sasl
    C2 - C: CAP REQ :multi-prefix draft/resume-0.5 sasl
    C2 - S: :irc.example.com CAP dan-backup-nick ACK :multi-prefix draft/resume-0.5 sasl
    C2 - S: :irc.example.com RESUME TOKEN JKbAypzFiovffzuD8VEfcs6bOLrXsSenxsyZNt8
    C2 - C: RESUME A8KgnZPYDaRiGMzZWLu2frVvtN7lbCxO3hTwGLO 2017-04-13T15:12:51.620Z
    C2 - S: :irc.example.com RESUME dan

Becomes

    C2 - C: RESUME A8KgnZPYDaRiGMzZWLu2frVvtN7lbCxO3hTwGLO 2017-04-13T15:12:51.620Z
    C2 - C: CAP LS
    C2 - C: NICK dan-backup-nick
    C2 - C: USER d * 0 :An example user!
    C2 - S: :irc.example.com RESUME dan
    C2 - S: :irc.example.com CAP * LS :multi-prefix draft/resume-0.5 sasl

If `RESUME` isn't supported anymore for some reason, it will get ignored silently, and that will be visible in the `CAP LS` response.

If it is supported, the first response from the server is the RESUME command, at which point the client can assume the old caps from that session are back

(optionally, the server could send a cap ack as if we just had requested it?)

If someone really wants to, they can send `CAP LS` first and then `RESUME` later somewhere around `NICK`/`USER`, but it seems more convenient to do it the other way around for the usual event-response flow.

Sending it in response to the activation of the cap (as the current version) is still a valid flow, just a higher latency one.

DanielOaks:
That flow definitely makes more sense, yeah

slingamn:
A note: one of the motivations for this change is to speed up reconnections of mobile clients that detach and wait for push notifications, but I'm not sure how viable it is for that purpose in its current form, since we'd have to keep the session information around for much longer than current implementations do (210 seconds in Oragono). This also has implications for presence and communication of presence to other clients.

Transfer of ISUPPORT (avoidable), MOTD (avoidable), and NAMES (probably unavoidable) were also suggested as reasons for slow mobile performance. It would be useful to have more details on mobile handshake performance (how much of it is roundtrips from the chattiness of the current registration sequence, how much of it is bandwidth).

edk0:
> 
> ```
> C2 - C: RESUME A8KgnZPYDaRiGMzZWLu2frVvtN7lbCxO3hTwGLO 2017-04-13T15:12:51.620Z
> C2 - C: CAP LS
> C2 - C: NICK dan-backup-nick
> C2 - C: USER d * 0 :An example user!
> C2 - S: :irc.example.com RESUME dan
> C2 - S: :irc.example.com CAP * LS :multi-prefix draft/resume-0.5 sasl
> ```
> 

Under the normal rules, C would then change nick to dan-backup-nick, which seems unideal.

slingamn:
In a potential low-latency version of the resume handshake, the server could send:

1. `RESUME SUCCESS`
2. `CAP LIST` (to compensate for any missed cap-notify)
3. `RESUME TOKEN` (the new resume token)
4. `JOIN` lines for all channels

and omit:

1. All of the conventional registration burst, including ISUPPORT and the MOTD
2. NAMES output (the client can re-request on demand, e.g., when the user tabs over to the channel)

slingamn:
Would it make sense to publish the session lifetime as a cap value?

DanielOaks:
I can't decide whether I like `resume` or not. The spec is pretty long (about 10 printed pages on A4 paper), or at least feels long for something that aims to solve a really simple issue ("Hey it's annoying when your connection drops and you need to use `/ns ghost`").

My specs tend to end up very long and wordy and precise, and a feature like this does require some specificity. But I feel like the current spec, as written, isn't as client-friendly as it could be. Server devs have seeeemed to like it though? (we have an impl, solanum has a pr, insp's been interested in implementing it).

Honestly it might just be that the spec as written is laid out very 'academically', whereas it should probably start up the top with a very quick rundown of 'here are a few flows inc. C2S proto stuff' that hotlinks to the actual message/etc definitions down below. Aiming to make it easier-to-understand for client devs. With a change like that, the length may not even be so much of an issue for me.

---

Now, onto some stuff that's `RESUME`-adjacent.

I like `BRB`. It's nice, it's shiny, it's (fairly) short and fairly specifically defined. And I do get the use case when it comes to installing upgrades (client/bot installs upgrade, sends a `BRB` and `RESUME`s when it spins back up), but it feels like there's two other use cases along similar sorts of lines that aren't addressed, so it sticks out:

- Server upgrades, network management and changes in general, servers needing to push clients actively to another server. Think `RPL_BOUNCE` but like, actually properly defined and all that junk. I think we have/had a spec for this floating about, and/or that maybe... plexus was looking into it? Maybe? My memory around that is hazy. This certainly wouldn't fit in this spec, but it's in the same areas that this spec touches, and a push-client-to-new-server spec/message/etc might use the resume feature.
- Mobile clients. In Ora we're experimenting with an `abbreviated-handshake` cap to speed up reattach for mobile clients. It doesn't really 'compete' or whatever with this spec, but 'fix mobile clients and irc' is a goal that's pretty often been pushed onto this spec and that people use to justify either liking (it solves this yay!) or not-liking (it doesn't solve this, boo!) this spec.

I don't mind `BRB` being a part of this spec very much (esp since it's like a paragraph of explanation if that), but it is an extra add-on to the original goal, so the goal stated in the spec should be updated to include a mention of it. I also want to make very clear that the above two use cases aren't things that this spec tries to touch in any way.

---

Another very specific point. This spec isn't 'first and foremost' designed to replay history. That's like an added bonus (since we're already sending a timestamp to notify other clients of how much history we lost, hell why can't the server just automagically replay that lost history to us). I think it's a bit too prominent in the spec and mentions of it should be dialled back, because this spec isn't for that. Or maybe the spec is for that and I need to accept it. Server implementors, thoughts?

---

Another point, implementors allowing `RESUME` across servers on a network. Obviously, this isn't always possible because S2S protocols haven't been designed with 'transferring active clients between servers' in mind, so implementing this takes some doing. I don't really think it's a knock against this spec but it is something that I've heard very consistently (why bother with this feature if it doesn't work across the network? it's useless then!). The current solanum pr only allows same-server resuming, the Insp plans last I heard were to allow it across the network, in Ora we only have one server so ¯\\\_(ツ)\_/¯

I'm interested in whether client devs think 'reconnect to the same server to use resume' complicates their reconnection logic significantly, or if there are other issues that make them not want to reconnect to the same server again as a rule (e.g. working around server issues, just not wanting to complicate their connection logic like that, etc). I've mentioned this and the general thoughts seem to be that it's either not enough of an issue for me to worry about or that it's a worthwhile complication. Further, would clients find a cap value like `resume=across-network` which indicates 'yes resume works across the network' actually useful?

slingamn:
I am opposed to ratification of `draft/resume-0.5` in its current form; in fact, I think clients and servers should not implement it.

The current spec tries to do too much and provides too little benefit to users, in exchange for a considerable increase in implementation complexity. In particular, the core purpose of the spec (as identified in its introduction) is essentially just to streamline or automate `NS GHOST`, i.e., the reclamation of stuck nicknames on reconnection. In the drafting process, this has become entangled with two unrelated concerns, neither of which is well addressed by the current text:

1. Giving users continuity of presence across disruptions to their connections. (In its current form, the proposal is capable of covering only very brief disruptions; the exact time window allowed is not visible to clients.)
2. Optimizing the handshake, in particular by removing the need for SASL. (In its current form, the proposal is not intended to reduce the bandwidth or the number of roundtrips in the handshake.)

On the server side, the spec is so difficult to implement that there are only two candidate implementations. One is Oragono, which doesn't have to consider the multiple-server case. The other implementation disallows resuming on a different server than the one you started on, which (a) increases the complexity of client reconnect implementations (b) is not currently exposed to clients.

There are workarounds for some of these problems, e.g., we could include both the server's extended ping timeout and whether all-server resume is supported in the CAP value. But I think at this point it's better to go back to the drawing board. For example:

1. oragono/oragono#1420 is a sketch of a proposal for a bandwidth-optimized handshake
2. If SASL performance (both the latency added by password checking, and the costs imposed on a centralized services node) is a concern, we could think about a proposal to issue "auth cookies" (most likely temporary passwords that could be submitted with SASL PLAIN, but which wouldn't have the same sensitivity and storage requirements as hashed user passwords)
3. If a server implementation is extending its S2S protocol to support transparently migrating a client between servers during `RESUME` (which, as I understand it, is a significant implementation barrier), it might be better to develop full "multiclient" capability instead (i.e., allowing multiple endpoint connections to the same nickname/presence).

slingamn:
Summarizing some discussion from #ircv3: the spec, as written, does not make IRC viable on mobile platforms (which are increasingly hostile to background TCP connections). So it's applicable mainly to connection disruptions and software restarts on desktops and servers. Moreover, not all implementations may be able to maintain history across spontaneous disruptions (as opposed to BRB).

So, the baseline supported use case is software restarts of desktop and server software. At which point the question is: is the (considerable) engineering effort involved in implementing this specification justified?

DanielOaks:
> In the drafting process, this has become entangled with two unrelated concerns, neither of which is well addressed by the current text: - Optimizing the handshake, in particular by removing the need for SASL

No, that's not the reason why SASL's disallowed with resuming a connection. The reason for that is just that... if you sasl, the connection registration process finishes, so you can't do it and also do something that hijacks the connection resumption process like resume. I'd be interested in client developer opinions on this, and how the spec handles this aspect.

> On the server side, the spec is so difficult to implement that there are only two candidate implementations.

This is a v3 draft, you really can't use 'number of implementations' as an indicator of spec worth/complexity. Hell, this is more implementations than most of our junk gets before it gets pushed to the site in draft form.

> But I think at this point it's better to go back to the drawing board. For example: [...]

1. I think these caps could be requested together.
2. SASL performance isn't a concern, the reason you can't do SASL and also resume is purely because of restrictions around the connection flow. And if I've screwed something up there, I'd be happy to resolve it.
3. Eh, maybe? It feels like it's much more difficult to dig into something like multiclient than it is to change active users between servers. Particularly since multiclient stuff relies on services authentication, and those other servers don't integrate as closely with their services packages as we do with ours in Ora.

> the spec, as written, does not make IRC viable on mobile platforms

Um. Yeah. The part of my comment which says _"'fix mobile clients and irc' is a goal that's pretty often been pushed onto this spec and that people use to justify either liking (it solves this yay!) or not-liking (it doesn't solve this, boo!) this spec. [...] I also want to make very clear that the above two use cases aren't things that this spec tries to touch in any way."_ indicates this.

> So it's applicable mainly to connection disruptions and software restarts on desktops and servers.

Yes. This is the point. I don't see why it couldn't be useful to mobile clients in the same way, just that they'd probably _also_ request whatever protocol-slimming-down cap we end up making for mobile clients.

> Moreover, not all implementations may be able to maintain history across spontaneous disruptions (as opposed to BRB).

Yes. This is the point. I don't expect implementations to suddenly start storing history when they haven't before -- that bit's purely opportunistic if it's available, and if it's not then cool other clients will know that the reconnecting client missed messages at least.

DanielOaks:
If someone is really bent on the 'this doesn't solve mobile, so it's not worth it' then what does a replacement for this spec which focuses on mobile look like? Does it aim at the same thing that this specification aims at, or is it pretty orthogonal to this?

The two alternate things which I've heard which would be mobile alternatives to this spec/functionality are:

- Something that speeds up connecting generally, see also [abbreviated-handshake](https://github.com/oragono/oragono/issues/1420). While this does better help things like mobile devices connecting, I don't really see it as 'competing' with this spec. I could see a device requesting both and the two inter-operating just fine.
- Something that builds a full persistent presence into the server/bouncer/etc side. While, yeah, a persistent presence and that kind of integration helps mobile clients very much (that's why so many mobile clients run through bouncers with history-grabbing, etc), I don't think big, large, traditional IRC servers are going to transition to that any time soon. For servers with persistent presence, this specification doesn't make sense. In a typical `client<->bouncer<->ircd` setup, I'd imagine the 'persistent presence' stuff to be built and work between `client<->bouncer`, and for this spec to work on `bouncer<->ircd`, generally.

slingamn:
>Something that builds a full persistent presence into the server/bouncer/etc side.

That's my position: doing this is necessary to support mobile, and having it would make RESUME redundant. Optimizing the handshake is just the icing on the cake.

>In a typical client<->bouncer<->ircd setup, I'd imagine the 'persistent presence' stuff to be built and work between client<->bouncer, and for this spec to work on bouncer<->ircd, generally.

I think we're on the same page here about use cases here. I just don't see bouncer<->ircd as something that's critical to streamline or optimize, particularly given the other limitations on functionality we're accepting.

>No, that's not the reason why SASL's disallowed with resuming a connection. The reason for that is just that... if you sasl, the connection registration process finishes, so you can't do it and also do something that hijacks the connection resumption process like resume.

As discussed, I believe this to be inaccurate.

DanielOaks:
As mentioned in the meeting, unless there's some more concrete client interest in this feature then I'm leaning very much towards killing it. As-is it probably isn't useful enough to justify the complexity.

If we do want to do something similar down the line as part of a client-redirection feature (e.g. a way for servers to push clients to new servers) then this'd probably be a workable base for that though, given it is fairly complete.

jwheare:
To clarify, I closed this in response to what looked like lack of interest from the proposers of the spec, and what appeared to be confused/conflicting ideas around what the spec is actually for. Apologies if that was hasty, I can reopen this if that's not the case.

From my own perspective as a client developer, I'm seeing resume as a solution to "faster and less noisy reconnect after brief/sudden loss of connectivity" or "brb for a software update restart". From that point of view, it seems like a cool concept, but not worth the length/complexity of this spec to achieve. And it's sounding like it will be hard to ratify as a result.

DanielOaks:
I probably would've closed it in a few days anyway.

With regards to 'lack of client interest', for me that's mostly just me informally mentioning it around a few client devs and them largely seeming meh or unenthused. Very unofficial, but yeah it affects my want to continue working on the spec.

If clients like the idea/spec in its current form, awesome. If they like the idea but are unenthused about the spec as it stands, cool I could work with that. If they aren't really into the idea at all, then eh.

Generally it's felt like server devs are pretty into this while client devs aren't. I think the spec right now is way too long for client devs to be interested in implementing it, and it's also not really written in a way that helps people understand how to implement it (this is my spec/tech writing nerd coming out but it's written in a very 'clinical' way that server devs may appreciate but I can't see clients liking at all). I could probably spend time working on this and trying to make it shorter/simpler, and improve the structure of the spec, and make sure there's much less confusion about what exactly this feature aims to solve. But fact is that right now I'm not enthusiastic enough to put the time into doing that.

kylef:
> Generally it's felt like server devs are pretty into this while client devs aren't. I think the spec right now is way too long for client devs to be interested in implementing it

Would server dev's be interested in migrate more? I see both resume and migrate as trying to do something very similar (allowing the other party to reconnect), just one is orchestrated from the server, and the other from the client.

As a client author, I'd happily implement migrate as I can see this as a gap from being able to keep networks up to date and secure from a maintenance point of view.

For resume, it doesn't solve any of my problems as it is today, but that's not to say it doesn't solve problems for all clients. Perhaps it could be useful for bouncers and similar which may want to perform any patching at a scale without disrupting their customers.

ailin-nemui:
as a client dev I like this extension idea

SadieCat:
>Would server dev's be interested in migrate more?

As a server developer no. It has an entirely dififerent use case and honestly it seems like it's trying to solve a problem that doesn't really exist (at least not for us).

kylef:
@SadieCat Orchestrating security patching/keep software up to date without disconnecting and disrupting ongoing conversations for connected clients is the use-case I imagine the most for migrate. I don't see this as an great user experiance at the moment.

SadieCat:
We've solved this problem by pushing TLS out to a separate process (in our case HAProxy).

kylef:
@SadieCat That doesn't count for all cases of security patching, the kernel/os, updating TLS libraries for active connections etc.

ValwareIRC:
:eyes:

PR #308, opened 2017-06-03
[WIP] Channel renaming extension.

SadieCat:
[Rendered view here.](https://github.com/SaberUK/ircv3-specifications/blob/rename/extensions/rename.md)

### DIFF
--- a/extensions/rename.md
+++ b/extensions/rename.md
@@ -1,0 +1,91 @@
+---
+title: IRCv3 `rename` Extension
+layout: spec
+work-in-progress: true
+copyrights:
+  -
+    name: "Peter Powell"
+    email: "petpow@saberuk.com"
+    period: "2017-2018"
+---
+
+## Notes for implementing work-in-progress version
+
+This is a work-in-progress specification.
+
+Software implementing this work-in-progress specification MUST NOT use the unprefixed `rename` capability name. Instead, implementations SHOULD use the `draft/rename` capability name to be interoperable with other software implementing a compatible work-in-progress version.
+
+The final version of the specification will use an unprefixed capability name.
+
+## Introduction
+
+Currently there is no reasonable way to rename channels. All existing methods involve closing down the channel and redirecting it to the a new one. This is undesirable as access and ban lists needs to be manually copied. It also causes problems for client logging as history is split across multiple files.
+
+This specification introduces a method of natively renaming channels that does not suffer from these faults.
+
+## Architecture
+
+### Capabilities
+
+This specification introduces the `draft/rename` capability. This capability informs the server that the client is capable of handling `RENAME` messages.
+
+### Messages
+
+This specification adds the `RENAME` C2S and S2C message. This message informs a client that the name of a channel has been changed.
+
+The `RENAME` message has between two and three parameters. The first parameter is the current channel name, the second parameter is the new channel name, and the optional third parameter is a reason for renaming the channel.
+
+Server implementations SHOULD NOT allow channels to be converted between types with the `RENAME` message. Client implementations SHOULD be able to handle renaming between channel types.
+
+### Numerics
+
+| No. | Label               | Format                                         |
+| --- | --------------------| ---------------------------------------------- |
+| 692 | `ERR_CHANNAMEINUSE` | `<nick> <new-channel> :Channel already exists` |
+| 693 | `ERR_CANNOTRENAME`  | `<nick> <old-channel> <new-channel> :<reason>` |
+
+### Examples
+
+Renaming a channel with a reason:
+
+    C: RENAME #boaring #boring :Typo fix
+    S: :nick!user@host RENAME #boaring #boring :Typo fix
+
+Renaming a channel without a reason:
+
+    C: RENAME #coding #programming
+    S: :nick!user@host RENAME #coding #programming
+
+Failing to rename a channel because the user does not have the appropriate privileges (this example is non-normative):
+
+    C: RENAME #aniki #aneki
+    S: :irc.example.com 482 nick #aniki :You must be a channel operator
+
+Failing to rename a channel because the new channel name already exists:
+
+    C: RENAME #evil #good :Don't be evil
+    S: :irc.example.com 692 nick #good :Channel already exists
+
+Failing to rename a channel because you can not convert between channel types (this example is non-normative):
+
+    C: RENAME #global %local
+    S: :irc.example.com 693 nick #global %local :You cannot change a channel type
+
+Failing to rename a channel because it has been renamed recently:
+
+    C: RENAME #magical-girls #witches
+    S: :irc.example.com 693 nick #magical-girls #witches :This channel has been renamed recently
+
+## Implementation Considerations
+
+Server implementations MUST implement a fallback method for legacy clients. This method SHOULD involve sending the legacy client a `PART` message for the old channel name, with the rename reason as the part message if given, followed by the usual messages that would be sent if the legacy client had joined the new channel normally (`JOIN`, `RPL_TOPIC`, `RPL_NAMREPLY`, etc) and finally a `MODE` message to restore any prefix modes that the legacy client has applied to it.
+
+Server implementations SHOULD implement `JOIN` redirection from the old channel to the new channel for as long as is deemed necessary.
+
+Server implementations SHOULD implement a cooldown system to prevent abuse.
+
+## Security Considerations
+
+Server implementations that link across a network MUST take measures to prevent channel name collisions. An example of such a method would be to use channel identifiers similar to how user identifiers are used to prevent nickname collisions in server-to-server protocols.
+
+Server implementations SHOULD limit the renaming of channels to privileged individuals in order to prevent abuse.
--- a/extensions/rename.md
+++ b/extensions/rename.md
@@ -78,7 +78,7 @@ Failing to rename a channel because it has been renamed recently:
 
 ## Implementation Considerations
 
-Server implementations MUST implement a fallback method for legacy clients. This method SHOULD involve sending the legacy client a `PART` message for the old channel name, with the rename reason as the part message if given, followed by the usual messages that would be sent if the legacy client had joined the new channel normally (`JOIN`, `RPL_TOPIC`, `RPL_NAMREPLY`, etc) and finally a `MODE` message to restore any prefix modes that the legacy client has applied to it.
+Server implementations MUST implement a fallback method for legacy clients. This method MUST involve sending the legacy client a `PART` message for the old channel name, with the rename reason as the part message if given, followed by the usual messages that would be sent if the legacy client had joined the new channel normally (`JOIN`, `RPL_TOPIC`, `RPL_NAMREPLY`, etc) and finally a `MODE` message to restore any prefix modes that the legacy client has applied to it.
 
 Server implementations MAY implement `JOIN` redirection from the old channel to the new channel for as long as is deemed necessary.
 
--- a/extensions/rename.md
+++ b/extensions/rename.md
@@ -80,12 +80,12 @@ Failing to rename a channel because it has been renamed recently:
 
 Server implementations MUST implement a fallback method for legacy clients. This method SHOULD involve sending the legacy client a `PART` message for the old channel name, with the rename reason as the part message if given, followed by the usual messages that would be sent if the legacy client had joined the new channel normally (`JOIN`, `RPL_TOPIC`, `RPL_NAMREPLY`, etc) and finally a `MODE` message to restore any prefix modes that the legacy client has applied to it.
 
-Server implementations SHOULD implement `JOIN` redirection from the old channel to the new channel for as long as is deemed necessary.
+Server implementations MAY implement `JOIN` redirection from the old channel to the new channel for as long as is deemed necessary.
 
-Server implementations SHOULD implement a cooldown system to prevent abuse.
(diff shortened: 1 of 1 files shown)


SaberUK:
>Is there a reference implementation for inspircd?

There will be soon&trade;. I just need to finish up a few things.

Zarthus:
I don't see anything wrong, but I feel like `rename` is kind of generic of a keyword and won't really scale if you ever expand this idea. Let's say in the future you want to tack on the ability to `archive` (read-only) channels or close them down entirely. To be consistent with this format you'd need to reserve `ARCHIVE` and `CLOSE`, and /close already is an alias in many IRC clients.

I would probably do something like `CHGCHAN` (consistent with `CHGHOST`) or `CHANRENAME`, and I agree it is bikeshedding to some degree, but it's something that should at least be mentioned if you want to expand this idea to the degree Slack already has.

extensions/rename.md
```diff
@@ -0,0 +1,93 @@
...
+
+This specification introduces a method of natively renaming channels that
+does not suffer from these faults.
+
+## Architecture
+
+### Capabilities
+
+This specification introduces the `draft/rename` capability. This capability
+informs the server that the client is capable of handling `RENAME` messages.
```
Zarthus: The draft command would be `draft/RENAME`, no?

Zarthus:
Would it be possible to rename a channel from `#chan` to `!chan`? On some IRCds this means the channel serves a different function and clients may not wish to adhere to channel redirection (let's say `#chan` is a normal channel and `!chan` is opless mode which does not support founders). 

Awfully channel specific, but perhaps worth mentioning in a security or nonnormative section.

extensions/rename.md
SaberUK: I didn't know we were doing draft prefixes for commands. @jwheare?

Zarthus: https://github.com/ircv3/ircv3-specifications/pull/292/files#diff-8397424fc8fee1b5d347e97c8787b2fbR14 I can't find the comment where it is initially implemented/suggested

jwheare: No need, this verb has a cap to enable it, unlike chathistory.

RyanSquared:
> Would it be possible to rename a channel from #chan to !chan?

To quote @SaberUK from IRC:

> 2017-06-04 09:23:12 SaberUK Zarthus: i'm thinking it should be a SHOULD NOT for servers

DanielOaks:
If a more specific error like `ERR_CHANOPRIVSNEEDED` can be sent in response to a rename request, what should servers do?

Should they send both `ERR_CHANOPRIVSNEEDED` followed by `ERR_CANNOTRENAME`, and similar for `ERR_CHANNAMEINUSE`. It seems appealing for clients/bots to be able to expect *either* a `RENAME` command or a `ERR_CANNOTRENAME` numeric in response, to know for sure whether it failed or succeeded.

Plays into the questions brought up in IRC around how to alert for errors with specific commands in general going forward. I'm just using the existing `ERR_UNKNOWNERROR` in my hacked-up impl, which includes the original sent command in it.

extensions/rename.md
```diff
@@ -0,0 +1,107 @@
...
+
+    C: RENAME #magical-girls #witches
+    S: :irc.example.com [ERR_CANNOTRENAME] nick #magical-girls #witches :This channel has been renamed recently
+
+## Implementation Considerations
+
+Server implementations MUST implement a fallback method for legacy clients.
+This method SHOULD involve sending a fake `PART` message for the old channel
+name and a `JOIN` message for the new channel name as well as a `MODE` message
+to restore any prefix modes that the legacy client has applied to it.
```
DanielOaks: It seems like this should be specified in greater detail as a normative thing. Should the `PART` message be the reason specified by the sending user, and a note as to the user who sent the rename request's name? It seems like they should also send the `NAMES` message and `TOPIC` ones if they normally send those, or it could even say something like 'just send everything as though it were a regular channel join' perhaps?

Specifying this in more detail might help reduce any possible confusion around here, and make sure servers stay consistent with how they implement it.

RyanSquared: As with the `MODE` for sending prefix modes, what about modes that exist otherwise, such as prefix ones such as `+b` and `+f` (ban and forward) and all other possible channel modes?

DanielOaks: I'd probably just send everything you'd normally send on a client joining the channel. Name lists, channel modes and all. If you'd normally send it, cool. If not, I'd leave it out.

jwheare: Yeah, the standard channel join burst is definitely required here, NAMES, TOPIC, etc.

SaberUK:
>If a more specific error like ERR_CHANOPRIVSNEEDED can be sent in response to a rename request, what should servers do?

Yeah, thats what my implementation sends. I'll amend the spec in a bit.

extensions/rename.md
SaberUK: Fixed, will push shortly.

SaberUK:
I think that should be everything other than deciding what numerics are to be used.

extensions/rename.md
```diff
@@ -0,0 +1,114 @@
...
+### Messages
+
+This specification adds the `RENAME` C2S and S2C message. This message informs
+a client that the name of a channel has been changed.
+
+The `RENAME` message has between two and three parameters. The first parameter
+is the current channel name, the second parameter is the new channel name, and
+the optional third parameter is a reason for renaming the channel.
+
+Server implementations SHOULD NOT allow channels to be converted between types
```
kylef: What is the reason behind preventing changing these types?

I apologise if this is mentioned somewhere, would (freenode) `##` and `#` could as different types?

extensions/rename.md
```diff
@@ -0,0 +1,114 @@
...
+
+Failing to rename a channel because it has been renamed recently:
+
+    C: RENAME #magical-girls #witches
+    S: :irc.example.com [ERR_CANNOTRENAME] nick #magical-girls #witches :This channel has been renamed recently
+
+## Implementation Considerations
+
+Server implementations MUST implement a fallback method for legacy clients.
+This method SHOULD involve sending the legacy client a `PART` message for the
```
kylef: Should the PART reason should contain the RENAME reason?

extensions/rename.md
DanielOaks: Nah, those would be the same type. Converting between types would be like converting a normal channel (`#`) to a local channel (`&`), and vice-versa. Different channel types tend to act in completely different ways, so I can understand this restriction.

kylef: Would `#&` and `##` be the same type? I though this could depend on server implementation details.

RyanSquared: Yes. The significant character is the first character. `&whatever` for some implementations is a perfectly valid channel name.

RyanSquared: Actually, as an example - znc has this channel concept called "partyline", where channels can be prepended with `~` and it is a different "type" of channel than the one for the IRC network you're connected to. In this case, it definitely wouldn't work because users in the "partyline" channels might not actually be able to join the channels on the server your znc session is connected to.

grawity: Yes and no... Servers can of course be written to do all sorts of crazy things (like making channels global depending on a mode) – the sky is the limit. But as far as the _documented_ IRC protocol goes, only the first character is recognized as type distinguisher, and IMHO it's perfectly valid for a spec to assume that.

extensions/rename.md
SaberUK: Yeah, i'll add that.

extensions/rename.md
```diff
@@ -0,0 +1,91 @@
...
+Failing to rename a channel because it has been renamed recently:
+
+    C: RENAME #magical-girls #witches
+    S: :irc.example.com 693 nick #magical-girls #witches :This channel has been renamed recently
+
+## Implementation Considerations
+
+Server implementations MUST implement a fallback method for legacy clients. This method SHOULD involve sending the legacy client a `PART` message for the old channel name, with the rename reason as the part message if given, followed by the usual messages that would be sent if the legacy client had joined the new channel normally (`JOIN`, `RPL_TOPIC`, `RPL_NAMREPLY`, etc) and finally a `MODE` message to restore any prefix modes that the legacy client has applied to it.
+
+Server implementations SHOULD implement `JOIN` redirection from the old channel to the new channel for as long as is deemed necessary.
```
jwheare: JOIN redirection is probably a MAY, "as long as is deemed necessary" is a bit wooly for a SHOULD anyway.

extensions/rename.md
```diff
@@ -0,0 +1,91 @@
...
+    S: :irc.example.com 693 nick #global %local :You cannot change a channel type
+
+Failing to rename a channel because it has been renamed recently:
+
+    C: RENAME #magical-girls #witches
+    S: :irc.example.com 693 nick #magical-girls #witches :This channel has been renamed recently
+
+## Implementation Considerations
+
+Server implementations MUST implement a fallback method for legacy clients. This method SHOULD involve sending the legacy client a `PART` message for the old channel name, with the rename reason as the part message if given, followed by the usual messages that would be sent if the legacy client had joined the new channel normally (`JOIN`, `RPL_TOPIC`, `RPL_NAMREPLY`, etc) and finally a `MODE` message to restore any prefix modes that the legacy client has applied to it.
```
jwheare: Any reason this fallback is a SHOULD? How else might a server implement it?

extensions/rename.md
```diff
@@ -0,0 +1,91 @@
...
+
+Server implementations SHOULD implement `JOIN` redirection from the old channel to the new channel for as long as is deemed necessary.
+
+Server implementations SHOULD implement a cooldown system to prevent abuse.
```
jwheare: Probably can be MAY as well.

extensions/rename.md
```diff
@@ -0,0 +1,91 @@
...
+
+## Security Considerations
+
+Server implementations that link across a network MUST take measures to prevent channel name collisions. An example of such a method would be to use channel identifiers similar to how user identifiers are used to prevent nickname collisions in server-to-server protocols.
+
+Server implementations SHOULD limit the renaming of channels to privileged individuals in order to prevent abuse.
```
jwheare: Also MAY I'd say.

extensions/rename.md
jwheare: That's fair, but not really backwards compatible. I guess the server can make that call but seems user hostile when an alternative that works exists.

jesopo:
Implemented in BitBot: https://github.com/jesopo/bitbot/commit/6d742f6a

RyanSquared:
Implemented (minimally) in Moon Moon: https://github.com/wiseguiz/Moon-Moon/commit/504e354e3e2dbf6907da8baa67021f9730b30209

DanielOaks:
This is implemented in Oragono (has been since 0.8.1 - 2017-06), but there's been a faaair few spec updates since then, will update to match the latest state of the PR.

extensions/rename.md
jesopo: shouldn't it also contain the new channel name?

jwheare:
Implemented in IRCCloud too. Spec still needs some adjustment, my prior review requesting changes still stands fwict.

extensions/rename.md
```diff
@@ -0,0 +1,91 @@
...
+    S: :irc.example.com 693 nick #global %local :You cannot change a channel type
+
+Failing to rename a channel because it has been renamed recently:
+
+    C: RENAME #magical-girls #witches
+    S: :irc.example.com 693 nick #magical-girls #witches :This channel has been renamed recently
+
+## Implementation Considerations
+
+Server implementations MUST implement a fallback method for legacy clients. This method MUST involve sending the legacy client a `PART` message for the old channel name, with the rename reason as the part message if given, followed by the usual messages that would be sent if the legacy client had joined the new channel normally (`JOIN`, `RPL_TOPIC`, `RPL_NAMREPLY`, etc) and finally a `MODE` message to restore any prefix modes that the legacy client has applied to it.
```
jwheare: Prompted by https://github.com/oragono/oragono/issues/300#issuecomment-471464285

MODE isn't needed here. Prefix modes come from NAMES, and the channel mode is not normally included in a join burst, clients will request it on join.

ilbelkyr:
As mentioned on IRC, it's probably a good idea to explicitly mention the possibility of a `RENAME` message that only changes casing to ensure this is properly handled by clients.

I feel like the requirement to provide a `PART`/`JOIN` fallback might be relaxed in this situation since it may be more disruptive than silently starting to use a different casing in messages from the server, although that's certainly debatable.

RyanSquared:
> I feel like the requirement to provide a PART/JOIN fallback might be relaxed in this situation since it may be more disruptive than silently starting to use a different casing in messages from the server, although that's certainly debatable.

I think this would break clients that don't care about casing (such as my bot), whereas the PART/JOIN solution would most likely work on all clients.

SaberUK:
>I think this would break clients that don't care about casing (such as my bot)

Such software is already broken as there is no guarantee the server will not change the case of a channel name (e.g. InspIRCd has a changecap module in extras).

jwheare:
@SadieCat let me know if you want me to take this over to tidy it up? Not sure if it already allows edits from maintainers.

dequis:
It says "from unknown repository" so I think you gotta close this PR and open a new one with a new source repo.

jwheare:
Replaced with #420 (blaze it)

PR #309, opened 2017-06-09
message-tags-3.2: Clarify how to handle lone backslashes

DanielOaks:
How to handle this was a question raised to me in email. I think this is the most appropriate way to handle this sort of case (since here, there's no actual `"Character"` as specified by the table, so nothing to output when it's unescaped).

It's an edge case, but it should help make implementations just a bit more compatible knowing how they should handle this specifically.

### DIFF
--- a/core/message-tags-3.2.md
+++ b/core/message-tags-3.2.md
@@ -67,6 +67,9 @@ The mapping between characters in tag values and their representation in `<escap
 Reason: more common URL-escaping eats more space, while IRC message's length is very limited.
 Also having semicolon as `\:` makes it easy to split the `<tags>` string by `;` first.
 
+If a lone `\` exists at the end of an escaped value (with no escape character following it), then there
+SHOULD be no output character. For example, the escaped value `"test\"` should unescape to `"test"`.
+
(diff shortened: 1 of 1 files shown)


jwheare:
Looks good to me, I've updated my parser without issues.

core/message-tags-3.2.md
```diff
@@ -67,6 +67,9 @@ The mapping between characters in tag values and their representation in `<escap
 Reason: more common URL-escaping eats more space, while IRC message's length is very limited.
 Also having semicolon as `\:` makes it easy to split the `<tags>` string by `;` first.
 
+If a lone `\` exists at the end of an escaped value (with no escape character following it), then there
+SHOULD be no output character. For example, the escaped value `"test\"` should unescape to `"test"`.
```
DarthGandalf: remove `"`, backticks are enough.

DanielOaks: Whoo, done!

DanielOaks:
Cool, are we happy to merge this?

SaberUK:
LGTM

PR #312, opened 2017-06-29
Amend echo-message to allow better spam filtering.

SaberUK:
Currently it is not permissible to respond to a spam message with a fake echo-message response. This relaxes the MUST requirement to a SHOULD to allow for this.

Thanks to @Robby- for raising this issue with me.


jwheare:
Seems good to me. Might be worth adding a justification in the main body of the spec rather than just the errata.

DarthGandalf:
Maybe s/spam filtering/filtering spam/g? I'm not native English speaker,
but for some reason that sounds better to me. In any case, LGTM

2017-11-27 17:27 GMT+00:00 James Wheare <notifications@github.com>:

PR #314, opened 2017-07-28
message-tags-3.3: Make the limit on servers adding tag data easier to read

lp0:
This can be read as "messages that exceed 510 bytes", which may be
interpreted as a confusing reference to RFC1459 message limits.


DanielOaks:
Thanks, reads good to me

PR #315, opened 2017-07-28
message-tags-3.3: Servers must remove tags without the client-only prefix by default

lp0:
Unless a tag is specified otherwise, this is a security flaw.

### DIFF
--- a/core/message-tags-3.3.md
+++ b/core/message-tags-3.3.md
@@ -135,9 +135,16 @@ Servers MUST reply with the `ERR_INPUTTOOLONG` (`417`) error numeric if a client
 ## Security considerations
 
 Client-only tags should be treated as untrusted data. They can contain any value
-and are not validated by servers in any way.
+and are not validated by servers in any way. The server MAY unescape tag values
+after receiving data and escape those values before sending them out.
 
-Some specifications may involve servers accepting client-initiated tags without the client-only prefix. Such tag values SHOULD be validated by the server before being sent with any response to a client, unless they are specifically specified to be untrusted data.
+Some specifications may involve servers accepting client-initiated tags without
+the client-only prefix, they SHOULD define a validation process to be performed
+by the server.
+
+Tags without the client-only prefix that are not specified to be relayed by the
+server or that fail validation MUST be removed by the server before being sent
+with any response to another client.
 
 ## Client implementation considerations
 


core/message-tags-3.3.md
```diff
@@ -135,9 +135,14 @@ Servers MUST reply with the `ERR_INPUTTOOLONG` (`417`) error numeric if a client
 ## Security considerations
 
 Client-only tags should be treated as untrusted data. They can contain any value
-and are not validated by servers in any way.
+and are not validated by servers in any way. The server MAY normalise values by applying the unescape
+and escape process.
```
DanielOaks: Hmm, I'm a little confused by how/when the server would apply the escape/unescape process itself. It seems like allowing this would just let servers randomly break incoming tagged messages by incorrectly applying escaping where it shouldn't exist?

Is this intended so that servers can essentially fix cases where clients have accidentally left a tag value as unescaped where they meant to escape it? If so then this seems like a fragile fix that'll have to depend on heuristically working out when to apply it.

core/message-tags-3.3.md
```diff
@@ -135,9 +135,14 @@ Servers MUST reply with the `ERR_INPUTTOOLONG` (`417`) error numeric if a client
...
 
-Some specifications may involve servers accepting client-initiated tags without the client-only prefix. Such tag values SHOULD be validated by the server before being sent with any response to a client, unless they are specifically specified to be untrusted data.
+Some specifications may involve servers accepting client-initiated tags without the client-only prefix,
+they SHOULD define a validation process to be performed by the server.
+
+Tags without the client-only prefix or that fail validation MUST be removed by the server before being
+sent with any response to another client.
```
DanielOaks: I can see what's meant here, but I think you could more clearly write it as something like this:

```
Tags without the client-only prefix that are unexpected or otherwise fail validation MUST be removed by the server before being sent with any response to another client.
```

As currently written, it feels like it's trying to tell servers to remove all tags without the client-only prefix in all cases, where I think above is more what's meant to be conveyed?

core/message-tags-3.3.md
dequis: Also normalization implies there's more than one way to do escaping, but the escapes specified in the spec are so limited that there's not much room to do that, if any. You would need to consider invalid inputs. So:

Less escapes than needed break parsing always, there's no reasonable way to recover or detect it. Unescaped semicolon ends the value, unescaped space ends the tag, unescaped CR/LF end the message.

Unknown escapes are a more interesting edge case. Does `escape(unescape("\a"))` turn into `a`, `\a` or `\\a`? Out of scope of this PR btw, just saying that it probably deserves more than an offhand "server MAY clean the mess".

lp0: This is intended to allow servers to unescape values when receiving data and then escape them again when sending.

escape(unescape("\\")) becomes ""
escape(unescape("\\a")) becomes "a"

core/message-tags-3.3.md
lp0: This is the same ambiguous wording that was there before.

The specification does not state what "unexpected" is or what "validation" is.


> it feels like it's trying to tell servers to remove all tags without the client-only prefix in all cases

That is exactly what it should do. In the absence of another specification defining a server-only tag that should be relayed from one client to others, the tag is "unexpected" and should be removed. Until such a specification is written, the server should remove all server-only tags.

core/message-tags-3.3.md
DanielOaks: In this case I think it should explicitly state "unescape tag values when receiving data and escape those values before sending them out". 'normalise' isn't clear and raises the above questions, stating the exact process exactly when it can be applied should remove any ambiguity.

PR #316, opened 2017-07-28
message-tags-3.3: Allow empty TAGMSG explicitly, clarify server removal of unknown tags

nomis:
Return the NEEDMOREPARAMS numeric if all tags are removed.

### DIFF
--- a/core/message-tags-3.3.md
+++ b/core/message-tags-3.3.md
@@ -136,12 +136,12 @@ Client-only tags should be treated as untrusted data. They can contain any value
 and are not validated by servers in any way. The server MAY unescape tag values
 after receiving data and escape those values before sending them out.
 
-Some specifications may involve servers accepting client-initiated tags without
-the client-only prefix, they SHOULD define a validation process to be performed
-by the server.
-
 Tags without the client-only prefix MUST be removed by the server before being relayed
 with any message to another client.
+
+Some specifications may involve servers accepting client-initiated tags without
+the client-only prefix, they MUST define a process to be performed by the server
+on these tags prior to their removal.
 
 ## Client implementation considerations
 
--- a/core/message-tags-3.3.md
+++ b/core/message-tags-3.3.md
@@ -136,6 +136,10 @@ Client-only tags should be treated as untrusted data. They can contain any value
 and are not validated by servers in any way. The server MAY unescape tag values
 after receiving data and escape those values before sending them out.
 
+Some specifications may involve servers accepting client-initiated tags without
+the client-only prefix, they SHOULD define a validation process to be performed
+by the server.
+
 Tags without the client-only prefix MUST be removed by the server before being relayed
 with any message to another client.
 
--- a/core/message-tags-3.3.md
+++ b/core/message-tags-3.3.md
@@ -229,3 +229,11 @@ A `TAGMSG` sent by a client with tags that exceed the size limit and rejected by
 
     C: @+tag1;+tag2;+tag[...];+tag5000 TAGMSG #channel
     S: :server.example.com 417 nick :Input line was too long
+
+---
+
+A `TAGMSG` sent by a client with an un-prefixed tag that has no specified behaviour. The server removes the tag before relaying the message
+
+    C: @unknown-tag TAGMSG #channel
+    S: :nick!user@example.com TAGMSG #channel
+
--- a/core/message-tags-3.3.md
+++ b/core/message-tags-3.3.md
@@ -105,7 +105,8 @@ Servers MAY apply moderation to this command using existing or newly specified m
 
 Servers MUST NOT deliver `TAGMSG` to clients that haven't negotiated the message tags capability.
 
-Servers SHOULD reject any `TAGMSG` command sent without tags. In this case, they MUST use the `ERR_NEEDMOREPARAMS` (`461`) error numeric.
+Servers MUST use the `ERR_NEEDMOREPARAMS` (`461`) error numeric to reject any `TAGMSG`
+command received without any client-only tags.
 
 See [`PRIVMSG` in RFC2812](https://tools.ietf.org/html/rfc2812#section-3.3.1) for more details on replies and examples.
 
@@ -138,13 +139,8 @@ Client-only tags should be treated as untrusted data. They can contain any value
 and are not validated by servers in any way. The server MAY unescape tag values
 after receiving data and escape those values before sending them out.
 
-Some specifications may involve servers accepting client-initiated tags without
-the client-only prefix, they SHOULD define a validation process to be performed
-by the server.
-
-Tags without the client-only prefix that are not specified to be relayed by the
-server or that fail validation MUST be removed by the server before being sent
-with any response to another client.
+Tags without the client-only prefix MUST be removed by the server before being relayed
+with any message to another client.
 
 ## Client implementation considerations
 
--- a/core/message-tags-3.3.md
+++ b/core/message-tags-3.3.md
@@ -105,9 +105,6 @@ Servers MAY apply moderation to this command using existing or newly specified m
 
 Servers MUST NOT deliver `TAGMSG` to clients that haven't negotiated the message tags capability.
 
-Servers MUST use the `ERR_NEEDMOREPARAMS` (`461`) error numeric to reject any `TAGMSG`
-command received without any client-only tags.
-
 See [`PRIVMSG` in RFC2812](https://tools.ietf.org/html/rfc2812#section-3.3.1) for more details on replies and examples.
 
 Clients that receive a `TAGMSG` command MUST NOT display them in the message history by default. Display guidelines are defined in the specifications of tags attached to the message.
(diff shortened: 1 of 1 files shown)


core/message-tags-3.3.md
```diff
@@ -135,9 +137,14 @@ Servers MUST reply with the `ERR_INPUTTOOLONG` (`417`) error numeric if a client
 ## Security considerations
 
 Client-only tags should be treated as untrusted data. They can contain any value
-and are not validated by servers in any way.
+and are not validated by servers in any way. The server MAY normalise values by applying the unescape
```
DanielOaks: These changes are already a part of #315, would be good to rebase them out if they're not desired here

core/message-tags-3.3.md
```diff
@@ -105,7 +105,9 @@ Servers MAY apply moderation to this command using existing or newly specified m
 
 Servers MUST NOT deliver `TAGMSG` to clients that haven't negotiated the message tags capability.
 
-Servers SHOULD reject any `TAGMSG` command sent without tags. In this case, they MUST use the `ERR_NEEDMOREPARAMS` (`461`) error numeric.
+Servers SHOULD reject any `TAGMSG` command sent without tags, or where there are no tags with the client-only prefix
+and all tags without the client-only prefix are removed by the server.
```
DanielOaks: Hmm, I think we can simplify this down a bit to something like:

> Servers SHOULD reject any `TAGMSG` command sent without tags, or where all tags on the message end up being removed by the server.

Even if it's not as explicit, I think it's clearer than the current text (mainly because I need to go back and re-read the current sentence a few times before I understand what's actually meant by it, even though I already know).

core/message-tags-3.3.md
lp0: This is intentional, I can't submit a PR on behaviour (#315) that doesn't exist.

core/message-tags-3.3.md
lp0: Ok. Thinking about this further, it should really be a MUST instead of a SHOULD.

Otherwise we just deliver useless empty tag messages and the client is unaware.

core/message-tags-3.3.md
```diff
@@ -105,7 +105,8 @@ Servers MAY apply moderation to this command using existing or newly specified m
 
 Servers MUST NOT deliver `TAGMSG` to clients that haven't negotiated the message tags capability.
 
-Servers SHOULD reject any `TAGMSG` command sent without tags. In this case, they MUST use the `ERR_NEEDMOREPARAMS` (`461`) error numeric.
+Servers SHOULD reject any `TAGMSG` command sent without tags, or where all tags on the message end up
+being removed by the server. In this case, they MUST use the `ERR_NEEDMOREPARAMS` (`461`) error numeric.
```
jwheare: Reword the first sentence:
```
Servers SHOULD reject any `TAGMSG` command received without tags, or where all tags are being removed
according to the [Security considerations](#security-considerations) section.
```

core/message-tags-3.3.md
jwheare: Delivering empty TAGMSG isn't a problem for clients. The messages are inert without tags. That said I don't feel too strongly about whether this is SHOULD or MUST territory. I think I was erring on the side of allowing server implementations more leeway for something harmless, but I can't think of a good reason to allow it.

core/message-tags-3.3.md
lp0: I've amended it to say MUST for both. I don't think we should specify this as optional behaviour for servers. Do any servers not respond to a PRIVMSG without text?

jwheare: Some servers accept privmsg without text, some respond with an error.

jwheare: Yeah, MUST is fine. When this was originally written there wasn't anything in the spec that meant the servers had to inspect tag data for size limits etc so it was more lenient on servers that wanted to be lazy. That's not an option now.

jwheare:
As this whole part is now a MUST, this can be simplified:

```
Servers MUST use the `ERR_NEEDMOREPARAMS` (`461`) error numeric to reject any `TAGMSG`
command received without any tags, or where all tags are removed according to
the [Security considerations](#security-considerations) section.
```

---

With this and #315, is it now mandatory for non-prefixed, client-initiated tags to always have a CAP? Otherwise a client won't know whether or not the server might respond with an error numeric.

The only such tag that's been defined so far is labeled-response, and that *does* use a CAP, because the client needs to behave differently depending on server support.

But I wonder, is there a chance we'd want to specify an unprefixed tag that clients can blindly send at a server, which may or may not deal with it, that would normally only need the root `message-tags` CAP enabled?

In such a case could it be up to the client to handle (and ignore) the error numeric? Or would we have to require a CAP for that spec?

Arguably part of the motivation for the `message-tags` CAP was to avoid necessarily needing CAPs for all tags.

lp0:
> With this and #315, is it now mandatory for non-prefixed, client-initiated tags to always have a CAP?
> Otherwise a client won't know whether or not the server might respond with an error numeric.

Yes, but this is a good thing because things like `labeled-response` tags are not sent to other clients. They need a specification to define special behaviour that the server performs.

> But I wonder, is there a chance we'd want to specify an unprefixed tag that clients can blindly send at a server, which may or may not deal with it, that would normally only need the root message-tags CAP enabled?

If the tags MAY be modified by the server, then use the client prefix.

If the tags MUST be handled specially by the server, then define a new CAP and don't use the client prefix. If the server doesn't support it the client will usually want to do something different, for that CAP is a lot better than "try it and see if it works".

jwheare:
Yeah, I think this is OK. For a potential spec where you might want to fire and forget non-critical metadata to the server, you could define an ISUPPORT key, rather than requiring a CAP, and also even use a custom command rather than TAGMSG.

DarthGandalf:
I disagree with `ERR_NEEDMOREPARAMS` for the removed tags. `ERR_NOTEXTTOSEND` looks more appropriate.
Also it clearly tells spammer whether his message was delivered or not. Does it need to be MUST rather than SHOULD?

nomis:
It's not about final delivery, it's for a technical rejection of the message when there are no tags.

If the server wants to reject it for other reasons it's still free to do so.

`PRIVMSG` does the same thing if you supply no message.

DarthGandalf:
Does `PRIVMSG` send `ERR_NEEDMOREPARAMS` if client sent the message, but server discarded its text?

nomis:
No, but I wouldn't it expect it to because the message itself is valid.

DarthGandalf:
But it's the same here, isn't it?
What kind of an invalid message do you have in mind? Do you have some example?

nomis:
Valid (assume the server rejects ```@account``` from a client and removes it):
```
PRIVMSG #example :
PRIVMSG #example :test
@+tag1=value;+tag2=value TAGMSG #example
@+tag1=value;+tag2=value;account=NickServ TAGMSG #example
```

Invalid (no message; no tags; no tags that weren't rejected by the server):
```
PRIVMSG #example
TAGMSG #example
@account=NickServ TAGMSG #example
```

I don't see what this has to do with spam. If the server wants to ignore valid messages there is nothing stopping it from doing so. The specification shouldn't attempt to specify every possible area where a specific spam filtering strategy could be used (i.e. the server is free to do whatever it wants, but the specification states what a normal client expects to happen).

nomis:
I've reworded this again to simply state that server-only tags MUST be removed before relaying tags in a message, and that a TAGMSG with no client-only tags MUST return an error.

Thus, server-only tags have no effect on the relayed message (subject to other specifications, which I don't feel we need to state here because removal only applies to relaying) or on whether or not it is accepted.

This doesn't prohibit specifications accepting server-only tags and then adding their own server-only tags on the relayed message (which could be the same tags with the same values, but at that point it may as well be a client-only tag).

core/message-tags-3.3.md
```diff
@@ -138,13 +136,8 @@ Client-only tags should be treated as untrusted data. They can contain any value
...
 
-Some specifications may involve servers accepting client-initiated tags without
-the client-only prefix, they SHOULD define a validation process to be performed
-by the server.
-
-Tags without the client-only prefix that are not specified to be relayed by the
-server or that fail validation MUST be removed by the server before being sent
-with any response to another client.
+Tags without the client-only prefix MUST be removed by the server before being relayed
+with any message to another client.
```
jwheare: I'm OK with empty TAGMSG being allowed, but I don't see why this needs to remove the explicit mention that valid unprefixed tags are allowed to be sent c2s.

We also should add an example of the client sending unprefixed tags that get removed and the server forwarding an empty TAGMSG to another client.

jwheare: I'm OK with empty TAGMSG being allowed, but I don't see why this needs to remove the explicit mention that valid unprefixed tags are allowed to be sent c2s.

We also should add an example of the client sending unprefixed tags that get removed and the server forwarding an empty TAGMSG to another client.

jwheare: Now this sentence doesn't make sense, maybe this instead:

> All other tags without a client-only prefix MUST be removed by the server before the message is relayed to other clients.

nomis: I've altered it again. My opinion is that server-only tags are never relayed to other clients.

A future specification is of course entitled to define receiving server-only tags and then relaying exactly the same values (but these should probably be client tags).

jwheare: Well, @label already exists, it's not relayed to *other* clients, but it isn't removed before sending back in an echo.

PR #327, opened 2017-11-08
Merge CAP 3.1, CAP 3.2, and cap-notify specs

DanielOaks:
tl;dr the fact that we have three separate specifications for capability negotiation is bad. It's difficult and annoying to look in three different specs to implement one feature.

This PR merges the specifications, delimiting the 3.2-specific stuff with the phrase _"Cap Version `302`"_.

The only thing I'm sorta iffy on is how to properly note that if a client sends `CAP LS 302`, they are considered to support `302` going forward. Right now this is the language I'm using:

> Clients that send `302` as the CAP LS version are presumed to support `CAP LS 302` features for the future life of the connection.

However, it looks like servers do drop the 302 version if the client then later sends a `CAP LS` command (without the 302). If this is what servers are doing these days I can modify the text there to say if the client sends a regular CAP LS, presume they support 301 (which is everything not specified to be 302), or some other similar language.

**note:** After merging this I'll need to change website stuff to match, so don't just update the git submodule after pulling this in.

[Formatted spec](https://github.com/DanielOaks/ircv3-specifications/blob/master+merged-cap/core/capability-negotiation.md)


DanielOaks:
One of the other big reasons for the merge is to properly rip out stuff like cap modifiers. If you're a client author and you read through the CAP spec, it's really easy to assume _"hey I need to implement cap modifiers"_. With the merged spec, since it's deprecated I've properly torn out the section and replaced it with [this](https://github.com/DanielOaks/ircv3-specifications/blob/master+merged-cap/core/capability-negotiation.md#capability-modifiers).

jwheare:
cap-notify should be fully described in the merged document, and the cap-notify page removed or replaced with a redirect to an anchor that describes how to enable it with cap 3.1

DanielOaks:
sounds good, will do

DanielOaks:
@jwheare now has `cap-notify` merged in. I think this method of delimiting whether `NEW`/`DEL` should be sent is appropriate:

> The NEW subcommand MUST ONLY be sent to clients that have negotiated CAP Version `302` or enabled the `cap-notify` capability.

In addition, it's reiterated in the proper `cap-notify` section just below.

DanielOaks:
Also rewrote the intro and threw my name up the top. It's a little ambitious, sure, but... the current intro is a bit of a garbage fire. Fine for the time, but these days (especially with the new context we have and all), this intro lays out why to use cap negotiation, what it's good for, and simply how to do it in a much simpler way.

syzop:
> However, it looks like servers do drop the 302 version if the client then later sends a CAP LS command (without the 302). If this is what servers are doing these days I can modify the text....

At UnrealIRCd if a client sends `CAP LS 302` and at a later point uses `CAP LS` (without the 302) we still assume 302. I think that's a fair assumption so I agree with your text:
> Clients that send `302` as the CAP LS version are presumed to support `CAP LS 302` features for the future life of the connection.

..but I'm fine with whatever you guys decide.

SaberUK:
>At UnrealIRCd if a client sends CAP LS 302 and at a later point uses CAP LS (without the 302) we still assume 302.

We (InspIRCd) also do this.

grawity:
I'd vote for preserving the already-negotiated version, if only to make `/quote CAP LS` less harmful. (There are tutorials online suggesting this to find out whether a server supports a feature, so it's much more likely to be invoked by a user.)

DanielOaks:
Awesome, thanks a bunch for weighing in \o/

core/capability-negotiation.md
```diff
@@ -0,0 +1,442 @@
...
+    CAP <nick> DEL :<extension 1> [<extension 2> ... [<extension n>]]
+
+The last parameter is a space-separated list of capabilities that are no longer available.
+
+Example:
+
+    Server: :irc.example.com CAP modernclient DEL :userhost-in-names multi-prefix away-notify
+
+Example showing new SASL mechanisms becoming available. This example requires the client to
+support both `CAP LS 302` and `batch`:
```
jwheare: Let's remove or rework this example. The batch isn't really helpful. This use case can be achieved without it, or even by the server sending a 908 RPL_SASLMECHS numeric.

jwheare: Discussion on that batch not being needed is here #168

jwheare:
Yeah, i said "or rework" cos the example scenario of changing SASL mechs would be good to have *somewhere*, but it probably fits better on the SASL spec. Can add that back there without the batch when we get to it.

DanielOaks:
Coolcool, sounds like it'd fit better there yeah.

DanielOaks:
Just curious, is there anything else we're waiting for on this or just reviews?

jwheare:
Sorry, I still need to get around to fully reviewing this.

core/capability-negotiation.md
```diff
@@ -0,0 +1,437 @@
...
+    Server: CAP * ACK :multi-prefix sasl
+
+### The CAP ACK subcommand
+
+The ACK subcommand is sent by the server to acknowledge a client-sent REQ, and let the client
+know that their requested capabilities have been enabled.
+
+This subcommand has a single parameter consisting of a space-separated list of capability
+names. Each capability name may be prefixed with a dash (`-`), indicating that this
+capability has been disabled as requested.
```
jwheare: This removes the following bit from the current spec:

> If an ACK reply originating from the server is spread across multiple lines, a client MUST NOT change capabilities until the last ACK of the set is received. Equally, a server MUST NOT change the capabilities of the client until the last ACK of the set has been sent.

Was that deliberate? As described it doesn't really make sense to me or seem possible to implement, so fair enough to drop it, but it needs an errata to explain why. If you removed it by mistake, any reason we should keep it?

core/capability-negotiation.md
```diff
@@ -0,0 +1,437 @@
...
+capability negotiation protocol supported by the client.
+
+Clients that send `302` as the CAP LS version are presumed to support `CAP LS 302` features
+for the future life of the connection.
+
+Clients that do not send any version number with CAP LS are presumed to not support these
+extra features.
+
+If a client has not indicated support for `CAP LS 302` features, the server MUST NOT send
+these new features to the client such as cap values and multiline replies.
```
jwheare: I think a complete summary of 302 features might make sense here. As written you need to skim the rest of the document for references to 302.

core/capability-negotiation.md
```diff
@@ -0,0 +1,437 @@
...
+
+    Client: CAP LS 302
+    Server: CAP * LS :multi-prefix
+
+#### Capability Values
+
+If the client supports CAP Version `302`, the server MAY specify additional data for each
+capability using the `<name>=<value>` format in `CAP LS` and `CAP NEW` replies.
+
+Each capability, if it supports a value, defines what this value means in its' specification.
```
jwheare: > *its* specification

core/capability-negotiation.md
```diff
@@ -0,0 +1,437 @@
...
+Clients that support capabilities but do not wish to enter negotiation SHOULD send CAP END
+upon connection to the server.
+
+### The CAP NEW subcommand
+
+The NEW subcommand MUST ONLY be sent to clients that have negotiated CAP Version `302` or
+enabled the `cap-notify` capability.
+
+The NEW subcommand signals that the server supports one or more new capabilities, and may be
+sent at any time. Clients that support `CAP NEW` messages should respond with a `CAP REQ`
```
jwheare: SHOULD

core/capability-negotiation.md
```diff
@@ -0,0 +1,437 @@
...
+message if they wish to enable one of the newly-offered capabilities.
+
+The format of a `CAP NEW` message is:
+
+    CAP <nick> NEW :<extension 1> [<extension 2> ... [<extension n>]]
+
+As with `LS`, the last parameter is a space-separated list of new capabilities that are now
+offered by the server. If the client supports CAP Version `302`, the capabilities SHOULD be
+listed with values, as they would be if the client sent `CAP LS 302`.
```
jwheare: > listed with values, as in the `CAP LS` response.

core/capability-negotiation.md
```diff
@@ -0,0 +1,437 @@
...
+    Server: :irc.example.com CAP tester ACK :extended-join away-notify
+
+### The CAP DEL subcommand
+
+The NEW subcommand MUST ONLY be sent to clients that have negotiated CAP Version `302` or
+enabled the `cap-notify` capability.
+
+The DEL subcommand signals that the server no longer supports one or more capabilities that
+have been advertised. Upon receiving a `CAP DEL` message, the client MUST treat the listed
+capabilities cancelled and no longer available. Clients SHOULD NOT send `CAP REQ` messages
```
jwheare: > *as* cancelled and no longer...

core/capability-negotiation.md
```diff
@@ -0,0 +1,437 @@
...
+to cancel the capabilities in `CAP DEL`, as they have already been canceled by the server.
```
jwheare: > cancelled

Brit vs American spelling. We don't have a style guide for this, but should be consistent within one document at least.

core/capability-negotiation.md
```diff
@@ -0,0 +1,437 @@
...
+These names are prefixed by a valid DNS domain name.
+
+For example: `znc.in/server-time`.
+
+In cases where the prefix contains non-ASCII characters, punycode MUST be used,
+e.g. `xn--e1afmkfd.org/foo`.
+
+Vendor-Specific capabilities should be submitted to the IRCv3 working group for consideration.
+
+### Standardized
```
jwheare: > Standardised

UK/US again. Change for consistency with the line below.

core/capability-negotiation.md
```diff
@@ -0,0 +1,437 @@
...
+    Server: CAP * NAK :multi-prefix sasl ex3
+
+### The CAP END subcommand
+
+The END subcommand signals to the server that capability negotiation is complete and requests
+that the server continue with client registration. If the client is already registered, this
+command MUST be ignored by the server.
+
+Clients that support capabilities but do not wish to enter negotiation SHOULD send CAP END
+upon connection to the server.
```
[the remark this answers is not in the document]
jwheare: Also, you removed it from the recommendations for cap negotiation.

core/capability-negotiation.md
```diff
@@ -0,0 +1,437 @@
...
+
+Vendor-Specific capabilities should be submitted to the IRCv3 working group for consideration.
+
+### Standardized
+
+Standardised capability names have no vendor namespace, and are listed on the
+[IRCv3 Capability Registry](http://ircv3.net/registry.html#capabilities). These capabilities
+also have one or more documents in the set of IRCv3 specifications describing how they work.
+
+Names in the IRCv3 Capability Registry are reserved for your capability.
```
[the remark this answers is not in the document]
SaberUK: I think the idea was at one point, before vendor caps existed, to have it so people could reserve cap names?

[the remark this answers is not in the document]
DarthGandalf: I think the idea was (and still is) that people shouldn't use unprefixed capabilities without reserving the name in the Registry.

core/capability-negotiation.md
DanielOaks: Yep, removed for the reasons you've stated. I'll add an errata to explain why

core/capability-negotiation.md
DanielOaks: I've added a table listing the features just below. Once it gets in I'll see how the names get added on the site and link to them from this table.

core/capability-negotiation.md
DanielOaks: Errata added to explain why that advice has been removed.

DarthGandalf: I somewhat disagree with this, because of the following in 3.1 spec:

> The capability identifier set must be accepted as a whole, or rejected entirely.

Client knows which capabilities it requested in REQ, so it can see that some of them are missing in ACK. If some caps are missing, it should wait for the next ACK.

DanielOaks: I don't really expect clients to keep that meticulous of an eye on what caps they send and get back, and then compare the lists together before enabling them. I think that expecting clients to do so, while a nice idea, won't actually end up being done in reality, and so having that paragraph isn't useful. Thoughts?

jwheare: Hmm, well I guess the client is supposed to send END after ACK, so really it should probably wait for all REQed caps to be acked.

DarthGandalf: Regarding usefulness of this feature: I don't really see how it's useful, so this is a mere speculation:  maybe server needs to do something asynchronously to enable a certain cap? In that case server would need to wait until all caps are enabled before it can send ACK. Now, it can ACK caps when they are prepared to be enabled. And after that, enable them when the last cap is asynchronously prepared to be enabled... So server needs to track the state anyway, uhh.

Well, this sentence is one of the reasons why ZNC requests capabilities from server one by one, waiting for ACK every time. Another reason is because it wants to know which capabilities are not enableable.

I'm mostly disagree here with bundling of two different things together: merge specs together to make them more understandable, and changing actual things (errata). If we want some errata, let's unbundle that in a separate pull request.

jwheare: So ZNC doesn’t trust that the caps listed in LS can be REQed? Seems bad to need to do it one by one. Or do you REQ blindly without LS?

Some clarifications will involve errata, but I don’t think these involve material changes to to the spec that should be relied on by implementations.

DarthGandalf: No, it starts with LS. But it's possible that the capability was removed from server after it sent response to LS, in which case REQuesting all LSed caps will result in NAK. And at that point cap-notify is not enabled yet, because ZNC doesn't request CAP LS 302, as of now.

This change won't break ZNC because it requests them one by one, but it may break other implementations which do rely on this behavior.

DarthGandalf: Here's a contrived example of such breakage:
cap `example/double` causes all characters to be doubled, e.g. `PPRRIIVVMMSSGG`
```
> CAP REQ :example/double multi-prefix
< CAP * ACK :example/double
< CCAAPP  **  AACCKK  ::mmuullttii--pprreeffiixx
```
Because client was waiting for all caps to be ACKed before using them, it won't understand second CAP ACK, which server sent using the capability which it enabled already.

DanielOaks: I think my main issue with the original statement isn't the usefulness of it – it's clearly useful for the server to be able to send all the `ACK` messages, and then after those are sent enable all the capabilities contained within. I think my main issue is just how difficult it is for clients to try to discover which `ACK`s it should respond to now, and which ones it should wait to enable when another message comes through.

As a client, I enable caps when I receive an `ACK` for those caps. Storing a list of the REQ's I've made, comparing them to the incoming `ACK`s and `NAK`s, and then waiting to enable stuff based on that seems somewhat fragile and is something that as a client I'd just avoid altogether by doing something like the above instead.

DarthGandalf: Yep, I agree, and I chose to enable them one by one instead, which is not as fragile.

I'm just saying that if you're changing this part of spec, new servers won't be compatible with some old clients anymore, if those clients did implement this feature.

DanielOaks: Regarding the bundling, originally I didn't consider this a change, more just a clarification of what's actually done in the real world. But if there's contention I can break it out into a separate PR.

DanielOaks: Coolcool, I'll re-add this in and we'll discuss it more in a separate PR

DanielOaks:
Alrighty, so I think this is all good to go, or at least get a second pass of review.

jwheare:
Yeah sorry, I wanted to do a print out review pass.

core/capability-negotiation.md
```diff
@@ -0,0 +1,457 @@
...
+- `CAP REQ` to blindly request a particular set of capabilities.
+
+Following this, the client MUST send the standard `NICK` and `USER` IRC commands.
+
+Upon receiving either a `CAP LS` or `CAP REQ` command during connection registration, the
+server MUST not complete registration until the client sends a `CAP END` command to
+indicate that capability negotiation has ended. This allows clients to request their
+desired capabilities before completing registration.
+
+Once capability negotiation has completed with a client-send `CAP END` command, registration
```
jwheare: > client-sent

core/capability-negotiation.md
```diff
@@ -0,0 +1,457 @@
...
+Example with capability negotiation, but where the client recognises no advertised caps: 
+
+    Client: CAP LS 302
+    Client: NICK dan
+    Client: USER d * 0 :This is a really good name
+    Server: CAP * LS :draft/example-1 draft/example-2
+    Client: CAP END
+    Server: 001 dan :Welcome to the Internet Relay Network dan
+    ...
+
```
jwheare: Add a blind REQ example?

core/capability-negotiation.md
```diff
@@ -0,0 +1,457 @@
...
+    Client: USER d * 0 :This is a really good name
+    Server: 001 dan :Welcome to the Internet Relay Network dan
+    ...
+
+## The CAP command
+
+The client capability negotiation extension is implemented by the addition of one command
+with several subcommands.  The command added is named `CAP`.  `CAP` takes a single
+required subcommand, optionally followed by a single parameter of space-separated capability
+identifiers.  Each capability in the list may be preceded by a capability modifier.
```
jwheare: delete everything after:
> optionally followed by a single parameter...
to the end of the paragraph.

Maybe add:
> Each subcommand defines any further parameters.

core/capability-negotiation.md
```diff
@@ -0,0 +1,457 @@
...
+
+The subcommands for `CAP` are: `LS`, `LIST`, `REQ`, `ACK`, `NAK`, and `END`.
```
jwheare: Add NEW and DEL and make this a table indicating which are client sent and which are server sent, maybe a column indicating what they're sent in response to. e.g.

| Subcommand | Client sent | Server sent (triggered by) |
|------------|-------------|----------------------------|
| LS         | *           | * (LS)                     |
| LIST       | *           | * (LIST)                   |
| REQ        | *           |                            |
| ACK        |             | * (REQ)                    |
| NAK        |             | * (REQ)                    |
| END        | *           |                            |
| NEW (302)        |             | *                          |
| DEL (302)        |             | *                          |

core/capability-negotiation.md
```diff
@@ -0,0 +1,457 @@
...
+
+The `LS`, `LIST`, `REQ`, `ACK` and `NAK` subcommands may be followed by a single parameter
+containing a space-separated list of capabilities.  If more than one capability is named,
+the RFC1459 designated sentinel (`:`) for a multi-parameter argument must be present. The
```
jwheare: (Also NEW and DEL). Or just rewrite as:

> Some subcommands may include a space-separated list of capabilities as their final parameter. If more than one capability is listed, they MUST be preceded by the final parameter indicator (`:`) according to standard encoding rules.

core/capability-negotiation.md
```diff
@@ -0,0 +1,457 @@
...
+list of capabilities MUST be parsed from left to right and capabilities SHOULD only be sent
+once per command. If a capability is sent multiple times, the last one received takes priority.
```
jwheare: Maybe move this to the CAP values section, as it doesn't really have any impact on valueless caps.

core/capability-negotiation.md
```diff
@@ -0,0 +1,457 @@
...
+should be a human-readable description of the error.
+
+Replies from the server must contain the client identifier name or asterisk if one is not yet
+available.
+
+Example:
+
+    Client: CAP FOO
+    Server: :example.org 410 * FOO :Invalid CAP command
+
```
jwheare: Include example of a server response, with nick/asterisk.

core/capability-negotiation.md
```diff
@@ -0,0 +1,457 @@
...
+The client MUST be able to use the `CAP` command anytime, even after registration.
```
jwheare: This is a requirement for server implementors, so rewrite as:

> The server MUST accept the `CAP` command at any time, even after registration.

core/capability-negotiation.md
```diff
@@ -0,0 +1,457 @@
...
+
+### The CAP LS subcommand
+
+The LS subcommand is used to list the capabilities supported by the server.  The client
+should send an LS subcommand with no other arguments to solicit a list of all capabilities.
+
+If a client issues an LS subcommand, registration must be suspended until an END subcommand
```
jwheare: Reversed responsibility again:

> If a server receives an `LS` subcommand while client registration is in progress, it MUST suspend registration until an `END` subcommand is received.

core/capability-negotiation.md
```diff
@@ -0,0 +1,457 @@
...
+is received.  If no capabilities are available, an empty parameter must be sent.
```
jwheare: > ...MUST be sent

core/capability-negotiation.md
```diff
@@ -0,0 +1,457 @@
...
+
+Example:
+
+    Client: CAP LS
+    Server: CAP * LS :multi-prefix sasl
+
```
jwheare: Add an empty parameter example.

core/capability-negotiation.md
```diff
@@ -0,0 +1,457 @@
...
+Each capability, if it supports a value, defines what this value means in its specification.
+
+Example:
+
+    Client: CAP LS 302
+    Server: CAP * LS :multi-prefix sasl=PLAIN,EXTERNAL server-time draft/packing=EX1,EX2
+
+#### Multiline replies to `CAP LS` and `CAP LIST`
+
+Servers MAY send multiple lines in response to `CAP LS`. If the reply contains multiple
```
jwheare: > or `CAP LIST`

then drop paragraph below this one.

core/capability-negotiation.md
```diff
@@ -0,0 +1,457 @@
...
+
+If a multi-line response to `CAP LIST` is sent, and the client supports CAP Version `302`,
+the server MUST delimit all replies except for the last one sent as noted above.
+
+Example with a client that **does not** support CAP version `302`:
+
+    Client: CAP LS
+    Server: CAP * LS :multi-prefix extended-join account-notify batch invite-notify tls
+    Server: CAP * LS :cap-notify server-time example.org/dummy-cap=dummyvalue example.org/second-dummy-cap
+    Server: CAP * LS :userhost-in-names sasl=EXTERNAL,DH-AES,DH-BLOWFISH,ECDSA-NIST256P-CHALLENGE,PLAIN
```
jwheare: Non-302 response shouldn't contain values.

core/capability-negotiation.md
```diff
@@ -0,0 +1,457 @@
...
+
+Example of a `LIST` reply with a client that supports CAP version `302`:
+
+    Client: CAP LIST
+    Server: CAP modernclient LIST * :example.org/example-cap example.org/second-example-cap account-notify
+    Server: CAP modernclient LIST :invite-notify batch example.org/third-example-cap
+
+### The CAP LIST subcommand
+
+The LIST subcommand is used to list the capabilities associated with the active connection.
```
jwheare: > ...used to list the enabled capabilities...

core/capability-negotiation.md
```diff
@@ -0,0 +1,457 @@
...
+The client should send a LIST subcommand with no other arguments to solicit a list of
+active capabilities.
+
+If no capabilities are active, an empty parameter must be sent.
+
+Example:
+
+    Client: CAP LIST
+    Server: CAP * LIST :multi-prefix
+
```
jwheare: Add an empty parameter example.

core/capability-negotiation.md
```diff
@@ -0,0 +1,457 @@
...
+### The CAP REQ subcommand
+
+The REQ subcommand is used to request a change in capabilities associated with the active
+connection.  Its sole parameter must be a list of space-separated capability identifiers.
```
jwheare: The wording for the description of the sole parameter here should be unified between LS and LIST, REQ, ACK, NAK, NEW, DEL.

LS and LIST don't mention the server sent aspect at all, except by chance in the example. NEW and DEL describe the format as "last parameter" and provide an explicit example, REQ and ACK describe the "sole parameter", NAK describes the "argument". I'd favour NEW/DEL style.

core/capability-negotiation.md
```diff
@@ -0,0 +1,457 @@
...
+Each capability identifier may be prefixed with a dash (`-`) to designate that the capability
+should be disabled.
+
+If a client requests a capability which is already enabled, or tries to disable a capability
+which is not enabled, the server MUST continue processing the REQ subcommand as though
+handling this capability was successful.
+
+The capability identifier set must be accepted as a whole, or rejected entirely.
+
+If a client issues a REQ subcommand, registration must be suspended until an END subcommand
```
jwheare: Reversed responsibility again:

> If a server receives a `REQ` subcommand while client registration is in progress, it MUST suspend registration until an `END` subcommand is received.

core/capability-negotiation.md
```diff
@@ -0,0 +1,457 @@
...
+If an ACK reply originating from the server is spread across multiple lines, a client MUST NOT
+change capabilities until the last ACK of the set is received. Equally, a server MUST NOT
+change the capabilities of the client until the last ACK of the set has been sent.
+
+### The CAP NAK subcommand
+
+The NAK subcommand designates that the requested capability change was rejected.  The server
+MUST NOT make any change to any capabilities if it replies with a NAK subcommand.
+
+The argument of the NAK subcommand MUST consist of at least the first 100 characters of the
```
jwheare: Anyone remember why 100? I guess a client could send a really long (+512) list of REQs, but not sure why we need to define specific additional ytruncation rules here.

core/capability-negotiation.md
```diff
@@ -0,0 +1,457 @@
...
+
+Example with following `REQ`:
+
+    Server: :irc.example.com CAP tester NEW :away-notify extended-join
+    Client: CAP REQ :extended-join away-notify
+    Server: :irc.example.com CAP tester ACK :extended-join away-notify
+
+### The CAP DEL subcommand
+
+The NEW subcommand MUST ONLY be sent to clients that have negotiated CAP Version `302` or
```
jwheare: s/NEW/DEL/

core/capability-negotiation.md
```diff
@@ -0,0 +1,457 @@
...
+capabilities as cancelled and no longer available. Clients SHOULD NOT send `CAP REQ`
+messages to cancel the capabilities in `CAP DEL`, as they have already been cancelled by
+the server.
+
+Servers MUST cancel any capability-specific behavior for a client after sending the
+`CAP DEL` message to the client.
+
+Clients MUST gracefully handle situations when the server removes support for any
+capability. If the client cannot continue to operate without a capability, disconnecting
+with an appropriate `QUIT` message is acceptable.
```
jwheare: When might this happen or be appropriate? Seems a bit weird to be suggesting this, a client is *always* free to disconnect at any time.

core/capability-negotiation.md
```diff
@@ -0,0 +1,457 @@
...
+
+Example:
+
+    Server: :irc.example.com CAP modernclient DEL :userhost-in-names multi-prefix away-notify
+
+## `cap-notify`
+
+The `cap-notify` capability indicates support for the `NEW` and `DEL` messages listed above.
+This capabilitiy MUST be implicitly enabled if the client requests `CAP LS` with a version
+of 302 or newer (`CAP LS 302`), as described above.
```
jwheare: "above" is quite far away, maybe name and link to the specific section.

core/capability-negotiation.md
```diff
@@ -0,0 +1,457 @@
...
+
+Further, the `cap-notify` capability MAY NOT be disabled if the client requests `CAP LS`
```
jwheare: Active voice, make it clear who's responsible.
> Further, servers MUST NOT disable the `cap-notify` capability if the client...

core/capability-negotiation.md
```diff
@@ -0,0 +1,457 @@
...
+with a version of 302 or newer.
+
+When implicitly enabled via this mechanism, servers MAY list the `cap-notify` capability
+in `CAP LS` and `CAP LIST` responses. Additionally client MAY request the capability with
```
jwheare: > Additionally clients...

core/capability-negotiation.md
```diff
@@ -0,0 +1,457 @@
...
+`CAP REQ`, and capable servers MUST accept and `CAP ACK` the request without side effects.
+This lets clients blindly enable this capability, regardless of it being implicitly
+enabled by the server.
+
+Clients that do not support CAP Version 302 MAY request the `cap-notify` capability
+explicitly. Such clients MAY disable the capability at any time.  This to ease the
```
jwheare: > Server MUST allow clients to disable the capability at any time

core/capability-negotiation.md
```diff
@@ -0,0 +1,457 @@
...
+adaptation of new features.
+
+When enabled, the server MUST notify the client about all new capabilities and about
+existing capabilities that are no longer available using the `NEW` and `DEL` subcommands
+listed above.
+
+## Rules for naming capabilities
+
```
jwheare: Should caps be treated as case sensitive? Are there rules, conventions for naming them? e.g. hyphen separated words?

core/capability-negotiation.md
```diff
@@ -0,0 +1,457 @@
...
+The full capability name MUST be treated as an opaque identifier.
+
+There are two capability namespaces:
+
+### Vendor-Specific
```
jwheare: Use headings that make more sense on their own.
> Vendor-specific capabilities

core/capability-negotiation.md
```diff
@@ -0,0 +1,457 @@
...
+These names are prefixed by a valid DNS domain name.
+
+For example: `znc.in/server-time`.
+
+In cases where the prefix contains non-ASCII characters, punycode MUST be used,
+e.g. `xn--e1afmkfd.org/foo`.
+
+Vendor-Specific capabilities should be submitted to the IRCv3 working group for consideration.
+
+### Standardised
```
jwheare: Standalone
> Standardised capabilities

core/capability-negotiation.md
```diff
@@ -0,0 +1,457 @@
...
+
+These names are prefixed by a valid DNS domain name.
+
+For example: `znc.in/server-time`.
+
+In cases where the prefix contains non-ASCII characters, punycode MUST be used,
+e.g. `xn--e1afmkfd.org/foo`.
+
+Vendor-Specific capabilities should be submitted to the IRCv3 working group for consideration.
+
```
jwheare: We need to include the `draft/` namespace as reserved for standardisation and listed in the registry.

core/capability-negotiation.md
dequis: No idea why 100. It goes all the way back to the 2005 [mitchell ietf draft][1]. The 2002 and 2003 drafts don't mention anything like this.

[1]: https://tools.ietf.org/html/draft-mitchell-irc-capabilities-00#section-3.5

The changelog entry `2005-01-21 KLM` says

```
      *  Explicitly note in the ABNF for NAK that the parameter is an
         opaque repeat of at least the first 100 characters of the
         argument to REQ.
```

The original text:

```
   [...]
   subcommand.  The argument of the NAK subcommand MUST consist of at
   least the first one hundred characters of the capability list in the
   REQ (Section 3.3) subcommand which triggered the NAK.

   ABNF [4] description of the NAK subcommand:

          nakcmd       =  "NAK" SP ":" acklist
                           ; acklist is at least 100 characters of the
                           ; capability list from the REQ
```

dequis:
Gave this a quick read. Noticed a typo and used a spellchecker to find the rest: `recieve capabilitiy verisons`

Would be nice if the errata section had more explicit mentions of the old modifiers (sticky and ack-required). Maybe also mention that they were deprecated since cap 3.2 (or "2015"?). Relvant issue is PR #122, but doesn't really have any useful discussion.

What's the plan to redirect links to the old specs?

DanielOaks:
Nice, thanks! I'll fixup all the comments. The old spec URLs will be redirected automagically thanks to this section in the header:

```yaml
redirect_from:
  - /specs/core/capability-negotiation-3.1.html
  - /specs/core/capability-negotiation-3.2.html
  - /specs/extensions/cap-notify-3.2.html
```

As far as replacing old spec links in the site / other specs with links to this combined spec instead, I'll do that manually after this is merged in (don't wanna have this PR touch like a million specs all at once if I can help it). Also requires actual site changes as well before this update goes live, so I'll do those after this is merged as well.

core/capability-negotiation.md
DanielOaks: Note to self, look at adding the version as a separate col instead to make things clearer (and link the table header to the section)

core/capability-negotiation.md
```diff
@@ -0,0 +1,437 @@
...
+Previous versions of this spec defined 'capability modifiers'. These are not in use by software and
+have been removed from this specification.
+
+Previous versions of this spec did not specify when the `<name>=<value>` format could be used. This
+was clarified to limit capability values to the `LS` and `NEW` subcommands.
+
+Previous versions of this spec did not link to the `cap-notify` specification (nor note the fact
+that it is automatically enabled for clients that enable CAP Version 302). The specification now
+links to the `cap-notify` spec and lets authors know about the implicit enabling of this capability.
```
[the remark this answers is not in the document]
DanielOaks: Resolved, now includes

core/capability-negotiation.md
```diff
@@ -0,0 +1,437 @@
...
+The capability identifier set must be accepted as a whole, or rejected entirely.
+
+If a client issues a REQ subcommand, registration must be suspended until an END subcommand
+is received.
+
+Example:
+
+    Client: CAP REQ :multi-prefix sasl
+    Server: CAP * ACK :multi-prefix sasl
+
```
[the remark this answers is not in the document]
DanielOaks: Now includes an example of this

core/capability-negotiation.md
```diff
@@ -0,0 +1,437 @@
...
+### Capability Modifiers
+
+Prior versions of capability negotiation included 'capability modifiers', which were characters
+prepended to capability names to indicate that they could not be disabled or required extra client
+acknowledgement to be enabled.
+
+These modifiers have been deprecated. Capabilities MUST NOT require an `ACK` from the client to be
+activated. Instead of indicating a capability as 'sticky', servers should instead send a `NAK` to
+any request to disable these capabilities.
```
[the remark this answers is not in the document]
DanielOaks: Removed these and references to 'em. They weren't used in real software anyway.

core/capability-negotiation.md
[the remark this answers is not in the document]
DanielOaks: I've rewritten this line in a new commit

core/capability-negotiation.md
```diff
@@ -0,0 +1,437 @@
...
+    period: "2014"
+    email: "attilamolnar@hush.com"
+  -
+    name: "Daniel Oakley"
+    period: "2017"
+    email: "daniel@danieloaks.net"
+---
+## What Client Capability Negotiation attempts to solve
+
+The IRC protocol is old, and changing it is difficult because you need to convince other
```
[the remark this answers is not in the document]
DanielOaks: Yeah rewrote this line, thanks for the yell

core/capability-negotiation.md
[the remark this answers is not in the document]
DanielOaks: Removed this advice, yeah doesn't make sense

core/capability-negotiation.md
DanielOaks: Fix'd

core/capability-negotiation.md
DanielOaks: reoslved

core/capability-negotiation.md
DanielOaks: fix'd

core/capability-negotiation.md
DanielOaks: dun

core/capability-negotiation.md
DanielOaks: fixd, thanks for the new language

core/capability-negotiation.md
DanielOaks: dun

core/capability-negotiation.md
DanielOaks: fixd

core/capability-negotiation.md
DanielOaks: added a name, will link later when it's live (will get the section link id then)

core/capability-negotiation.md
DanielOaks: Roughly wrote this out, hopefully this works.

core/capability-negotiation.md
DanielOaks: Wrote a basic few sentences to list these conventions

core/capability-negotiation.md
DanielOaks: hmm, I'm not sure how this relates to valueless caps. in general it seems strange to say that if a cap is sent multiple times, the last one takes priority anyway. why is this recommended?

if I was the client getting a list of caps from the server, the last one would be the one that is applied, but if I was a server getting, say, a `REQ` from the client then I'd act on the first one in the list...

we should make this much more explicit, instead of just shifting it around. what is this sentence actually trying to say to vendors?

core/capability-negotiation.md
DanielOaks: Added an example with a nickname

core/capability-negotiation.md
DanielOaks: Yeah this seems weird, I can't think of anything that would trigger this but it feels like there could be some examples. Dunno if it's harmful at all, so I'm fine to keep it in or remove it. Seems to be here more for safety's sake than anything else, "hey clients if you ever do run into this weird case, disconnecting is appropriate"

core/capability-negotiation.md
DanielOaks: this feels weird. I guess it's explicitly mentioned for... reasons of making sure the client can match up its' REQs to NAKs, but... yeah it feels strange. I can go either way on it, unless we get some more info from someone that actually uses this explicitly?

core/capability-negotiation.md
DanielOaks: Unified the language here, should be more consistent now.

DanielOaks:
Alright, so this is now at the point where all review comments have been addressed. The only comments that do still exist are ones where there's a bit of a discussion to be had, etc.

DanielOaks:
And all the other review comments are now addressed, whoo.

jwheare:
Looks good for merge.

jwheare:
@DanielOaks if you're happy, feel free to do the honours.

DanielOaks:
Thanks much, all the back-and-forth and great reviews keep me honest ;)

Issue #19, opened 2013-10-25
Length of tags is unspecified


Elizafox:
Hello,

I have noticed a glaring flaw in the spec for IRCv3. It is obvious that this flaw is intentional and is supposed to be a feature, but I'm going to point out why it's a bad idea.

Namely, the flaw involves the length of tags being unspecified.

"BUT FIXED-LENGTH BUFFERS SUCK"

No, they prevent an attacker from clogging the server with a tag that could be of theoretically unbound length, not to mention simplify the handling of buffers and parsing.

I know that the authors of the spec seem to be from the ZNC team and they have fancy C++ string functions and all that, but C doesn't like unbounded buffers. It means parsing and buffering becomes harder than it has to.

Even ignoring the previous issue, unbounded tags are a terrible idea. The obvious workaround is "clip at a sane limit."

But what should this limit be? Implementation-defined is not acceptable - it needlessly shifts the burden onto client authors for no reason except for server developer laziness. @kaniini thinks, based on discussions with him on IRC, it should be bound at 512. This still keeps the 512 char limit for messages, but also reserves 512 chars for tags. I think this is a sensible solution and makes it relatively quick and easy to implement tags if you update your parse() and add a separate buffer for them.

Thoughts? Feelings? Rotten tomatoes?

PR #328, opened 2017-11-09
Define space-separated options parameter in WEBIRC

DanielOaks:
As discussed in https://github.com/ircv3/ircv3-ideas/issues/12 , this PR adds an optional space-separated parameter used for additional flags/options.

### DIFF
--- a/extensions/webirc.md
+++ b/extensions/webirc.md
@@ -21,20 +21,30 @@ When a user connects through an indirect connection to the IRC server, the user'
 ## Format
 The `WEBIRC` command MUST be the first command sent from the WebIRC gateway to the IRC server and MUST be sent before capability negotiation.
 
-The `WEBIRC` command takes four parameters: `password `, `gateway`, `hostname`, and `ip`.
+The `WEBIRC` command may take five parameters: `password `, `gateway`, `hostname`, `ip`, and `options`.
 - `password` Password authenticating the WebIRC gateway to the IRC server which must be agreed upon ahead of time
 - `gateway` WebIRC gateway service name
 - `hostname` User's resolved hostname
 - `ip` IPv4 or IPv6 address of the connecting user. If the IP address cannot be resolved to a DNS name, the IP address MUST be sent as both the `hostname` and `ip`.
+- `options` is an optional set of space-separated arguments detailing other information about the connection.
 
 If forward and reverse DNS do not match, the IP address SHOULD be sent as the `hostname` and the unverified `hostname` SHOULD NOT be sent (see [Security Considerations](#security-considerations)). If the connection fails to the IRC server, the WebIRC gateway MUST return an `ERROR` with an appropriate message and terminate the connection.
 
 Note: Previous implementations referred to the `gateway` field as `user`. This change is for documentation clarity only and maintains compatibility with all existing implementations.
 
+### Options
+The `options` parameter is a space-separated set of additional arguments, each argument having the form `<name>[=<value>]`.
+
+These options are defined and may be sent by clients while connecting:
+
+- `secure`: This flag indicates that the client has a TLS-secured connection to the gateway. Servers MUST ONLY treat the connection as secure if this flag is sent and the connection from the gateway to the server is also secure with TLS or similar.
+- `remote_port=<port>`: This flag indicates the remote port the client has connected to the gateway from.
+- `local_port=<port>`: This flag indicates the port the gateway accepted the client connection on (e.g. `6697`, `6667`).
+
 ### Examples
 Generic format.
 
-    WEBIRC password gateway hostname ip
+    WEBIRC password gateway hostname ip [:option1 option2...]
 
 IP address resolves to hostname.
 
@@ -43,6 +53,14 @@ IP address resolves to hostname.
 IP address does not resolve to hostname.
 
     WEBIRC hunter2 ExampleGateway 198.51.100.3 198.51.100.3
+
+Secure connection.
+
+    WEBIRC hunter2 ExampleGateway 198.51.100.3 198.51.100.3 secure
+
+Secure connection with ports passed through.
+
+    WEBIRC hunter2 ExampleGateway 198.51.100.3 198.51.100.3 :secure local_port=6697 remote_port=21726
 
 Error from invalid password.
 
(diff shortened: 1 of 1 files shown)


extensions/webirc.md
```diff
@@ -21,20 +21,30 @@ When a user connects through an indirect connection to the IRC server, the user'
 ## Format
 The `WEBIRC` command MUST be the first command sent from the WebIRC gateway to the IRC server and MUST be sent before capability negotiation.
 
-The `WEBIRC` command takes four parameters: `password `, `gateway`, `hostname`, and `ip`.
+The `WEBIRC` command may take five parameters: `password `, `gateway`, `hostname`, `ip`, and `options`.
```
SaberUK: "takes between four and five parameters" sounds more specific to me.

DanielOaks: yeah this was the only thing I was iffy on, will fix it

extensions/webirc.md
```diff
@@ -21,20 +21,30 @@ When a user connects through an indirect connection to the IRC server, the user'
...
 
+### Options
+The `options` parameter is a space-separated set of additional arguments, each argument having the form `<name>[=<value>]`.
+
+These options are defined and may be sent by clients while connecting:
+
+- `secure`: This flag indicates that the client has a TLS-secured connection to the gateway. Servers MUST ONLY treat the connection as secure if this flag is sent and the connection from the gateway to the server is also secure with TLS or similar.
+- `remote_port=<port>`: This flag indicates the remote port the client has connected to the gateway from.
+- `local_port=<port>`: This flag indicates the port the gateway accepted the client connection on (e.g. `6697`, `6667`).
+
```
SaberUK: Some minor bikeshedding:

- My WIP implementation of this uses "tls" rather than "secure" as "secure" is fairly ambiguous term. Thoughts?
- If we were to use - rather than _ (i.e. local-port) then it would be more consistent with tags.

DanielOaks: yeah my impl uses `tls` as well. I could see it either way, will see if anyone else wants it to be `tls` rather than `secure`. I do like `-` better than `_` here, the name just flows more nicely imo, will change that over

DanielOaks:
q: do people prefer the option name `tls` or `secure` to indicate tls-secured connections from the client to the gateway?

SaberUK:
Also re: tls, I was wondering if having the TLS version and cipher suite name in the value might be useful. It would allow the server to potentially block WEBIRC clients that are connected using a weak TLS version and/or cipher suite.

DanielOaks:
Eh, could be useful but I think more than anything it just complicates things at this early-ish stage. I'd suggest leaving as-is and later on if gateways and servers really want it we can add it in.

jwheare:
Worth adding a note to make sure that servers can handle valueless options gaining a value with future extensions.

DanielOaks:
@jwheare Thanks, added in that commit

extensions/webirc.md
```diff
@@ -21,20 +21,32 @@ When a user connects through an indirect connection to the IRC server, the user'
 ## Format
 The `WEBIRC` command MUST be the first command sent from the WebIRC gateway to the IRC server and MUST be sent before capability negotiation.
 
-The `WEBIRC` command takes four parameters: `password `, `gateway`, `hostname`, and `ip`.
+The `WEBIRC` command takes four or five parameters: `password `, `gateway`, `hostname`, `ip`, and `options`.
```
Zarthus: is it still a parameter if there's a `:` per the described synax:
````
WEBIRC password gateway hostname ip [:option1 option2...]
````

File: extensions/webirc.md [hunk 4]
```diff
@@ -44,6 +56,14 @@ IP address does not resolve to hostname.
 
     WEBIRC hunter2 ExampleGateway 198.51.100.3 198.51.100.3
 
+Secure connection.
+
+    WEBIRC hunter2 ExampleGateway 198.51.100.3 198.51.100.3 secure
+
+Secure connection with ports passed through.
+
+    WEBIRC hunter2 ExampleGateway 198.51.100.3 198.51.100.3 :secure=examplevalue local-port=6697 remote-port=21726
```
Zarthus: You didn't say secure could take a value

extensions/webirc.md
DanielOaks: yep, it's a trailing parameter

extensions/webirc.md
DanielOaks: [This line](https://github.com/ircv3/ircv3-specifications/pull/328/files#diff-407b2883275563cf9468d4506dae17ceR44) indicates that servers MUST be able to handle keys that don't have values right now gaining values in the future. This example explicitly includes a value to make sure servers don't trip over and can comply with that statement.

DanielOaks:
Placing this on hold until we have an encoding decided for these keys and their values.

syzop:
I'm a bit confused because in https://github.com/ircv3/ircv3-ideas/issues/12#issuecomment-343267337 you said:
> We can continue churning and working on #19,  in the meantime ircv3/ircv3-specifications#328 can be worked on and merged, and then when we've got an appropriate escaping method we can put that into the WebIRC spec as well.

I agree with that. Yet, this pull request is closed with the exact opposite reasoning (?)

DanielOaks:
Lots of back-and-forth and yelling in the channel. For what it's worth, I still agree 100% with the advice given in this PR and think it can be done no problem without having an encoding decided, but yeah I won't re-introduce it formally until the above is done.

DanielOaks:
Note: `local-port` and `remote-port` are ones that were thrown around ages back as ideas iirc. `secure` is actively used in Unreal today, and what was sorta decided on already so cool.

jwheare:
Anything needed to merge this?

DanielOaks:
fwiw I don't have anything else to add, looks like no opinions/comments on the escaping but if implementers are happy with it and how it directly tries to share the existing method, cool.

SaberUK:
:+1:

This matches what UnrealIRCd implements and what InspIRCd implements in Git.

extensions/webirc.md
prawnsalad: Just for completeness, all implementations that I know of (inspircd, unreal, webircdgateway) are using 'secure' rather than 'tls'.

DanielOaks: ye our attempt was all in vain, at least the spec still says `secure`~

prawnsalad:
👍
All good from the webirc gateway point of view

DanielOaks:
This is getting merge. If you want a say, speak up now or forever hold your peace

PR #330, opened 2017-11-13
Add `migrate` extension

aji:
I'm working on an implementation over at https://github.com/aji/charybdis-migrate (No code yet. The code is all on my hard disk and is half-written, but my approach is outlined in the document.)

I think `migrate` addresses a big gap in the IRC client-server protocol.

`migrate` is designed to be dead simple for clients to implement, at the expense of more complex server-side implementations, especially for s2s protocols like TS6 that have no central authority to coordinate something like this. The idea is that if it's easy for a client to implement, then more clients will implement it and server authors won't feel like their effort is wasted by implementing it.

I want to hear people's thoughts so I don't waste my own time putting together a proof-of-concept in charybdis-migrate.

### DIFF
--- a/extensions/migrate.md
+++ b/extensions/migrate.md
@@ -1,0 +1,321 @@
+# IRCv3 `migrate` Extension
+
+*Copyright &copy; 2016 Alex Iadicicco <https://github.com/aji>*
+
+*Copyright &copy; 2016 Justin Kaufman <https://github.com/akaritakai>*
+
+*Unlimited redistribution and modification of this document is allowed provided
+that the above copyright notice and this permission notice remains intact*
+
+This document describes the `migrate` IRCv3 client capability and associated
+protocol, which allows servers to migrate client sessions between servers, with
+appropriate client cooperation.
+
+Clients that support the migration protocol outlined in this document should
+register the `migrate` client capability. If the server supports client
+migrations (even if the network is not in a configuration to support
+migrations), it should accept the capability.
+
+# Protocol overview
+
+From the client's perspective, a migration works as follows. The details of this
+protocol are described later in this section.
+
+1. The server sends `MIGRATE START` to indicate that it wants to migrate the
+   client to a different server. Included in this message are a *resume token*
+   and *confirm token* that will be used when connecting to the new server.
+
+2. The client confirms the migration request with `MIGRATE OK`. After sending
+   `MIGRATE OK` the client should stop sending normal IRC traffic, as it will
+   not be processed. If for some reason the client does not wish to migrate,
+   then the migration fails. After receiving `MIGRATE OK`, the server can begin
+   coordinating the migration with the new server.
+
+3. After the server has finished coordinating the migration with the new server,
+   it sends `MIGRATE PROCEED` and closes the connection. If the connection is
+   closed before receiving `MIGRATE PROCEED`, the migration fails.
+
+4. On connecting to the new server, instead of using the normal connection setup
+   protocol (`NICK`, `CAP`, etc.) the client sends a `MIGRATE RESUME` message
+   containing this migration's resume token. Refer to later documentation for
+   the exact semantics of this message.
+
+5. The new server sends `MIGRATE CONFIRM`, containing this migration's confirm
+   token. If the confirm token does not match, the migration fails.
+
+6. The client sends `MIGRATE END`, and can start using the connection as it
+   would have before sending `MIGRATE OK`.
+
+Roughly, this looks as follows:
+
+    s1 <- PRIVMSG #chan :message 1
+    s1 -> :s1 MIGRATE START s2 +6697 abc def :migrate now
+    s1 <- PRIVMSG #chan :message 2
+    s1 <- MIGRATE OK abc
+    s1 -> :s1 MIGRATE PROCEED abc
+      (client disconnects from s1, connects to s2:6697 using TLS)
+    s2 <- MIGRATE RESUME abc
+    s2 -> :s2 MIGRATE CONFIRM def
+    s2 <- MIGRATE END
+    s2 <- PRIVMSG #chan :message 3
+
+Other users in `#chan` will see messages 1 2 and 3 as if the client's connection
+was left untouched.
+
+## Client-initiated migrations
+
+The previous example describes a migration initiated by a server, but clients
+can request migrations as well using the `MIGRATE REQUEST` message. This message
+contains connection details for a server the client wishes to migrate to. If the
+server is willing to perform the migration, then it starts with a
+`MIGRATE START` message using the normal semantics of a server-initiated
+migration.
+
+## Failures
+
+Migrations can fail for a number of reasons:
+
+* The client does not wish to migrate.
+
+* The client does not send `MIGRATE OK` in a timely fashion.
+
+* The server does not send `MIGRATE PROCEED` in a timely fashion.
+
+* The client can't connect to the new server.
+
+* The new server does not understand the `MIGRATE` message.
+
+* The new server does not recognize the client's resume token.
+
+* The new server presents an incorrect confirm token.
+
+If at any point the client is not sure how to proceed, the client should close
+the connection immediately and discard all internal state, in the same manner as
+an `ERROR` message or other server-initiated disconnection.
+
+## Timeouts
+
+Migrations should have an associated timeout. We recommend a timeout similar to
+a ping timeout, e.g. a few minutes. This way, if the client fails the migration,
+the effect is similar to a connection lost for networking reasons.
+
+## Server-to-server considerations
+
+The following protocol is suggested for server-to-server protocols that are
+based on a spanning tree. This protocol is simply advisory and is not part of
+the `migrate` specification:
+
+1. The old server (**OLD**) initiates a migration to a new server (**NEW**) by
+   sending `MIGRATE START` to the client. (Servers may choose to perform a
+   commitment-free negotiation of a possible migration before this step, to
+   reduce the chance of the migration failing.)
+
+2. The client accepts the migration with `MIGRATE OK`. In a single message,
+   **OLD** broadcasts a "flip" message that indicates **NEW** is now responsible
+   for the client. Servers receiving this message should immediately start
+   routing messages for the migrated user to **NEW**.
+
+  * The "flip" message will propagate outward from **OLD**, moving through the
+    tree toward **NEW**. If a server along the path between **OLD** and **NEW**
+    (including **NEW**) has outdated information and incorrectly routes a
+    message toward **OLD**, the message will eventually reach a server **X**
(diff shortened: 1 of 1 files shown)


extensions/migrate.md
```diff
@@ -0,0 +1,321 @@
...
+migrations (even if the network is not in a configuration to support
+migrations), it should accept the capability.
+
+# Protocol overview
+
+From the client's perspective, a migration works as follows. The details of this
+protocol are described later in this section.
+
+1. The server sends `MIGRATE START` to indicate that it wants to migrate the
+   client to a different server. Included in this message are a *resume token*
```
DarthGandalf: and address of the new server

extensions/migrate.md
```diff
@@ -0,0 +1,321 @@
...
+      (client disconnects from s1, connects to s2:6697 using TLS)
+    s2 <- MIGRATE RESUME abc
+    s2 -> :s2 MIGRATE CONFIRM def
+    s2 <- MIGRATE END
+    s2 <- PRIVMSG #chan :message 3
+
+Other users in `#chan` will see messages 1 2 and 3 as if the client's connection
+was left untouched.
+
+## Client-initiated migrations
```
DarthGandalf: The previous text should have a title Server-initiated migrations

extensions/migrate.md
```diff
@@ -0,0 +1,321 @@
...
+   would have before sending `MIGRATE OK`.
+
+Roughly, this looks as follows:
+
+    s1 <- PRIVMSG #chan :message 1
+    s1 -> :s1 MIGRATE START s2 +6697 abc def :migrate now
+    s1 <- PRIVMSG #chan :message 2
+    s1 <- MIGRATE OK abc
+    s1 -> :s1 MIGRATE PROCEED abc
+      (client disconnects from s1, connects to s2:6697 using TLS)
```
DarthGandalf: This doesn't allow client to connect to s2 first, and then disconnect from s1, which will result in some downtime. I don't know how important that is.

extensions/migrate.md
```diff
@@ -0,0 +1,321 @@
...
+
+Clients can send `MIGRATE REQUEST` to suggest a migration.
+
+    MIGRATE REQUEST cactus.desert :Free pizza
+
+The parameters are
+
+1. `REQUEST`, indicating the `MIGRATE REQUEST` operation
+
+2. The network-internal name of the server the client wishes to be migrated to.
```
DarthGandalf: Where to find such name?

extensions/migrate.md
```diff
@@ -0,0 +1,321 @@
...
+
+    :saguaro.desert MIGRATE START cactus.desert-network.com +6697 5AEAD4D0
+    9DD9F7F5 :Server going down for maintenance
+
+The parameters are
+
+1. `START`, identifying the command
+
+2. The client-reachable address of the server that the client will eventually be
+   migrated to. This can be either a domain name or an IP address, but should be
```
DarthGandalf: IP address in text format, I assume, not DCC format?

extensions/migrate.md
```diff
@@ -0,0 +1,321 @@
...
+
+*Unlimited redistribution and modification of this document is allowed provided
+that the above copyright notice and this permission notice remains intact*
+
+This document describes the `migrate` IRCv3 client capability and associated
+protocol, which allows servers to migrate client sessions between servers, with
+appropriate client cooperation.
+
+Clients that support the migration protocol outlined in this document should
+register the `migrate` client capability. If the server supports client
```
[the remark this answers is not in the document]
aji: I can be more precise here.

extensions/migrate.md
aji: Ah, I hadn't intended for this line to list everything in the message. I'll add a bit at the end that mentions the other fields.

extensions/migrate.md
aji: Hm, I'll have to think about the implications of that one. It should be fine but I want to give it some thought first.

extensions/migrate.md
aji: Well, "client-initiated migrations" are basically clients requesting that the server begin a server-initiated migration, and they're intended to be issued manually. I'll rephrase the title of this section to something like "Manually requesting a migration", or possibly omit it entirely. (Server authors will add appropriate additional commands for users and operators to trigger migrations so I'm not sure there's much value in specifying those commands.)

extensions/migrate.md
aji: This is the name in /map, /links, /whois, etc. The spec tries to distinguish between those names and DNS names, because there's not always a 1-1 mapping.

extensions/migrate.md
aji: Yes, text format. I'll clarify.

extensions/migrate.md
```diff
@@ -0,0 +1,321 @@
...
+protocol are described later in this section.
+
+1. The server sends `MIGRATE START` to indicate that it wants to migrate the
+   client to a different server. Included in this message are a *resume token*
+   and *confirm token* that will be used when connecting to the new server.
+
+2. The client confirms the migration request with `MIGRATE OK`. After sending
+   `MIGRATE OK` the client should stop sending normal IRC traffic, as it will
+   not be processed. If for some reason the client does not wish to migrate,
+   then the migration fails. After receiving `MIGRATE OK`, the server can begin
```
ProgVal: Should it reply explicitely it does not want to migrate? (eg. `MIGRATE NAK`)

extensions/migrate.md
```diff
@@ -0,0 +1,321 @@
...
+# Protocol overview
+
+From the client's perspective, a migration works as follows. The details of this
+protocol are described later in this section.
+
+1. The server sends `MIGRATE START` to indicate that it wants to migrate the
+   client to a different server. Included in this message are a *resume token*
+   and *confirm token* that will be used when connecting to the new server.
+
+2. The client confirms the migration request with `MIGRATE OK`. After sending
```
ProgVal: What about `MIGRATE ACK`, to use the same vocabulary as the cap negotiation specs?

extensions/migrate.md
```diff
@@ -0,0 +1,321 @@
...
+migrations), it should accept the capability.
+
+# Protocol overview
+
+From the client's perspective, a migration works as follows. The details of this
+protocol are described later in this section.
+
+1. The server sends `MIGRATE START` to indicate that it wants to migrate the
+   client to a different server. Included in this message are a *resume token*
+   and *confirm token* that will be used when connecting to the new server.
```
ProgVal: You should specify the format of the message with these two tokens. I think it is rather clear what you mean, but since you are writing a specification, it is better to be explicit.

extensions/migrate.md
ProgVal: and port

extensions/migrate.md
```diff
@@ -0,0 +1,321 @@
...
+   it sends `MIGRATE PROCEED` and closes the connection. If the connection is
+   closed before receiving `MIGRATE PROCEED`, the migration fails.
+
+4. On connecting to the new server, instead of using the normal connection setup
+   protocol (`NICK`, `CAP`, etc.) the client sends a `MIGRATE RESUME` message
+   containing this migration's resume token. Refer to later documentation for
+   the exact semantics of this message.
+
+5. The new server sends `MIGRATE CONFIRM`, containing this migration's confirm
+   token. If the confirm token does not match, the migration fails.
```
ProgVal: Could you state explicitely what "the migration fails" implies at this point?

extensions/migrate.md
```diff
@@ -0,0 +1,321 @@
...
+The parameters are
+
+1. `START`, identifying the command
+
+2. The client-reachable address of the server that the client will eventually be
+   migrated to. This can be either a domain name or an IP address, but should be
+   reachable by the client.
+
+3. The port that the client should use when connecting to the new server. An
+   ASCII plus character (`+`) before the port indicates that TLS should be used.
```
ProgVal: What happens if the old server uses STS, but does not use a `+` here?

extensions/migrate.md
```diff
@@ -0,0 +1,321 @@
...
+This message should be sent as the first message when opening a connection with
+the new server, in place of normal registration messages like `NICK` and `CAP`.
+
+To ensure that the new server is prepared to accept the migration, clients
+should also send normal registration commands after `MIGRATE RESUME`:
+
+    MIGRATE RESUME 5AEAD4D0
+    CAP LS
+    NICK aji
+    USER alex * * :Alex
```
ProgVal: This contradicts what you wrote above: 

> 4. On connecting to the new server, instead of using the normal connection setup
>   protocol (`NICK`, `CAP`, etc.) the client sends a `MIGRATE RESUME` message

extensions/migrate.md
```diff
@@ -0,0 +1,321 @@
...
+   coordinating the migration with the new server.
+
+3. After the server has finished coordinating the migration with the new server,
+   it sends `MIGRATE PROCEED` and closes the connection. If the connection is
+   closed before receiving `MIGRATE PROCEED`, the migration fails.
+
+4. On connecting to the new server, instead of using the normal connection setup
+   protocol (`NICK`, `CAP`, etc.) the client sends a `MIGRATE RESUME` message
+   containing this migration's resume token. Refer to later documentation for
+   the exact semantics of this message.
```
ProgVal: Wouldn't it be required to do another capability negotiation, in case the new server does not support the same set of capabilities?

ProgVal:
This is a great idea. As a client author, I really like how easy it will be for us to implement it.
Client-initiated is also a nice addition that could be useful in specific contexts.

extensions/migrate.md
ProgVal: and port

extensions/migrate.md
RyanSquared: I'm not sure how it would "fail" otherwise, other than that it just wouldn't happen and the server would keep waiting on it.

extensions/migrate.md
RyanSquared: Additionally, could this instead use the "resume" spec (currently draft)? That might also make it easier on both clients implementing this protocol, and servers implementing "resume" if this is already implemented.

jesopo:
Huge +1 for this spec - seamlessly being able to push people off a leaf so it can be rebooted is a very cool idea.

That said, I really think this should be changed to leverage `resume` (#306) instead of having 2 systems that use `"resume tokens"`

k4bek4be:
This discussion looks dead for now. What about standarizing JUMPSERVER instead (or additionally)? It uses 010 `RPL_REDIR` as `:irc.some.net 010 nick otherserver.some.net 6697 :Please use this Server/Port instead` and looks like some clients and servers already support it. Won't be so seamless but easy on both sides. Only somehow make it aware about different kinds of connections (TLS, websocket etc).

PR #335, opened 2018-02-12
message-tags: Clarify how to handle invalid escapes

DanielOaks:
Pretty simple change. Just clarifies how to handle invalid escapes. From the looks of it, a fair bit of software already assumes this case is treated in this way, and it's the assumption I've been working under for a long while now.


jwheare:
This will need rebasing onto the new merged/unversioned version.

jwheare:
Rebased onto the new version in #373

Issue #278, opened 2016-11-16
Non-normative sections in the metadata spec


attilamolnar:
Sections to work on:
 * Security considerations
 * Implementation considerations
 * Moderation considerations

jwheare:
These will probably take a very similar form to the message-tags-3.3 sections:
https://github.com/ircv3/ircv3-specifications/blob/8eb2bd8b21f1216a21ffd88eb9b7cf94057c27af/core/message-tags-3.3.md#security-considerations

jwheare:
Closing as obsolete.

Issue #265, opened 2016-07-15
Unbundle specifications from versioned releases


attilamolnar:
Specifications should be released when they are ready. There is little point in bundling a bunch of unrelated specifications together, and it is also slowing down the ecosystem because finished minor specifications are being held back because work is underway on an unrelated major specification.

Related: https://github.com/ircv3/ircv3-specifications/issues/221#issuecomment-183365232

jwheare:
My thoughts:

This should apply from the 3.3 batch of specifications going forward. i.e. we should drop the use of identifiers like 3.x entirely.

The only purpose this served was to indicate that the previous version in some way was a prerequisite for support. Specifications instead should list their dependencies explicitly. This also removes the need for the Base vs Optional distinction. If a "base" specification is needed, it can be listed as such, rather than implied by a version number.

Some examples:
* STS depends on Capability negotiation 3.2, CAP LS 302 (due to CAP values)
* The `message-tags` CAP depends on the 3.2 Message Tags spec.

In the case of `message-tags`, I think it's also worth including the parts of the previous spec that still apply, and mark the previous version as deprecated/obsolete, to avoid implementors having to consult a chain of specs in order to understand the latest version.

This shouldn't apply where a spec merely uses a mechanism specified elsewhere (e.g. specs don't all need to include the details of CAP negotiation)

In summary:

* Drop version numbers
* Drop optional/base distinction
* List dependencies explicitly
* Release specifications independently when they're ratified

jwheare:
This is replaced by #329

PR #339, opened 2018-04-11
[WIP] Merged metadata rewrite

jwheare:
This is a work in progress. Currently representing the changes from #245 and #250. This can probably replace those issues.

This still needs massaging and editing, some things might be in a weird order, BUT this should contain everything you need to implement the updated metadata spec and matches what we've implemented in IRCCloud.

The best way to improve this spec is to try implementing it now, rather than wait for it to be finished. That way we can identify any issues and fix them.

This will eventually exist at a new URL, without 3.2 in it, and probably without errata, but not renaming it for now to allow a degree of diff readability,

Here are the TODO items from #250 (most of which are done)

* [ ] #278 Additional non-normative sections
* [ ] Allow ordered but async processing of requests (database storage, etc)
* [x] Update numerics to match [inspircd implementation](https://github.com/inspircd/inspircd/blob/3d52a33d3c650292b33825ce3458456790149caa/src/modules/m_ircv3_metadata.cpp#L35-L39)
* [x] Rewrite to include the parts of 3.2 that still apply, clarifying if necessary and mention that 3.2 is deprecated (see #279)
* [x] Add a requirement that clients should expect to receive notifications even if they haven't subscribed to any key, e.g. if a server needs to update a client's own metadata, e.g. on oper, or auth, or if an oper sets a key on another user.
* [x] Drop the 3.3 versioning. (see #265)
* [x] Change ERR_KEYINVALID to drop the message reason and include the invalid key as the last arg since it could contain spaces
* [x] Clarify ERR_KEYNOPERMISSION to expand "not settable by the requesting user" with "or otherwise disallowed by the server"
* [x] Allow hyphens in key names
* [x] Rename CAP to just `metadata` (no need for version number, values are about more than notifications)
* [ ] Possibly specify a cap value that advertises a limit of key and/or value names instead of ISUPPORT?

### DIFF
--- a/core/metadata-3.2.md
+++ b/core/metadata-3.2.md
@@ -15,6 +15,18 @@ copyrights:
     period: "2018"
     email: "james@irccloud.com"
 ---
+
+## Notes for implementing work-in-progress version
+
+This is a work-in-progress specification.
+
+Software implementing this work-in-progress specification MUST NOT use the
+unprefixed `metadata` capability name. Instead, implementations SHOULD
+use the `draft/metadata` capability name to be interoperable with other
+software implementing a compatible work-in-progress version.
+
+The final version of the specification will use an unprefixed capability name.
+
 ## Introduction
 
 It is generally useful to associate metadata with one's IRC presence, e.g. to
@@ -31,8 +43,7 @@ this purpose.
 
 All metadata subcommands will flow through the outgoing `METADATA` verb.
 
-This specification adds the `metadata-notify` capability for notifications.
-Please see the 'Metadata Notifications' section for more information.
+This specification adds the `metadata` capability.
 
 For the purposes of this specification, 'targets' are entities for which
 metadata may be set. Such entities MUST be a valid nickname or channel. As a
@@ -47,6 +58,16 @@ attempts with `ERR_KEYNOPERMISSION`.
 Implementations MAY provide a mechanism for limiting visibility of certain
 metadata to certain users. Such mechanism is implementation-defined;
 for instance, it may depend on some permission level or a flag.
+
+### Capability value
+
+The value of the `metadata` capability introduced in this document MUST
+have the following format:
+
+    [<anything>,]maxsub=<N>[,<anything>]`
+
+`<N>` is an integer indicating the maximum number of keys a client is allowed
+to be subscribed to.
 
 ### METADATA GET
 
@@ -134,6 +155,83 @@ using `ERR_KEYNOPERMISSION` with an asterisk (`*`) in the `<Key>` field.
 
 *Errors*: `ERR_KEYNOPERMISSION`
 
+### METADATA SUB
+
+This subcommand is used to subscribe to metadata keys.
+
+Syntax:
+
+    METADATA * SUB <key1> [<key2> ...]
+
+The server MUST reply with zero or more numerics of the following types in any
+order: `RPL_METADATASUBOK`, `ERR_KEYINVALID`, `ERR_KEYNOPERMISSION` and zero or
+one `ERR_METADATATOOMANYSUBS` numeric.
+The server MUST end the reply with one `RPL_METADATAEND` numeric.
+
+The server MUST process the keys in the given order. This is critical when
+determining which keys the client gets subscribed to in case the server limits
+the number of keys the client can subscribe to.
+
+If the client is subscribed to too many keys then the server MUST include a
+`ERR_METADATATOOMANYSUBS` numeric in its reply and not process any further keys
+in the command.
+
+If the client successfully subscribes to a key it is not subscribed to then the
+key MUST appear in a `RPL_METADATASUBOK` reply numeric.
+
+If the client tries to subscribe to a key it is already subscribed to then the
+client remains subscribed to the key. In case such a key is processed in a
+request the key MUST appear at least once in a `RPL_METADATASUBOK` numeric in
+the reply.
+
+Servers MUST respond to requests to subscribe to a key whose name is invalid
+with a `ERR_KEYINVALID` numeric.
+
+Servers MAY additionally respond to requests to subscribe to a key that the
+client has no privilege to access with a `ERR_KEYNOPERMISSION` numeric.
+Even if a server does that, the subscription MUST still be successful.
+This is because in this case the `ERR_KEYNOPERMISSION` numeric only serves as a
+warning, indicating that the client will not receive METADATA events about this
+key unless it acquires the necessary (implementation defined) privileges later.
+
+### METADATA UNSUB
+
+This subcommand is used to unsubscribe from metadata keys.
+
+Syntax:
+
+    METADATA * UNSUB <key1> [<key2> ...]
+
+The reply of the server MUST be zero or more numerics of the following types
+in any order: `RPL_METADATAUNSUBOK`, `ERR_KEYINVALID`.
+Then the server MUST end the reply with one `RPL_METADATAEND` numeric.
+
+If a client successfully unsubscribes from a key it is subscribed to then the
+key MUST appear in a `RPL_METADATAUNSUBOK` reply numeric.
+
+If a client tries to unsubscribe from a key that it is not subscribed to then
+the client remains not subscribed to the key and the key MUST appear at least
+once in a `RPL_METADATAUNSUBOK` numeric in the reply.
+
+Servers MUST respond to requests to subscribe to a key whose name is invalid
+with a `ERR_KEYINVALID` numeric.
+
+### METADATA SUBS
+
+This subcommand can be used to get a list of keys which the client is
+subscribed to.
+
+Syntax:
+
+    METADATA * SUBS
+
+The server MUST reply with zero or more `RPL_METADATASUBS` numerics and then
+one `RPL_METADATAEND` numeric.
+
+The replied `RPL_METADATASUBS` numerics, collectively, MUST contain all keys
+the client is subscribed to exactly once and MUST NOT contain keys the client
+is not subscribed to.
+
 ### METADATA SYNC
 
 This subcommand requests the full synchronization of metadata associated with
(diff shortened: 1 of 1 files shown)


core/metadata-3.2.md
```diff
@@ -41,6 +59,16 @@ Implementations MAY provide a mechanism for limiting visibility of certain
...
 
+### Capability value
+
+The value of the `metadata` capability introduced in this document MUST
+have the following format:
+
+    [<anything>,]maxsub=<N>[,<anything>]`
+
+`<N>` is an integer indicating the maximum number of keys a client is allowed
+to be subscribed to.
```
DanielOaks: `s/is allowed to/can/`? reads a bit easier imo

core/metadata-3.2.md
```diff
@@ -41,6 +59,16 @@ Implementations MAY provide a mechanism for limiting visibility of certain
 metadata to certain users. Such mechanism is implementation-defined;
 for instance, it may depend on some permission level or a flag.
 
+### Capability value
+
+The value of the `metadata` capability introduced in this document MUST
+have the following format:
+
+    [<anything>,]maxsub=<N>[,<anything>]`
```
DanielOaks: the `<anything>` feels really sketchy. Can we instead specify this the same way we specify tokens in the STS spec [here](https://ircv3.net/specs/extensions/sts.html#details)?

jwheare:
I’m basically gonna rewrite everything so maybe hold off on wording reviews for a bit (already changed most of these)

DanielOaks:
Ah fairfair, sounds good!

core/metadata-3.2.md
```diff
@@ -121,6 +155,146 @@ using `ERR_KEYNOPERMISSION` with an asterisk (`*`) in the `<Key>` field.
...
+The `<key>` parameter of this numeric is the first key that the client was not
+subscribed to.
+
+## Notification Mechanics (TODO merge with Metadata Notifications)
+
+A client can either be subscribed to a key, or not subscribed to it.
+
+The server MUST allow a client to subscribe to any valid keys, even to
+privileged keys when the client has no privilege to access that key at
+the time of subscription.
```
SaberUK: Why?

jwheare: Presumably because the privilege may change and then they can start getting notifications.

jwheare:
OK, reviews welcome on the Mechanisms section. Still need to rewrite the opening preamble, the subcommand headings and revisit examples.

core/metadata-3.2.md
```diff
@@ -20,26 +40,139 @@ This proposal aims to codify one mechanism for working with metadata: metadata
...
+
+Errors:
+
+* `ERR_TARGETINVALID` when a client refers to an invalid target.
+* `ERR_KEYINVALID` when a client refers to an invalid key.
+* `ERR_KEYNOPERMISSION` when a client attempts to access or set a key on a target without sufficient permission.
+
+### Keys and values
+
+Keys MUST be restricted to the ranges `A-Z`, `a-z`, `0-9`, and `_.:-` and are case-insensitive. Values are unrestricted, except that they MUST be UTF-8.
```
DanielOaks: hmm... we should specify that the first character of a key can't be `:`, for obvious reasons.

DanielOaks:
So with the rewrite, it looks like there's now a `METADATA` message sent by the server. Is that message replacing the existing `RPL_KEYVALUE` numeric? It looks like it does in SYNC, but the other commands still refer to `RPL_KEYVALUE`.

core/metadata-3.2.md
DanielOaks: Makes sense to me. If the key name's valid, just throw it in the sync list no matter what. For example, if they setup all their metadata stuff on connecting and then after connect they oper up.

core/metadata-3.2.md
```diff
@@ -20,26 +40,139 @@ This proposal aims to codify one mechanism for working with metadata: metadata
...
+
+
+TODO is this still required? include it for back compat?
+If `METADATA` is supported, it MUST be specified in `RPL_ISUPPORT` using the
+`METADATA` key. Servers MAY specify a limit on the number of explicitly-set
+keys per-user; the format in that case MUST be `METADATA=<integer>`, where
+`<integer>` is the limit.
+
+
+### Server notification
```
DanielOaks: "Server notification" is a phrase I've never seen before in any IRC docs or specs, and it honestly sounds pretty confusing. Can we change how this is referred to, for example 'Server message' or something else that's more familiar?

core/metadata-3.2.md
```diff
@@ -20,26 +40,139 @@ This proposal aims to codify one mechanism for working with metadata: metadata
...
+
+### Capability
+
+The `metadata` capability and value indicates that a server supports metadata, and communicates any configuration value keys that may affect the level of support.
+
+The ABNF format of the `metadata` capability and value is: (TODO verify ABNF)
+
+    capability ::= 'metadata' ['=' value]
+    value      ::= value_item [',' value_item]*
+    value_item ::= ( 'maxsub=' integer ) |  ( 'maxkey=' integer )
```
DanielOaks: If we're intending this to be extensible we should probably do something more along the lines of:

```
    capability ::= 'metadata' ['=' tokens]
    tokens     ::= token [',' token]*
    token      ::= key ['=' value]
    key        ::= <sequence of a-zA-Z0-9.-_ >
    value      ::= // whatever
```
or whatevs. and then just below define the keys and their allowed values. for the ISUPPORT key we can just say use the same `tokens` format, `METADATA ['=' tokens]` etc.

core/metadata-3.2.md
```diff
@@ -20,26 +40,139 @@ This proposal aims to codify one mechanism for working with metadata: metadata
...
+    METADATA <Target> <Key> <Visibility> <Value>
+
+`Target` MUST be any valid nickname or channel name.
+
+`Key` MUST be a valid key name
+
+`Visibility` MUST be an asterisk (`*`) for keys visible to everyone, or an implementation-defined value which describes the key's visibility status; for instance, it MAY be a permission level or flag. (TODO should we define a prefix like STATUSMSG for this value?)
 
-All metadata subcommands will flow through the outgoing `METADATA` verb.
+`Value` MUST be any UTF-8 encoded value.
```
DanielOaks: "MUST be any" sounds a bit weird. can we just simplify this down to just _"`Value` MUST be in UTF-8 encoding"_?

DanielOaks:
This PR hasn't touched it, but under `METADATA CLEAR`, it says:
```
The server MUST respond with one `RPL_KEYVALUE` event per cleared key
```

This is a little confusing to read off the bat. We should specify there that the `RPL_KEYVALUE` should be sent with no value to indicate the key's been cleared.

jwheare:
The notification isn’t new. It used to be enabled by the optional metadata-notify cap. Same as away/account notify. Keyvalue is the still the reply for queries.

DanielOaks:
Ahh, I see... in that case can we just rip out the `METADATA` message wholesale and make everything send `RPL_KEYVALUE` instead? It's confusing to have a message and a numeric that send exactly the same data (down to the exact params).

core/metadata-3.2.md
timofonic: I'm just a mere IRC user that wants progress over the IRC protocol. I'm just happy to see research and development over it, I consider it needs to adapt and improve to be relevant and be able to regain user base again against very popular but proprietary chat (+voip +easy image and file sharing, +server logging, +integration with other services such as GitLab and GitHub and more, +extremely easy to use and popular clients in major platforms to the point of being foolproof...) SaaS solutions. 

Anyway, I'm very curious and usually read a lot of documentation and books, including many stuff related to IRC and even it's related IETF RCF documents. 

From my very extremely limited Point Of View: I consider your argumentation seems reasonable to me. Please take it with a very tiny grain of salt, anyway.

Thanks to everyone for improving IRCv3×! I hope many great ideas get approved and implemented by mostly all IRC/chat clients/bouncers/matrix/psyc2-gnunet etc

core/metadata-3.2.md
timofonic: Extensibility sounds nice. I just hope clients get really aware of those and implement them.  Any strategy plan to avoid fragmentation and incompatibility?

jesopo:
Would it be worth making considerations for server validated/provided keys? e.g. client cert fingerprint? this might need something like what message-tags do (prefix key with `+` to denote client-provided) though.

k4bek4be:
I have almost done a server side implementation of this capability. (I'll soon be starting the client-side work.) Here are my comments and own solutions of problems:
- I have added a new numeric, 776, ERR_METADATAINVALIDSUBCOMMAND, "%s :invalid metadata subcommand" for unknown subcommands, like it's done with the ERR_INVALIDCAPCMD.
- I do not understand the purpose and usage of the RPL_WHOISKEYVALUE.
- I am sending ERR_METADATASYNCLATER not only after a SYNC or JOIN, but also after SUB (notifying the subsriber about users that have the key already set), or even without any client command (a netjoin generating massive amounts of metadata notifications). I've done this in the purpose of avoiding "max sendq exceeded" kills.
- The SYNC command target is ignored. This command just sends out the notifications that were queued before. In the ERR_METADATASYNCLATER, i'm returning either a valid target (one of the queued ones), or "*".
- I have used two caps, "draft/metadata=maxsub=XX" and "draft/metadata-notify-2", and the isupport tag METADATA=(max subscriptions).
- Only the "*" visibility is used, as all keys are visible to everyone. However, user keys are hidden from users that aren't on any common channel, and channel keys are hidden from users that haven't joined the channel.
- As suggested by other people, I allowed the yet-unregistered (NICK/USER) users to submit the SUB and SET metadata commands. There is no word about such feature in the document.

k4bek4be:
A little update on the previous.
We have have the metadata support active for few days now. There are no significant problems detected.
- I have stopped using the ERR_METADATASYNCLATER. Instead, the server is sending the queued data periodically (every 1.5 seconds) without having the client to send the SYNC subcommand. I think it's a cleaner solution. The SYNC command now has no effect.
- Metadata for logged in users is stored by account name basis, and restored for users logging into the account. It allows sessions running on different computers (or metadata-incompatible clients) to retain all the data. The stored value can be dropped by usual "METADATA * SET key-name" command.

SadieCat:
Is there anything that is still blocking this?

jwheare:
Blocked on available time/energy from someone who can own the editing process. It's a relatively higher complexity spec and I'm still not sure if all the remaining flaws have been addressed. @k4bek4be kindly submitted a PR to my fork that I haven't been able to review https://github.com/jwheare/ircv3-specifications/pull/1

I'm open to anyone else who can commit to taking it forward.

jwheare:
@jesopo and @DanielOaks expressed an interest in taking over ownership of this.

The main issues are

* Review https://github.com/jwheare/ircv3-specifications/pull/1
* Continue/finish rearranging/deduping the words to make sense
* Consider any remaining scaling issues

@justjanne also expressed an interest in implementing this in Quassel. Implementations are probably the best way to tease out issues.

DanielOaks:
As an FYI, I'm working on this over here: https://github.com/DanielOaks/ircv3-specifications/blob/metadata/core/metadata.md

Will submit as a PR sometime soon, wanna find any possible pain points in it by implementing it in Ergo.

jwheare:
Closed in favour of #501

PR #341, opened 2018-07-21
EXTJWT command for integrating external web services

prawnsalad:
This spec provides a way for web services hosted externally to an IRC server to authenticate users that are connected to the IRC server by making use of the standard JWT tokens (https://jwt.io/).

This allows a web service to do things such as:
* Granting admin access to a networks wiki page if a user has +o on the network
* Granting write access to a channels wiki page if the user has +o in the channel
* Automatically creating a user account if the user is logged into the IRC server and has an account name

For a more indepth example we could use the free audio/video conference service - Jitsi Meet. This service has built in JWT verification in that an application can send a user to a URL that contains a JWT token, and if the Jitsi Meet server verifies this token successfully, the user is granted access to that conference room.
When an IRC client wants to join a conference room, it would first call `EXTJWT #testchannel` to receive a JWT token from the IRCd. The client would then open a browser window navigating to the Jitsi Meet URL while passing that token. It is up to the client to decide how and where to use this token, eg. via a "Jitsi Call" button for example.

### DIFF
--- a/extensions/extjwt.md
+++ b/extensions/extjwt.md
@@ -21,7 +21,7 @@ Once in use, this:
 
 ### How the proof works
 
-The IRC network creates the proof by generating JWT tokens and sending them to the client. The client may use this token to open an external service with the token in its URL.
+The IRC network creates the proof by generating signed JWT tokens and sending them to the client. The client may use this token to open an external service with the token in its URL.
 
 The JWT token (https://jwt.io/) is an encoded JSON payload that is signed with a shared secret string between the IRC server and the external service. The JSON payload consists of known properties (claims) that include:
 * `exp` `1529917513` Expiry time for this token. Usually less than 1 minute from the token generation.
@@ -42,9 +42,9 @@ Only one new command is introduced in this extension, `EXTJWT`.
 
 Syntax: `EXTJWT [channel]`
 
-Response syntax: `EXTJWT <requested_target> [*] <JWT_token>`
+Response syntax: `EXTJWT <requested_target> [*] <jwt_token>`
 
-The client may send `EXTJWT` or `EXTJWT *` to the server to request a new JWT token. The server must then reply with `EXTJWT *` and a JWT token as its second parameter, containing the following claims that are relevant to the client at that time:
+The client MAY send `EXTJWT` or `EXTJWT *` to the server to request a new JWT token. The server MUST then reply with `EXTJWT *` and a JWT token as its jwt_token parameter containing the following claims that are relevant to the client at that time:
 
 * `exp` Number; Unix timestamp for when this token expires. Usually less than 1 minute from the token generation.
 * `iss` String; The server name that generated this token.
@@ -52,18 +52,18 @@ The client may send `EXTJWT` or `EXTJWT *` to the server to request a new JWT to
 * `account` String; The account name of the user that requested this token. Empty if not available.
 * `net_modes` []String; An array of user modes the IRCd may want to disclose. Eg, if the user is an operator.
 
-The command must also support a single parameter of a channel name. Eg. `EXTJWT #channel`. The server must then reply with the channel name as its first parameter and the JWT token containing the above claims and also the following claims relevant to the channel at that time:
+The command MUST also support a single parameter of a channel name. Eg. `EXTJWT #channel`. The server MUST then reply with the channel name as its requested_target parameter, the JWT token containing the above claims and also the following claims relevant to the channel at that time:
 
 * `channel` String; The channel name this token is related to.
 * `joined` Boolean; True if the client that requested this token is joined to the channel.
 * `time_joined` Number; The time in which the user joined the channel.
 * `modes` []String; An array of the channel modes the client has in this channel.
 
-The IRC server must include the above claims but may include any extra claims.
+The IRC server MUST include the above claims but MAY include any extra claims.
 
 #### Handling long responses
 
-In some cases the encoded token may be longer than the maximum line length allowed between the client and server. In this case, the first parameter of the response must be `*` to indicate that further data will follow. The final chunk of the response sent to the client must not include `*` as the first parameter.
+In some cases the encoded token may be longer than the maximum line length allowed between the client and server. In this case, the first parameter of the response MUST be `*` to indicate that further data will follow. The final chunk of the response sent to the client MUST NOT include `*` as the first parameter.
 
 Eg:
 ~~~
--- a/extensions/extjwt.md
+++ b/extensions/extjwt.md
@@ -33,7 +33,7 @@ The JWT token (https://jwt.io/) is a base64 encoded JSON payload that is signed 
 
 When an external service is opened with this token in its URL, the external service verifies that the token has not been tampered with by one of two methods:
 1. If the external service has a shared secret for this IRC network, it can now verify the token directly.
-2. The external service can make a HTTP GET request to the URL given in the `vfy` claim, replacing `%s` with the token string. The request MUST be responded to with a HTTP 200 status if the token has been verified, or a HTTP 403 status if the token is invalid.
+2. The external service can make a HTTP GET request to the URL given in the `vfy` claim, replacing `%s` with the token string. The request MUST be responded to with a HTTP 200 status if the token has been verified, or a HTTP 401 status if the token is invalid.
 
 Once successfully verified, the external service can then use the available claims in the token to create any required user accounts and log the user in automatically.
 
--- a/extensions/extjwt.md
+++ b/extensions/extjwt.md
@@ -34,7 +34,7 @@ When an external service is opened with this token in its URL, the external serv
 
 ## Usage
 
-If the feature is available on the IRC server, the `EXTJWT` token is added to its ISUPPORT list.
+If the feature is available on the IRC server, the `EXTJWT=1` token pair is added to its ISUPPORT list where `1` denotes the version of this feature.
 
 Only one new command is introduced in this extension, `EXTJWT`.
 
--- a/extensions/extjwt.md
+++ b/extensions/extjwt.md
@@ -24,11 +24,11 @@ Once in use, this:
 The IRC network creates the proof by generating signed JWT tokens and sending them to the client. The client may use this token to open an external service with the token in its URL.
 
 The JWT token (https://jwt.io/) is an encoded JSON payload that is signed with a shared secret string between the IRC server and the external service. The JSON payload consists of known properties (claims) that include:
-* `exp` `1529917513` Expiry time for this token. Usually less than 1 minute from the token generation.
+* `exp` `1529917513` Expiry time for this token.
 * `iss` `"irc.example.org"` The server name that generated this token.
 * `sub` `"somenick"` The nick of the user that generated this token.
 * `account` `"somenick"` The account name of the user that generated this token. Empty if not logged in.
-* `net_modes` `["o"]` User modes the IRCd wishes to disclose. Eg, if the user an operator.
+* `net_modes` `["o"]` User modes the IRCd wishes to disclose. Eg, if the user is an operator.
 
 When an external service is opened with this token in its URL, the external service verifies that the token has not been tampered with using its pre-configured secret string and can then use the available claims to create any required user accounts and log the user in automatically.
 
@@ -46,7 +46,7 @@ Response syntax: `EXTJWT <requested_target> [*] <jwt_token>`
 
 The client MAY send `EXTJWT` or `EXTJWT *` to the server to request a new JWT token. The server MUST then reply with `EXTJWT *` and a JWT token as its jwt_token parameter containing the following claims that are relevant to the client at that time:
 
-* `exp` Number; Unix timestamp for when this token expires. Usually less than 1 minute from the token generation.
+* `exp` Number; Unix timestamp for when this token expires. See below for notes on the expiry claim.
 * `iss` String; The server name that generated this token.
 * `sub` String; The nick of the client that requested this token.
 * `account` String; The account name of the user that requested this token. Empty if not available.
@@ -71,6 +71,12 @@ Eg:
 [S -> C] EXTJWT #channel * iY2xhaW0yIjoic29tZSBsb25nIHZhbHVlIiwiY2xhaW0zIjoic29tZSBsb25nIHZhbHVlIiwiY2xhaW00Ijoic29tZSBsb25nIHZhbHVlIiwiY2xhaW01Ijoic29tZSBsb25nIHZhbHVlIiwiY2xhaW02Ijoic29tZSBsb25nIHZhbHVlIiwiY2xhaW03Ijoic29tZSBsb25nIHZhbHVlIiwiY2xhaW04Ijoic29tZSBsb25nZXIgdmFsdWUgdG8gbWFrZSBzdXJlIHRoaXMgdG9rZW4gaXMgdG9vIGxvbmc
 [S -> C] EXTJWT #channel gdG8gc2VuZCBvbiBvbmUgSVJDIDUxMiBjaGFyYWN0ZXIgbGluZSJ9.c9_pKy1jFsDeevja7o6spPa-JUyzg4z4k3A65fxwZWw
 ~~~
+
+#### Notes on the token expiry claim
+
+When generating a token, the expiry (exp) claim must be configured with enough length of time for the token to be used but also be short enough that the token does not last indefinately, leaving the user with a valid token after the user has left a channel or changed its network modes.
+
+One minute is usually enough time for the client to receive the token from the `EXTJWT` command and then open the external service webpage, however, non-webpage based services may require a longer expiration depending on its implementation.
 
 ## Examples
 
--- a/extensions/extjwt.md
+++ b/extensions/extjwt.md
@@ -26,7 +26,7 @@ The IRC network creates the proof by generating signed JWT tokens and sending th
 The JWT token (https://jwt.io/) is an encoded JSON payload that is signed with a shared secret string between the IRC server and the external service. The JSON payload consists of known properties (claims) that include:
 * `exp` `1529917513` Expiry time for this token. Usually less than 1 minute from the token generation.
 * `iss` `"irc.example.org"` The server name that generated this token.
-* `nick` `"somenick"` The nick of the user that generated this token.
+* `sub` `"somenick"` The nick of the user that generated this token.
 * `account` `"somenick"` The account name of the user that generated this token. Empty if not logged in.
 * `net_modes` `["o"]` User modes the IRCd wishes to disclose. Eg, if the user an operator.
 
@@ -48,7 +48,7 @@ The client MAY send `EXTJWT` or `EXTJWT *` to the server to request a new JWT to
 
 * `exp` Number; Unix timestamp for when this token expires. Usually less than 1 minute from the token generation.
 * `iss` String; The server name that generated this token.
-* `nick` String; The nick of the client that requested this token.
+* `sub` String; The nick of the client that requested this token.
 * `account` String; The account name of the user that requested this token. Empty if not available.
 * `net_modes` []String; An array of user modes the IRCd may want to disclose. Eg, if the user is an operator.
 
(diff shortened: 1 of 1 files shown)


extensions/extjwt.md
```diff
@@ -0,0 +1,111 @@
...
+
+### The EXTJWT Command
+
+Syntax: `EXTJWT [channel]`
+
+Response syntax: `EXTJWT <requested_target> [*] <jwt_token>`
+
+The client MAY send `EXTJWT` or `EXTJWT *` to the server to request a new JWT token. The server MUST then reply with `EXTJWT *` and a JWT token as its jwt_token parameter containing the following claims that are relevant to the client at that time:
+
+* `exp` Number; Unix timestamp for when this token expires. Usually less than 1 minute from the token generation.
```
ProgVal: + UTC

extensions/extjwt.md
```diff
@@ -0,0 +1,111 @@
...
+
+## Usage
+
+If the feature is available on the IRC server, the `EXTJWT` token is added to its ISUPPORT list.
+
+Only one new command is introduced in this extension, `EXTJWT`.
+
+### The EXTJWT Command
+
+Syntax: `EXTJWT [channel]`
```
ProgVal: should be: `EXTJWT ( <channel> | * )`

extensions/extjwt.md
```diff
@@ -0,0 +1,111 @@
...
+* `iss` String; The server name that generated this token.
+* `nick` String; The nick of the client that requested this token.
+* `account` String; The account name of the user that requested this token. Empty if not available.
+* `net_modes` []String; An array of user modes the IRCd may want to disclose. Eg, if the user is an operator.
+
+The command MUST also support a single parameter of a channel name. Eg. `EXTJWT #channel`. The server MUST then reply with the channel name as its requested_target parameter, the JWT token containing the above claims and also the following claims relevant to the channel at that time:
+
+* `channel` String; The channel name this token is related to.
+* `joined` Boolean; True if the client that requested this token is joined to the channel.
+* `time_joined` Number; The time in which the user joined the channel.
```
ProgVal: What should its value be if `joined` is False?

The best solution would probably be `join` and `time_joined` as a single optional attribute, so the receiver of the JWT does not need to sanity-check this.

extensions/extjwt.md
```diff
@@ -0,0 +1,111 @@
...
+* `modes` []String; An array of the channel modes the client has in this channel.
+
+The IRC server MUST include the above claims but MAY include any extra claims.
+
+#### Handling long responses
+
+In some cases the encoded token may be longer than the maximum line length allowed between the client and server. In this case, the first parameter of the response MUST be `*` to indicate that further data will follow. The final chunk of the response sent to the client MUST NOT include `*` as the first parameter.
```
ProgVal: You should use the same mechanism as [`AUTHENTICATE`](https://ircv3.net/specs/extensions/sasl-3.1.html), for consistency between IRCv3 specs.

extensions/extjwt.md
```diff
@@ -0,0 +1,111 @@
...
+* `nick` String; The nick of the client that requested this token.
+* `account` String; The account name of the user that requested this token. Empty if not available.
+* `net_modes` []String; An array of user modes the IRCd may want to disclose. Eg, if the user is an operator.
+
+The command MUST also support a single parameter of a channel name. Eg. `EXTJWT #channel`. The server MUST then reply with the channel name as its requested_target parameter, the JWT token containing the above claims and also the following claims relevant to the channel at that time:
+
+* `channel` String; The channel name this token is related to.
+* `joined` Boolean; True if the client that requested this token is joined to the channel.
+* `time_joined` Number; The time in which the user joined the channel.
+* `modes` []String; An array of the channel modes the client has in this channel.
```
ProgVal: Maybe add a way to disclose modes with arguments? eg. `+I` to tell the mask that allows the user to be in that channel.

extensions/extjwt.md
```diff
@@ -0,0 +1,111 @@
...
+
+Response syntax: `EXTJWT <requested_target> [*] <jwt_token>`
+
+The client MAY send `EXTJWT` or `EXTJWT *` to the server to request a new JWT token. The server MUST then reply with `EXTJWT *` and a JWT token as its jwt_token parameter containing the following claims that are relevant to the client at that time:
+
+* `exp` Number; Unix timestamp for when this token expires. Usually less than 1 minute from the token generation.
+* `iss` String; The server name that generated this token.
+* `nick` String; The nick of the client that requested this token.
+* `account` String; The account name of the user that requested this token. Empty if not available.
+* `net_modes` []String; An array of user modes the IRCd may want to disclose. Eg, if the user is an operator.
```
ProgVal: Some generic net modes should probably added to this spec.

extensions/extjwt.md
```diff
@@ -0,0 +1,111 @@
...
+
+#### A client logged into the IRC server with operator privileges
+~~~
+[C -> S] EXTJWT *
+[S -> C] EXTJWT eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJleHAiOjE1Mjk5MTc1MTMsImlzcyI6ImlyYy5leGFtcGxlLm9yZyIsIm5pY2siOiJzb21lbmljayIsImFjY291bnQiOiJzb21lbmljayIsIm5ldF9tb2RlcyI6WyJvIl19.NREHeoO-aewAry44erDgCHuVmUW9zyJjG05mJYCXXfs
+~~~
+
+Where the replied token is decoded into:
+~~~json
+{"exp":1529917513,"iss":"irc.example.org","nick":"somenick","account":"somenick","net_modes":["o"]}
```
ProgVal: `"o"` looks like a channel op, you should probably rename it. A non-single-letter name would be nice too (let's not repeat the mistakes of the past).

extensions/extjwt.md
```diff
@@ -0,0 +1,111 @@
...
+
+#### A client connected to the IRC server without a registered account or operator privileges
+~~~
+[C -> S] EXTJWT *
+[S -> C] EXTJWT * eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJleHAiOjE1Mjk5MTc1MTMsImlzcyI6ImlyYy5leGFtcGxlLm9yZyIsIm5pY2siOiJzb21lbmljayIsImFjY291bnQiOiIiLCJuZXRfbW9kZXMiOltdfQ.Vkm2XJXHz6rkq-R93fJUp88kNmAU9J65w46ZsQLjJrY
+~~~
+
+Where the replied token is decoded into:
+~~~json
+{"exp":1529917513,"iss":"irc.example.org","nick":"somenick","account":"","net_modes":[]}
```
ProgVal: Why not allow `"account": null`, or making that attribute optional?

ProgVal:
Looks like a cool idea overall. Even if it does not get implemented by IRC daemons (or networks), this could be partially provided by (non-privileged) bots.

extensions/extjwt.md
DanielOaks: This is the mechanism that `CAP LS 302` uses, so there's consistency between that and EXTJWT at least.

ProgVal: Indeed. As the `CAP LS 302` way is much easier to implement, I agree this mechanism should be used.

extensions/extjwt.md
prawnsalad: Unix timestamp is consistent with other parts such as RPL_TOPICWHOTIME so we don't want to go changing that.

extensions/extjwt.md
prawnsalad: These modes are the user modes the IRC server has set. Ie. `o` for operator. If an IRCd introduces named modes then the named modes will be placed in here too.

extensions/extjwt.md
```diff
@@ -0,0 +1,116 @@
...
+
+Only one new command is introduced in this extension, `EXTJWT`.
+
+### The EXTJWT Command
+
+Syntax: `EXTJWT ( <channel> | * )`
+
+Response syntax: `EXTJWT <requested_target> [*] <jwt_token>`
+
+The client MAY send `EXTJWT` or `EXTJWT *` to the server to request a new JWT token. The server MUST then reply with `EXTJWT *` and a JWT token as its jwt_token parameter containing the following claims that are relevant to the client at that time:
```
SaberUK: Is there any reason to have two command syntax variants for a non-channel JWT (no parameters and `*`)?

extensions/extjwt.md
```diff
@@ -0,0 +1,116 @@
...
+
+* `exp` Number; Unix timestamp for when this token expires. See below for notes on the expiry claim.
+* `iss` String; The server name that generated this token.
+* `sub` String; The nick of the client that requested this token.
+* `account` String; The account name of the user that requested this token. Empty if not available.
+* `net_modes` []String; An array of user modes the IRCd may want to disclose. Eg, if the user is an operator.
```
SaberUK: Bikeshed: `net_modes` should be referred to as `user_modes` and `modes` (below) should be `channel_modes`.

SaberUK:
How should a server behave if there was an issue with generating the JWT for some reason?

SaberUK:
Is it useful to add a context parameter to the command so that JWTs can be generated with different secrets for integrating with multiple services? e.g. `EXTJWT #channel servicename`.

prawnsalad:
@SaberUK I could imagine that use case, yes. If I make the `*` in `EXTJWT *` required instead of optional as you questioned in another comment, then the service name could be an optional second parameter, `EXTJWT * [servicename]` / `EXTJWT #channel [servicename]`

DanielOaks:
I like this, sounds like a nice way to let nets setup external services without giving db access and similar. From the looks of it, it's not useful for services being setup without the network's prior OK (agreement on the secret, etc), and that's intended?

Will look at writing up an implementation of this in Ora. The `servicename` stuff being discussed above sounds especially useful, once that change is resolved I'll write up our impl.

prawnsalad:
I'm about to update this draft spec with some minor tweaks that's come up during testing since this was first implemented.
* Adding the optional `[servicename]` argument as mentioned above.
* Renaming `net_modes` and `modes` with `umodes` and `cmodes` respectively.
* Replacing `iss` with `iat` (issued at) so that the external service can decide how long the token should be valid for. Different services may require different lengths of time.

It has been mentioned by a few people that usage would be fairly limited since third parties would need to be configured with JWT secrets from the IRCd. To get around this I could add an optional `verify` claim in the token that contains a URL. This will allow any external service to make a simple HTTP call to verify a token they received from the network.
eg.
1. A client receives a JWT token containing `"verify":"https://irc.foo.net/extjwtverify/%s"`.
2. The client opens a third party service that accepts the token.
3. The service makes a HTTP call to the verify URL replacing `%s` with the token
4. The verify URL replies with either a `200` or `403` HTTP status if the token is valid.

Advantages:
- The IRCd JWT secrets will never need to be shared to third party services
- No previous agreement needs to be configured between the IRCd and third party services
- The URL is determined by the IRCd and can be configured as fits. eg. it may use a separate host instead of the IRCd directly to verify tokens.

extensions/extjwt.md
```diff
@@ -12,71 +12,82 @@ copyrights:
...
 * `account` `"somenick"` The account name of the user that generated this token. Empty if not logged in.
-* `net_modes` `["o"]` User modes the IRCd wishes to disclose. Eg, if the user is an operator.
+* `umodes` `["o"]` User modes the IRCd wishes to disclose. Eg, if the user is an operator.
 
-When an external service is opened with this token in its URL, the external service verifies that the token has not been tampered with using its pre-configured secret string and can then use the available claims to create any required user accounts and log the user in automatically.
+When an external service is opened with this token in its URL, the external service verifies that the token has not been tampered with by one of two methods:
+1. If the external service has a shared secret for this IRC network, it can now verify the token directly.
+2. The external service can make a HTTP GET request to the URL given in the `vfy` claim, replacing `%s` with the token string. The request MUST be responded to with a HTTP 200 status if the token has been verified, or a HTTP 403 status if the token is invalid.
```
kylef, line 36:
```suggestion
2. The external service can make a HTTP GET request to the URL given in the `vfy` claim, replacing `%s` with the token string. The request MUST be responded to with a HTTP 200 status if the token has been verified, or a HTTP 401 status if the token is invalid.
```

403 usually indicates that the token is valid, but doesn’t have access to the particular resource (AuthN vs AuthZ).

Secondly, can we relax the 200 response to something such as a “successful” response. If the API doesn’t have a response body then it might want to return a 204 which would be against this spec’s MUST requirement.

RyanSquared:
What happens if a user forges a JWT and sets the "vfy" URL to something like, for example, `"https://gib_200_always.fusionscript.info/%s"`?

I'd suggest instead changing the "vfy" format to instead be the following format:

`"https://%s/extjwtverify?t=%s"`

This would take the "iss" field as the first input, and the token as a whole as the second field.

Additionally, from what I can tell, it's possible to downgrade HTTPS to HTTP using the same method?

extensions/extjwt.md
```diff
@@ -0,0 +1,138 @@
...
+
+#### A server responding with a verification (`vfy`) claim
+~~~
+[C -> S] EXTJWT #channel forum.example.org
+[S -> C] EXTJWT #channel forum.example.org eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJleHAiOjE1Mjk5MTc1MTMsImlzcyI6ImlyYy5leGFtcGxlLm9yZyIsInZmeSI6Imh0dHBzOi8vaXJjLmV4YW1wbGUub3JnL2V4dGp3dD90PSVzIiwic3ViIjoidGVzdG5pY2siLCJhY2NvdW50IjoidGVzdG5pY2siLCJ1bW9kZXMiOltdLCJjaGFubmVsIjoiI2NoYW5uZWwiLCJqb2luZWQiOjE1Mjk5MTc1MDEsImNtb2RlcyI6WyJvIl19.WhZpr-m2v2T6y1tlLFukX3pvk4k-tiQLzlW-poR74fk
+~~~
+
+Where the replied token is decoded into:
+~~~json
+{"exp":1529917513,"iss":"irc.example.org",vfy":"https://irc.example.org/extjwt?t=%s","sub":"testnick","account":"testnick","umodes":[],"channel":"#channel","joined":1529917501,"cmodes":["o"]}
```
DanielOaks, line 137: json's a tiny bit mangled here, should be `,"vfy"` instead of `,vfy"`

RyanSquared:
```
02:19 <LordRyan> a user-supplied path is bound to have issues.
02:20 <LordRyan> i'd suggest a verify subdomain and a verify path, if you want to go that route.
02:20 <LordRyan> that way it is impossible to spoof the origin.
02:20 <LordRyan> "origin" => "issuer"
02:21 <LordRyan> so { verify_subdomain => "verify", iss => "example.com", verify_path => "/verify_extjwt?t=%s" }
02:21 <LordRyan> would go to https://verify.example.com/verify_extjwt?t=<token>
02:21 <LordRyan> and add in an enforcement that verify_path MUST be preceded by a /, enforced in any implementation that uses this system
02:21 <LordRyan> because otherwise you can just pass ".gib_200_always.example.com"
02:23 <LordRyan> and if you do it this way you can still do a "configuration-less", i.e. this gets added to meet.jit.si and they don't need to know anything about the server, they can just do the request.
```

extensions/extjwt.md
```diff
@@ -55,8 +55,7 @@ The client MAY send `EXTJWT` or `EXTJWT *` to the server to request a new JWT to
 The command MUST also support a single parameter of a channel name. Eg. `EXTJWT #channel`. The server MUST then reply with the channel name as its requested_target parameter, the JWT token containing the above claims and also the following claims relevant to the channel at that time:
 
 * `channel` String; The channel name this token is related to.
-* `joined` Boolean; True if the client that requested this token is joined to the channel.
-* `time_joined` Number; The time in which the user joined the channel.
+* `joined` Number; Unix timestamp of the time in which the client the channel. 0 if not joined.
```
SadieCat, line 68:
```suggestion
* `joined` Number; Unix timestamp of the time at which the client joined the channel. 0 if not joined.
```

SadieCat:
As I mentioned on IRC last night, what do people think about sending the token using an `Authorisation` header instead of the %s thing?

RyanSquared:
I would be fine with that. That means you only need to specify an optional "issuer subdomain" and an optional "issuer verification path" which can both be static strings.

RyanSquared:
I see some continued comments on IRC (both within IRCv3 and when discussing with some infosec friends) where the reason for this validation URL is misunderstood. I think it would be good to put some emphasis on only using the validation URL **IF** there is no possible shared secret. Most importantly, if there is a shared secret and it fails, I think that the validation URL **MUST NOT** attempted, as the shared secret is a more secure method and should take higher priority.

RyanSquared:
Additionally, some conversations that prawnsalad and I had about this:

```
02:00 <prawnsalad> LordRyan: if a user tampers with the token then it won’t match the hash any more
02:01 <LordRyan> and so the user makes a new hash
02:01 <LordRyan> i wasn't even thinking "tamper" i was thinking "forge" as in make it themselves
02:02 <LordRyan> if you're using vfy as a backup for not using a signature then there's no integrity in that signature and your ONLY verification is that URL in "vfy"
02:02 <LordRyan> anything including "vfy" can be forged
02:06 <prawnsalad> hm no. a jwt token is "$header.$payload.<hmac($header+"."+$payload . $secret)>"
02:06 <LordRyan> yes
02:06 <prawnsalad> to forge it you would need to know the secret that the ircd has
02:06 <LordRyan> why?
02:06 <LordRyan> this is supposed to be used if the service ALSO doesn't know the secret.
02:07 <LordRyan> you can just make it up. the service doesn't care.
02:10 <LordRyan> btw prawnsalad this issue also came up in a similar context in webauthn domain phish+MITMing so the request for the WebAuthn challenge must include the origin domain, kinda like a "client-side" salt.
02:11 <LordRyan> prawnsalad: let me phrase it this way: how does the party using the JWT for authentication know that you're not spoofing it if they don't have the secret either?
02:13 <prawnsalad> yea i get what you mean now, a rather than a vfy url it would be safer to have that pre set with the third party so there is only ever 1 verify url then
02:14 <LordRyan> or like i suggested on the GitHub, tie the issuer into the vfy url
02:14 <LordRyan> though that has the same issue, now that i think about it
```

RyanSquared:
Discussion on IRC brought up usage of an asymmetric key, with the public version stored in a standardized location in the issuer. This resolves both the issue of the external service not having a shared secret and the issue of the client being able to spoof the endpoint used for authentication.

slingamn:
What is the use case for including the channel join time in the token?

slingamn:
From discussion in #kiwiirc: we're tentatively proposing that server implementations that do not store channel join time information can send `joined: 1`, which indicates that the user is joined to the channel but the join time is unavailable for whatever reason.

SadieCat:
What is the reason for this change exactly?

slingamn:
I'd prefer not to add it to my implementation, given that there is no use case for it right now.

slingamn:
Clarification questions that came up during Dan's implementation:

1. Is the signing algorithm always `HS256`?
2. What should happen when the service name is not specified, or is `*`? Should the server be configured with a default secret key for signing these tokens?

prawnsalad:
> 1. Is the signing algorithm always `HS256`?

JWT tokens allow you to use any algo your implementation can support. As long as you can create + verify your own token, you're all good.

> 2. What should happen when the service name is not specified, or is `*`? Should the server be configured with a default secret key for signing these tokens?

A default should be used, yes. This default would be the most used as clients making use of EXTJWT wouldn't know what services your IRCd has running or what name to use for them unless the network has a dedicated client for it.

prawnsalad:
@RyanSquared 

> Discussion on IRC brought up usage of an asymmetric key, with the public version stored in a standardized location in the issuer. This resolves both the issue of the external service not having a shared secret and the issue of the client being able to spoof the endpoint used for authentication.

Other than the service being able to cache a pub key from the issuer, what other benefit would this have? Both algo types will need a request back to the issuer either for the pub key or verification as far as I understood.

Also to note, the external service is not required to know the shared secret for a token. While services run by the same IRCd network may share the secret which saves the verification request trip, third parties can make the verification request without knowing any secrets.

RyanSquared:
> Both algo types will need a request back to the issuer either for the pub key or verification as far as I understood.

For pubkey based verification you can just put a static key in a directory. For verifying the JWT itself, you'd have to implement something server side. I think a majority of hosts would rather have a static file option.

RyanSquared:
Can we add the supported services to the ISUPPORT token in any way? Something like, `EXTJWT=V=1&S=nextcloud,jitsi`

k4bek4be:
> `EXTJWT=V:1&S:nextcloud,jitsi`

Any explanation for this example, please?

slingamn:
+1 for @RyanSquared's suggestion (or some variant of it). It definitely seems useful to have the named services be discoverable.

k4bek4be:
I have created an implementation for the EXTJWT command. I'm supporting HS, ES and RS type tokens, with method and key/password selectable per service. Users on our network are already using the `HS384` method.

The specification does not mention any error handling. That's what I've devised:
- No arguments: send `ERR_NEEDMOREPARAMS`.
- The first parameter is neither an existing channel name nor `*`: send `ERR_NOSUCHNICK`
- The second parameter does not match any configured service name: send `:irc.example.com FAIL EXTJWT NO_SUCH_SERVICE :No such service`
- Token generation failed for some other reason: send `:irc.example.com FAIL EXTJWT UNKNOWN_ERROR :Failed to generate token`

The `vfy` claim can be specified in configuration, separately for each service, and the IRC server is not handling the verification, requiring administrator to set up some separate verification service.

k4bek4be:
This draft is now shipped as optional feature with unrealircd 6.0.0.

jwheare:
Closed in favour of #547

PR #342, opened 2018-10-30
websocket spec

prawnsalad:


jwheare:
Nice one. It'd be worth fleshing out advice on converting to UTF8 and managing invalid data from other clients connecting to the network. E.g. using the replacement character or treating all invalid data as latin1, or what have you.

extensions/websocket.md
```diff
@@ -0,0 +1,38 @@
...
+
+Because web browsers are pushing more towards a UTF8 default encoding for HTML5 documents and scriptings languages, and these web browsers are the primary use case for websockets, IRC servers MUST encode messages as UTF8 before sending them to the client. Ignoring this may result in the browser closing the websocket connection on invalid UTF8 sequences.
+
+## Message syntax
+
+The syntax of IRC lines MUST NOT be changed. However, IRC servers and clients MUST NOT include trailing new lines in their websocket messages as these are obsolete by websockets native messages.
+
+## Client example
+~~~
+let socket = new Websocket('https://ws.example.org');
```
SaberUK:
```suggestion
let socket = new WebSocket('wss://irc.example.org:6698');
```

extensions/websocket.md
```diff
@@ -0,0 +1,38 @@
...
+    email: darren@kiwiirc.com
+---
+
+## Introduction
+
+Web based IRC clients are becoming more common and they all increase the demand for websocket connectivity to IRC networks.
+
+This specification provides a common implementation for web and other websocket clients to connect to IRC networks using common and standard development tools used on web pages, with a focus on a simple client facing interface.
+
+## Websocket features and encoding
```
SaberUK: Nit: WebSocket not Websocket

extensions/websocket.md
```diff
@@ -0,0 +1,38 @@
...
+
+Websockets offer different transports of communication, binary and strings. With the consideration that the IRC protocol is entirely lines of text, we make use of the message based text implementation that websockets natively provide. These are UTf8 encoded text messages where one websocket message consists of a single complete IRC message line.
+
+The internal websocket protocol handles message fragmentation and termination via frames, which is out of scope for this document.
+
+Because web browsers are pushing more towards a UTF8 default encoding for HTML5 documents and scriptings languages, and these web browsers are the primary use case for websockets, IRC servers MUST encode messages as UTF8 before sending them to the client. Ignoring this may result in the browser closing the websocket connection on invalid UTF8 sequences.
```
grawity: > IRC servers MUST encode messages as UTF8 before sending them to the client.

An IRC server can determine that an incoming message isn't valid UTF-8, but how is it supposed to know the actual original encoding to translate from?

IIRC this was the primary argument for using binary mode with IRC, rather than text mode. (It seems that's what [InspIRCd already does][1].)

[1]: https://github.com/inspircd/inspircd/blob/master/src/modules/m_websocket.cpp#L356

prawnsalad:
> How should a client discover that a server supports websocket? Should it try to connect, and if failed, fallback to plain IRC? Maybe some CAP which advertises that the server supports it? Or a DNS record?

@DarthGandalf Web clients can't connect to plain sockets so it needs to know the WebSocket address up front. And clients that connect via a plain socket has no need to discover if a WebSocket option is available as they are already connected.

extensions/websocket.md
prawnsalad: This is a topic probably worth mentioning in the spec.

Personally I'm in favour of sticking to standard web ports 80/443 and heavily suggesting this specifically. Mostly because browsers block some ports in the IRC range and not all net access points allow connections outside of the usual web ports when behind a proxy of some sort (hotel, corp environment). This could unknowingly prevent some clients from connecting at all while others are fine. The can be prevented all together by sticking to 80/443.

Then for the subdomain, we could also suggest one (`ws.example.org` would be good as many use this format already) but networks are free to choose one as they see fit.

SISheogorath: Staying with 443 in the example. The RFC itself says that 443 is the default port for secure websockets: https://tools.ietf.org/html/rfc6455#section-1.7

For the dns name, I actually don't care.

extensions/websocket.md
Herringway: I would expect a configuration option for fallback encoding, because most encodings are impossible to detect.

Or possibly the option to not detect the encoding and just filtering/replacing the illegal characters

ProgVal: Can't we just ban non-UTF-8 from the IRC protocol once and for all?

dwfreed:
For a reverse proxy adding WebSocket support to an ircd that doesn't have it, as well as a client that wishes to support both plain socket and WebSocket transport, this isn't ideal.

A reverse proxy would have to scan the lines coming from the ircd to strip the CRLF and frame the WebSocket messages appropriately; this would noticeably increase the CPU demands of the reverse proxy for no real benefit that couldn't be done on the client side.

A client that wishes to be transport-agnostic would have to have special handling for WebSockets, rather than being able to treat them as a normal socket for most intents and purposes (and most WebSocket implementations provide an interface that is compatible with the language's stdlib socket implementation, so using WebSockets would be mostly a drop-in replacement).

WebSockets should just be an alternate means of transporting IRC lines, and not an ever-so-slightly different protocol.

syzop:
First some history of my take on websockets, since occasionally some people think I am against broad websocket adoption, while I have in fact been one of the main proponents. I think this will help to understand my position, Kiwi's position and how we came to the present day situation. After this comes another post which is more technical regarding this draft proposal.

In January 2017 the first UnrealIRCd stable version shipped with websocket support. At that time I thought this was a feature that would be "big". It would allow webclients to connect easily to irc servers. I envisioned a case where admins would not have to configure anything and it would "just work" from any browser. All this without WEBIRC or specific allow block configuration and without websocket-to-IRC gateways in-between.

There had already been an InspIRCd module written by a 3rd party module writer, even going back to 2014 if I'm not mistaken, so credit goes to him for being first. But no major IRCd shipped with websocket support at that time, and I think even until May 2019 when InspIRCd 3 was released.

So I thought this would be "big", but things remained silent for half a year...

Then, @prawnsalad approached me in August 2017 and we talked about a number of technical details. First of all, and this is very similar to what @dwfreed touches on, I always considered websockets to be an extra layer wrapped around the IRC protocol. So it did send \r\n and such in the socket. And theoretically, or I thought so anyway, a complete websocket frame could even contain half of an IRC message (without \n) only to be continued in a next websocket frame, or even two IRC messages in one websocket frame (eg: ```:n!u@h PRIVMSG #test :hi\r\n:other!user@host QUIT :..\r\n```).
On the request of @prawnsalad I ensured that each websocket would always be 1 line and I changed the implementation that the traffic would no longer end in \r\n. This all to make things easy for client developers.

Then later on - or actually this happened the same day - came the other discussion about handling of invalid UTF8 in IRC traffic. For those that do not know the problem it is as follows:
IRC accepts a stream of bytes, terminated by the \n command, which is considered 1 line, or 1 command. It is worth to mention that IRC predates UTF8 (even RFC1459 was published only months after UTF8 was first presented). At various places these relaxed rules and allowing of non-UTF8 traffic mean that you can use certain character sequences that are "invalid UTF8 sequences". I will touch on more details in a later post, but first:
This is really a problem for websockets since the websocket RFC645 states (if you use the TEXT frame type):

> 8.1.  Handling Errors in UTF-8-Encoded Data
> 
>    When an endpoint is to interpret a byte stream as UTF-8 but finds
>    that the byte stream is not, in fact, a valid UTF-8 stream, that
>    endpoint MUST _Fail the WebSocket Connection_.  This rule applies
>    both during the opening handshake and during subsequent data
>    exchange.

So that means a malicious (or non-malicious) client writing an invalid UTF8 sequence in a message to a channel will result in all websocket users in that channel to be disconnected. Not good.
The 3rd party InspIRCd websocket module at that time and the UnrealIRCd module (fortunately before the stable release in Jan 2017) recognized this problem and changed the frame type from TEXT to BINARY to make sure this does not occur. This because the BINARY frame allows you to send any binary data, including "invalid UTF8".
Now, this does mean (among other things) that on the client these invalid UTF8 sequences need to be handled.

So it is clear that there can be invalid UTF8 sequences in general IRC traffic, the question was who should do this sanitizing, and that is where @prawnsalad and I disagreed at the time.
He had arguments why it would be best for the server to handle this and I had arguments of why it would be best for the clients to handle this. I don't think any of the two sides are wrong here. What also came in to play, or placed a significant role, is the options and limitations we both had (server and client) to handle the problem/sanitizing/conversion.
Just as a reference, IRC web clients CAN do the conversion, as for example this client shows: http://bramka.pirc.pl/bramka/
Similarly, IRC servers CAN do the conversion, as for example InspIRCd 3 does.

Anyway, in the end we were stuck with a situation where KiwiIRC, which I think looks like a nice client, could not communicate with UnrealIRCd websockets because of the decision by developers on both sides. This was really disappointing for me, and I think to the Kiwi devs too. I had this vision of websockets being used and booming, and Kiwi being a major web client not being able to communicate over websockets with UnrealIRCd, hindered this adoption significantly.

I will do a post later this morning to explain my thoughts and the effort I did recently regarding implementing the UTF8 sanitizing in UnrealIRCd (so, on the server side).

syzop:
In February of this year I was helped by someone with writing a sanitize UTF8 routine for incorporation in UnrealIRCd. This would filter out "invalid UTF8 byte sequences" with a replacement character, similar to what InspIRCd 3 does. To be honest I don't yet know if that is sufficient for websocket client developers or if they would need more than that. But, when working on that, I noticed that this would lead to a number of issues, some more serious than others, that do not seem to be mentioned anywhere.

Perhaps before we start:

- An example "invalid byte sequence" would be the bytes ```0xC3 0x28```
- The "UTF8 replacement character" would be U+FFFD which in bytes is ```0xEF 0xBF 0xBD```

Now, some examples of the problems that arise with sanitizing, starting with a low-impact example:

**Example 1: channel messages**
It could be that someone is talking in some non-utf8-charset in a channel (say, Cyrillic in some of the many Cyrillic codepages that exist), then utf8 capable clients would have no idea what this person is saying. This is a normal limitation and well, it does not cause a security issue or problematic handling of commands. It's just the reading by the end-user that is a problem. Garbled text or invalid byte sequences.. there would be little difference between the two: in both cases the text was unreadable text and it is still unreadable text for the end-user. This is sortof already discussed here in this thread and elsewhere. Some people have even proposed automatic character set detection and conversion. I don't think that's a viable option to be honest and it will always be guesswork anyway, but okay.. Anyway, this 1st example is not really the topic of what this post is about. It is just to illustrate that this is the least of my concerns at the moment.

**Example 2: Channels with invalid utf8 sequences**
Most IRC servers allow channels to contain "invalid utf8 byte sequences". Say, JOIN #test followed by the two bytes 0xC3 and 0x28.
The problem is that for a websocket user, at least if the server does the filtering, then it will see a join to #test followed by the "replacement character" (which is not 0xC3 0x28).
Now, you may wonder how a websocket user may end up in such a channel in the first place, if they can never send the JOIN. Well, it can happen, for example by auto-join or by a channel redirecting to another channel (channel redirects in some shape or form are widely implemented amongst irc servers).
A specific example would be:
- Create a channel #welcome and a channel #fun + 0xC3 + 0x28.
- Make a channel link, for example ```MODE #welcome +lL 1 #fun + 0xC3 + 0x28```
- Invite the websocket user or ask them to join #welcome  
- The websocket user will be joined to ```#fun + 0xC3 + 0x28``` due to the redirect.
- The websocket user can now never part this channel, since their PART would be not for the channel ```#fun + 0xC3 + 0x28``` but for ```#fun + 0xEF + 0xBF + 0xBD``` which is not the same channel.
Or:
- It can even become more fun if you have a websocket client in two of such channels with invalid byte sequences that resolve back to the same identical sanitized string. Then the client sees joins that could be duplicates, parts of people that were never there, and so on.

**Example 3: Nick names containing invalid utf8**
Some IRC servers may allow international nick names in a character set different than UTF8. The attack here is similar to the previous example but is even more severe:
- Say, channel #welcome is a channel where the websocket client has ops in
- Then someone joins with a nick with an "invalid utf8 character sequence"
- This user can then never be kicked by the op with the websocket connection, since their kick command would use the replacement character (it doesn't know any better) and not the real byte sequence of the nick.
Or:
- Similar to example 2, two nicks may end up resolving to the same sanitized string, which may result - again all this in the eyes of the websocket user - in double joins and then if the first one of those users parts you will have a ghost user: the other one is still in the channel but is not in your nick list.

**Example 4: ban lists**
A ban, except or invex is usually only loosely checked by IRCds and may be permitted to contain invalid utf8 byte sequences. These would be bans/exempts/invexes that the websocket user would be unable to remove since the command to unset would, again, be for a different entry since it does not (and cannot) know the original byte sequence.

**More?**

There may be other places. In the end, it all comes down to the websocket user having a different "view" so to say, compared to other users. When doing the "cleaning" of invalid utf8 sequences on the server, they loose the ability to send commands back with those (invalid) byte sequences, whether it's a PART, a MODE, or something else.

So does this bring back the whole BINARY vs TEXT frame argument? I don't know.
I can only say that at least theoretically, when sending BINARY websocket frames (like UnrealIRCd currently does), the client at least theoretically still has access to the original data and could do the PART, MODE, KICK or whatever command with the correct byte sequence.
Perhaps in practice they may end up with the exact same situation / problem. This is where client devs should chime in, since I don't have much experience with websockets on the client side, or javascript at all for that matter.

jwheare:
Thanks for writing this up.

Can 2 and 3 not be prevented on the IRC server by restricting valid nick/channel characters, something that as far as I'm aware is already done by 99% of irc servers.

4 seems minor, the person who set the ban can presumably remove it, and again, if nicks are restricted this is probably not an issue.

That being said, I do think that a binary websocket is probably the better solution, since it doesn't lead to data simply being thrown away irretrievably. Clients can decode binary and decide how they want to deal with encodings and perform sanitisation/substitution as they please. If they want to reject invalid UTF8 they can, but if they'd rather allow it to be handled, they aren't prevented from doing so by the server.

However, I have no real intention to implement websocket support in my client.

jwheare:
Alternatively, allow a text websocket interface only on servers that force valid UTF8 everywhere.

SaberUK:
InspIRCd v3 defaults to UTF-8 being required although we have an option to opt-out of that behaviour.

syzop:
> Can 2 and 3 not be prevented on the IRC server by restricting valid nick/channel characters, something that as far as I'm aware is already done by 99% of irc servers.

Indeed that can be prevented by that measure, so that is feasible.

Regarding the figures (%) for channels: I don't know about InspIRCd (ah @SaberUK just commented, it depends on the version and settings), but UnrealIRCd allows it and ircd-hybrid as well when I read the code correctly. So, translating that into servers (via ircstats.org) that means 53-71% of the servers. Regardless of the exact figure: a lot of servers.

> That being said, I do think that a binary websocket is probably the better solution, since it doesn't lead to data simply being thrown away irretrievably. Clients can decode binary and decide how they want to deal with encodings and perform sanitisation/substitution as they please. If they want to reject invalid UTF8 they can, but if they'd rather allow it to be handled, they aren't prevented from doing so by the server.

Yeah, that's my general thought as well. It's best not to throw away information in general. But.. yeah. Let's hear from others :)

> Alternatively, allow a text websocket interface only on servers that force valid UTF8 everywhere.

Again, true. Unfortunately I think it's too early to limit IRC to UTF8 only. I think - but that is a guess based on no particular data - many people would not do that and thus it would seriously limit/hinder websocket adoption, while I in fact would like to see websockets grow a lot :)

I'll let others comment for a while before I respond again. Only fair and I probably wrote too much already for today ;)

prawnsalad:
A good writeup. Since the insp3 release it has spurred up a lot of websocket client dev activity, both for using kiwiirc and interfacing directly using an irc lib or directly.

So the main points I had from ~a year ago that would make binary websockets more difficult to use.
1. Web devs (which are mostly the people that are going to be using this) tend to get tripped up a lot when it comes to binary messages, while receiving a text message avoids all that headache.
  While this is simply the case that the devs just need to learn how to handle it - most don't. This is proven by the fact that Unreal has had support for a while now and I've seen many people trying to use it directly and getting tripped up so they just use something else instead. While with inspircds text based websocket it just works how they expect and they go on their way and use it. This barrier will never cause the websocket feature to be big as you hoped.
  (I don't mean to talk most web devs down here... but it's the reality.)

2. Websockets will almost always be used on a HTML5 page which defaults to UTF8 encoding (for example, the entire Kiwi project is UTF8 encoded along with all it's translations). If we decode a message to an incompatible encoding then displaying that message on the page may cause broken texts. Mixing encodings isn't that great on the page document. We could of course try to reencode to match the pages encoding, but then this limits browser support and.. following onto point 3...

3. Performance of switching encodings in the browser is awful. It is already difficult to keep performance at a suitable level while parsing IRC messages and it's formatting, and adding the reencoding on top of that will start causing browsers to lock and battery life to diminish because of the fact. This is then blamed on the client for being slow as users don't realise what's actually needed to parse their messages technically.

Points 2 and 3 are the main issues as to why I haven't implemented support in Kiwi personally, though I have always mentioned to people that ask about it that I'm willing to accept PRs even if I wouldn't enable it myself. This is also the reason why I don't think it should be available in the spec as it would just make IRC from a users point of view bad.

Simultaneously accepting/receiving binary and text messages would be fine as long as both are expected to be UTF8 encodings, that way the client can decode incoming binary messages and ignore the failed decodings without having to try other encodings, or just use the text it receives if a text message.

syzop:
It seems like I misinterpreted @SaberUK comments earlier, so this is a real issue today with Kiwi and Insp3? (at least the 2nd and 4th example)

syzop:
Thanks, @prawnsalad for your take on this. I am seeing this mostly as a response to my 1st post here. It's good for everyone to see the concerns of client devs.

Could you respond to my 2nd post? I mention a couple of examples there that are caused by the filtering on the server-side. (And what you write does not resolve these, so therefore I'm assuming you were only responding to the 1st post)

It's a bit long read, I realize, no immediate hurry ;)

finalhit:
@syzop how about implementing the same thing:

> InspIRCd v3 defaults to UTF-8 being required for WebSocket users although we have an option to opt-out of that behaviour and send binary frames instead of text frames. We also handle receiving both binary and text frames.

Plus filter out invalid UTF8 byte sequences server side.

extensions/websocket.md
```diff
@@ -0,0 +1,38 @@
...
+
+Websockets offer different transports of communication, binary and strings. With the consideration that the IRC protocol is entirely lines of text, we make use of the message based text implementation that websockets natively provide. These are UTF8 encoded text messages where one websocket message consists of a single complete IRC message line.
+
+The internal websocket protocol handles message fragmentation and termination via frames, which is out of scope for this document.
+
+Because web browsers are pushing more towards a UTF8 default encoding for HTML5 documents and scriptings languages, and these web browsers are the primary use case for websockets, IRC servers MUST encode messages as UTF8 before sending them to the client. Ignoring this may result in the browser closing the websocket connection on invalid UTF8 sequences.
+
+## Message syntax
+
+The syntax of IRC lines MUST NOT be changed. However, IRC servers and clients MUST NOT include trailing new lines in their websocket messages as these are obsolete by websockets native messages.
```
hhirtz, line 28: > IRC servers and clients MUST NOT include trailing new lines in their websocket messages 

I disagree with this, because it would conflict with RFC 2812 and RFC 1459.

Those RFCs state that IRC messages MUST end with `\r\n`, and that IRC is agnostic of the transport layer used (right now most use TCP, but WebSockets are allowed too).  This was made so IRC client libraries can be transport agnostic, and don't have to deal with special cases like messages over WebSockets not ending with `\r\n`.

Ideally WebSockets should just be another mean of transportation, and not an entirely new specification, as dwfreed said.  Clients and servers should threat WebSocket payloads as parts of a stream of data, like you'd do with TCP segments (not dealing with the text vs binary issue here).  This is what I think the RFCs had in mind anyway.

hhirtz: Please note that I'm not against a "SHOULD" recommendation against putting several messages, or half a message in a WebSocket payload, but please let every IRC message be ended with their beloved `\r\n`.

jwheare: The new lines are used with TCP to mark the end of the line. Websockets have their own framing that renders that unnecessary and they’d just have to be stripped out before further processing anyway.

There’s already precedent for modifying the RFC message format in new extensions, I don’t see why CRLF is so special that it needs to be retained.

SadieCat: >This was made so IRC client libraries can be transport agnostic, and don't have to deal with special cases like messages over WebSockets not ending with \r\n.

You don't need to deal with anything if you handle your line splitting separate from your message parsing and only implement it in your TCP transport layer.

hhirtz: Hey!

> The new lines are used with TCP to mark the end of the line. 

That's where I disagree.  Newlines are used to mark the end of an *IRC message*.  Saying "*WebSockets already have their own framing, so when I implement IRC on top I don't have to put `\r\n` at the end*" is like saying "*my VPN-based layer 2 protocol have its own thing, so when I implement the Internet Protocol (layer 3) on top of it I don't have to put the destination address in the header*".

Maybe the example doesn't make much sense, but the point is to not mix the transport layer (TCP, WebSockets) with the application layer (IRC).

prawnsalad: > the point is to not mix the transport layer (TCP, WebSockets) with the application layer (IRC).

The RFCs mixed transport and application layers together as it was such a simple protocol back then. The application layer does not require `\r\n`, this is used as part of the transport to separate messages - framing. `\r\n` adds no meaning to a complete message.

emersion:
Should we register an IRC sub-protocol as per https://tools.ietf.org/html/rfc6455#section-11.5?

prawnsalad:
Is there anything else needed for this to be merged now that IRCDs have settled+implemented this?

k4bek4be:
I think it should be pointed out in the spec that both server and client MUST accept both binary-type and text-type websocket frames (regardless of what type it has decided to send) to be compatible with existing implementations.

I was surprised by Oragono sending text packets to me even though i've been sending binary, and I have since then modified the client to accept both.

prawnsalad:
@k4bek4be the spec outlines why simple text websocket messages are being used. All implementations that I'm aware of are compatible with this doc, if there are some that don't then it would be good to mention them so someone can reach out to them?

k4bek4be:
I am using binary-type messages on our network (server and client), because the network supports non-websocket connections too. Limiting websocket clients to valid UTF-8 would make some content created by non-websocket users unreachable for these clients (for example chanops using websockets would not be able add text bans for some invalid-utf8-spam), and we do not want to support users encountering such problems.
Of course someone can connect their own client sending text-type messages and I'd be happy to allow it. But in such cases I'll direct any complaining users to the client's developer. And, undoubtly, that client must understand received binary frames too to work at all.

prawnsalad:
Assuming that the client webpage is UTF8 encoded as this is the default and highly pushed by web bodies for some time now, if the client receives a different encoding via binary websocket then the client would have to try to re-encode it to UTF8 anyway for it to be displayed or handled anywhere sensibly. If it tries to display it in the client then the page risks becoming malformed and the document then has undefined behaviour - including a risk of injecting executable code.

Now if a web client has to re-encode every message it receives which is a relatively heavy operation in javascript this then causes an unavoidable resource drain causing battery life issues on user devices. It is far more efficient for the IRCd to ensure valid UTF8 messages than impact all users this way.

From a usability standpoint, web devs getting their feet wet with implementing a simple IRC client on their webpage will generally not have had to deal with text encodings before since the web has moved faster than IRC and settled on UTF8 for the most part. So they would then need to learn and be aware about encodings, how to ensure UTF8 strings and why they shouldn't display invalid sequences otherwise they risk security issues on their page without even knowing about it.

From a calendar standpoint, it's 2020.

It's entirely your choice if you only want to support users that will display all data sent to them in their client, but you would be putting them at a security risk if you do so.

emersion:
Is there anything else that still needs discussion?

slingamn:
Should we have some recommendations (possibly non-normative) about interoperability?

For context, Oragono currently has two modes of operation (that can be toggled in the config file). One (the recommended default) preemptively rejects non-UTF8 lines that come in from non-websocket clients (sending, e.g., `FAIL PRIVMSG INVALID_UTF8`). The other accepts those messages, then silently drops them from being relayed to websocket clients.

emersion:
>One (the recommended default) preemptively rejects non-UTF8 lines that come in from non-websocket clients (sending, e.g., FAIL PRIVMSG INVALID_UTF8). The other accepts those messages, then silently drops them from being relayed to websocket clients.

FWIW, soju replaces invalid UTF-8 characters with the Unicode replacement char `\uFFFD` when relaying messages to WebSocket clients.

slingamn:
I don't think we should specify the use of the replacement character, but I think we should include language that mentions it as a possibility.

emersion:
This new version looks good to me.

slingamn:
Next steps on this?

jwheare:
Can you sum up a list of who has implemented this and if there are any differences between implementations, binary vs text etc?

hhirtz:
non-exhaustive list of impls (those i know of)

Severs:

- [oragono] drops non-utf8 messages
- [soju] replaces invalid characters with U+FFFD
- [unrealircd] replaces invalid characters with U+FFFD

Clients:

- [gamja]
- [kiwiirc]

[oragono]: https://github.com/oragono/oragono/
[soju]: https://sr.ht/~emersion/soju/
[unrealircd]: https://github.com/unrealircd/unrealircd/
[gamja]: https://sr.ht/~emersion/gamja/
[kiwiirc]: https://github.com/kiwiirc/kiwiirc

prawnsalad:
+ inspircd, + kiwibnc, + webircgateway if that one counts

slingamn:
>Can you sum up a list of who has implemented this and if there are any differences between implementations, binary vs text etc?

Just checking that we're all on the same page: this version of the spec mandates the use of text messages. Any implementation using binary messages is non-compliant and incompatible.

dwfreed:
> > Can you sum up a list of who has implemented this and if there are any differences between implementations, binary vs text etc?
> 
> Just checking that we're all on the same page: this version of the spec mandates the use of text messages. Any implementation using binary messages is non-compliant and incompatible.

That's the point; he wants to know what implementations already exist that are incompatible.  I'll reply later with my evaluation of the two websocket server implementations I have to deal with: solanum (and by extension charybdis) and OFTC node-ircws.

SadieCat:
InspIRCd sends text frames by default (replacing invalid UTF-8 with the replacement character) but has a config option to send binary frames instead (which are identical to that which is sent to normal IRC users).

dwfreed:
As of this writing, this is the state of compliance with the spec in this PR with the software I have to deal with:

solanum/charybdis (no changes in WebSockets since forking):
* Always sends text frames
* Always sends one IRC protocol message per WebSocket frame
* [non-compliant] Sends IRC protocol messages as-is, without ensuring they are UTF-8 clean (Latin1 goes through as Latin1, browser kills the socket)
* [non-compliant] Sends \r\n to clients, requires client to send at least \n just like a TCP connection

OFTC node-ircws:
* Always sends text frames
* [non-compliant] May send multiple IRC protocol messages in a single WebSocket frame
* Ensures everything encodes to UTF-8, uses replacement character for invalid UTF-8 bytes
* [non-compliant] Sends \r\n to clients, requires client to send at least \n just like a TCP connection

Within solanum, we've discussed removing the built-in websocket support and recommending using something to handle the conversion between WebSockets and regular sockets.  No final decision has been made on that yet.

jwheare:
Yeah I just wanted to make sure there weren't any outstanding compatibility issues. binary vs text was the main one a while ago, and I wasn't sure if that had been resolved in a way that everyone was on board with. Seems like it has been.

@dwfreed the newline incompatibilities are interesting though. Seems like clients will have to strip/split on newlines within frames and always send a newline within a frame to be compatible with server implementations.

That directly contradicts the spec. Are those non-compliant servers going to fix those issues or or do we need to relax the spec?

slingamn:
I'm bearish on relaxing the property that frames must contain exactly one IRC line, or on requiring clients to send newlines.

I'm more sympathetic to a Postel relaxation of the current language on newlines (i.e., allowing servers and clients to send terminating newlines with the frame, but not allowing either to require them to be present). (But it seems like this may not significantly help anyone?)

SadieCat:
>allowing servers and clients to send terminating newlines with the frame, but not allowing either to require them to be present

This breaks compatibility with implementations which follow the spec as-is.

slingamn:
I'd prefer to ratify the spec as written. Questions:

1. Is node-ircws https://github.com/tjfontaine/node-ircws exactly, or does OTFC have a fork?
1. What clients are currently deployed to interoperate with charybdis/solanum and node-ircws? Would they require changes to bring them into compatibility with this spec?
1. What level of incompatibility in the wild is a release blocker for this spec?

jwheare:
Incompatibility in the wild might signal that the spec isn't fit for purpose for some implementors. That's why I asked if those non-compliant servers planned to fix it or if they implemented it incorrectly for any specific reason.

slingamn:
Here's my understanding:

1. There is a nontrivial limitation imposed by this spec: only UTF-8 is supported. In the past, people have expressed doubts about whether this limitation was acceptable. However, it appears that debate has largely converged in favor of requiring UTF-8.
2. Incompatibilities like `\r\n` in frames, multiple lines in frames, etc. do not reflect limitations imposed by this spec; it's just that in the absence of a ratified spec, we ended up with a plurality of behaviors in the wild.

slingamn:
We've identified two noncompliant server implementations. Does anyone know what the client implementations are that interoperate with them?

prawnsalad:
I'm not sure specifically which ones but there were a few basic websocket<>tcp gateways built years back and added to networks hoping that clients would use them, but they just never got used or updated as implementations started to come together. Probably worth keeping in mind.

slingamn:
Looks like OTFC's client is based on https://github.com/qwebirc/qwebirc/ ?

RyanSquared:
| Software | Incompatibilities with the specifications |
| -------- | ----------------- |
| Oragono | Has two modes for UTF8: One FAILs, one drops to websocket clients |
| InspIRCd | Accepts binary frames, can be configured to send binary frames |
| Solanum | Does not ensure that IRC messages are UTF-8 clean, sends \r\n to clients, requires \n from clients |
| node-ircws | Can send multiple IRC messages per frame, sends \r\n to clients, requires \n from clients |

I believe all of the above are incompatible with the current specification. I think we need to merge *some* kind of specification now or we're going to have more incompatibilities in the future due to lack of specification.

DanielOaks:
Yeah I think we're (Ora) just gonna kill non-utf8 entirely, or at least kill it while a websocket listener is enabled.

RyanSquared:
Should be worth noting in some kind of documentation that enabling websockets means enforcing UTF-8.

SadieCat:
> InspIRCd 	Accepts binary frames, can be configured to send binary frames

Accepting binary frames as well as text frames is not incompatible with the spec. The default behaviour for sending frames is also spec compliant.

RyanSquared:
> we require that all IRC messages MUST be sent as WebSocket text messages, where each message consists of a single complete IRC message line.

I thought this meant servers could only send WebSocket "text" messages, and that binary frames are not allowed.

SadieCat:
>I thought this meant servers could only send WebSocket "text" messages

This is the default behaviour. You specifically need to set an option to make it non-spec compliant which is clearly marked as such.

slingamn:
I maintain that the (non-default) Oragono behavior of accepting non-UTF8, then relaying it only to non-WebSocket clients, is compliant with the spec as written.

The question remains whether the spec should be changed to disallow this. @RyanSquared suggests that if a message is delivered to non-WebSocket clients, WebSocket clients MUST receive some indication that it was sent (e.g., a version of the message using the Unicode replacement character). I don't really agree but I have no strong objections either, so I'm willing to go with whatever the consensus ends up being.

slingamn:
@dwfreed IIRC you had some more thoughts on this?

dwfreed:
Sorry, life has been super busy lately.

OFTC's current webchat uses https://github.com/oftc/node-ircws/ for the IRC to WebSocket/socket.io conversion, and https://github.com/oftc/qwebirc-websocket/ as the frontend, which is a patched version of qwebirc to add socket.io support.

OFTC is currently working on a replacement for node-ircws, and we're nearing completion.  It is intended to be a transparent transport of IRC, as node-ircws is, but takes it a step further and does not attempt to make non-UTF-8 messages UTF-8 clean.  It has the same incompatibility properties as node-ircws with this PR, and adds one more: if a chunk contains any character that is not valid UTF-8, that chunk will be sent as a binary message.  (It will be doing strictly WebSocket, whereas node-ircws also does socket.io.)

The primary design goal of the project was to do as little work and store as little state as possible, and be as transparent as possible.  Cleaning non-UTF-8 cannot be performed losslessly, and this has impacts on communities that choose to use their traditional encoding instead of UTF-8.  We believe the client should be able to choose how it wishes to handle this situation, opening up the opportunity to allow the end user to choose how they wish to handle it on a per-target basis.

The plan is to patch either the qwebirc or kiwiirc frontend to support the new WebSocket gateway.  We will also be submitting the completed project to solanum, and if approved by the freenode half of the team, it will become the recommended WebSocket gateway to use with solanum and solanum's own WebSocket support will be removed. (There are issues with it internally such that just killing it is honestly the better solution.)

prawnsalad:
@dwfreed just a heads up that kiwiirc natively supports plain websocket out of the box without any patches. Just set `startupOptions.websocket: '/path/to/ws/server'` in the config.

Unless you want to parse the binary messages that is, which then you would need to incorporate UTF8 parsers and replacement characters otherwise you end up with security issues.

dwfreed:
@prawnsalad you know browsers can do decoding of WebSocket binary messages already, right? FileReader API readAsText method accepts a Blob just the same as a File.

prawnsalad:
@dwfreed just to confirm, if freenode staff approves of your implementation as you suggest, thats what would be going live on the freenode network despite what happens with this draft? Apologies if I misunderstood this

dwfreed:
I don't know what would happen with freenode the network.  Their approval insofar as solanum goes would only affect what the solanum project suggests people use for a websocket gateway.  They would be free to run whatever they like for their own needs (but if they approve it, it wouldn't surprise me if they ran it as well at some point in the future; dogfooding and all that).

slingamn:
>It has the same incompatibility properties as node-ircws with this PR, and adds one more: if a chunk contains any character that is not valid UTF-8, that chunk will be sent as a binary message.

I'm still thinking about the compatibility implications here (my initial reaction is that they are very alarming, possibly enough to exclude the possibility of a consensus spec), but I wanted to ask up front: how committed are you to sending multiple IRC messages per WS frame?

hhirtz:
What is the use-case of binary messages?

Ultimately, WebSockets messages are to be shown on a web page, which only supports UTF-8. Therefore, there will always be a binary->text conversion, be it on the client (which i suppose replaces invalid chars with a replacement char) or on the server (which may do what it wants).

Sorry if this is has already been answered.

slingamn:
We had an extensive discussion about this in #ircv3. Here's my understanding of the state of play:

1. Some people want to be able to use binary messages (for lossless transport of non-UTF8 encodings)
2. Some people want to be able to use text messages (for simplicity of implementation)
3. Any proposal that requires one or the other, or even makes one mandatory and the other optional, is unviable
4. Any compromise proposal will have to allow implementations to choose one or both (consequently, it will be possible for two implementations to comply with the spec and be mutually incompatible, because one only speaks binary and the other only speaks text)

The proposed mechanism for this compromise is [WebSocket subprotocols](https://developer.mozilla.org/en-US/docs/Web/API/WebSockets_API/Writing_WebSocket_servers#subprotocols). Clients will connect to the server and use the `Sec-WebSocket-Protocol` header to request at least one of the `ircv3-text` and `ircv3-binary` subprotocols (if the client supports both, it sends its preferred subprotocol first). If the server supports at least one of the requested subprotocols, it will reply with the preferred supported protocol. If it supports neither, it will reply without a `Sec-WebSocket-Protocol` header.

If this negotiation process succeeds, all messages sent MUST be of the negotiated type (text for `ircv3-text`, binary for `ircv3-binary`). As a transition mechanism, if the client does not request a subprotocol, the server MAY implicitly assume a preferred subprotocol and send messages of that type. Similarly, if the server replies without a `Sec-WebSocket-Protocol` header, the client MAY implicitly assume a preferred subprotocol and send messages of that type. If either side receives a message of unexpected type, it SHOULD disconnect.

As for framing and line termination:

1. If `ircv3-text` is negotiated (explicitly or implicitly), then each WebSocket message MUST contain exactly one IRC message, without the terminating `\r\n`
2. If `ircv3-binary` is negotiated (explicitly or implicitly), then each WebSocket message MUST contain exactly one IRC message (I'm not sure which way to go on `\r\n` here)

SadieCat:
I'm happy to implement WebSocket subprotocol support.

emersion:
Using subprotocols sounds OK to me.

Bikeshedding:

> To avoid potential collisions, it is recommended to use names that contain the ASCII version of the domain name of the subprotocol's originator.

Maybe we should name the subprotocols something like `text.ircv3.net` and `binary.ircv3.net`.

jwheare:
> Maybe we should name the subprotocols something like text.ircv3.net and binary.ircv3.net.

WFM (for the record that quote is from https://tools.ietf.org/html/rfc6455#section-1.9)

Also of note for future reference:

> Subprotocols can be versioned in backward-incompatible ways by changing the subprotocol name, e.g., going from "bookings.example.net" to "v2.bookings.example.net".  These subprotocols would be considered completely separate by WebSocket clients.  Backward-compatible versioning can be implemented by reusing the same subprotocol string but carefully designing the actual subprotocol to support this kind of extensibility.

jwheare:
We should also definitely register the names with IANA once ratified.

slingamn:
I've adopted this suggestion and edited the comment to reflect it.

slingamn:
OK, so, the thing that's really bothering me about the proposed compromise is that there is no baseline standard of interoperability: two clients can comply with the spec and not be able to talk to each other. Here's a counterproposal:

1. Servers MUST support both text and binary frames. This does not require any subprotocol negotiation; you can accept both on the same socket. Message interpretation is unambiguous on the server side. (Servers that support, e.g., #432, are still free to reject binary frames that happen to contain non-UTF8. Servers that support non-UTF8 will be required to munge outgoing lines sent as text frames, probably using the replacement codepoint, but this work can be offloaded to a gateway.)
2. Client implementations pick either text or binary, then use that, without negotiation.
3. Servers and clients MUST NOT send `\r\n` (so current text-based clients remain correct)

Thoughts on this? If it has sufficient buy-in here I'll try and take this back to OFTC.

slingamn:
OK, so, the thing that's really bothering me about the proposed compromise is that there is no baseline standard of interoperability: two clients can comply with the spec and not be able to talk to each other. Here's a counterproposal:

1. Servers MUST support both text and binary frames. (Servers that support, e.g., #432, are still free to reject binary frames that happen to contain non-UTF8. Servers that support non-UTF8 will be required to munge outgoing lines sent as text frames, probably using the replacement codepoint, but this work can be offloaded to a gateway.)
2. Client implementations negotiate either `text.ircv3.net` or `binary.ircv3.net` as a subprotocol, then send and receive frames of that type. In the absence of subprotocol negotiation, the default is text (so current text-based clients remain correct).
3. Servers and clients MUST NOT send `\r\n` (so current text-based clients remain correct)

Thoughts on this? If it has sufficient buy-in here I'll try and take this back to OFTC.

prawnsalad:
Just to note here; While I don't have absolute numbers, since insp+unreal got text websocket support there has been a lot of networks that have started using them directly with kiwi. The ease of just throwing up a static web client on a webserver and pointing it to their IRCds has been a major boon for them. Just going from the #kiwiirc support channel I would say at least 30+ networks have asked about it over time so it would be nice if we could not break things for them when they update their IRCds in future.

emersion:
We could just specify that for backwards-compatibility, no subprotocol means text?

slingamn:
That would be somewhat simpler than what I've proposed. I thought we might allow autodetection as well in the absence of a subprotocol. Here's my draft:

https://gist.github.com/slingamn/94933b3436d784399b7faa99d9324088

jesopo:
I'd like to not have to rehash this discussion again with no new insights

edk0:
We'd already come to a compromise, and I'm disturbed that some parties would discard that progress and repeat the same argument. From the point of view of an existing network with non-UTF8 users. UTF8-only clients are *already* incompatible, and requiring servers that support non-UTF8 messages to provide UTF8-only websocket is effectively demanding that we support mutually incompatible client protocols for mutually incompatible clients.

emersion:
> We could just specify that for backwards-compatibility, no subprotocol means text?

After discussion in #ircv3, it seems that a better path forward would be to make the server's behavior implementation-defined when no sub-protocol is negotiated. That would still allow servers to assume text when no subprotocol is negotiated to support some of the existing websocket clients.

slingamn:
I changed the legacy MUST to a MAY in accordance with this recommendation.

jwheare:
Here's what i think has the most consensus. Binary and Text are both first class supported subprotocols. Neither is mandated, and neither has a mandated legacy fallback.

If implementations want to auto detect/character munge/replace etc, that's up to them.

We can describe the pros and cons for each subprotocol, but I don't think a recommendation either way makes sense given the split in opinion on this.

Please +1 or -1 if you approve so we can test my consensus assumption. This is not a voting system.

If we do have consensus, we can move on and publish a spec along those lines. If not, it's probably because I misunderstood something, and we'll try again. If it's still at am impasse, I will invoke "[The Chair](https://ircv3.net/charter)" in an appropriate manner.

> The chair has ultimate responsibility for moving discussions forward and will make decisions in cases where consensus isn’t possible, including returning proposals to the drawing board where necessary.

jwheare:
An alternative that I presumed was off the table has been proposed:

Require servers to always support a binary baseline, but clients are not required to negotiate it as a subprotocol. This at least provides a better interoperability story.

A concern was also raised about breaking existing text-only implementations that don't support subprotocol negotiation. I don't think that is an issue with these proposals. Existing implementations can continue to operate as they currently do and either assume text if no subprotocol is negotiated, or perform heuristics to detect text vs binary.

The subprotocol is just intended to make this explicit and obvious for future implementations.

slingamn:
To advance this we would need to check with the main server implementations that they're OK with this:

- [ ] Unreal
- [ ] Inspircd
- [ ] Solanum (ACK in #ircv3)
- [ ] Oragono (ACK from me)

SadieCat:
We're fine with it.

syzop:
Nice to see this moving forward. We have been suggesting KiwiIRC for a while now in our changelog and websocket docs, and we are seeing more web clients that support direct websocket connections (and more users using them).

> Here's what i think has the most consensus. Binary and Text are both first class supported subprotocols. Neither is mandated, and neither has a mandated legacy fallback.
> If implementations want to auto detect/character munge/replace etc, that's up to them.

Agreed.

> Require servers to always provide a binary baseline. This at least provides a better interoperability story.

Sounds good, at UnrealIRCd we will always make ```Sec-WebSocket-Protocol``` with the "binary" variant available.

The "text" variant may or may not be available in UnrealIRCd, depending on the configuration. For example, if the server is configured to use non-UTF8 extended nick character sets, which [pose an interoperability and security problem](https://github.com/ircv3/ircv3-specifications/pull/342#issuecomment-491252301) for type "text", then "text" won't be available.

In [the UnrealIRCd websocket documentation](https://www.unrealircd.org/docs/WebSocket_support) we will change the meaning of listen::websocket::type from "the websocket type" to "the *default* websocket type".

Sorry, I did not read the full thread, so I apologize if it has already been answered. Just two things come to mind to throw in the spec:
* Of course, for a while clients will encounter server implementation that will just ignore ```Sec-WebSocket-Protocol``` since this header is not parsed in current versions of IRCds. Clients will have to deal with getting back the wrong type of websockets (eg: hang up the connection or something else, up to them).
* Let us IRCd devs know what needs to be done if a client requests a method that we do not support (this can only happen for "text"), eg should we send some HTTP error code back, which one, etc.

emersion:
> Let us IRCd devs know what needs to be done if a client requests a method that we do not support (this can only happen for "text"), eg should we send some HTTP error code back, which one, etc.

FWIW, the JS API will [fail the WebSocket connection](https://fetch.spec.whatwg.org/#concept-websocket-establish) if the client requested only "text" but the server doesn't support it:

> If protocols is not the empty list and extracting header list values given `Sec-WebSocket-Protocol` and response’s header list results in null, failure, or the empty byte sequence, then fail the WebSocket connection.

So it should be OK for the server to just indicate that no sub-protocol was negotiated if none of the client subprotocols are supported.

Note, if the client requested `["text", "binary"]`, the server can just fallback to "binary" if it doesn't support "text".

slingamn:
>FWIW, the JS API will fail the WebSocket connection if the client requested only "text" but the server doesn't support it:

Good find, thanks. This is good overall, but has annoying implications for backwards compatibility of new clients with old servers: attempting to negotiate will cause the connection to fail.

I suppose on failure, the client could try reconnecting with no subprotocol, and then see if the server's legacy fallback works?

jwheare:
Looks like this has approval from server devs. Can we get this wording changed so that servers MUST implement binary.

I think it's probably appropriate to say that servers SHOULD also implement text, because it's probably a good thing to promote. The client decoding situation for invalid utf8 is still not great, you're basically stuck with the replacement character or raw byte arrays.

Pros/cons of each should be in here, sub-protocol stuff, and suggestions for how to deal with clients and servers that don't support sub-protocols, in a non-normative section.

@prawnsalad let me know if you're happy to make these changes, or if someone else should take over this PR.

slingamn:
I'll do it.

prawnsalad:
Now that binary messages is required here, the security considerations could do with some notes on using non-utf8 encodings within a webpage if using binary, since that opens developers up to RCE and other DOM injection issues over using text format.

jwheare:
I think this is only potentially an issue if using the raw bytes. When decoding to text with either the `FileReader.readAsText()` or `Blob.text()` APIs, you'l get the replacement character. Binary clients might need to use the raw bytes if referring to non-utf8 nicks/channels when communicating back to the server, but not for display/DOM-rendering. I wouldn't be opposed to a sentence on that.

jwheare:
That said, it would be helpful if you could point to a specific proof of concept/documentation of the threat you have in mind.

prawnsalad:
It's difficult to come up with a specific example as it's an indepth topic and would would take time to come up with something, but things like when using `readAsText(data, 'GBK')` returns the string with GBK encoding, which when added to a UTF8 page (the default page encoding for HTML5) allows some characters to be altered in different cases. Here's some examples for this specific encoding combination eating and altering characters for strings, though I'm sure this would work in a page too with something like `el.innerHtml = 'some text:' + makeHtmlSafe(textFromReadAsText);` since javascript strings are utf8/16 https://ahuigo.github.io/b/sec/security-sql#/ (yes this is SQL, and mostly talks about quotes and back slashes, but given a HTML context there could be characters that do similar things. This was just a quick google search away)

I wouldn't trust just dumping strings into the DOM if the page and string encodings do not match, even if ran through some sort of HTML sanitisation since they mostly replace certain chars unrelated to encoding nor would I feel comfortable leaving this open in a spec. The only way to do this safely IMO is to suggest always re-encoding the IRC string to the same encoding the page is before adding it to the DOM. Which.. is UTF8 by default.

I'm happy for someone that knows more about mixing encodings more than I do to show and correct me if I'm wrong though.

jwheare:
`readAsText(data, 'GBK')` will *decode* `data` from GBK to a javascript string. Once decoded, it can be html escaped with no issues.

That SQL link is talking about tricking known insecure PHP escaping functions which isn't really relevant here.

prawnsalad:
Hm, right, MDN docs says the encoding is for the *returned data*, but trying it out now it's actually for the source data encoding as you say.

Last thing then for anyone that's implemented this new draft, how were you sending different encodings from the client? Any code snippets would be helpful there.

syzop:
> > Let us IRCd devs know what needs to be done if a client requests a method that we do not support (this can only happen for "text"), eg should we send some HTTP error code back, which one, etc.
> 
> FWIW, the JS API will [fail the WebSocket connection](https://fetch.spec.whatwg.org/#concept-websocket-establish) if the client requested only "text" but the server doesn't support it:
> 
> > If protocols is not the empty list and extracting header list values given `Sec-WebSocket-Protocol` and response’s header list results in null, failure, or the empty byte sequence, then fail the WebSocket connection.
> 
> So it should be OK for the server to just indicate that no sub-protocol was negotiated if none of the client subprotocols are supported.
> 
> Note, if the client requested `["text", "binary"]`, the server can just fallback to "binary" if it doesn't support "text".

So, just to be clear: if a client requests a method we cannot support, we send back a HTTP status code 101 in a header without Sec-WebSocket-Protocol and continue the websocket connection with binary encoding. I think it would be wise to mention that in the specification, especially the latter since otherwise it would be undefined which protocol to use. One can reason it by deduction but better to be 100% explicit about it.

( I was thinking it would be better to send some specific HTTP error code instead, so clients could better differentiate between failed 'text' support and some other error (eg: websockets are not supported at all by the server), but it seems the above (so using 101) is what the RFC considers correct and perhaps only in an utopian world we could differentiate anyway so nevermind. )

slingamn:
>I think it would be wise to mention that in the specification, especially the latter since otherwise it would be undefined which protocol to use.

As I understand the specification as written, this is in fact left implementation-defined. (My implementation falls back to text.)

Is this how other people understood the intention of the spec? I can add clarifying language.

>I was thinking it would be better to send some specific HTTP error code instead, so clients could better differentiate between failed 'text' support and some other error (eg: websockets are not supported at all by the server), but it seems the above (so using 101) is what the RFC considers correct and perhaps only in an utopian world we could differentiate anyway so nevermind.

As I understand it, the proper mechanism to indicate this is the subprotocol field itself --- the client can decide to close the connection if it didn't get its subprotocol (as the JS API does), or it can try to proceed.

slingamn:
Candidate clarification in my fork: https://github.com/slingamn/ircv3-specifications/commit/8b3560e37d59796efc85167a221f162ba5555ac3

jwheare:
I just tested in Chrome what happens if you request a protocol that the server doesn't support (i.e. it doesn't send a `Sec-WebSocket-Protocol` header back). The browser aborts the request, closes the connection, prints a specific error to the console and sends an indistinguishable generic error event to the JS api

```
WebSocket connection to 'wss://irc.example.com/' failed: Error during WebSocket handshake: Sent non-empty 'Sec-WebSocket-Protocol' header but no response was received
```

The "no response" there refers to the header specifically. It aborts before reading the server's response, so even if the server does respond in this case, the lack of that header means no further response is read.

The actual response body, code, etc, do not matter as they're ignored, but a server could print something informational and return a 400 "bad request" response if it wanted to.

So yes, this means that clients wishing to support legacy websocket servers that *don't* support subprotocols will need to retry without a protocol header if they get an error event.

Servers wishing to support legacy websocket *clients* can just choose which protocol they want to use.

slingamn:
I have a (hopefully correct) implementation of the latest draft on my stage for testing: `wss://files.stronghold.network:8097/`

prawnsalad:
Quick poke for anyone thats pushing for binary use - how have you implemented or tested the client side for binary support? The only way I can find to support different encodings so far is to use a client side lib such as iconv-lite which is incredibly slow and bloons my web app file size by 50% - which is obviously very bad for web pages. Anything I'm missing?

prawnsalad:
@dwfreed @syzop I think you both had experimented with the javascript side of this, any input on the last comment re. encoding?

Unless I'm missing something I still can't find a way to natively send different encodings to the IRCd. Since mentioning this back in 2019, browsers have actively *removed* multi encoding support in some of its functions. If this can't be found then supporting multiple encodings should not be a baseline and should only be an optional thing if clients want that large performance hit.

slingamn:
>If this can't be found then supporting multiple encodings should not be a baseline and should only be an optional thing if clients want that large performance hit.

Is this recommendation actionable with respect to the current proposal? As I understand it, the current proposal makes binary messages the baseline, but does not mandate that clients support encodings other than UTF-8. Clients are still free to either:

1. Support text messages only
2. Support binary messages while only accepting and emitting UTF-8

prawnsalad:
@slingamn the reasoning so far for wanting binary was to support multiple encodings. But if we can't actually fully support multiple encodings client-side natively then we shouldn't be suggesting it as the default way in a spec.

slingamn:
>@slingamn the reasoning so far for wanting binary was to support multiple encodings. But if we can't actually fully support multiple encodings client-side natively then we shouldn't be suggesting it as the default way in a spec.

Again, it just seems to me like the conclusion doesn't follow? Under this proposal, it is not the default (in any sense) that clients have to support multiple encodings. There is a sense in which binary websockets are now the default, and clients using them have to deal with the possibility of receiving non-UTF8 data from the server --- but they don't have to "support" it, they can refuse to accept those lines.

Making text the interoperable baseline was rejected, so the choice is between this proposal and having no baseline at all. This seems strictly better to me.

dwfreed:
The reason for the binary protocol was to put the encoding decisions entirely on the client, which is better equipped to make those decisions, as it is possible for it to ask the user how they wish to handle it, within the client's capabilities.  If a client wishes to only support UTF-8 in the binary protocol, that is the client's choice.  A client may also choose to handle any encoding coming from the server, but only give the server user input encoded as UTF-8.  It could also choose to support decoding from and encoding to any arbitrary encoding on a per target or even per line basis.  The point of the binary protocol is to make these choices possible, because there is no one correct answer for all situations.

Keep in mind that this spec is not solely for web clients, even if web clients will be the most frequent user.  Any client may implement WebSocket support, and there are decoding/encoding libraries for any language.  So JS in a browser may be garbage for encoding support, but there are plenty of other languages and environments that don't have this problem.

SadieCat:
I'm unsure why a non-web client would want to implement WebSocket support given that the only reason that WebSockets exist is because web clients can't do normal sockets.

dwfreed:
Restrictive firewall bypass?  Obfuscation of the type of connection to the outside observer?  Silly networks that only allow WebSocket connections?  It doesn't even have to be implemented in a traditional client natively; I'm sure I could throw together a TCP to WebSocket translator in something like 100 lines of Python.  There might even already exist such a translator that is unaware of the application layer protocol it is carrying.  If it is known that one WebSocket implementation expects an Origin header in the HTTP request as part of the WebSocket handshake, then it's likely implementing clients will provide a way for users to specify its value if needed.

DarthGandalf:
>> Silly networks that only allow WebSocket connections?
>
> I hope this doesn't and won't exist.

It exists already. For example, consider running full VM inside a web page in a browser. The only way how it can access external network by TCP is via WebSockets, which can then connect to e.g. websockify daemon, or, in this case, to IRC server directly.

emersion:
It's not possible to directly connect to an IRC server following this specification, because of the framing issue I mentioned above. The VM use-case you describe must connect through the generic WebSocket proxy, then to the TCP IRC server.

prawnsalad:
@dwfreed 

> The reason for the binary protocol was to put the encoding decisions entirely on the client, which is better equipped to make those decisions, as it is possible for it to ask the user how they wish to handle it, within the client's capabilities

The clients available decisions are "Use UTF8" or "Read but not write other encodings". The latter being broken, actively discouraged by web bodies and would cause more confusion than it solves to users.

I don't see an issue with including the binary support and explaining it's limitations, but with the two choices of text or binary, we should not have a half broken option as `MUST` and the other fully working option as `SHOULD` be supported. That's backwards. They should either be switched around or both be required - either way works. You do say yourself now that the binary support is just for keeping clients options open and I agree, but the default and most used use-case (by far) is what this spec is trying to keep as simple as possible and should be fully working. Anything else to cover minor use-cases are better suited to be optional and non-default as thats getting into ymmv territory.

I'm still not entirely sure how (or if) I would implement the binary support in my web client as supporting it removes so much effort and time put into reducing the apps file size and explodes the size by 50% again, causing it to load slower, especially on mobile devices. Other web clients will more than double their app sizes (and I'm including all UI+graphics+stylings in this :x )

prawnsalad:
After reading #ircv3 today, it is obvious there is a big disconnect from the original intent of this spec and what has already been decided to go forward on some networks. The current spec breaks all existing live implementations and will kill performance from web client load time and UI rendering in cases, and as mentioned on IRC the powerful networks have 0 intention to budge. The original idea of "connect to this websocket" has become "hey connect to this websocket but only if x , y, z, and you must learn a,b,c first". Therefore I can't be behind this one. It degrades functionality in almost every way on the current webircgateway proxy a lot of networks are currently using.

In my own web clients I don't know how this could be implemented without effecting everybody, but I'm willing to try and find a way to make it possible once it's ratified.

@slingamn if you could please remove my name from this spec, that would be appreciated.

jwheare:
> The current spec breaks all existing live implementations

Misunderstanding. It does not. Existing implementations can carry on supporting whichever format they currently do.

> will kill performance from web client load time and UI rendering

Overstatement. Only if those clients choose to *optionally* do extra work to support servers that were never going to send utf-8 text frames anyway. Those clients can very easily just not support those servers. This is no different from all the servers that have no websocket support at all.

> the powerful networks have 0 intention to budge

This isn't about powerful entities muscling things to get their way. It's a compromise to allow *everyone* to get their way while still using a shared spec. Any alternative solution would only lead to multiple non-standard competing implementations.

> The original idea of "connect to this websocket" has become "hey connect to this websocket but only if x , y, z, and you must learn a,b,c first".

Misrepresentation. You already needed to know beforehand that a server supports websockets. Now, you get a mechanism (subprotocol negotiation) to know more about what exactly that server supports, and can act accordingly. There is no additional load *imposed* on clients, just an option to do extra work if they so choose.

> It degrades functionality in almost every way on the current webircgateway proxy a lot of networks are currently using.

I can't see anything in the spec that would force webircgateway to work in any substantially different way that would affect any of the clients/networks already using it.

> In my own web clients I don't know how this could be implemented without effecting everybody,

You don't need to implement binary at all, carry on as you already are.

syzop:
I've gone through the current spec and it looks good to me. I also think it's fine to have only a client example with 'text', no problem.
As far as I'm concerned the spec is good to go and can be merged. I'd be happy to implement it in UnrealIRCd.

jwheare:
Yup LGTM

DanielOaks:
Here's a client example that connects to both binary and text servers (with very basic utf8 decoding).
```js
let socket = new WebSocket('wss://ws.example.org', ['text.ircv3.net', 'binary.ircv3.net']);
socket.addEventListener('open', () => {
  socket.send('USER 0 0 0 0');
  socket.send('NICK bob');
});
socket.addEventListener('message', async e => {
  var line = e.data;
  if (line instanceof Blob) {
    line = await e.data.text();
  }
  processLine(line);
})
```

DanielOaks:
@jwheare can we click merge on this? it's short, sweet, hits a balance that means we won't end up with incompatible server impls - let's put something on the website that we can point implementors towards since we're fairly happy that it satisfies our needs now.

slingamn:
This currently has no published client implementations, which seems like a merge blocker. (I think for Kiwi or Gamja to implement this, they would need fallback code for the case where the server fails to advertise subprotocols.)

DanielOaks:
I've built a little IRC WebSocket test page over here, lets you connect and then type+receive lines: https://ircdocs.horse/tools/wstester

SadieCat:
The InspIRCd testnet should support websocket subprotocols now.

syzop:
Nice stuff @SadieCat and @DanielOaks. I've implemented Sec-WebSocket-Protocol as well, will be in UnrealIRCd 5.0.10 release.

syzop:
Also a small note to client devs on browser behavior:
I noticed that Firefox with @DanielOaks test page continues connecting even if you tell it to use eg ```text.ircv3.net``` and the server does not support (nor send back) a ```Sec-WebSocket-Protocol``` header.
Chrome fails the connection though, as others indicated.

SadieCat:
@syzop That seems to be a Firefox bug. I've been wondering if the server should send a HTTP error (412 Precondition Failed?) if no subprotocols match.

syzop:
wss://irc2.unrealircd.org/ is up for testing. Using https://ircdocs.horse/tools/wstester from @DanielOaks it supports all 3 options.

> @syzop That seems to be a Firefox bug. I've been wondering if the server should send a HTTP error (412 Precondition Failed?) if no subprotocols match.

Yeah, I mentioned that too in https://github.com/ircv3/ircv3-specifications/pull/342#issuecomment-790612318. I'm fine with your suggestion, code 412.

It is clear that when the client is asking for something and the server cannot offer it, the request should not continue. If you ask only for binary then you don't want text (and vice-versa). Instead of relying on the client to terminate things because "that is correct" (like Chrome does) it is probably better if we terminate the connection from the server side. Especially now that we know some major browser(s) just continue instead of closing. So I'd be in favor of returning HTTP status code 412 unless it creates some side effect that is unwanted.

Would appreciate feedback from client devs on this.

slingamn:
That sounds reasonable to me as long as it doesn't impact efforts to provide compatibility with legacy code. I think the only salient case is: we shouldn't recommend that the server fail the connection in the case where the client fails to request a subprotocol, only in the case where they requested subprotocols but no compatible subprotocol is available.

SadieCat:
As there hasn't really been any client amovement with subprotocols I decided to commit support for it under vendored names (text.inspircd.org, binary.inspircd.org) and they'll be in the next InspIRCd release at the end of the month. The testnet will support both that and the IRCv3 names as soon as it finishes rebuilding.

emersion:
Feedback from a client dev: I won't support binary frames.

https://lists.sr.ht/~emersion/public-inbox/%3C162961968276.10797.10454955895150869027-0%40git.sr.ht%3E#%3CrKvDkwJh3-RXxYLNsfOKAEdBRjOMRpayatTWY4X9K-TpjAsjdrLWJT4eUq4FC5uQxHP4OupTpc-2PdJXODzmAsRNTcX_62ugX-9Q01emX-8=@emersion.fr%3E

jesopo:
this is a thorough disappointing and ignorant stance. I'm not interested in reiterating for the millionth time why some servers can't sent text frames, I'm just going to say you've disappointed me

justjanne:
For us as desktop client, the spec won't change anything, we'd just support it the way we'd support any other transport protocol, while keeping the entire parsing and message handling code identical.

For us, that means this spec has one single benefit, better NAT/firewall traversal capabilities, which is quite useful in itself, even though it's not what this spec once promised to be.

emersion:
Just to make it clearer what my motivations are: let's stop pretending web clients will ever support anything other than UTF-8. Web browsers only expose UTF-8 APIs. There is an impedance mismatch between IRC and the web, and the server is the best spot to resolve that mismatch.

DanielOaks:
> and the server is the best spot to resolve that mismatch.

But they won't. Reading the comments from server devs in the issue would be good here.

We've been tossing this ball back and forth for at least eight months now, and what we roughly settled on hasn't changed in five months.

At this point we're not in the 'discuss and argue that servers should be doing that translation and always sending/receiving text frames' stage, we're in the "Unreal and Ergo have shipped production implementations of what we've standardised on, Insp's implementation matches that (except using temp names for now), the other large networks are OK with this spec as-is, and a number of the server developers on board have wholeheartedly said that they won't implement text frames" stage.

I know that getting everyone to agree is something that we try to do, but here it's not going to happen and the servers seem to have converged onto an approach, the one described in this spec as it exists right now. It's not worth continuing this line of discussion because the implementations have or are getting written and the servers have basically agreed on this approach. If clients don't want to implement binary frames, they're totally free to do that (the spec's written to allow that!), but I don't think badgering servers to tell them that is gonna do anything but make us all even more weary and irritable than we already are.

emersion:
OK. @jesopo made it sound like there was still need for discussion. If the discussion is over and server devs can't agree with client devs, then I think we're done here. Why is this still not merged?

BTW, if a client or server implements a spec sitting in a PR and then complains that the spec has changed, that's on them.

slingamn:
>Why is this still not merged?

Lack of full client implementations: we don't have a published client implementation that attempts subprotocol negotiation.

Would it be hard to add this to gamja? I think you'd just need to initially request `text.ircv3.net`, then retry without a subprotocol if the connection fails.

SadieCat:
FWIW once this spec is ratified we are likely to remove support for our non-IRCv3 websocket mode (i.e. using the default from the config [usually text] where no subprotocol is set) in our v4 dev branch.

slingamn:
Summarizing some discussion about backwards compatibility mechanisms for JS clients that support only text frames:

1. Chrome and Firefox have different behaviors when `text.ircv3.net` is requested but the server doesn't support it: Chrome fails the connection (as required by the spec) with code [1006](https://developer.mozilla.org/en-US/docs/Web/API/CloseEvent), Firefox proceeds (incorrectly). This is now [mozilla bug #1727118](https://bugzilla.mozilla.org/show_bug.cgi?id=1727118).
2. The following state machine should work: redefine a single attempt to connect to the server as the following algorithm:
    1. Attempt to connect with `text.ircv3.net`
    2. On receiving a close event with code 1006 before any open event, attempt to connect with no subprotocol
    3. If that fails, consider the connection attempt a failure
    4. If the server ever sends a binary frame, consider the connection attempt a failure (this should not be considered fatal, because one could get a transient 1006 due to connectivity issues, then reconnect without a subprotocol and have the server default to binary)

emersion:
I have implemented this logic in gamja: https://git.sr.ht/~emersion/gamja/log/ircv3-websocket

I'll wait for this PR to be merged before integrating the changes in my master branch.

k4bek4be:
> I have implemented this logic in gamja: https://git.sr.ht/~emersion/gamja/log/ircv3-websocket
Works for me.

emersion:
Can we merge this, or are we waiting for something?

extensions/websocket.md
awfulcooking: I have a new IRC friend who wants to connect to modern IRC using old computer hardware, including the Apple II, DOS-era machines, etc.

Detection / negotiation are a different issue, but IMO non-UTF-8 support shouldn't be deprecated.

(I know granted they aren't likely to choose a WebSocket transport for their Apple II client, but still, on the larger topic...)

emersion:
@slingamn @SadieCat @DanielOaks @jesopo Please give ACKs if you're happy with this spec.

jesopo:
I'm no longer interested in this spec, count me as an abstention

slingamn:
I'm happy with the state of the published server and client implementations. I'd like to get an ACK from someone affiliated with Libera, OFTC, or Solanum.

SadieCat:
LGTM.

edk0:
LC & Solanum approve

jwheare:
OK I'll give this another look over and merge tomorrowish. Thanks all, it's been a ride.

extensions/websocket.md
```diff
@@ -0,0 +1,49 @@
...
+socket.addEventListener('open', () => {
+	socket.send('USER 0 0 0 0');
+	socket.send('NICK bob');
+});
+socket.addEventListener('message', event => processLine(event.data));
+~~~
+
+## Security considerations
+
+Clients and servers SHOULD impose limits on the maximum size of messages they will accept, in order to prevent denial-of-service attacks. The limits SHOULD reflect the increased total line lengths allowed by the [`message-tags`](./message-tags) specification.
```
jwheare, line 39: Fix link for consistency with other inter-spec refs in the repo

```suggestion
Clients and servers SHOULD impose limits on the maximum size of messages they will accept, in order to prevent denial-of-service attacks. The limits SHOULD reflect the increased total line lengths allowed by the [`message-tags`](../extensions/message-tags.html) specification.
```

jwheare (changes requested):
Just some style edits/suggestions, but nothing blocking. @slingamn if you have quick solutions/responses then cool but if nothing obvious springs to mind just mark as resolved.

extensions/websocket.md
```diff
@@ -0,0 +1,49 @@
...
+
+---
+
+## Introduction
+
+This specification describes a mechanism for WebSocket-based clients (in particular, web-based IRC clients running in the browser) to connect to IRC networks.
+
+## WebSocket features and encoding
+
+WebSocket is a message-based protocol, i.e., it handles message fragmentation and termination internally. It offers both text (UTF-8 encoded) and binary (8-bit clean) message types; each has advantages and disadvantages as a transport for IRC. Binary messages can transport any valid IRC line, while text messages can only transport lines that are entirely UTF-8. Consequently, binary frames are necessary to fully support IRC communities that use encodings other than UTF-8. On the other hand, given that web browsers are pushing increasingly towards UTF-8 as the default encoding for HTML5 documents and scripting languages, text messages make it simpler to achieve a correct implementation on the client side.
```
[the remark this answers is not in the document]
emersion: "arbitrary bytes"?

jwheare (approved):
LGTM, thanks.

slingamn:
Alpha release of a spec-compliant gateway: https://github.com/ergochat/webircproxy

PR #349, opened 2018-11-04
Add the CHATHISTORY extension specification

MuffinMedic:
Reopened from #292 with a new base branch.

[Rendered view available here.](https://github.com/MuffinMedic/ircv3-specifications/blob/chathistory/extensions/chathistory.md)

### DIFF
--- a/extensions/chathistory.md
+++ b/extensions/chathistory.md
@@ -20,37 +20,47 @@ When a client with the above mentioned capabilities requests `chathistory` conte
 The `server-time` SHOULD be the time at which the message was received from the IRC server and the `batch id` a unique ID for the entire batch. `draft/label` SHOULD be included and MUST be a unique ID used to identify the `chathistory` request and any replies. A `draft/msgid` to identify each individual message MUST be the `draft/msgid` included when each message was first received from the IRC server.
 
 ### `CHATHISTORY` Command
-`CHATHISTORY` content can be requested by the client to the server by sending the `CHATHISTORY` command to the server. A `batch` MUST be returned by the server. If no content exists to return, an empty batch SHOULD be returned to avoid the client waiting for a reply. Command support is sent to the client as the RPL_ISUPPORT 005 numeric `:irc.host 005 nick CHATHISTORY=max_message_count :are supported by this server`
+`CHATHISTORY` content can be requested by the client to the server by sending the `CHATHISTORY` command to the server. A `batch` MUST be returned by the server. If no content exists to return, an empty batch SHOULD be returned to avoid the client waiting for a reply. Command support is sent to the client as the RPL_ISUPPORT 005 numeric `:irc.host 005 nick CHATHISTORY=max_messages :are supported by this server`
 
-Both the `message_count` and `max_message_count` MUST be integers. `message_count` MUST be a non-zero and `max_message_count` MUST be greater than or equal to zero. The client SHOULD not request a `message_count` with an absolute value greater than the `max_message_count` parameter sent in the command.
+Both the `limit` and `max_messages` MUST be integers, and `max_messages` MUST be greater than or equal to zero. The client SHOULD NOT request a `limit` with an absolute value greater than the `max_messages` parameter sent in the command. If the `limit` is 0 or no `limit` is provided, the number of messages equal to `max_messages` MUST BE returned.
 
-Both the `start` and `end` parameters support `draft/msgid` and `timestamp`. The `end` parameter also accepts a `message_count`. If the number of lines between the `start` and `end` parameters exceeds the `max_message_count`, the server SHOULD return a number of lines equal to the `max_message_count` and the appropriate warning as described below. A `max_message_count` of 0 indicates that no maximum exists.
-
-A positive `message_count` or tag `draft/msgid` prepended with a `+` indicates the client is requesting content after the `start`. A negative  `message_count` or `draft/msgid` tag prepended with a `-` indicates the client is requested content before the specified `start`.
+Both the `start` and `end` parameters support `draft/msgid` and `timestamp`. If the number of lines between the `start` and `end` parameters exceeds the `max_messages`, the server SHOULD return a number of lines equal to the `max_messages` and the appropriate warning as described below. A`limit` or `max_messages` of 0 indicates that no limit or maximum exists.
 
 The `target` parameter specifies a single channel or query from which history SHOULD be retrieved. Wildcards or multiple targets are not supported.
+
+#### Context
+
+The following keywords are used to describe how the server should return messages relative to the `timestamp(s)` or `draft/msgid(s)` given.
+
+`BEFORE` get up to `limit` number of messages before the given `timestamp` or `draft/msgid`. The `limit` MUST BE positive.
+
+`AFTER` get up to `limit` number of messages after the given `timestamp` or `draft/msgid`. The `limit` MUST BE positive.
+
+`LATEST` get the most recent (up to `limit`) messages that have been sent since the given `timestamp` or `draft/msgid`. The `limit` MUST BE positive.
+
+`AROUND` get a number of messages before and after the `timestamp` or `draft/msgid`. The `limit` MUST BE positive.
+
+`BETWEEN` get up to `limit` number of messages between the given `timestamps` or `draft/msgids`. The `limit` MAY positive or negative. If the `limit` is positive, messages will be returned in ascending order starting at the oldest message. If the `limit` is negative, messages will be returned in decending order starting at the newest message. 
 
 #### Format
 `chathistory` uses the following generic format:
 
-    @draft/label=ID CHATHISTORY <target> <start> <end>
+    @draft/label=ID CHATHISTORY <target> <context> <start> <end> [<direction>] [<limit>]
 
 The `chathistory` content can requested using timestamps:
 
-    @draft/label=ID CHATHISTORY <target> timestamp=YYYY-MM-DDThh:mm:ss.sssZ message_count=<message_count>
+    @draft/label=ID CHATHISTORY <target> <context> timestamp=YYYY-MM-DDThh:mm:ss.sssZ [<direction>] [<limit>]
 
 Alternatively, content can be requested using a `draft/msgid`:
 
-    @draft/label=ID CHATHISTORY <target> draft/msgid=<message_id> timestamp=<timestamp>
+    @draft/label=ID CHATHISTORY <target> <context> draft/msgid=<message_id> timestamp=<timestamp>
 
-Content can also be requested up to a specified timestamp or `draft/msgid` in place of the `message_count`. The start and end parameter types do not have to match:
-
-    @draft/label=ID CHATHISTORY <target> timestamp=<timestamp> +draft/msgid=<message_id>   
+For `BEFORE`, `AFTER`, `LATEST`, and `AROUND`, a single `timestamp` or `draft/msgid` is required.  For `BETWEEN`, both a `start` and `end` `timestamp` and/or `draft/msgid` is required. The  `limit` is optional for all contexts.
 
 #### Errors and Warnings
 If the server receives a `CHATHISTORY` command with missing parameters, the `NEED_MORE_PARAMS` error code SHOULD be returned.
 
-If the number of lines between the `start` and `end` parameters exceeds the `max_message_count`, warn code `MAX_MSG_COUNT_EXCEEDED` SHOULD be returned. The command SHOULD continue to be processed as described above.
+If the number of lines between the `start` and `end` parameters exceeds the `max_messages`, warn code `MAX_MSG_COUNT_EXCEEDED` SHOULD be returned. The command SHOULD continue to be processed as described above.
 
 If the target has no `chathistory` content to return or the user does not have permission to view the requested content, `NO_SUCH_NICK` or `NO_SUCH_CHANNEL` SHOULD be returned accordingly.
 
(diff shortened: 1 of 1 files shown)


MuffinMedic:
@jwheare Last 2 commits should solve both the stated issues.

DanielOaks:
Nice, it feels like a change of `context` -> `subcommand` makes sense, since we use that terminology heaps already in other specs

extensions/chathistory.md
```diff
@@ -0,0 +1,106 @@
...
+Alternatively, content can be requested using a `draft/msgid`:
+
+    @draft/label=ID CHATHISTORY <target> <context> draft/msgid=<message_id> timestamp=<timestamp>
+
+For `BEFORE`, `AFTER`, `LATEST`, and `AROUND`, a single `timestamp` or `draft/msgid` is required.  For `BETWEEN`, both a `start` and `end` `timestamp` and/or `draft/msgid` is required. The  `limit` is optional for all contexts.
+
+#### Errors and Warnings
+If the server receives a `CHATHISTORY` command with missing parameters, the `NEED_MORE_PARAMS` error code SHOULD be returned.
+
+If the number of lines between the `start` and `end` parameters exceeds the `max_messages`, warn code `MAX_MESSAGES_EXCEEDED` SHOULD be returned. The command SHOULD continue to be processed as described above.
```
DanielOaks: Ehh, this doesn't feel too appropriate as a warning. This is gonna be a pretty standard occurrence when using this subcommand (and honestly the case of 'more playback exists' is gonna happen in other subcommands too).

Maybe this would make sense as a `MORE_HISTORY_EXISTS` or something better-named instead, and not really either a warning or an error.

extensions/chathistory.md
```diff
@@ -0,0 +1,106 @@
...
+
+    @batch=ID;draft/msgid=ID;time=YYYY-MM-DDThh:mm:ss.sssZ :nick!ident@host PRIVMSG target :message
+
+    @batch=ID;draft/msgid=ID;time=YYYY-MM-DDThh:mm:ss.sssZ :nick!ident@host NOTICE target :message
+
+    @batch=ID;draft/msgid=ID;time=YYYY-MM-DDThh:mm:ss.sssZ :nick!ident@host PRIVMSG target :ACTION message
+
+    :irc.host BATCH -ID
+
+    @draft/label=ID :nick!ident@host ERR CHATHISTORY ERR_CODE
```
DanielOaks: Clients should be able to associate the `WARN_CODE` / `ERR_CODE` with the specific chathistory return that's failing/warning. Feels like they should instead be inside the batch, rather than separately outside it. There's no compatibility issues with including 'em inside the batch because the only clients that'll receive these messages are ones that already support `CHATHISTORY`.

extensions/chathistory.md
```diff
@@ -0,0 +1,106 @@
...
+`LATEST` get the most recent (up to `limit`) messages that have been sent since the given `timestamp` or `draft/msgid`. The `limit` MUST BE positive.
+
+`AROUND` get a number of messages before and after the `timestamp` or `draft/msgid`. The `limit` MUST BE positive.
+
+`BETWEEN` get up to `limit` number of messages between the given `timestamps` or `draft/msgids`. The `limit` MAY positive or negative. If the `limit` is positive, messages will be returned in ascending order starting at the oldest message. If the `limit` is negative, messages will be returned in decending order starting at the newest message. 
+
+#### Format
+`chathistory` uses the following generic format:
+
+    @draft/label=ID CHATHISTORY <target> <subcommand> <start> <end> [<direction>] [<limit>]
```
jwheare: This doesn’t really make sense as a generic description of the commands. Not all of them will take start and end or direction. And when direction is available on a command it needs to be explicit not optional.

I also just want to stop using confusing terms like start and end. Each subcommand probably needs its own syntax description.

extensions/chathistory.md
jwheare: We also need to define the top level err/warn more fully. Do they allow just free form text as a generic alert or do they always need a command and error code?

extensions/chathistory.md
```diff
@@ -0,0 +1,106 @@
...
+layout: spec
+work-in-progress: true
+copyrights:
+  -
+    name: "Evan Magaliff"
+    period: "2017"
+    email: "evan@muffinmedic.net"
+---
+## Description
+This document describes the format of the `chathistory` extension. The `chathistory` extension uses the [chathistory][batch/chathistory] batch type, adding a client-side command for requesting `chathistory` content from the server.
```
DanielOaks: "for requesting chat history content from"

the word "`chathistory`" by itself doesn't mean anything in that phrase, so splitting the words up makes sense

extensions/chathistory.md
```diff
@@ -0,0 +1,106 @@
...
+
+The `chathistory` extension adds an optional `draft/msgid` to the `chathistory` batch reply.
+
+Clients MUST support the [`batch`][batch] and [`server-time`][server-time] capabilities. Clients SHOULD support the [`draft/labeled-response`][draft/labeled-response] and [`draft/message-tags`][draft/message-tags] capabilities.
```
DanielOaks: "To use this extension, clients MUST...", and "Clients SHOULD also support..."? As currently written no reason's given as to why clients MUST and SHOULD support those capabilities.

extensions/chathistory.md
```diff
@@ -0,0 +1,106 @@
...
+
+When a client with the above mentioned capabilities requests `chathistory` content from the server (using the `CHATHISTORY` command outlined below), the server SHOULD return to the client a single batch containing raw IRC lines starting with and exluduing the `start` parameter and up to and including the message at the `end` parameter. The raw IRC lines SHOULD be formatted and returned to the client as they were originally, with the addition of the above capability tags.
```
DanielOaks: "the above mentioned capabilities" can just be "these capabilities", since they were just mentioned. In addition, this paragraph explicitly says what to do when the client does have these other caps and requests chathistory... what should happen if the client doesn't have these caps but requests chathistory?

exluduing -> excluding

extensions/chathistory.md
```diff
@@ -0,0 +1,106 @@
...
+
+The `server-time` SHOULD be the time at which the message was received from the IRC server and the `batch id` a unique ID for the entire batch. `draft/label` SHOULD be included and MUST be a unique ID used to identify the `chathistory` request and any replies. A `draft/msgid` to identify each individual message MUST be the `draft/msgid` included when each message was first received from the IRC server.
```
DanielOaks: This feels weirdly lacking context. I can half understand what's being referred to here with each key/tag, but... this needs to sit alongside an example which clearly displays each of these in action, so implementers can get a sense of how these are used (and where they are used) in the real world.

prawnsalad: `received from the IRC server` should be `received at the IRC server`? `from` could mean different things here such as when the IRC server originally sent the line to the client which may not be the case.

extensions/chathistory.md
```diff
@@ -0,0 +1,106 @@
...
+
+### `CHATHISTORY` Command
+`CHATHISTORY` content can be requested by the client to the server by sending the `CHATHISTORY` command to the server. A `batch` MUST be returned by the server. If no content exists to return, an empty batch SHOULD be returned to avoid the client waiting for a reply. Command support is sent to the client as the RPL_ISUPPORT 005 numeric `:irc.host 005 nick CHATHISTORY=max_messages :are supported by this server`
```
DanielOaks: "Chat history content can be..." - as noted above, it's not a proper noun so let's go the more generic descriptor.

It feels like the ISUPPORT list doesn't fit too well here, in the opening paragraph description of the command. Feels like we should split this out into its own section (describing just the RPL_ISUPPORT token and `max_messages`, with no description of `limit`).

extensions/chathistory.md
```diff
@@ -0,0 +1,106 @@
...
+
+Both the `limit` and `max_messages` MUST be integers, and `max_messages` MUST be greater than or equal to zero. The client SHOULD NOT request a `limit` with an absolute value greater than the `max_messages` parameter sent in the command. If the `limit` is 0 or no `limit` is provided, the number of messages equal to `max_messages` MUST BE returned.
+
+Both the `start` and `end` parameters support `draft/msgid` and `timestamp`. If the number of lines between the `start` and `end` parameters exceeds the `max_messages`, the server SHOULD return a number of lines equal to the `max_messages` and the appropriate warning as described below. A`limit` or `max_messages` of 0 indicates that no limit or maximum exists.
+
+The `target` parameter specifies a single channel or query from which history SHOULD be retrieved. Wildcards or multiple targets are not supported.
```
DanielOaks: SHOULD? This language makes it seem like the history could come from any channel or query the server pleases. It feels like what's meant here is just that `target` is where the history will be retrieved from – and that we should leave the possible warning/error conditions to elsewhere in the document.

extensions/chathistory.md
jwheare: From is correct. Msgids are supplied by servers.

jwheare:
Closing some tabs, so just leaving a note that the latest changes to this are based on https://gist.github.com/DanielOaks/c104ad6e8759c01eb5c826d627caf80d

extensions/chathistory.md
DanielOaks: #357 submitted, specifies how I think these sort of response messages should work.

jesopo:
I'd like to suggest this become a `CAP` rather than just `ISUPPORT` - so that servers implementing `resume` (#306) can decide whether to send unprompted chat playback or to wait for the client to request it by `chathistory`

DanielOaks:
Mentioned by Alex on IRC: would also be useful for bouncers to know, since presumably if a client supported chathistory they could do opportunistic loading of channel history after connecting rather than needing all the history of every channel being sent their way automagically.

jwheare:
Reposting from here https://github.com/ircv3/ircv3-specifications/pull/306#issuecomment-464015738
> CAP is probably the best bet, even though it might seem weird that it's mainly needed to suppress the auto-playback feature behaviour that isn't chathistory.

RyanSquared:
Should there be a reply for SUBCOMMAND to make this future-proof in case we add something in the future but the server doesn't support that subcommand?

extensions/chathistory.md
```diff
@@ -0,0 +1,106 @@
...
+
+The following subcommands are used to describe how the server should return messages relative to the `timestamp(s)` or `draft/msgid(s)` given.
+
+`BEFORE` get up to `limit` number of messages before the given `timestamp` or `draft/msgid`. The `limit` MUST BE positive.
+
+`AFTER` get up to `limit` number of messages after the given `timestamp` or `draft/msgid`. The `limit` MUST BE positive.
+
+`LATEST` get the most recent (up to `limit`) messages that have been sent since the given `timestamp` or `draft/msgid`. The `limit` MUST BE positive.
+
+`AROUND` get a number of messages before and after the `timestamp` or `draft/msgid`. The `limit` MUST BE positive.
```
xPaw: What issue does `AROUND` solve? It's not obvious to me when I would want to use this command.

extensions/chathistory.md
```diff
@@ -0,0 +1,106 @@
...
+
+`BETWEEN` get up to `limit` number of messages between the given `timestamps` or `draft/msgids`. The `limit` MAY positive or negative. If the `limit` is positive, messages will be returned in ascending order starting at the oldest message. If the `limit` is negative, messages will be returned in decending order starting at the newest message.
```
xPaw: The positive/negative limit seems like an unnecessary complication to the spec. It's hard to tell what's even supposed to happen if there are more messages in response than the specified limit.

extensions/chathistory.md
xPaw: To add to that, what does direction even do here?

There are BEFORE/AFTER commands, the BETWEEN command describes that limit can be positive or negative.

extensions/chathistory.md
xPaw: I agree the error needs to be inside of the batch, that way you can at least "continue".

extensions/chathistory.md
```diff
@@ -0,0 +1,106 @@
...
+
+## Security Considerations
+Secure identification of users and clients MUST exist in order to ensure that users cannot obtain history they are not authorized to view. Use of account names, internal account identifiers, or certificate fingerprints SHOULD be strongly considered when matching content to users. The server MUST verify the current user's identity matches that of the desired content. This information is not sent as part of the `CHATHISTORY` command and MUST be validated via other means, such as those stated above. Access MUST be checked first and return a `NO_SUCH_NICK` or `NO_SUCH_CHANNEL` error as appropriate. If no authorization error exists, the server can check for returnable content. If no returntable content is found, the server MUST send a `NO_TEXT_TO_SEND` error. The server MUST NOT expose the existence of valid targets to unauthorized users.
+
+While a `max_messages` of 0 MAY be used to indicate no limit exists, servers SHOULD set and enforce a reasonable `max_messages` and properly throttle `CHATHISTORY` commands to prevent abuse.
```
xPaw: Does `max_messages` specify how many messages are stored on the server, how many messages can be requested with a command (the limit param), or both?

If it's how many messages are stored, the note about enforcing a reasonable limit seems moot.

extensions/chathistory.md
RyanSquared: It seems like an odd case where a client requests "around 50" but the server only has 20 above and 30 after.

jwheare: From IRC:

Sunday, February 24th, 2019
13:45:35 <@jwheare> xPaw: AROUND is to enable stuff like search in clients. Search index returns a matching id/timestamp, useful to show context around it.
13:45:47 <@jwheare> Or like a highlights browser
14:11:17 <@xPaw> are you saying it should request AROUND for each search match?
16:25:11 <@jwheare> xPaw: it could do, possibly on demand
16:25:24 <@jwheare> (lazily)
16:26:35 <@xPaw> That would be ridiculously inefficient. Are you talking about server side searching in this case? (Which isn't defined in that spec)
16:28:27 <@jwheare> not necessarily that inefficient. search was just an example. anything that lets you jump back to a point in time in a client. it would be useful to load in context around that point

prawnsalad:
@MuffinMedic are you planning on making any changes based on the above comments? If not then someone can fork it and continue improving the spec

MuffinMedic:
@prawnsalad It's probably better if someone forks it. I no longer have the time to work on this.

prawnsalad:
Ok cool, I'll take a look at this PR later this week as I'm in the middle of an implementation anyhow

SaberUK:
@prawnsalad If you have a more updated version of this would it be possible for you to make it public so that I can implement it? Thanks.

slingamn:
What's the current state of the discussion around private messages?

From a client-side perspective, private messages with different nicks are different "buffers" or "targets". But when I implemented this on the server side, I assumed that all private messages sent to the user were a single target, addressable under the user's own nickname.

This solution has one advantage: it solves the problem of receiving a private message while offline from a new sender. With separate targets for each sender, the client doesn't know to request messages from that sender. With a single target for the recipient, the client can request the entire buffer and then display all the messages appropriately.

I propose making this official: all private messages sent to, or from, the user should be under the same target, either the user's nickname or something like `*self`.

slingamn:
Something else that's come up: the use of msgids in any ordering-based query (before/after/latest/between, anything other than "around" really) is problematic because under server-to-server linking, there is no global ordering of msgids, i.e., no guarantee that the client's observed msgid order matches the server-time order, or matches the order the messages were stored in the history buffer.

This is a problem for walltimes too, but at least with walltimes you can do something like "add a 30-second buffer" and then deduplicate on the client side by msgid.

slingamn:
Let's pick a new owner for this? Now that message IDs are ratified, it would be really nice to move forward here.

prawnsalad:
@slingamn I'm going over this today. Things have been busy so it had dropped down my todo

prawnsalad:
Current overhaul: https://gist.github.com/prawnsalad/4a6cc3cf8c427cadc6bb1d492d7ee29c

I'm yet to proofread this a little closer but the general outline is there. It's a rather large cleanup and change from this existing draft as there was so many confusing parts, so I'll create a new PR for us to work from once I've read over it.

jwheare:
Closing in favour of #393

PR #352, opened 2019-01-10
Add `chathistory` to the list of batch types

Ascrod:
This is a small fix to add a link to the `chathistory` page onto the `batch` extension page.


DanielOaks:
Thanks! We may replace this with a link to the more general registry page later on

PR #358, opened 2019-02-05
Clarify that all `CAP` command replies can be multi-line

jesopo:
As suggested in the third paragraph here https://github.com/ircv3/ircv3-specifications/blob/f06aa95a/core/capability-negotiation.md#the-cap-ack-subcommand...

> If an ACK reply originating from the server is spread across multiple lines, a client MUST NOT change capabilities until the last ACK of the set is received. Equally, a server MUST NOT change the capabilities of the client until the last ACK of the set has been sent.

`CAP ACK` lines can be split across multiple lines, suggesting it behaves similarly to `CAP LS` and `CAP LIST` which would also mean `CAP NACK` would behave this way and it would make sense that `CAP NEW` would too.


jesopo:
closed in favour of issue ircv3/ircv3-specifications#359

PR #361, opened 2019-02-12
SETNAME Spec

justjanne:
## In Short
* Introduces a new `setname` capability
* Introduces a new bidirectional `SETNAME` command

## Rationale

Realnames have historically been immutable for a connection, which has
historically been accepted, as they were rarely used and usually not as present
in client UIs. Still, multiple IRCds have provided independent, non-standard
commands to update realnames.

Nowadays, multiple clients show realnames directly in chat, next to the nick
name, or use information from the realname to provide enhancements, such as
using emails in the realname to provide gravatar based avatars, leading to
higher demand for this functionality than ever.

### DIFF
--- a/extensions/setname.md
+++ b/extensions/setname.md
@@ -73,3 +73,25 @@ modified to support it. The proper way to do so is this:
 
 2) Update the realname portion of data structures and process channel users as
    appropriate.
+
+## Examples
+
+**Client-to-Server**
+
+Complex realname with spaces
+
+    SETNAME :Bruce Wayne <bruce@wayne.enterprises>
+    
+Simple Realname
+
+    SETNAME Batman
+
+**Server-to-Client**
+
+Complex realname with spaces
+
+    :batman@~batman!bat.cave SETNAME :Bruce Wayne <bruce@wayne.enterprises>
+    
+Simple Realname
+
+    :batman@~batman!bat.cave SETNAME Batman
--- a/extensions/setname.md
+++ b/extensions/setname.md
@@ -44,11 +44,11 @@ negotiation time.
 When enabled, clients are allowed to send a `SETNAME` message to notify servers
 of their intent to update their realname.
 
-This client-to-server `SETNAME` message looks as follows:
+This client-to-server `SETNAME` coammdn looks as follows:
 
     SETNAME :realname goes here
 
-This message represents the intent to change the realname. The only, trailing
+This coammdn represents the intent to change the realname. The only, trailing
 parameter is the new realname. Until the server confirms this change, the
 client MUST assume that no change has happened yet.
 
--- a/extensions/setname.md
+++ b/extensions/setname.md
@@ -22,23 +22,17 @@ The final version of the specification will use an unprefixed capability name.
 
 ## Motivation
 
-Realnames have historically been immutable. This has never been much of an
-issue, as they were rarely used and usually not as present in client UIs.
-Still, multiple IRCds have provided independent, non-standard
-commands to update realnames.
-
-Nowadays, this has changed. Multiple clients show realnames directly in chat
-(e.g. next to the nickname) or use information from the realname to provide
-user-visible enhancements (e.g. using emails in the realname to provide
-gravatar based avatars).
-
-Due to this, the ability to change the realname has become relevant.
+Historically, a user's realname could only be set on the initial connection
+handshake. However, multiple IRC servers have provided independent non-standard
+commands to update the realname without reconnecting. This specification
+describes a standardised behaviour based on these existing implementations.
 
 ## Description
 
 The `draft/setname` client capability allows clients to change their realname
-(GECOS) on an active connection. It also allows servers to directly inform clients about such a change.
-This avoids a client reconnect when updating this value. 
+(GECOS) on an active connection. It also allows servers to directly inform
+clients about such a change.
+This avoids a client reconnect when updating this value.
 
 This capability MUST be referred to as `draft/setname` at capability
 negotiation time.
@@ -66,7 +60,8 @@ change.
 Servers MUST check the realname for validity (taking length limits into
 consideration). If they accept the realname change, they MUST send the
 server-to-client version of the `SETNAME` message to all clients in common
-channels, as well as to the client from which it originated, to confirm the change has occurred.
+channels, as well as to the client from which it originated, to confirm the
+change has occurred.
 
 The `SETNAME` message MUST NOT be sent to clients which do not have the
 `draft/setname` capability negotiated.
@@ -85,21 +80,22 @@ modified to support it. The proper way to do so is this:
 1) Enable the `draft/setname` capability at capability negotiation time during
    the login handshake.
 
-2) On receipt of a server-to-client `SETNAME` message, update the realname portion of data structures and process channel users as
-   appropriate.
+2) On receipt of a server-to-client `SETNAME` message, update the realname
+   portion of data structures and process channel users as appropriate.
 
 ## Errors
 
-The server MUST use the standard replies extension to notify the client of failed `SETNAME`
-commands.
+The server MUST use the standard replies extension to notify the client of
+failed `SETNAME` commands.
 
-If the server rejects the realname as a result of a validation failure, it MUST send a `FAIL`
-message with the `INVALID_REALNAME` code.
+If the server rejects the realname as a result of a validation failure, it MUST
+send a `FAIL` message with the `INVALID_REALNAME` code.
 
     FAIL SETNAME INVALID_REALNAME :Realname is not valid
-    
+
 If the server rejects the change for any other reason, it MUST send a `FAIL`
-message with the `CANNOT_CHANGE_REALNAME` code and an appropriate description of the reason:
+message with the `CANNOT_CHANGE_REALNAME` code and an appropriate description of
+the reason:
 
     FAIL SETNAME CANNOT_CHANGE_REALNAME :Slow down your realname changes
     FAIL SETNAME CANNOT_CHANGE_REALNAME :Cannot change realname while banned from a channel
@@ -109,14 +105,14 @@ message with the `CANNOT_CHANGE_REALNAME` code and an appropriate description of
 Complex realname with spaces
 
     C: SETNAME :Bruce Wayne <bruce@wayne.enterprises>
-    S: :batman@~batman!bat.cave SETNAME :Bruce Wayne <bruce@wayne.enterprises> 
-    
+    S: :batman@~batman!bat.cave SETNAME :Bruce Wayne <bruce@wayne.enterprises>
+
 Simple realname
 
     C: SETNAME Batman
     S: :batman@~batman!bat.cave SETNAME Batman
-    
+
 Example with realname rejected by the server
 
     C: SETNAME :Heute back ich, morgen brau ich, übermorgen hol ich der Königin ihr Kind; ach, wie gut, dass niemand weiß, dass ich Rumpelstilzchen heiß!
-    S: FAIL SETNAME INVALID_REALNAME :Realname is not valid 
+    S: FAIL SETNAME INVALID_REALNAME :Realname is not valid
--- a/extensions/setname.md
+++ b/extensions/setname.md
@@ -85,7 +85,7 @@ modified to support it. The proper way to do so is this:
 1) Enable the `draft/setname` capability at capability negotiation time during
    the login handshake.
 
-2) Update the realname portion of data structures and process channel users as
+2) On receipt of a server-to-client `SETNAME` message, update the realname portion of data structures and process channel users as
    appropriate.
 
 ## Errors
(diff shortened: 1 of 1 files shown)


extensions/setname.md
```diff
@@ -0,0 +1,75 @@
...
+negotiation time.
+
+When enabled, clients are allowed to send a `SETNAME` message to notify servers
+of their intent to update their realname.
+
+This client-to-server `SETNAME` message looks as follows:
+
+    SETNAME :realname goes here
+
+This message represents the intent to change the realname. The only, trailing
```
DanielOaks: final parameter* rather than trailing. just like with `USER`, the final param doesn't necessarily need to be a trailing one.

extensions/setname.md
```diff
@@ -0,0 +1,75 @@
...
+
+Nowadays, multiple clients show realnames directly in chat, next to the nick
+name, or use information from the realname to provide enhancements, such as
+using emails in the realname to provide gravatar based avatars, leading to
+higher demand for this functionality than ever.
+
+## Description
+
+The `draft/setname` client capability allows clients to change their realname
+(GEICOS) as well as allowing servers to directly inform clients about such a
```
DanielOaks: think this should be GECOS rather than GEICOS

justjanne: Oops, thanks for finding that typo!

extensions/setname.md
DanielOaks: alternatively, having an example that very explicitly signifies that the `:` on the final param isn't required would work too, e.g. something like:

    C: SETNAME myrealname
    S: nick!user@host SETNAME myrealname

extensions/setname.md
```diff
@@ -0,0 +1,75 @@
...
+change  without having to send a fake quit and join and without clients having
+to actually disconnect and reconnect. 
+
+This capability MUST be referred to as `draft/setname` at capability
+negotiation time.
+
+When enabled, clients are allowed to send a `SETNAME` message to notify servers
```
DanielOaks: hmm, chould we refer to it as a 'command' when talking about the client-to-server side? using 'message' is technically correct syntax as per the RFCs, but it feels like saying _"send a `SETNAME` command"_ here (and below when talking about the c2s message) would make it a lil clearer for client authors implementing this

extensions/setname.md
DanielOaks: resolved, thanks!

DanielOaks:
Nice, I like it.

The `draft/setname` capability. I presume that the s2c `SETNAME` message will only be sent to clients that have negotiated that capability? (and clients that have not negotiated it will not receive the `SETNAME` message).

With servers validating the new realname, if a given realname is too long it feels like the server should probably just silently truncate it, matching what servers already do when getting it initially via `USER`.

If the server would otherwise reject the change (maybe because the client's been changing it too fast, it's being used to bypass a channel ban like how nick changes are restricted, or the value's just bad), it might make sense to explicitly send a response. Maybe a good chance to use the proposed [`FAIL` message](https://github.com/ircv3/ircv3-specifications/pull/357), something like this:

```
FAIL SETNAME CANNOT_CHANGE_REALNAME :Slow down your realname changes
FAIL SETNAME CANNOT_CHANGE_REALNAME :Cannot change realname while banned from a channel
FAIL SETNAME INVALID_REALNAME :Realname is not valid
```

jesopo:
Very simple implementation in BitBot as of jesopo/bitbot@acbe02ba - let me know when there's a test server to try it on

extensions/setname.md
```diff
@@ -0,0 +1,112 @@
...
+    period: "2019"
+    email: "janne@kuschku.de"
+---
+
+## Notes for implementing work-in-progress version
+
+This is a work-in-progress specification.
+
+Software implementing this work-in-progress specification MUST NOT use the
+unprefixed `setname` tag name. Instead, implementations SHOULD use
```
DanielOaks: s/tag name/capability name/, same for the next line

extensions/setname.md
```diff
@@ -0,0 +1,112 @@
...
+## Description
+
+The `draft/setname` client capability allows clients to change their realname
+(GECOS). It also allows servers to directly inform clients about such a change.
+This avoids clients having to actually disconnect and reconnect. 
+
+This capability MUST be referred to as `draft/setname` at capability
+negotiation time.
+
+When enabled, clients are allowed to send a `SETNAME` command to notify servers
```
DanielOaks: If the cap isn't enabled, can the client not send a `SETNAME` command? If the client can't use the command, should the command `FAIL` with a more general error code like `CAP_NOT_NEGOTIATED` - message of `"Capability 'draft/setname' is required to use this command, and is not negotiated"` maybe?

jwheare: No, client can always send the command, but won't get a response.

DanielOaks:
PR written up for Oragono [here](https://github.com/oragono/oragono/pull/372), pending the review q above right now it fails if the sending user hasn't negotiated the cap.

extensions/setname.md
DanielOaks: Whichever way is intended, spec language should be clarified to spell it out

jwheare:
Should there be a NAMELEN isupport key?

extensions/setname.md
justjanne: Considering the way IRCds implement it currently, I’d suggest that clients can always send it, but they won’t get a response (as unreal/inspircd already implement it).

DanielOaks: Slight nibble: The sentence here says _"When enabled, clients are allowed to send a `SETNAME` command..."_. If it's actually able to send the command all the time, should we change this sentence to reflect that? I like the clarification below, makes sense.

DanielOaks:
On how existing impls handle the realname being too long:

Insp:
> `user->WriteNotice("*** SETNAME: Real name is too long");`
> `maxreal` is 128 by default

Unreal:
> `sendnotice(sptr, "*** /SetName Error: \"Real names\" may maximum be %i characters of length", REALLEN);`
> `REALLEN` is 50

So they both explicitly send a notice to the user and fail out. Cool.

Zarthus:
Does it make sense to call this SETNAME over SETREALNAME or SETGECKOS?

When I think of `setname` I think of `/nick`, personally. Alternatively to be consistent with `/nick` you could just call it `/realname` too

DanielOaks:
SETNAME is already in use by Insp and Unreal, if we can make sure this is compatible and use the same command name that'd be cool. If we could invent our own from scratch though I'd def go for `SETREAL` or some permutation ;)

jesopo:
+1 for the command to be `REALNAME` not `SETNAME`

jwheare:
I'd favour not changing a command that users might be accustomed to just for the sake of it.

jwheare:
This is now deployed on the IRCCloud client, and our teams/slack/bnc servers. @jesopo you should already have access to the team server.

SaberUK:
>`/SETNAME` is already in use by Insp and Unreal,

InspIRCd uses SETNAME (change own) and CHGNAME (change other) for this purpose, I believe Unreal does the same.

SaberUK:
Suggestion: if a client has cap `extended-join` support but not `setname` the server SHOULD do a join/part like with CHGHOST.

jesopo:
BitBot's implementation test against irccloud teams - worked perfectly for both sides.

extensions/setname.md
```diff
@@ -46,6 +46,11 @@ Servers advertising the `draft/setname` capability MUST implement the command
 as specified here, and MUST support usage of the command even while the
 capability is not negotiated.
 
+Servers advertising the capability MUST also provide a `NAMELEN` token in
+`RPL_ISUPPORT` with the maximum allowed length for realnames. They MUST NOT
+apply a different maximum allowed length than the one advertised in
+`RPL_ISUPPORT`.
```
jwheare: The second sentence sounds a bit redundant, can probably drop it.

extensions/setname.md
```diff
@@ -0,0 +1,123 @@
...
+as specified here, and MUST support usage of the command even while the
+capability is not negotiated.
+
+Servers advertising the capability MUST also provide a `NAMELEN` token in
+`RPL_ISUPPORT` with the maximum allowed length for realnames.
+
+If a client sends a `SETNAME` command without having negotiated the capability,
+the server SHOULD handle it silently, as the historic implementations did.
+
+Absence of this capability MUST NOT cause the client to prevent the user from
```
xPaw: This sentence effectively duplicates the one above:

> and MUST support usage of the command even while the
capability is not negotiated.

justjanne: What would you then suggest as alternative to replace these paragraphs:

> Servers advertising the `draft/setname` capability MUST implement the command as specified here, and MUST support usage of the command even while the capability is not negotiated.

> If a client sends a `SETNAME` command without having negotiated the capability, the server SHOULD handle it silently, as the historic implementations did.

DanielOaks: I'd honestly just leave this paragraph out, both of those paragraphs make different points but do explicitly say that the command is able to be used when the cap isn't negotiated.

justjanne: One of them says that the command MUST be usable, (but how servers should react to it is still unclear), the other says how they SHOULD react to it. Both are relevant, in a way.

DanielOaks: Yepyep. With the first one saying _"MUST support usage of the command even while the capability is not negotiated."_ makes this paragraph redundant. This doesn't add any information, state any new points, or state it in a more understandable way really (that initial line lays it out pretty well).

DanielOaks: so the main issue here is just the structure of the sentences was a bit confusing and presented a little backwards (as happens when we get different feedback as we push a spec through the process), made for some misreadings and made it unclear as to why the different sentences were included.

if we re-include text to this effect, a structure along these lines should be less confusing for readers:

>>>
Servers advertising the draft/setname capability MUST support usage of the command even while the capability is not negotiated. Clients MUST NOT prevent users from manually using the command while the capability is not negotiated.

If a client sends a SETNAME command without having negotiated the capability, the server SHOULD handle it silently (with no response), as historic implementations did.

jwheare: Sounds good to me.

DanielOaks:
Looks good to me, nice work.

extensions/setname.md
```diff
@@ -0,0 +1,121 @@
+---
+title: IRCv3 `setname` Extension
+layout: spec
```
jwheare: Needs the wip tag

```suggestion
layout: spec
work-in-progress: true
```

extensions/setname.md
```diff
@@ -0,0 +1,122 @@
...
+
+1) Enable the `draft/setname` capability at capability negotiation time during
+   the login handshake.
+
+2) Update the realname portion of data structures and process channel users as
+   appropriate.
+
+## Errors
+
+The server MUST use the FAIL spec to notify the client of failed `SETNAME`
```
jwheare: Add a link to https://github.com/ircv3/ircv3-specifications/pull/357 and change:

```suggestion
The server MUST use the standard replies extension to notify the client of failed `SETNAME`
```

extensions/setname.md
```diff
@@ -0,0 +1,122 @@
...
+commands.
+
+If the server rejects the realname due to invalidity, it MUST send a FAIL
```
jwheare:
```suggestion
If the server rejects the realname as a result of a validation failure, it MUST send a `FAIL`
```

extensions/setname.md
```diff
@@ -0,0 +1,122 @@
...
+message of the following format:
+
+    FAIL SETNAME INVALID_REALNAME :Realname is not valid
+    
+If the server rejects the change for any other reason, it MUST send a FAIL
```
jwheare:
```suggestion
If the server rejects the change for any other reason, it MUST send a `FAIL`
```

extensions/setname.md
```diff
@@ -0,0 +1,122 @@
...
+2) Update the realname portion of data structures and process channel users as
+   appropriate.
+
+## Errors
+
+The server MUST use the FAIL spec to notify the client of failed `SETNAME`
+commands.
+
+If the server rejects the realname due to invalidity, it MUST send a FAIL
+message of the following format:
```
jwheare:
```suggestion
message with the `INVALID_REALNAME` code.
```

extensions/setname.md
```diff
@@ -0,0 +1,122 @@
...
+
+    FAIL SETNAME INVALID_REALNAME :Realname is not valid
+    
+If the server rejects the change for any other reason, it MUST send a FAIL
+message indicating this:
```
jwheare:
```suggestion
message with the `CANNOT_CHANGE_REALNAME` code and an appropriate description of the reason:
```

extensions/setname.md
jwheare: Still needs the link added.

jwheare:
Thanks, this is mergeable as draft now imo.

Issue #293, opened 2017-01-03
Event playback (JOIN/PART/QUIT/NICK/etc as part of chathistory)


dequis:
The chathistory batch is currently limited to PRIVMSG and NOTICE and has no way to some events because they may be processed by the client in a way that affects the present channel state.

* JOIN
* PART
* QUIT
* NICK
* KICK
* MODE
* TOPIC

And I'm probably missing something. Other pieces of information that may be missing in a replay are:

* user prefixes (`@` before nicks)
* away state (may matter if any clients color nicks differently)
* real name (sometimes the real name is displayed right next to the nick)
* hostmask (right click to ban?)

Not sure about the importance of these last few.

From #156, describing the issue:

> >Could this also handle joins/parts/quits/nicks? These are currently problematic due to being global (a replayed NICK would affect global state).
>
>JOINs are problematic too, even though they are not global. In order to send them, server needs to send outdated list of users in channel in state how it was at beginning of the playback. It's possible to implement of course, but still harder than just applying batch tag to PRIVMSGs.
That's why ZNC's buffextras sends JOINs, QUITs, etc. as PRIVMSGs too, originating from *buffextras, and client-side script can render them as "proper" join instead.

See also current discussion at #292

Some ideas:

* A new capability and a message tag marking the event as playback. The events would be sent the same way, but clients are expected to avoid modifying the state and only display the message.
* Same as above, but rely on the chathistory batch to mark the event as playback.
* A new capability to allow sending a new verb that wraps the original event. For example: `:source EVENT JOIN #channel`.

    Might make it easier for client developers to make it a purely visual thing. This is my favorite idea so far.

* No new capabilities, but send buffextras-style PRIVMSGs with message tags describing the event in a machine-readable way (if the message tags cap is enabled).

    Would be compatible with current implementations of chathistory that display a few messages before the user joined. Allows clients to be very lazy. But there's no real need to think in terms of legacy compatibility this time.

* Complementing any of the above, send extra user metadata like prefix as message tags applying to replayed privmsgs.

DanielOaks:
If I could put forth a rough suggestion, perhaps something along these lines would work?

## `event-playback` extension
If the `event-playback` capability is enabled, the `chathistory` batch can contain messages other than `PRIVMSG` and `NOTICE`. This allows servers and bouncers to play back channel joins, parts, and similar messages for clients to display.

[...]

PR #362, opened 2019-02-12
Allow history playback of arbitrary events (channel join/part, quits, etc)

DanielOaks:
The diff is a bit screwy because of the rename, the changes can be seen [here](https://github.com/ircv3/ircv3-specifications/compare/b6d6ee7d6fc10ad70d99c8b135519147fb81aaae...DanielOaks:master+event-playback) though.

Brought up and discussed fairly extensively in #293, there are some events that servers may wish to replay as part of the `chathistory` batch. I know that in Ora, we want to replay joins and parts as they can be useful for working out context.

Right now, the `chathistory` batch only supports `PRIVMSG` and `NOTICE` messages, which makes sense. This spec tries to extend the batch so that clients who support it can be sent any abritrary messages as part of the batch.

At the same time, we want to make totally sure that clients aren't going to run into some use case and accidentally parse an old message as something new. Because of this, as recommended by @dequis in the linked issue, we replay these events as a new message called `CHEVENT` (chat history event). Doing it in this way simplifies client implementations because they can simply catch the `CHEVENT` message and go on to displaying the contents, and it also means that clients have no way of accidentally parsing a replayed event as a real one no matter what strange edge cases they run into.

Feedback would be much appreciated!


RyanSquared:
I think there should be a point about clients being able to either handle or ignore verbs that they don't understand, because I'm quite sure a client author will forget about possible cases they missed otherwise. For example, someone could forget about AWAY or CHGHOST - or not interact with them at all - when writing clients.

DanielOaks:
Suresure, makes sense

slingamn:
Sorry I'm a bit late to this: we're talking about making CHATHISTORY a cap for #306, so are there any downsides to folding this directly into the CHATHISTORY cap?

DanielOaks:
I think it makes sense for event playback to be its own very explicit capability, yeah. Getting `PRIVMSG`/`NOTICE` from played-back history is really standard and makes a lot of sense (heck, bouncers have been doing that for decades), but getting any sort of generic messages in playback (`TOPIC`/etc) is a pretty new feature. Having it be a specific capability, and also using the `CHEVENT` or a similar message to prevent clients from accidentally interpreting incoming replayed messages as real ones is a decent compromise. Right now I don't see a way of rolling it nicely into a more general chat-history capability which meaningfully simplifies things for clients.

dequis:
So, given that this is a new cap, shouldn't it be a standalone spec in its own file instead of changes to a batch type?

DanielOaks:
It's a new cap that directly changes the batch type. We could have it as a separate spec but it would point towards the chathistory batch spec, and just the same the chathistory batch spec would point towards it as well. Open to it, would like to hear other peeps thoughts before going through with it though.

SaberUK:
FWIW the reason I called this batch `chathistory` rather than `history` was so that a separate `history` spec could be created if people wanted something with full playback later.

DanielOaks:
I'll make this a separate batch type and capability. tl;dr if the client's negotiated the `history` (or whatever) capability, and a case exists where the server would send the client a `chathistory` batch, it may instead send it a `history` batch which includes `HEVENT` messages (still doing that to make absolutely sure they aren't parsed raw no matter what happens to the batch framing or whatever).

DanielOaks:
Alright, `chathistory` batch type put back to stock. The `history` batch type (enabled with the `event-playback` capability) is a separate batch type that can also contain `HEVENT` messages. The reason why `HEVENT` messages are used instead of anything else is to ensure that clients parse them as historical messages, even if the batch framing gets stuffed up and even if their clients are hardwired to read all incoming messages and just do actions and modify state based on the verb.

slingamn:
Suggestion: if you attach a msgid to a JOIN or PART, you should be able to attach the same msgid to the HEVENT that replays it.

jwheare:
That’s probably covered already by the msgid spec 

> However, if a message is re-transmitted as-is, for example with the chathistory batch type, the ID SHOULD be reused. As a result, clients MUST accept shared IDs.

DanielOaks:
Yeah I'd consider that to be correct behaviour. I'll add a note to the spec stating it explicitly though (just as 'any tags on the original message SHOULD be supplied as-is on the `HEVENT` message replaying it', or something like that).

extensions/batch/history.md
```diff
@@ -0,0 +1,89 @@
...
+
+
+## Examples
+
+### Channel Message
+
+    :irc.host BATCH +sxtUfAeXBgNoD history #channel
+    @batch=sxtUfAeXBgNoD;time=2015-06-26T19:40:31.230Z :foo!foo@example.com PRIVMSG #channel :I like turtles.
+    @batch=sxtUfAeXBgNoD;time=2015-06-26T19:43:53.410Z :bar!bar@example.com NOTICE #channel :Tortoises are better.
+    @batch=sxtUfAeXBgNoD;time=2015-06-26T19:44:23.423Z :bar!bar@example.com HEVENT PART #channel :Tortises for life!
```
slingamn: This typo may be intentional but it is somewhat distracting ;-)

slingamn:
I just noticed that TAGMSG is specified as requiring a HEVENT instead of TAGMSG. What's the rationale for that?

prawnsalad:
As discussed in #ircv3 yesterday, this whole HEVENT verb not only complicates parsing on the client side and is very odd, but is also redundant. Clients must opt-in to see this batch type so they will already know to only use them for display purposes.

When it comes to parsing the batch, there is no difference vs parsing chathistory which means most logic will simply be `if(batchType === 'chathistory' || batchType === 'history') { handle both }`. This combined with the fact that both history and chathistory batch types serve the same purpose - to replay conversation history - and also formatted the same, why not simply have this `event-playback` cap as the opt-in to extend the chathistory batch with events? The end result bing a more concise and simpler history spec.

slingamn:
I propose that the suggestion in the above comment should be adopted, and the `event-playback` capability should be folded into #393.

DanielOaks:
yeah, it'd be best to handle it that way. totally agree that `HEVENT` makes things annoying for no good reason really.

slingamn:
I believe this is fully absorbed into #393 at this point and can be closed.

PR #363, opened 2019-02-13
Add `brb` extension

jesopo:
Born out of the need to restart BitBot during rapid development - this spec allows clients to use a `BRB` command that piggybacks off `resume` to instantly close a connection while the server holds the session available to `RESUME`.

### DIFF
--- a/extensions/brb.md
+++ b/extensions/brb.md
@@ -50,11 +50,13 @@ closes - in which case it will be used for a `QUIT` reason from the client.
 
 #### Success
 
-    BRB SUCCESS
+    BRB SUCCESS <time>
 
-Used to tell the client that their `BRB` command was successful.
+`<time>` is a positive integer, representing the window of time the client has
+to reconnect.
 
-This is to be immediately followed by the server terminating the TCP connection.
+Used to tell the client that their `BRB` command was successful. This is to be
+immediately followed by the server terminating the TCP connection.
 
 #### Failure
 
--- a/extensions/brb.md
+++ b/extensions/brb.md
@@ -22,7 +22,7 @@ The final version of the specification will use an unprefixed capability name.
 
 ## Introduction
 
-Occasionally, clients will need to restart in order to update servers and, at 
+Occasionally, clients will need to restart in order to update software and, at
 the moment, this will cause a `QUIT` to be sent out for the user on all their 
 networks (unless the client supports in-place upgrading e.g. weechat.)
 
--- a/extensions/brb.md
+++ b/extensions/brb.md
@@ -1,0 +1,86 @@
+---
+title: IRCv3 `brb` Extension
+layout: spec
+extends: resume
+copyrights:
+  -
+    name: "Jess"
+    period: "2019"
+    email: "jess@jesopo.uk"
+---
+
+## Notes for implementing work-in-progress version
+
+This is a work-in-progress specification.
+
+Software implementing this work-in-progress specification MUST NOT use the
+unprefixed `brb` capability name. Instead, implementations SHOULD
+use the `draft/brb` capability name to be interoperable with other
+software implementing a compatible work-in-progress version.
+
+The final version of the specification will use an unprefixed capability name.
+
+## Introduction
+
+Occasionally, clients will need to restart in order to update servers and, at 
+the moment, this will cause a `QUIT` to be sent out for the user on all their 
+networks (unless the client supports in-place upgrading e.g. weechat.)
+
+With the introduction of the `resume` spec, we have an opportunity to leverage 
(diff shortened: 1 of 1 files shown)


extensions/brb.md
```diff
@@ -0,0 +1,93 @@
...
+Software implementing this work-in-progress specification MUST NOT use the
+unprefixed `brb` capability name. Instead, implementations SHOULD
+use the `draft/brb` capability name to be interoperable with other
+software implementing a compatible work-in-progress version.
+
+The final version of the specification will use an unprefixed capability name.
+
+## Introduction
+
+Occasionally, clients will need to restart in order to update servers and, at
```
DanielOaks: "clients will need to restart in order to update servers" this feels a bit confused, aren't clients restarting to update themselves?

extensions/brb.md
```diff
@@ -0,0 +1,93 @@
...
+The final version of the specification will use an unprefixed capability name.
+
+## Introduction
+
+Occasionally, clients will need to restart in order to update servers and, at 
+the moment, this will cause a `QUIT` to be sent out for the user on all their 
+networks (unless the client supports in-place upgrading e.g. weechat.)
+
+With the introduction of the `resume` spec, we have an opportunity to leverage 
+the ability to resume an old session to achieve this but we need a way to
```
DanielOaks: "to achieve this", may just be because the above paragraph doesn't go too in-depth into the actual solution but it feels like the meaning of 'this' here isn't really defined. Doesn't really clearly explain the issue currently (disruption to the client upgrading and others on the network feels like the main issue?) and how this spec is designed to solve that problem

DanielOaks: From this: It feels like the 'this' is mostly-silent and un-disruptive software upgrades. If this is what's intended, could we add a sentence to the end of the first paragraph saying something along the lines of... _"This specification introduces a feature that allows clients to do these upgrades in a mostly-silent way, using the `resume` feature."_ Spells out the purpose of the document more cleanly and to-the-point

extensions/brb.md
```diff
@@ -0,0 +1,93 @@
...
+use the `draft/brb` capability name to be interoperable with other
+software implementing a compatible work-in-progress version.
+
+The final version of the specification will use an unprefixed capability name.
+
+## Introduction
+
+Occasionally, clients will need to restart in order to update software and, at
+the moment, this will cause a `QUIT` to be sent out for the user on all their 
+networks (unless the client supports in-place upgrading e.g. weechat.)
```
DanielOaks: We should avoid mentioning software names explicitly, just leave it at "...supports upgrading in-place)"

extensions/brb.md
```diff
@@ -0,0 +1,93 @@
...
+
+With the introduction of the `resume` spec, we have an opportunity to leverage 
+the ability to resume an old session to achieve this but we need a way to 
+client-initiate a clean disconnect that holds a session open (`resume` is 
+designed solely for ping timeouts.)
+
+## Dependencies
+
+Clients wishing to use this capability must also support the `draft/resume-0.3`
```
DanielOaks: must -> MUST

extensions/brb.md
```diff
@@ -0,0 +1,93 @@
...
+#### `BRB` Command
+
+This command MAY ONLY be sent by a client that has previously negotiated the 
+`resume` capability.
+
+    BRB <reason>
+
+`<reason>` is a trailing argument that will be used as an `AWAY` message for the 
+client until either they resume the session or the window of time for resumption 
+closes - in which case it will be used for a `QUIT` reason from the client.
```
DanielOaks: Should add another paragraph explaining the reconnection step. Maybe something like: _"Following this command, the server will hold the client's session open for some length of time. When the client is ready to resume their session, they will reconnect and use the [`RESUME` command](resume.html) with the resume token they had received earlier."_

extensions/brb.md
```diff
@@ -0,0 +1,93 @@
...
+capability and the `FAIL` message.
+
+## Architecture
+
+### Commands
+
+#### `BRB` Command
+
+This command MAY ONLY be sent by a client that has previously negotiated the 
+`resume` capability.
```
DanielOaks: "has previously negotiated the `resume` capability, and has received and stored their resume token"

just to make it clear that they need to get and keep the token in order for the BRB to be effective

prawnsalad:
Is this not what RESUME on its own is for already? ie. client disconnects and has a set timeframe to reconnect and RESUME its session.

jesopo:
> Is this not what RESUME on its own is for already? ie. client disconnects and has a set timeframe to reconnect and RESUME its session.

`resume` currently only works for clients that don't disconnect cleanly - e.g. ping timeout.

Issue #324, opened 2017-10-07
CAP chghost & self


syzop:
I just added ```CAP chghost``` support in UnrealIRCd which made me wonder about the following:
Would it be acceptable, if person X's  username or hostname changes, for a server to also send the CHGHOST message to person X itself? The [specification](http://ircv3.net/specs/extensions/chghost-3.2.html) isn't clear on this.

For client coders this would provide easy notification of user/host changes of not only others but also themselves. For self host changes there is also numeric 396, but there isn't one for username changes. Now, I don't need or want one either, if using "self"  CHGHOST notification would be permitted. It would provide a simple and singular way to notify the client of a user/host change.

The only downside I can think of is that clients may not expect it for source == self. I highly doubt it will cause any issues, though.

Opinions?

DanielOaks:
fwiw I think this makes sense to do and allow

jwheare:
Works for me. Anyone care to PR a suggested wording change and errata?

jwheare:
https://github.com/unrealircd/unrealircd/commit/b7b8e41ac471d700be7d49255a40866dba374c88

For reference, this change was made in Elemental-IRCd too https://github.com/Elemental-IRCd/elemental-ircd/commit/36f09896a19864ad9c6839e628827ea55e7568ab in October 2015 after the initial implementation in September (https://github.com/Elemental-IRCd/elemental-ircd/commit/e84f541ac893daa818a9e8a92034fb824f1e5a88) and they were (I believe) the original proposers of the cap in #17 in 2013. So this was perhaps the original intent. Maybe @Xe or @ariscop can confirm.

Hybrid followed with the same in November 2015 https://github.com/ircd-hybrid/ircd-hybrid/commit/7bc493be91f0338e0c9f9658f540446b49f0027b 

Hybrid/Elemental both use `sendto_common_channels_local` which includes the user.

Charybdis did not include this change in their implementation in November 2015 https://github.com/charybdis-ircd/charybdis/commit/4a5330bb4d2a04c238767213d5621e46d468c7bd (`sendto_common_channels_local_butone` excludes the user) and nor did InspIRCd in December 2015 https://github.com/inspircd/inspircd/commit/ecca6c2300056f1cff1a9c7cac03939ea3f6f9cc#diff-a786522c317191b607a58037bedff213 (`WriteNeighboursWithExt` excludes the user)

[...]

PR #364, opened 2019-02-16
chghost-3.2.md: clarify specification

RyanSquared:
~~TODO add in comment that servers SHOULD send a CHGHOST to the client~~

This PR implements suggestions from #324

### DIFF
--- a/extensions/chghost-3.2.md
+++ b/extensions/chghost-3.2.md
@@ -15,39 +15,39 @@ copyrights:
 ## Introduction
 
 The `chghost` client capability allows servers to directly inform clients about
-clients changing their host and/or user without having to simulate the client
-disconnecting. This is useful for servers implementing virtual hosts or masks
-and helps with reducing the amount of information spent on a client's UI.
+clients changing their host or user without having to simulate the client
+reconnecting. This is useful for servers implementing virtual hosts or masks
+and helps reduce clutter on the client's UI.
 
 ## The `chghost` capability
 
-When a client changes their user and/or host, servers MUST send the `CHGHOST`
+When a client changes their user or host, servers MUST send the `CHGHOST`
 message to clients who have enabled the `chghost` capability and either share
 the same channel as the target client or have the client on a `MONITOR` list.
 Servers SHOULD additionally send the `CHGHOST` message to the client whose
-user and/or host has changed if the client supports the `chghost` capability.
+user or host has changed if the client supports the `chghost` capability.
 
 When the capability is not enabled for other clients who share channels with or
 monitor the changed client, servers SHOULD send messages to simulate the client
 reconnecting. This allows clients to keep their user state up to date. For
-shared channels, the simulated events SHOULD include appropriate QUIT, JOIN and
-MODE commands, to restore membership and user channel modes. For monitored
-clients, the events SHOULD include appropriate RPL_MONOFFLINE and RPL_MONONLINE
-numerics.
+shared channels, the simulated events SHOULD include appropriate `QUIT`, `JOIN`
+and MODE commands, to restore membership and user channel modes. For monitored
+clients, the events SHOULD include appropriate `RPL_MONOFFLINE` and
+`RPL_MONONLINE` numerics.
 
-The server MUST send a `CHGHOST` message to a client, but MUST defer doing so
-until both a successful `NICK` command and a successful `USER` command have
-been received by the server. The server CAN choose to defer it until after
-registration is completed, for example if a valid SASL authentication during
-client registration triggers an assignment of a virtual host. If the server
-does not defer it until registration is completed, and either the user or the
-host of the client changes, the server MUST send down a new `CHGHOST` message.
+The server MUST send a `CHGHOST` message to a client, but defer doing so until
+both successful `NICK` and `USER` commands have been received by the server.
+The server MAY choose to defer it until after registration is completed, for
+example if a valid SASL authentication during client registration triggers an
+assignment of a virtual host. If the server does not defer it until
+registration is completed, and either the user or the host of the client
+changes, the server MUST send a new `CHGHOST` message.
 
 ## The `CHGHOST` message
 
 The `CHGHOST` message is as follows:
 
-    :nick!user@old_host.local CHGHOST new-user new_host.local
+    :nick!old-user@old_host.local CHGHOST new-user new_host.local
 
 The `new-user` parameter represents the user's "username" or "ident" which may
 or may not have changed in the CHGHOST process.
--- a/extensions/chghost-3.2.md
+++ b/extensions/chghost-3.2.md
@@ -12,13 +12,27 @@ copyrights:
     email: "ryan@hashbang.sh"
 ---
 
-The chghost client capability allows servers to directly inform clients about a
-host or user change without having to send fake quits or joins. This capability
-MUST be referred to as `chghost` at capability negotiation time.
+## Introduction
 
-When the capability is enabled, servers MUST send the `CHGHOST` message to
-clients sharing the same channel as the target client, informing those clients
-that the user or host of the target client has changed.
+The `chghost` client capability allows servers to directly inform clients about
+clients changing their host and/or user without having to simulate the client
+disconnecting. This is useful for servers implementing virtual hosts or masks
+and helps with reducing the amount of information spent on a client's UI.
+
+## The `chghost` capability
+
+When a client changes their user and/or host, servers MUST send the `CHGHOST`
+message to clients who have enabled the `chghost` capability and either share
+the same channel as the target client or have the client on a `MONITOR` list.
+Servers SHOULD additionally send the `CHGHOST` message to the client whose
+user and/or host has changed if the client supports the `chghost` capability.
+
+When the capability is not enabled for clients who share the same channel,
+servers should fall back to having the client `QUIT` the server, rejoin all
+channels, and re-establish the channel and user modes the client had before, as
+though the client had reconnected.
+
+## The `CHGHOST` message
 
 The `CHGHOST` message is as follows:
 
@@ -30,16 +44,13 @@ or may not have changed in the CHGHOST process.
 The `new_host.local` parameter represents the new hostname for the user which
 may or may not have changed in the CHGHOST process.
 
-When the capability is not enabled for clients who share the same channel,
-servers should fall back to having the client `QUIT` the server, rejoin all
-channels, and re-establish the channel and user modes the client had before, as
-though the client had reconnected.
+## Fallback for not supporting the `chghost` capability
 
     :nick!user@old_host.local QUIT :Changing hostname
     :nick!new-user@new_host.local JOIN #ircv3
     :ircd.local MODE #ircv3 +v :nick
 
-# Errata
+## Errata
 
 * Previous versions of this specification used confusing descriptions and have
 since been rewritten to include a simpler description and example.
--- a/extensions/chghost-3.2.md
+++ b/extensions/chghost-3.2.md
@@ -20,11 +20,6 @@ When the capability is enabled, servers MUST send the `CHGHOST` message to
 clients sharing the same channel as the target client, informing those clients
 that the user or host of the target client has changed.
 
-Servers SHOULD also send the `CHGHOST` command to the client who changed their
-host, as well as the previously-unspecified `396` numeric, whose exact value
-count changes per implementation (from two to three arguments), but always
-places the client's new host in the second-to-last position.
-
 The `CHGHOST` message is as follows:
 
     :nick!user@old_host.local CHGHOST new-user new_host.local
--- a/extensions/chghost-3.2.md
+++ b/extensions/chghost-3.2.md
@@ -40,9 +40,9 @@ clients, the events SHOULD include appropriate RPL_MONOFFLINE and RPL_MONONLINE
 numerics.
 
 The server MUST send a `CHGHOST` message to a client, but MUST defer doing so
-until both a successful `NICK` command a successful `USER` command are received
-by the server. The server CAN choose to defer it until after registration is
-completed.
+until both a successful `NICK` command and a successful `USER` command have
+been received by the server. The server CAN choose to defer it until after
+registration is completed.
 
 ## The `CHGHOST` message
 
--- a/extensions/chghost-3.2.md
+++ b/extensions/chghost-3.2.md
@@ -6,41 +6,30 @@ copyrights:
     name: "Christine Dodrill"
     period: "2013"
     email: "xena@yolo-swag.com"
+  -
+    name: "Ryan"
+    period: "2016"
+    email: "ryan@hashbang.sh"
 ---
-The chghost client capability allows a server to directly inform clients about a
-host or user change without having to send a fake quit and join. This capability
+
+The chghost client capability allows servers to directly inform clients about a
+host or user change without having to send fake quits or joins. This capability
 MUST be referred to as `chghost` at capability negotiation time.
 
-When enabled, clients will get the CHGHOST message to designate the host of a
+When enabled, clients will get the `CHGHOST` message to designate the host of a
 user changing for clients on common channels with them.
 
-The CHGHOST message is one of the following:
+The `CHGHOST` message is as follows:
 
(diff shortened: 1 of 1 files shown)


extensions/chghost-3.2.md
```diff
@@ -6,41 +6,30 @@ copyrights:
...
-    :nick!user@host CHGHOST newuser new.host.goes.here
+The `CHGHOST` message is as follows:
 
-This second and third form should only be seen on IRC daemons that support
-changing the user field of a user.
+    :nick!user@host CHGHOST new-user new.host.goes.here
 
-In order to take full advantage of the CHGHOST message, clients must be modified
-to support it. The proper way to do so is this:
+The field represented by `new-user` represents the user's "username" or "ident"
```
DanielOaks: The `new-user` parameter represents...

extensions/chghost-3.2.md
```diff
@@ -6,41 +6,30 @@ copyrights:
...
+which may or may not have changed in the CHGHOST process.
 
-1) Enable the chghost capability at capability negotiation time during the
-   login handshake.
+The field represented by `new.host.goes.here` represents the new hostname for
```
DanielOaks: The `new.host.goes.here` parameter represents...

extensions/chghost-3.2.md
```diff
@@ -6,41 +6,30 @@ copyrights:
...
 ---
-The chghost client capability allows a server to directly inform clients about a
-host or user change without having to send a fake quit and join. This capability
+
+The chghost client capability allows servers to directly inform clients about a
+host or user change without having to send fake quits or joins. This capability
 MUST be referred to as `chghost` at capability negotiation time.
 
-When enabled, clients will get the CHGHOST message to designate the host of a
+When enabled, clients will get the `CHGHOST` message to designate the host of a
```
DanielOaks: When this capability is enabled*

Specifically because the more explicit 'enable the chghost capability' note below has been removed

DanielOaks:
Nice, I like it

SaberUK:
If we're sending the CHGHOST back to a client now should we omit the `RPL_YOURDISPLAYEDHOST` numeric?

RyanSquared:
Question:

Should MONITOR targets also be included on this list? My thoughts on it are that servers SHOULD send it to clients who monitor the changed client.

RyanSquared:
> If we're sending the CHGHOST back to a client now should we omit the `RPL_YOURDISPLAYEDHOST` numeric?

I did not realize that was a thing. I didn't find it in 1459, 2812, or the modern ircdocs project. I'm thinking that clients should support `CHGHOST` *as well* in case any clients rely on it, and because `CHGHOST` also lets you change the ident field of the user.

extensions/chghost-3.2.md
```diff
@@ -16,20 +16,37 @@ The chghost client capability allows servers to directly inform clients about a
...
 MUST be referred to as `chghost` at capability negotiation time.
 
-When enabled, clients will get the `CHGHOST` message to designate the host of a
-user changing for clients on common channels with them.
+When the capability is enabled, servers MUST send the `CHGHOST` message to
+clients sharing the same channel as the target client, informing those clients
+that the user or host of the target client has changed.
+
+Servers SHOULD also send the `CHGHOST` command to the client who changed their
+host, as well as the previously-unspecified `396` numeric, whose exact value
```
DanielOaks: We can't... specify that implementers send a numeric without being clear about the syntax of that numeric. And especially if that numeric already has conflicting definitions between existing vendors. For what it's worth, [info](https://defs.ircdocs.horse/defs/numerics.html#rpl-visiblehost-396).

If we want to allow servers to keep sending their `RPL_VISIBLEHOST`/`RPL_YOURDISPLAYEDHOST `/`RPL_HOSTHIDDEN` numerics we should just not mention it at all and leave it up to the server, especially since newer implementations shouldn't need to pick between existing conflicting numeric definitions.

extensions/chghost-3.2.md
```diff
@@ -16,20 +16,37 @@ The chghost client capability allows servers to directly inform clients about a
...
+The `new-user` parameter represents the user's "username" or "ident" which may
+or may not have changed in the CHGHOST process.
+
+The `new_host.local` parameter represents the new hostname for the user which
+may or may not have changed in the CHGHOST process.
 
-The field represented by `new-user` represents the user's "username" or "ident"
-which may or may not have changed in the CHGHOST process.
+When the capability is not enabled for clients who share the same channel,
+servers should fall back to having the client `QUIT` the server, rejoin all
```
DanielOaks: It should be made extremely clear that this is a **fake** leaving and rejoining the channel. May be worth discussing whether recommending `PART` is nicer as the server won't also need to do fake MONITOR events and such? I wonder how existing servers handle this and whether this behaviour matches that in-the-wild behaviour.

extensions/chghost-3.2.md
RyanSquared: That's fair, this was a complicated thing I found incredibly hard to word anyways. I'll remove it tomorrow.

extensions/chghost-3.2.md
RyanSquared: I was going to bring it up - and thought I already had - but I was thinking that it should act as a full disconnect from the client in a way that MONITOR detects them going offline and coming back online, as that would probably be the only way to send the information. I'm open to discussing whether or not it should act as a `/join 0` or a full disconnect and reconnect.

SaberUK:
InspIRCd's implementation is updated: https://github.com/inspircd/inspircd/commit/b6e548d8dfb933d8329ea7aed2357c88ecaf2e79

extensions/chghost-3.2.md
jwheare: We could maybe mention how CHGHOST differs from the numeric (e.g. 396 doesn't consistently report changes to ident) and say that servers MAY continue to send it alongside CHGHOST without mandating it.

jwheare: I want to avoid servers going "oh I can just replace 396 with CHGHOST now" because some clients might still be relying on the numeric.

extensions/chghost-3.2.md
```diff
@@ -6,41 +6,47 @@ copyrights:
     name: "Christine Dodrill"
     period: "2013"
     email: "xena@yolo-swag.com"
```
Xe: Please update this to me@christine.website, this email address is going to be discontinued.

extensions/chghost-3.2.md
RyanSquared: Important to note that existing implementations (Freenode at the very least) treat it like a full disconnect at the very least for the channels that the user is in - treating it like a QUIT rather than a PART.

extensions/chghost-3.2.md
```diff
@@ -5,42 +5,43 @@ copyrights:
...
-This message represents that the user identified by nick!user@host has changed
-host to another value. The first parameter is the user of the client. The
-second parameter is the new host the client is using.
+The chghost client capability allows servers to directly inform clients about a
+host or user change without having to send fake quits or joins. This capability
+MUST be referred to as `chghost` at capability negotiation time.
 
-On irc daemons with support for changing the user portion of a client, the
-second form may appear:
+When the capability is enabled, servers MUST send the `CHGHOST` message to
```
DarthGandalf: While I understand what it's trying to say, what it actually says is quite different: "at the time of enabling the capability, server MUST send something to other clients"

extensions/chghost-3.2.md
DanielOaks: > Upon a client changing their details with the `USERHOST` command, other clients sharing a channel with that receive the updated username and hostname. Clients that have the capability negotiated receive a `USERHOST` message indicating the new username and hostname of the given client. Clients that do not have the capability negotiated MUST see the given client `QUIT` and proceed to reconnect with the new username and hostname.

I believe this explains how clients should act nicely. It explains both cases in one place rather than at either end of the document.

I'd replace the _"When the capability is enabled, servers MUST"_ paragraph near the top with this rewritten one here. After that, I'd remove this _"When the capability is not enabled"_ paragraph entirely.

extensions/chghost-3.2.md
```diff
@@ -5,42 +5,54 @@ copyrights:
...
+Servers SHOULD additionally send the `CHGHOST` message to the client whose
+user and/or host has changed if the client supports the `chghost` capability.
 
-This message represents that the user identified by nick!user@host has changed
-host to another value. The first parameter is the user of the client. The
-second parameter is the new host the client is using.
+When the capability is not enabled for clients who share the same channel,
+servers should fall back to having the client `QUIT` the server, rejoin all
+channels, and re-establish the channel and user modes the client had before, as
+though the client had reconnected.
```
jwheare: > When the capability is not enabled for other clients who share channels with or monitor the changed client,  servers SHOULD send messages to simulate the client reconnecting. This allows clients to keep their user state up to date. For shared channels, the simulated events SHOULD include appropriate `QUIT`, `JOIN` and `MODE` commands, to restore membership and user channel modes. For monitored clients, the events SHOULD include appropriate `RPL_MONOFFLINE` and `RPL_MONONLINE` numerics.

jwheare: > When the capability is not enabled for other clients who share channels with or monitor the changed client, servers SHOULD send messages to simulate the client reconnecting. This allows clients to keep their user state up to date. For shared channels, the simulated events SHOULD include appropriate QUIT, JOIN and MODE commands, to restore membership and user channel modes. For monitored clients, the events SHOULD include appropriate RPL_MONOFFLINE and RPL_MONONLINE numerics.

SaberUK:
It is possible for a user to receive `CHGHOST` and possibly others before registration if the user sends something like the following:

```
=> CAP REQ :account-notify chghost sasl
=> AUTHENTICATE PLAIN
(the user is sucessfully authenticated here as "wibble")
=> :something!something@example.com CHGHOST wibble users.example.com
```

This is a problem as the user has not send a `NICK` or `USER` message then they won't have a nickname or username. On InspIRCd this defaults to their UID for their nickname and unknown for their ident but this is non-ideal.

Anyone have an idea about what we should do in this case? Its not like `RPL_YOURDISPLAYEDHOST` where we can just replace the nick with `*`.

RyanSquared:
I'm not sure what the best idea would be for this, given this PR was only meant to fix wording in a way that was compatible with the previous version of the specification. I also do not have any thoughts on what would be an ideal solution. Maybe we should take a look at existing implementations and see how many could/would be affected by this - I think the biggest issue here would be parsing the prefix.

RyanSquared:
@SadieCat did this change at all? and is it possible to just defer sending it until the client registration is complete?

SadieCat:
We no longer send CHGHOST until the user has sent both NICK and USER.

RyanSquared:
Waiting on response from @jwheare regarding doubled phrasing of deferring until both successful "NICK" and "USER".

extensions/chghost-3.2.md
```diff
@@ -5,42 +5,68 @@ copyrights:
...
-host or user change without having to send a fake quit and join. This capability
-MUST be referred to as `chghost` at capability negotiation time.
 
-When enabled, clients will get the CHGHOST message to designate the host of a
-user changing for clients on common channels with them.
+## Introduction
+
+The `chghost` client capability allows servers to directly inform clients about
+clients changing their host and/or user without having to simulate the client
+disconnecting. This is useful for servers implementing virtual hosts or masks
```
jwheare, line 19: *reconnecting

extensions/chghost-3.2.md
```diff
@@ -5,42 +5,68 @@ copyrights:
...
-The chghost client capability allows a server to directly inform clients about a
-host or user change without having to send a fake quit and join. This capability
-MUST be referred to as `chghost` at capability negotiation time.
 
-When enabled, clients will get the CHGHOST message to designate the host of a
-user changing for clients on common channels with them.
+## Introduction
+
+The `chghost` client capability allows servers to directly inform clients about
+clients changing their host and/or user without having to simulate the client
```
jwheare, line 18: Not a fan of and/or. Just "or" is fine. "either their host or user" works too.

extensions/chghost-3.2.md
```diff
@@ -5,42 +5,68 @@ copyrights:
...
+disconnecting. This is useful for servers implementing virtual hosts or masks
+and helps with reducing the amount of information spent on a client's UI.
```
jwheare, line 20: Not sure what this means. maybe: "and helps reduce clutter in client UI" ?

extensions/chghost-3.2.md
```diff
@@ -5,42 +5,68 @@ copyrights:
...
+Servers SHOULD additionally send the `CHGHOST` message to the client whose
+user and/or host has changed if the client supports the `chghost` capability.
 
-    :nick!user@host CHGHOST user new.host.goes.here
+When the capability is not enabled for other clients who share channels with or
+monitor the changed client, servers SHOULD send messages to simulate the client
+reconnecting. This allows clients to keep their user state up to date. For
+shared channels, the simulated events SHOULD include appropriate QUIT, JOIN and
+MODE commands, to restore membership and user channel modes. For monitored
+clients, the events SHOULD include appropriate RPL_MONOFFLINE and RPL_MONONLINE
```
jwheare, line 35: backticks on QUIT, JOIN, MODE, RPL_MONOFFLINE, RPL_MONONLINE

extensions/chghost-3.2.md
```diff
@@ -5,42 +5,68 @@ copyrights:
...
+numerics.
 
-This message represents that the user identified by nick!user@host has changed
-host to another value. The first parameter is the user of the client. The
-second parameter is the new host the client is using.
+The server MUST send a `CHGHOST` message to a client, but MUST defer doing so
```
jwheare, line 38: No need for redundant MUST:

> The server MUST defer sending a `CHGHOST` message to a client until...

extensions/chghost-3.2.md
```diff
@@ -5,42 +5,68 @@ copyrights:
...
+until both a successful `NICK` command and a successful `USER` command have
```
jwheare, line 39: > both successful `NICK` and `USER` commands have...

extensions/chghost-3.2.md
```diff
@@ -5,42 +5,68 @@ copyrights:
...
+been received by the server. The server CAN choose to defer it until after
```
jwheare, line 40: s/CAN/MAY

extensions/chghost-3.2.md
```diff
@@ -5,42 +5,68 @@ copyrights:
...
+registration is completed, for example if a valid SASL authentication during
+client registration triggers an assignment of a virtual host. If the server
+does not defer it until registration is completed, and either the user or the
+host of the client changes, the server MUST send down a new `CHGHOST` message.
```
jwheare, line 44: s/send down/send/

extensions/chghost-3.2.md
```diff
@@ -5,42 +5,68 @@ copyrights:
...
-On irc daemons with support for changing the user portion of a client, the
-second form may appear:
+## The `CHGHOST` message
 
-    :nick!user@host CHGHOST newuser host
+The `CHGHOST` message is as follows:
 
-If specified, a client may also have their user and host changed at the same
-time:
+    :nick!user@old_host.local CHGHOST new-user new_host.local
```
jwheare, line 50: nuh should be old_user

extensions/chghost-3.2.md
RyanSquared: I think the examples have a case where both are changed so I'm fine with this. I don't think there's any real implementations where this can happen but it might be good just in case.

RyanSquared: Replaced all instances.

jwheare:
The two tilde examples and preamble can probably be replaced by something like this (and no need to repeat in the fallback section)

> Note that servers implementing ident might choose to prefix a username with a tilde character to denote missing confirmation from an ident server. This MUST be considered part of the username and included in any CHGHOST messages where relevant.

jwheare:
@RyanSquared do you want me to take over this PR?

jwheare:
18:51:42 <LordRyan> jwheare: i would appreciate it, i haven't had time to work on any of my open things.

jwheare:
Replaced with #423

PR #365, opened 2019-02-16
labeled-response.md: typos

trobotham:
Fix labeled-response typos.

### DIFF
--- a/extensions/labeled-response.md
+++ b/extensions/labeled-response.md
@@ -13,7 +13,7 @@ copyrights:
     email: "james@irccloud.com"
   -
     name: "Terrance Robotham"
-    period: "2010"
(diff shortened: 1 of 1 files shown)


extensions/labeled-response.md
```diff
@@ -11,6 +11,10 @@ copyrights:
     name: "James Wheare"
     period: "2016"
     email: "james@irccloud.com"
+  -
+    name: "Terrance Robotham"
+    period: "2010"
```
dgw: Should this be 2019?

jwheare:
Thanks for the fixes but could you remove the copyright addition. For two typos it’s not worth the complications down the line if we want to relicense this as eg something more open.

PR #367, opened 2019-02-21
Update echo-message to require servers to echo TAGMSGs as well.

SaberUK:


DanielOaks:
Yeah I think this makes sense to do (of course, so long as the client being replayed to has the `message-tags` capability negotiated). `TAGMSG` is basically the same as PRIVMSG and NOTICE for intents and purposes here, and considering only new clients will get this behaviour there's no worries.

Should we add a quick mention about needing the `message-tags` capability to be negotiated? We should definitely add an errata item noting the change. Nice PR~

jwheare:
This needs to be clarified in the message-tags spec primarily.

jwheare:
Oh, this is already mentioned in message-tags:

> This means for example, honouring channel membership, modes, echo-message, …

jwheare:
Echo message change added in #373. This isn't really an errata, more new information from another spec.

Issue #369, opened 2019-02-22
Expected behaviour when querying CAP LS after negotiation.


jwheare:
If a client negotiates CAP LS 302 during registration but then later sends a plain CAP LS, should the server downgrade them to 3.1 and stop sending cap notify?

I would say not, spec should probably forbid downgrades explicitly.

DanielOaks:
Ora assumes you support 302 forever and never downgrades. Makes sense to assume that I think?

Issue #359, opened 2019-02-05
Clarify whether multi-line CAP replies applies to ACK/NAK/NEW


jesopo:
Given the 3rd paragraph here https://github.com/ircv3/ircv3-specifications/blob/f06aa95a/core/capability-negotiation.md#the-cap-ack-subcommand

> If an ACK reply originating from the server is spread across multiple lines, a client MUST NOT change capabilities until the last ACK of the set is received. Equally, a server MUST NOT change the capabilities of the client until the last ACK of the set has been sent.

and this conversation https://github.com/ircv3/ircv3-specifications/pull/327#discussion_r156116778

It seems appropriate to discuss a clarification on whether multi-line functionality (as-per-spec for `CAP LS` and `CAP LIST`) apply to other `CAP` command replies too.

If it *does* apply to `CAP ACK`, it really should apply to `CAP NAK` and `CAP NEW` too.

jwheare:
I'm not sure it's safe to suddenly change whether other subcommands can use the continuation functionality. If a server were to make that change, existing clients implemented to the previously valid spec would fail unexpectedly.

What is probably needed is some clear advice on how exactly to deal with long/potentially split REQ/ACKs.

Something like "clients should bear in mind the line length limit, the prefix in use by the server and how many CAPs they're requesting when sending a REQ. If needed, multiple REQs should be sent to avoid truncation in both sending and in receiving the NAK/ACK replies."

The concept of an "ACK spread across multiple lines" isn't really a thing fwict, most implementations will either truncate the REQ or the ACK if they would result in a line >512.

jwheare:
Some other discussion of this from the past: https://github.com/ircv3/ircv3-specifications/pull/327#discussion_r156116778

DanielOaks:
Yeah let's not change existing ACK/NAK/NEW replies. Advice along the lines of jw's suggestions makes sense as advice to client authors.

dequis:
This removed part of the spec kind of helped avoid the need of sending multi-line ack by explicitly saying servers can truncate it and clients shouldn't expect it to be complete: https://github.com/ircv3/ircv3-specifications/commit/d81c6e1e81a96325eadc457cb3961cbe35231806

DanielOaks:
fwiw I think a server truncating an ACK is doing something very wrong (partly why I removed that block of text). I know my client-side stuff treats only the contents of the received ACKs and NAKs from the server as canonical, so any truncation would result in really confusing stuff.

[...]

PR #371, opened 2019-02-22
Cap Clarifications

### LINKED ISSUE #368: Expected forward compat behaviour with CAP LS
What should a client that supports CAP LS 302 do if we in future introduce CAP LS 303. Should it assume 3.2 or 3.1 style responses and cap-notify?

DanielOaks:
Clarifies some issues brought up with the CAP spec. Specifically, fixes #368 and #369, and adds some language along the lines of that suggested by jw in #359.

These clarifications should make CAP more future-proof and consistent.

I'm not quite happy with the _"approaching the 510 byte limit"_ fragment of text in particular, but will continue editing it up. If anyone has recommendations, open to 'em!

### DIFF
--- a/core/capability-negotiation.md
+++ b/core/capability-negotiation.md
@@ -28,11 +28,11 @@ copyrights:
     email: "attilamolnar@hush.com"
   -
     name: "Daniel Oakley"
-    period: "2017-2018"
+    period: "2017-2019"
     email: "daniel@danieloaks.net"
   -
     name: "James Wheare"
-    period: "2017-2018"
+    period: "2017-2019"
     email: "james@irccloud.com"
 ---
 ## What Client Capability Negotiation attempts to solve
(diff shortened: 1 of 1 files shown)


jwheare:
The downgrade stuff should probably be a SHOULD/SHOULD NOT. There’s probably a bunch of implementations that get this “wrong” so I dunno if we can retroactively make this a hard requirement. Dunno, thoughts?

DanielOaks:
Yeah that makes sense, I'll update that language

core/capability-negotiation.md
```diff
@@ -208,10 +208,30 @@ When CAP Version `302` is enabled, the client also implicitly indicates support
 `cap-notify` capability listed below, and support for the relevant `NEW` and `DEL`
 subcommands.
 
+The CAP version number argument MUST be treated as a digit by the server. If the client's
+CAP version is not supported by the server, the server MUST enable the features it does
+support below the given version. For example, if a server supports `302` and `303`
+features, and a client sends version `304`, the server would enable those `302` and `303`
+features (just as it would do if a client sent the `303` version number). However, if the
+server also supported a `306` feature, it whould NOT enable this feature for the client.
```
xPaw: whould?

core/capability-negotiation.md
```diff
@@ -208,10 +208,30 @@ When CAP Version `302` is enabled, the client also implicitly indicates support
 `cap-notify` capability listed below, and support for the relevant `NEW` and `DEL`
 subcommands.
 
+The CAP version number argument MUST be treated as a digit by the server. If the client's
```
grawity: Hmm, it already consists of three digits, so "the CAP version argument MUST be treated as a number" would be more correct (…or "as a decimal integer", depends on how specific you want it).

DanielOaks: Makes sense. I'll just leave it as 'number', that should be fine

jwheare:
Is this good to go?

Issue #370, opened 2019-02-22
Explicitly state that c2c tags MUST be forwarded on TAGMSG


jwheare:
It’s kind of heavily implied but the spec only says PRIVMSG and NOTICE as is.

jwheare:
Done in #373

Issue #334, opened 2018-01-07
Clarify whether the tags `a` and `a=` are the same


DanielOaks:
When parsing tag strings, are `@a=;b` and `@a;b` the same? It feels like software should treat them exactly the same, but making this explicit in the spec would help simplify implementations and make sure clients know they're identical.

jwheare:
From IRC
```
10:32:59 <@jwheare> dan-: fwiw I treat a as a=null and a= as a=‘’
10:33:13 <@dan-> jwheare: I do the same in all my code
```

grawity:
I usually do the same. But, not all languages can distinguish empty string from unset, so IMHO, although clients MAY parse `a` differently from `a=`, tag specifications MUST NOT assign different meanings to them.

nomis:
The specification implies that `a=` is not valid (```<tag> ::= <key> ['=' <escaped value>] ```) because `escaped value` is mandatory.

In charybdis, `a` is a `NULL` value and `a=` is `""`.

CarrotCodes:
Kale does the same as the above - tags are `Map<String, String?>` where `a;b` is `a to null, b to null` and `a=;b` is `a to "", b to null`. Definitely needs clarifying though.

DarthGandalf:
ZNC parses `a` and `a=` the same way, as empty string. When serializing, it writes `=` only if the value is not empty.

ilbelkyr:
@nomis `escaped value` is mandatory, but is defined as

    <sequence of any characters except NUL, CR, LF, semicolon (`;`) and SPACE>

I believe that "sequence of" doesn't necessarily imply "non-empty", but it is perhaps a little unclear.

nomis:
If "sequence of" is not "non-empty" then an empty tag name would be valid.

ilbelkyr:
Good point. It's probably sensible to clarify this, then. (I don't have a strong opinion either way, although @grawity's suggestion appears reasonable.)

jwheare:
#351 got merged but we should check through and see if anything should be added to address this.

jwheare:
This should be resolved by #373

Issue #297, opened 2017-01-07
Rewrite draft/message-tags to be self-contained


jwheare:
At the moment it depends on having already read the message-tags-3.2 spec. It should instead be rewritten as a superset, and the individual 3.2 tags can point to it instead as they're still valid in the new format.

We can then consider removing the 3.2 spec from the site, or prominently marking it as obsolete.

Also, remove 3.3 references, in line with #265.

Add some clarifying language to address #307.

This is a blocker for ratifying the draft.

jwheare:
Should also add something about how tags replace the need for future specs to use CTCP.

lp0:
Please define:
 * What clients should do when the server's tags are too long?
 * What clients should do if the server's response interleaves client and server tags, because the tag data limits are hard to apply in this scenario?

jwheare:
> What clients should do when the server's tags are too long?

We could add something like:

`If a server sends a message with more tag data than the allowed limit, clients MUST ignore the message`

> What clients should do if the server's response interleaves client and server tags, because the tag data limits are hard to apply in this scenario?

Either clients don't need to care, or they can consume server tags up to the 510 limit or the first received +client tag, then continue consuming client tags up the 4094 limit.

lp0:
> Clients MUST ignore the message
That would be ok. It's not nice but there's not much else that can be done. It's hard to define where the RFC1459 message is in the line when this happens.

We could potentially state that clients should find the `' '` (if it exists before the end of the overall line limit) and then parse the RFC1459 message (ignoring all the tags).

> Either clients don't need to care
This is my preferred option as it's simpler for clients than having to do lots of string size checking - they should just parse any tags up to the 4605 byte limit.

jwheare:
The merged spec got done in #351 but might be some clarifications that got missed. Need to double check the points raised here.

jwheare:
Further clarifications added in #373

PR #373, opened 2019-02-27
Message tags updates

jwheare:
This addresses comments from #297 #334 #335 #367 #370 

I've deliberately only added errata for stuff that would have applied to the 3.2 spec.

Edit: this also covers additional typos and comments raised below:
* Clarification around empty/missing tag values
* Clarification around dupe handling and validation of tag key names
* Draft tags

### DIFF
--- a/extensions/message-tags.md
+++ b/extensions/message-tags.md
@@ -107,7 +107,7 @@ with a plus sign (`+`). Servers MAY send client-only tags that weren't provided 
 The client-only tag prefix allows servers to safely relay untrusted client tags,
 keeping them distinct from unprefixed tags that carry verified meaning.
 
-Client-only tags MUST be relayed on `PRIVMSG` and `NOTICE` messages, and MAY be relayed on other messages.
+Client-only tags MUST be relayed on `PRIVMSG`, `NOTICE` and `TAGMSG` messages, and MAY be relayed on other messages.
 
 Any server-sent tags attached to messages MUST be included before client-only
 tags to prevent them from being pushed outside of the byte limit.
--- a/extensions/message-tags.md
+++ b/extensions/message-tags.md
@@ -190,6 +190,8 @@ Servers MUST reply with the `ERR_INPUTTOOLONG` (`417`) error numeric if a client
     417    ERR_INPUTTOOLONG
           ":Input line was too long"
 
+If a server sends a message with more tag data than the allowed limit, clients MAY ignore the message.
+
 ## Security considerations
 
 Client-only tags should be treated as untrusted data. They can contain any value
--- a/extensions/message-tags.md
+++ b/extensions/message-tags.md
@@ -50,9 +50,7 @@ The message pseudo-BNF, as defined in [RFC 1459, section 2.3.1][rfc1459] is exte
 
 The ordering of tags is not meaningful.
 
-Individual tag keys MUST only be used a maximum of once per message. Clients
-receiving messages with more than one occurrence of a tag key SHOULD discard all
-but the final occurrence.
+Individual tag keys MUST only be used a maximum of once per message. Implementations receiving messages with more than one occurrence of a tag key name SHOULD disregard all but the final occurrence.
 
 Implementations MUST interpret empty tag values (e.g. `foo=`) as equivalent to missing tag values (e.g. `foo`). Specifications MUST NOT differentiate meaning between tags with empty and missing values. Implementations MAY normalise tag values by converting the empty form to the missing form, but MUST NOT convert values from missing to empty, to prevent size limit issues.
 
--- a/extensions/message-tags.md
+++ b/extensions/message-tags.md
@@ -54,6 +54,8 @@ Individual tag keys MUST only be used a maximum of once per message. Clients
 receiving messages with more than one occurrence of a tag key SHOULD discard all
 but the final occurrence.
 
+Implementations MUST interpret empty tag values (e.g. `foo=`) as equivalent to missing tag values (e.g. `foo`). Specifications MUST NOT differentiate meaning between tags with empty and missing values. Implementations MAY normalise tag values by converting the empty form to the missing form, but MUST NOT convert values from missing to empty, to prevent size limit issues.
+
 ## Escaping values
 
 The mapping between characters in tag values and their representation in `<escaped value>` is defined as follows:
@@ -327,6 +329,9 @@ no escape character. This was added to help consistency across implementations.
 
 Previous versions of this spec did not specify how to handle invalid escapes. This was
 clarified to help consistency across implementations.
+
+Previous versions of this spec did not specify the difference between empty and missing
+tag values.
 
 [rfc1459]: http://tools.ietf.org/html/rfc1459#section-2.3.1
 [privmsg]: https://tools.ietf.org/html/rfc2812#section-3.3.1
--- a/extensions/message-tags.md
+++ b/extensions/message-tags.md
@@ -44,8 +44,8 @@ The message pseudo-BNF, as defined in [RFC 1459, section 2.3.1][rfc1459] is exte
     <tag>           ::= <key> ['=' <escaped_value>]
     <key>           ::= [ <client_prefix> ] [ <vendor> '/' ] <key_name>
     <client_prefix> ::= '+'
-    <key_name>      ::= <sequence of letters, digits, hyphens ('-')>
-    <escaped_value> ::= <sequence of any characters except NUL, CR, LF, semicolon (`;`) and SPACE>
+    <key_name>      ::= <non-empty sequence of ascii letters, digits, hyphens ('-')>
+    <escaped_value> ::= <sequence of zero or more bytes except NUL, CR, LF, semicolon (`;`) and SPACE>
     <vendor>        ::= <host>
 
 The ordering of tags is not meaningful.
--- a/extensions/message-tags.md
+++ b/extensions/message-tags.md
@@ -121,6 +121,8 @@ This means client-only tags that aren't specified in the IRCv3 extension registr
 use a vendor prefix and SHOULD be submitted to the IRCv3 working group for consideration
 if they are appropriate for more widespread adoption. See [Rules for naming message tags](#rules-for-naming-message-tags).
 
+Client-only tags are intended to replace the use of future [CTCP commands][ctcp].
+
 ### The `TAGMSG` tag-only message
 
        Command: TAGMSG
@@ -328,3 +330,4 @@ clarified to help consistency across implementations.
 [privmsg]: https://tools.ietf.org/html/rfc2812#section-3.3.1
 [statusmsg]: https://tools.ietf.org/html/draft-hardy-irc-isupport-00#section-4.18
 [registry]: https://ircv3.net/registry.html#tags
+[ctcp]: https://tools.ietf.org/html/draft-oakley-irc-ctcp-02
--- a/extensions/message-tags.md
+++ b/extensions/message-tags.md
@@ -52,6 +52,8 @@ The ordering of tags is not meaningful.
 
 Individual tag keys MUST only be used a maximum of once per message. Implementations receiving messages with more than one occurrence of a tag key name SHOULD disregard all but the final occurrence.
 
+Implementations MUST treat tag key names as opaque identifiers and MUST NOT perform any validation that would reject the message if an invalid tag key name is used. This allows future modifications to the tag key name format.
+
 Implementations MUST interpret empty tag values (e.g. `foo=`) as equivalent to missing tag values (e.g. `foo`). Specifications MUST NOT differentiate meaning between tags with empty and missing values. Implementations MAY normalise tag values by converting the empty form to the missing form, but MUST NOT convert values from missing to empty, to prevent size limit issues.
 
 ## Escaping values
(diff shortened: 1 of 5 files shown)


extensions/message-tags.md
```diff
@@ -54,6 +54,8 @@ Individual tag keys MUST only be used a maximum of once per message. Clients
 receiving messages with more than one occurrence of a tag key SHOULD discard all
 but the final occurrence.
 
+A tag with an empty value (e.g. `foo=`) MAY be considered equivalent to a tag missing a value (e.g. `foo`). Specifications MUST NOT assign different meanings to tags with empty or missing values.
```
nomis: I think this should be a `MUST` and (to make the message smaller) the `foo` form should be stated as the canonical form.

nomis:
The specification needs to clarify that tag names cannot be empty, and that tag values can be.

Currently both tag names and tag values are either both implicitly non-empty or implicitly allowed to be empty.

extensions/message-tags.md
jwheare: Imagine a basic parser looking like this:

```python
tags = 'foo;bar='
[part.split('=') for part in tags.split(';')]
# [['foo'],['bar','']]
```

Saying that implementations MAY consider empty/missing as equivalent *allows* a normalisation step without *requiring* it, and saying that specs MUST NOT differentiate makes that safe.

Changing the MAY to a MUST seems like forcing extra awkwardness on implementations for no real gain. Implementations might decide there's a gain *for them* but the spec certainly doesn't need it.

nomis: It's not incompatible with a MUST to parse it like example but then check the values properly if they get compared.

DanielOaks: As currently written it feels like it's saying that specs can't assign different meanings to a tag when it has a value vs the same tag when it's not given a value at all, when what's actually intended is that specs can't treat the `a=` and `a` case differently. Maybe we could use the term _"tags with empty **vs** missing values"_ or similar to make it clear what we mean here?

For what it's worth I don't mind making it a MUST to ensure they're treated the same way, but I'm unsure about specifying which one software has to send out.

extensions/message-tags.md
```diff
@@ -54,6 +54,8 @@ Individual tag keys MUST only be used a maximum of once per message. Clients
 receiving messages with more than one occurrence of a tag key SHOULD discard all
 but the final occurrence.
 
+Implementations MAY consider a tag with an empty value (e.g. `foo=`) as equivalent to a tag missing a value (e.g. `foo`). Specifications MUST NOT assign different meanings to tags with empty or missing values.
```
jwheare: How about

```suggestion
Implementations MUST interpret empty tag values (e.g. `foo=`) as equivalent to missing tag values (e.g. `foo`). They MAY normalise such tags to a single form. Specifications MUST NOT differentiate meaning between tags with empty and missing values.
```

DanielOaks: Perfect, well said

extensions/message-tags.md
nomis: There is a potential issue with a client tag `+a` being rewritten by a server to `+a=` and then the whole set of client tags may not fit in the outbound buffer. We should state that servers MUST NOT rewrite relayed client tags with a missing value from `+a` to `+a=`. This also applies more generally (i.e. not adding spurious backslashes).

jwheare: The size limits already forbid this. Let’s not bog down specs by repeating things covered elsewhere ad nauseam to cover minor edge cases with non compliant implementations.

PS also this thread applies to outdated code so is hidden, worth replying elsewhere.

nomis: "The server MAY unescape tag values after receiving data and escape those values before sending them out."

> The size limits already forbid this.

Where is this specified?

> non compliant implementations

It's not non-compliant; `+a` is equivalent to `+a=` so a server implementation that finds it easier to always output `+a=` can do so.

(These comments aren't hidden on the PR page.)

jwheare: Read the size limit section. Servers and clients are forbidden from overflowing the output buffers. Hence, non-compliant.

nomis: I'm not suggesting the server would overflow it, but could be forced to omit a client tag when it doesn't fit on the end because of missing values being rewritten as empty values.

DanielOaks: I don't think this possibility is worth us worrying about at all in the specification.

jwheare: Should be resolved by https://github.com/ircv3/ircv3-specifications/pull/373/commits/ed7ad5bb33f835708f3383d52c01643c4e4e1898

unattributed:
A few things which should be clarified:

It's only specified what _clients_ should do with duplicate tag keys. Since CTC tags exist as well now, servers also need to deal with this, so the wording should be changed to the more generic "implementation".

It isn't specified what implementations should do upon receiving a message with unnamed tags (e.g. `foo;=bar` or `baz;;qux`). Should they discard only that field, ignore all tags for the message, ignore the message entirely, or do something else instead?

I'd personally be fairly lenient with this (as seems to be the general precedent with IRC software) and just ignore the individual tag. This is the behaviour used by InspIRCd. I have not checked other implementations.

It's also not specified what implementations should do upon receiving tag names that aren't valid per spec. Should their format be validated by implementations and invalid tags be ignored? Discarding non-conforming tags would artificially reduce compatibility with non-conforming (yet still unambiguous) specifications, and implementations only handle the tags they understand anyway. It's also unclear what it entails for a key to be "treated as an opaque identifier" and if/how that relates to this.

I currently treat the tag name as a blob, with the exception of the client prefix. This is the same (as far as I can tell, from a cursory look at the code) behaviour used by InspIRCd. I have not checked other implementations.

Finally (albeit not a clarification), "A TAGMSG sent by a client with tags that exceed the size limit and rejected by the server" is not valid grammar.

DanielOaks:
> It's also not specified what implementations should do upon receiving tag names that aren't valid per spec.
On this point, the _"treated as an opaque identifier"_ text fairly specifically means that clients and servers don't validate names outside checking for the client-only prefix (`+`) to decide whether to treat it as one of those tags. The software shouldn't know whether it's 'valid' as per the spec because it shouldn't be doing any sort of validation at all, only treating them as opaque identifiers (matches your experience with implementations and how my software handles it). Not sure how you'd note this better in the spec without complicating it unduly.

Agreed that all other points could use some clarification.

extensions/message-tags.md
```diff
@@ -44,8 +44,8 @@ The message pseudo-BNF, as defined in [RFC 1459, section 2.3.1][rfc1459] is exte
     <tag>           ::= <key> ['=' <escaped_value>]
     <key>           ::= [ <client_prefix> ] [ <vendor> '/' ] <key_name>
     <client_prefix> ::= '+'
-    <key_name>      ::= <sequence of letters, digits, hyphens ('-')>
-    <escaped_value> ::= <sequence of any characters except NUL, CR, LF, semicolon (`;`) and SPACE>
+    <key_name>      ::= <non-empty sequence of ascii letters, digits, hyphens ('-')>
+    <escaped_value> ::= <sequence of zero or more utf8 characters except NUL, CR, LF, semicolon (`;`) and SPACE>
```
DarthGandalf: This retroactively breaks networks which use non-utf8 charset

jwheare: Which existing tags allow non utf8?

DarthGandalf: In the value? Any non-standard one.

But I was more refering to the concept of requiring UTF-8. E.g. in ZNC charset is applied to the whole line, before parsing of tags even begins. I hope you don't want UTF-8 to be encoded inside some other encoding.

jwheare: How is znc going to handle client sent utf8 in future?

jwheare: Also, we don’t need to care about non standard implementations.

DarthGandalf: IRC is defined in terms of octets, not UTF-8 text. If we are to change that, that should be done explicitly, in a separate spec, not retroactively baked into some obscure line in one of unrelated specs.

> How is znc going to handle client sent utf8 in future?

There are settings for that.

jwheare: I am focused on new parts of the protocol here, not changing things retroactively. I want to ensure everyone can assume a sane encoding for new features, so we can avoid some of the short sightedness of the past. This seemed like a good place to enforce this for new tags. What would be a more appropriate way of doing this?

jwheare: > There are settings for that.

Does this mean the default would already be “broken” behaviour as you call it? I’m not sure how this changes anything in that case.

DarthGandalf: > What would be a more appropriate way of doing this?

Instead of changing things subtly and retroactively, change them explicitly. I generally agree with usage of UTF-8 everywhere. But need to write a separate spec which defines UTF-8 as the encoding, which discusses how to enable this (cap? implicitly?), and how it should interact with other protocol features, with existing software out there.

> Does this mean the default would already be “broken” behaviour as you call it?

Define "broken".

> I’m not sure how this changes anything in that case.

Here's an example.

Client <--- UTF-8 ---> ZNC <--- KOI8-R ---> Server

If client sends a message, ZNC interprets it as UTF-8, but sends it to server as KOI8-R. If server sends a message, ZNC decodes it from KOI8-R, and sends to client as UTF-8.

Now imagine that the message had some tags, and the value of one of characters in some tag is outside of ASCII range. What exactly is supposed to happen here?

jwheare: Broken was your word, I’m not sure how you’re defining it.

With your example, what would you do in znc right now if that happens?

DarthGandalf: Now the whole line gets reencoded, including tags. So by changing the requirement for tags to be utf-8, you're retroactively breaking existing software, which suddenly wouldn't be conformant anymore.

jwheare: The KOI8-R "server" you're talking about is actually just an encoding unaware relay. The KOI8-R data you're dealing with is in message bodies that users on that server have decided to use amongst each other. The clients of those users could already start sending utf8 data in tags while still allowing KOI8-R in messages. The server could also add utf8 tag data of its own if it wanted. Your whole line encoding might already not deal with that very well so it's already a "broken" situation you will likely need to address. And if you can assume tags are always utf8, this is actually much easier for you to address.

All I'm proposing is that defined, specified new tag values MUST be utf8. This is designed to make things easier. I don't think the existing tags we have (server-time, account-tag, batch) are at much risk of containing non-utf8 data.

DarthGandalf: > The KOI8-R "server" you're talking about is actually just an encoding unaware relay. 

Unfortunately, it's not. Server actually has several ports, and uses different encoding on each. So client connected on port 6667 and client connected on port 6668 use different encodings, while still talking to each other.

> The server could also add utf8 tag data of its own if it wanted

So where is the split between utf8 and the other encoding? End of tags? Again, this needs a separate spec and a separate discussion.

jwheare: Ok so in that case, your russian re-encoding server will have to deal with this, not you. And if it's in the spec, that's doable. I think we can agree this is a rather niche case, isn't it?

Why is the tags spec not a good place to specify the encoding of tags?

jwheare:
Addressed points raised by @unattributed. This reiterates that tag key names are opaque and forbids validation. This avoids issues like we had where implementations were validating vendor prefixes as dns names when we started using the draft/ prefix.

extensions/message-tags.md
jwheare: I've removed the "utf8" bit from this PR for now and will move it its own PR.

jwheare:
Barring any further comments, I'll merge this after the weekend.

unattributed:
@DanielOaks I disagree; an implementation could be parsing individual tags using a regex such as `(\+?(<HOST>/)?[[:alnum:]]+)(=.*)?`, which would work perfectly fine with any conformant specification, but subtly break on anything else. I'd personally read "opaque" as "should not derive meaning from this" (e.g. not resolving the vendor prefix as a domain name), which doesn't exclude implementations from parsing things in such a manner. (Yes, I'm aware that this would be very brittle, and I'm not advocating for anyone to do this.)

@jwheare Looks good to me, although I'd remove "Full tag names, including any vendor prefixes MUST be treated as an opaque identifier." since it's already handled (more thoroughly) by your change.

DarthGandalf:
Thanks, LGTM

jwheare:
@unattributed I think it's ok to repeat that point when discussing tag naming.

I've also added a section on draft names, adapted from https://github.com/ircv3/ircv3-specifications/blob/e784d1b615d08577d4881a403031d84240e82e59/core/capability-negotiation.md#draft-capabilities

unattributed:
@jwheare Sure, I'm not strongly opinionated either way.

unattributed:
Actually, there still ought to be clarification on how empty key names/fields should be treated. I listed some potential options in my earlier comment, but feel that it would probably be simplest to just leave them be completely. It fits with the rule of not validating key names, is lenient (which fits with the general approach that IRC software takes towards handling non-conforming implementations), can still be unambiguously parsed (an empty field is just a tag with an empty name and no value, after all), and is the least logic to implement.

I raised this point initially when it was still unclear as to how to handle non-conforming tag names. With that out of the way, the motivation for having this clarified as well is somewhat reduced. It should probably still be done, however, to remove any remaining uncertainty, since at least one major implementation (InspIRCd) does discard tags with no name.

jwheare:
Since I added that key names are non empty sequences, this is already covered by the guidance for invalid keys, they mustn’t be validated. That said, they’d be impossible to meaningfully handle so stripping them would present no issues. I don’t think it’s worth adding an exception to the spec to cover this.

PR #374, opened 2019-03-06
[message-tags] Specify that tag values are utf8

jwheare:
This is a one word change but needs some discussion (and possibly needs expanding with a clarifying paragraph)

Following on from discussion in https://github.com/ircv3/ircv3-specifications/pull/373#pullrequestreview-209467284 (click `show resolved` to expand) and ircv3/ircv3-ideas#38 this proposal aims to enforce utf8 in at least one, relatively newly introduced section of the protocol.

This would allow us to assume an interoperable baseline for future tag specs, even if we can't for freeform text that has a long legacy of mixed encodings where the user is on their own to decide.

From an implementation point of view, I don't think this should present a major issue. Tags already need to be parsed separately from the rest of the message. For clients that handle multiple encodings, or do things like attempt to parse as UTF8 with a fallback to latin1, they can simply skip this step for tag parsing. For implementations that only support utf8 already, this would change nothing.

This change would mean that any data stored by a server in a different encoding, would need to be re-encoded for message tags. For a contrived example, a server that allowed non-utf8 account names would need to re-encode them for the `account-tag` spec. I think this possibility is sufficiently unlikely and niche as to not be a significant concern for interoperability. And the benefits of establishing a standard encoding far outweigh it.

You might ask why we can't just enforce utf8 everywhere. I think that would be a far more difficult task, and cause much more significant interop issues with legacy. There might be scope for a separate ISUPPORT/cap to enforce that, but it's tricky because users are the ones that decide on their encoding, and any negotiation between client and server wouldn't be done with user intervention. Besides, servers typically are not encoding-aware (nor do they want to be) so it's effectively a client-to-client decision, in which case, a cap isn't really the right solution.

### DIFF
--- a/extensions/message-tags.md
+++ b/extensions/message-tags.md
@@ -54,6 +54,8 @@ Individual tag keys MUST only be used a maximum of once per message. Implementat
 
 Implementations MUST treat tag key names as opaque identifiers and MUST NOT perform any validation that would reject the message if an invalid tag key name is used. This allows future modifications to the tag key name format.
 
+Tag values MUST be encoded as UTF8. This ensures a shared interoperable baseline for data exchange. If tag values are encountered that cannot be decoded as UTF8, implementations MAY drop the value entirely, or substitute replacement bytes in place of invalid data.
+
 Implementations MUST interpret empty tag values (e.g. `foo=`) as equivalent to missing tag values (e.g. `foo`). Specifications MUST NOT differentiate meaning between tags with empty and missing values. Implementations MAY normalise tag values by converting the empty form to the missing form, but MUST NOT convert values from missing to empty, to prevent size limit issues.
 
 ## Escaping values
@@ -342,6 +344,8 @@ clarified to help consistency across implementations.
 
 Previous versions of this spec did not specify the difference between empty and missing
 tag values.
+
+Previous versions of this spec did not specify the UTF8 encoding for tag values
 
 [rfc1459]: http://tools.ietf.org/html/rfc1459#section-2.3.1
 [privmsg]: https://tools.ietf.org/html/rfc2812#section-3.3.1
--- a/extensions/message-tags.md
+++ b/extensions/message-tags.md
@@ -54,7 +54,7 @@ Individual tag keys MUST only be used a maximum of once per message. Implementat
 
 Implementations MUST treat tag key names as opaque identifiers and MUST NOT perform any validation that would reject the message if an invalid tag key name is used. This allows future modifications to the tag key name format.
 
-Tag values MUST be encoded as UTF8. This ensures a shared interoperable baseline for data exchange. If tag values are encountered that cannot be decoded as UTF8, implementations MAY drop the value entirely, or substitute replacement bytes in place of invalid data.
+Tag values MUST be encoded as UTF8. This ensures a shared interoperable baseline for data exchange. If tag values are encountered that cannot be decoded as UTF8, implementations MAY drop the value entirely but MUST NOT substitute replacement bytes in place of invalid data.
 
 Implementations MUST interpret empty tag values (e.g. `foo=`) as equivalent to missing tag values (e.g. `foo`). Specifications MUST NOT differentiate meaning between tags with empty and missing values. Implementations MAY normalise tag values by converting the empty form to the missing form, but MUST NOT convert values from missing to empty, to prevent size limit issues.
 
--- a/extensions/message-tags.md
+++ b/extensions/message-tags.md
@@ -54,7 +54,7 @@ Individual tag keys MUST only be used a maximum of once per message. Implementat
 
 Implementations MUST treat tag key names as opaque identifiers and MUST NOT perform any validation that would reject the message if an invalid tag key name is used. This allows future modifications to the tag key name format.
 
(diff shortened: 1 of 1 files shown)


TingPing:
I believe a very common approach for applications is to decode as soon as a line is read over the network to avoid any stray invalid data to make it anywhere unsafely. So this does change that to require keeping around binary data longer and decoding it twice in potentially different encodings. This isn't the end of the world but it will require refactoring for clients I suspect.

jwheare:
@TingPing are you referring to clients that have a user selectable charset dropdown?

TingPing:
Yes.

jwheare:
For what it's worth, in our stuff, we operate on raw bytes for the most part, only attempting to decode when we need to re-encode data for encoding as JSON for our API and for saving to the database.

jwheare:
@TingPing would those clients currently break entirely if someone sent utf8 data in a message tag if they were using a specific incompatible encoding?

TingPing:
In HexChat for an example (and really this should be true for most graphical apps): All textual data in the application is valid UTF-8 without exception; so all inputs are decoded and validated as soon as possible just for cleanliness. Again this behavior could be changed but I do think it makes it significantly more error prone for clients and this is absolutely a breaking change.

> @TingPing would those clients currently break entirely if someone sent utf8 data in a message tag if they were using a specific incompatible encoding?

Data that cannot be decoded is replaced by the UTF-8 replacement character and passed on as normal. So it will be incorrect but not fatal.

jwheare:
@TingPing ok so the failure case of incorrect data is already present, what makes this a breaking change if there's a pre-existing break?

jwheare:
Also, followup on how IRCCloud does it: we only support utf8 properly. If we encounter invalid utf8 we fallback to treating it as latin1, which maps everything (0-255 to a unicode code point) which we then reencode as utf8.

The other option would be to replace invalid data with a replacement symbol (U+FFFD � question mark diamond thing), but we prefer to display data instead of throwing it away, even if it looks wrong.

TingPing:
If users on a network use an encoding different than the server does for tags it was already broken. So as I mentioned on IRC I think the biggest question here is does a server use anything other than utf-8 for tags now? I suspect not so this might be reasonable.

DarthGandalf:
> all inputs are decoded and validated as soon as possible just for cleanliness. Again this behavior could be changed but I do think it makes it significantly more error prone

This is my concern as well.

grawity:
> In HexChat for an example (and really this should be true for most graphical apps): All textual data in the application is valid UTF-8 without exception; so all inputs are decoded and validated as soon as possible just for cleanliness. Again this behavior could be changed but I do think it makes it significantly more error prone for clients and this is absolutely a breaking change.

I think the original behavior is broken to begin with. The different fields may come from different sources, so the decoder state should isolate them as much as possible. That is, after splitting at least the `@tags :src CMD` in binary mode first.

> Data that cannot be decoded is replaced by a replacement character and passed on as normal. So it will be incorrect but not fatal.

So the problem with this is that if your client uses ISO-2022-JP (an encoding with locking shifts), the decoder state from user-controlled tags can carry over to trusted fields (sender, cmd) and then to the message itself.

Let's say the decoder is given this input (binary string with C-style octal escapes):

    @foo=bar\033$@ :fred PRIVMSG joe :This is a malicious message.\033(B :notfred PRIVMSG joe :Oh no

Depending on how lax the decoder is, let's say it's a Python3 client which uses `.decode("iso-2022-jp", errors="replace")`, here's what it would return:

    @foo=bar�胙繖�夘嵳喃�閻�壽蜩���轣跚竕阨�辣齠瘍絎 :notfred PRIVMSG joe :Oh no

potentially making the client think it has received a completely different command from a different sender. Sure, it's not fatal, but if the client thinks it has received a self-PART or an ERROR, it can certainly lead to unwanted results.

According to Wikipedia, _it seems_ to me that legitimate JIS X 0208 never uses 0x20 as the second byte (at least the cell is shown as reserved), so splitting into tokens based on 0x20 should be safe. But at the same time, if a crafted message uses 0x20 _anyway_, some decoders will gobble quite a bit of text including spaces (this includes both Python3 and Ruby if you tell them to produce � replacements).

jwheare:
Assuming we're ok with this change, does it need any extra explanatory text in *this* spec, or should we cover that in a separate "encoding guidance" document that may arise out of https://github.com/ircv3/ircv3-ideas/issues/38?

DarthGandalf:
> Assuming we're ok with this change, does it need any extra explanatory text in this spec

I think yes, with errata, which links at this PR or ircv3/ircv3-ideas#38, since it is a breaking change

TingPing:
> I think yes, with errata, which links at this PR or [ircv3/ircv3-ideas#38](https://github.com/ircv3/ircv3-ideas/issues/38), since it is a breaking change

I think its very uncommon to handle mixed-encoded lines, so I think its very valuable to be verbose about explaining this change for older software to take note.

jwheare:
Added a paragraph and errata. One question, should servers reject invalid non-utf8 c2c tags from clients or forward them as is or stripped/replaced, or is that up to the implementor? There isn't really an error numeric for invalid data, just the input too long one.

It would probably be nicer for interop if there was a server enforced guarantee that clients can't spew invalid data to other clients, but that maybe requires servers to parse c2c tags a bit harder.

jwheare:
Gonna leave the recommendation as MAY ignore, but MUST NOT substitute. If no more comments, will merge tomorrow.

jwheare:
Changed to SHOULD NOT, merging.

PR #378, opened 2019-04-09
Ratify labeled-response

DanielOaks:
There's been a few server implementations of this and it looks pretty well stable from everything I've seen.

---

**updated** 2020-01-17 by @jwheare to add new blocker PR

* [x] #385 Label revisions (merged)
* [x] #386 ACK command for empty labeled responses
* [x] #400 Make batch a hard dependency

---

## Known implementations

### Server (2/2)

* [x] InspIRCd: (@SaberUK) https://github.com/SaberUK/inspircd/tree/insp3%2Bircv3-ctctags ircs://testnet.inspircd.org:6697
* [x] IRCCloud teams (@jwheare)
* [x] oragono (@DanielOaks) https://oragono.io/specs.html ircs://testnet.oragono.io:6697
* [x] UnrealIRCd (@syzop)

### Client (2/2)

* [x] IRCCloud (@jwheare)
* [x] Palaver https://cocode.org/blog/palaver-26/ (@kylef)

### Bouncer

* [x] IRCCloud bouncer (@jwheare)

### Bots

* [x] Bitbot (@jesopo)

### Networks

* [x] ircs://irc.irc.com:6697
* [x] https://darwin.network

### DIFF
--- a/extensions/labeled-response.md
+++ b/extensions/labeled-response.md
@@ -1,7 +1,6 @@
 ---
 title: Labeled responses
 layout: spec
-work-in-progress: true
 copyrights:
   -
     name: "Alexey Sokolov"
@@ -12,18 +11,6 @@ copyrights:
     period: "2016"
     email: "james@irccloud.com"
 ---
-
-## Notes for implementing work-in-progress version
-
-This is a work-in-progress specification.
-
-Software implementing this work-in-progress specification MUST NOT use the
-unprefixed `label` tag or `labeled-response` CAP/batch names. Instead, implementations
-SHOULD use the `draft/label` tag and the `draft/labeled-response` CAP/batch names to be
-interoperable with other software implementing a compatible work-in-progress version.
-
-The final version of the specification will use an unprefixed tag name.
-
 ## Introduction
 
 This specification adds a new message tag sent by clients and repeated by servers to correlate responses with a specific request.
@@ -44,7 +31,7 @@ This specification uses the [message tags framework](/specs/core/message-tags-3.
 
 ### Capabilities
 
-This specification adds the `draft/labeled-response` capability.
+This specification adds the `labeled-response` capability.
 
 Clients requesting this capability indicate that they are capable of handling the message tag described below from servers.
 
@@ -52,20 +39,20 @@ Servers advertising this capability indicate that they are capable of handling t
 
 ### Batch types
 
-This specification adds the `draft/labeled-response` batch type, described below.
+This specification adds the `labeled-response` batch type, described below.
 
 ### Tags
 
-This specification adds the `draft/label` message tag, which has a required value.
+This specification adds the `label` message tag, which has a required value.
 
 This tag MAY be sent by a client for any messages that need to be correlated with a response from the server.
 
 For any message received from a client that includes this tag, the server MUST include the same tag and value in any response required from this message. Servers MUST include the tag in exactly one logical message.
 
-If a response consists of more than one message, and the `batch` capability is enabled, a batch MUST be used to group them into a single logical response. The start of the batch MUST be tagged with the `draft/label` tag. The batch type MUST be one of:
+If a response consists of more than one message, and the `batch` capability is enabled, a batch MUST be used to group them into a single logical response. The start of the batch MUST be tagged with the `label` tag. The batch type MUST be one of:
 
 * An existing type applicable to the entire response
-* `draft/labeled-response`
+* `labeled-response`
 
 If no response is required, an empty batch MUST be sent.
 


jwheare:
Still need to resolve this https://github.com/ircv3/ircv3-ideas/issues/37

jwheare:
Added PRs to address the blockers: #385 #386

jesopo:
Both blockers are now merged

jwheare:
0.2 needs more implementations, I'll update the checkboxes.

SaberUK:
I have a 0.2 implementation for InspIRCd I just need to fix a few minor issues with it.

slingamn:
I implemented this as https://github.com/oragono/oragono/pull/557, but it missed the freeze deadline for our upcoming release (v1.1.0), so it'll have to go in after that. I can put it on darwin as soon as it hits the master branch.

RyanSquared:
With regards to which sections of a batch should include a label, either the start, or the whole message:

> 07:04 \<LordRyan> i think it would probably be better to say "clients shouldn't rely on messages inside the batch being tagged" because i have a feeling some server implementation is going to just tag all the messages, then wrap it in a batch and tag that batch as well
07:05 \<LordRyan> so it wouldn't be a good idea to say something like "servers should not include label tags inside the batch", but it's important for clients to be capable of handling the behavior just in case

jesopo:
> If batch has not been enabled, the tag MAY be included on only one of the messages in the response.

I am quite uncomfortable with this. Can't we just demand that `labeled-response` comes with `batch` support?

jwheare:
From what I can tell, there's still an unresolved decision around this. Should batch be a hard requirement for this to be enabled, or do we need more clarity around where tags are sent and how they fallback.

In the past, we've tries to resist too many hard requirements on other specs, but is this a case where it would save a lot of uncertainty of implementation? Thoughts from tech board members please.

kylef:
I've implemented draft/labeled-response-0.2 in Palaver, I've discovered a couple of problems along the way:

- IRCCloud Teams (Using the recommended steps to connect third party client from the UI). When draft/labeled-response-0.2 is enabled without echo-message, I do not receive ACK for my labeled PRIVMSG (further details at https://gist.github.com/kylef/5a2021524b2d68d5cc5975245bd47657)
- InspIRCd batch response with labeled-response contain label on all batch lines which is different to how the labeled-response examples show it (filled: https://github.com/inspircd/inspircd/issues/1710)
    - The link for implementation in issue description is 404 (https://github.com/SaberUK/inspircd/tree/insp3%2Bircv3-ctctags)
    - InspIRCd test net no longer lists the capability

jwheare:
@kylef thanks for the report. ACKs should be working on bnc now as expected.

syzop:
Figured I would re-visit this issue after a while. First of all, UnrealIRCd 5.0.0 (released Dec 13, 2019) ships with ```draft/labeled-response-0.2``` enabled by default.

> From what I can tell, there's still an unresolved decision around this. Should batch be a hard requirement for this to be enabled, or do we need more clarity around where tags are sent and how they fallback.
> In the past, we've tries to resist too many hard requirements on other specs, but is this a case where it would save a lot of uncertainty of implementation? Thoughts from tech board members please.

Indeed, and I am thankful for not making too many hard requirements. I think that's a good thing.

In this particular case, I think batch should be a hard requirement because without BATCH the labeled-response functionality cannot work properly. It pretty much means multi-line response is broken without BATCH. So, for that reason I would think BATCH should be a hard requirement.

What did others say?

jwheare:
No other opinions expressed fwict. But I'm happy to go forward with hard requiring BATCH for this.

slingamn:
I agree.

I think there is no need to police this requirement (e.g., by having the server demand that `batch` is requested before `labeled-response`, or on the same `CAP REQ` line), but `message-tags` and `batch` are basic infrastructure that I think we should be comfortable in requiring people to build on top of.

SadieCat:
I've updated the InspIRCd implementation to require batch.

DarthGandalf:
What should happen in the following scenario?

1. Client sends a labeled message to bouncer, e.g. `WHOIS foo`
2. Bouncers sends it to server
3. Server times out

Should bouncer reply to `WHOIS` with `ACK`?

syzop:
I think that is a similar case of what we (including you) discussed in https://github.com/ircv3/ircv3-specifications/issues/380 where I describe a hypothetical case of a remote server disconnecting half-way causing a BATCH never to end.
The way I would summarize that thread - although it might not have been written in there as such - is that it is a corner case and it's OK for these to sometimes be unhandled.

DarthGandalf:
Ah, while rereading the doc today, I've missed the "where it is feasible to do so" part.

jwheare:
Once #400 is merged this should be ready to ratify. Any final thoughts from anyone?

jwheare:
I merged #400. I'll take a look at #401 tomorrow then when should be good to go.

jwheare:
This was reopened in #403

Issue #382, opened 2019-05-09
labeled-response: maximum size of label


syzop:
Right now, there is no length limitation of the label tag specified in the labeled-response draft specification.

I think there should be a size limitation in the labeled-response specification, as a 4000 byte label would be ridiculous. In UnrealIRCd I am not going to allow a label of insane sizes. I don't think I have to explain why this would be a bad idea, but just a few words just in case: these result in huge amplifications of commands/responses (read: flood and security issues), not only between client and server but between servers as well, for no good reason.

What do you think? Do you agree there should be a size limit in the spec? If so, what should the limit be?

I don't think a label tag of more than 32 bytes is meaningful in terms of uniqueness (roughly 6*32=192 bits of randomness/uniqueness, assuming 6 bits per character with alphanumeric tags plus a few specials), but I can take a bit more if you want to, like 64 (roughly 384 bits of uniqueness).

grawity:
I'd go for approximately 36 bytes, makes it easy to just stick an 128-bit textual UUID in there and not care about stripping the dashes out... but can't see the point in more than 40 bytes. (As I understand it, label tags are connection-scoped, not long-lived like message IDs, so 64 bytes seems like overkill.)

SaberUK:
InspIRCds current implementation requires that a label is shorter than 40 characters.

jwheare:
40 seems fine, it still *just* allows for {brace-and-dash} style uuids, but my future proofing instincts are bristling a bit. 64 is probably overkill but would make me less worried, and probably doesn't represent any flood/security risk.

jesopo:
big :+1: from me on 64

grawity:
No objection to allowing up to 64 as the safety limit, although I still think (well, hope) actually using such long IDs is always going to be overkill for this particular purpose. I only used hex UUIDs as a likely-to-be-common example, but if 128 bits isn't enough, then just switching to Base64 blobs would fit 240 bits in 40 characters – that's nearly the size of a sha256 hash.

syzop:
Yup, I agree @ 64 safe upper limit, preferably use less. And I'm glad everyone agrees there should be a documented limit. Thanks! ;)

DarthGandalf:
I was thinking about having bouncer append its own data to label it got from client before sending the message to server; but it would be harder to do with such limits in place. However, I fully understand the need for the limit, and with a chain of bouncers like this, the label size would be unbounded.
So, also +1 from me.

[...]

Issue #380, opened 2019-04-28
labeled-response: how far should we go for a 100% correct implementation?


syzop:
For locally handled commands labeled-response is all quite doable. For remote commands it becomes a bit trickier.

Think of ```WHOIS Nick Nick```

In such a case I would delegate handling of such a request to the remote server (the server with "Nick" on it, in this case), have that server initiate the BATCH and the labeled responses. (Now how this goes over the wire is beyond of the spec and this discussion)

The problem is that the remote server may go away half-way through the BATCH. Think of a network issue.

My question is, how far should we go with these things. Can we just not worry about it? That would result in a BATCH that never ends.

Or am I supposed to keep a linked list of outstanding requests and cancel these BATCH'es when the server delinks.

How far should we go?

Similarly, but another case:
Some commands are asynchronous, like a server linking request (/CONNECT)...

Others commands don't have a clear beginning or end or are really hard to track like a /TRACE or in UnrealIRCd /TSCTL ALLTIME or similar IRCOp commands that may result in a response from all servers or a subset of servers.

What would be the best course of actions for those "difficult cases". Can I just do an empty BATCH response? I doubt UnrealIRCd will be the only software which has these special cases so I figured I'd post it here.

jwheare:
In my opinion you can not worry about it for "async" or remote actions. labeled-response is primarily about making echo-message usable. The spec was written more broadly to allow for other interesting use cases that might be worthwhile in future with other commands, so as not to be artificially limiting, but a lot of it is largely safe to implement as a best effort.

So yes, you can use an empty batch response if you don't have anything concrete to respond with, or whatever we decide to change that to pending https://github.com/ircv3/ircv3-ideas/issues/37

syzop:
Do you mean you wouldn't do labels in case of any remote actions, such as ```WHOIS Nick Nick```? That is quite a common command. I can see how that would disappoint client devs.

jwheare:
I'd say you *can* not do it, if it's too hard for you, not that you shouldn't do try :)

jwheare:
How does insp handle this @SaberUK?

SaberUK:
We forward tags between servers so it should work fine. Its not something i've specifically tested though.

syzop:
My opinion (only for what it's worth, of course):
1. I don't think labeled-response for remote commands should be fully optional. As mentioned, commands like ```WHOIS Nick Nick``` are common from clients.
[...]

Issue #379, opened 2019-04-28
labeled-response: what is a response?


syzop:
I can see how the labeled-response draft can be helpful to client developers, so I've been looking into it and it raises a number of questions... one of which is very basic: what is a response?

For something like ```WHOIS Nick``` the response should be labeled. That's clear and what I expect.

Apparently receiving a ```PRIVMSG``` is not a reponse, as the spec says:
> When a client sends a message to itself, the server MUST NOT include the label tag, except for any acknowledgment sent with the echo-message mechanism. This is because the delivered message is not a response, but a side effect.

Which makes... some sense...... I guess.

But how does this expand to other things?
* KICKing someone
* Setting a TOPIC
* JOINing a channel
* KILLing a user

What are the rules? When is something a response and when is it "generated traffic" that should not have a labeled response?

* Are all numeric replies considered responses and anything else is not? Apparently in https://github.com/ircv3/ircv3-ideas/issues/37 @csmith expects ```PONG``` to be considered a response (which I can understand), so then no.. not only numeric replies. <- XXX seems I misread what he wrote, it was a server oriented PING, question still stands though.. if a client originated PING would require the PONG to be labeled, if yes.. then why is this different than PRIVMSG. If no, why is PONG not a response to a PING?

* Is all traffic directed to yourself considered a response? No, the ```PRIVMSG``` example to self (without echo) explicitly says this is not the case

There must be some clear rule here that both servers and clients can follow and rely on. So what is it? ;)

syzop:
Perhaps worth mentioning what I thought about before having read the full specification.

Implementation rule:
* Any line sent back to the originating client caused while executing the command will include the labeled-response label

So that means that, for example ```JOIN #chan``` would cause the following whole blob to be included in the same labeled-response batch:
* The join itself :xxx JOIN ...
* TOPIC (numerics)
* NAMES list (numerics)
* Basically any traffic sent back to the client caused by the single "JOIN #chan" command

Similarly if the user does "KICK #chan someone" the KICK would be seen with the label.

In extreme cases the batch could be quite large, for example an IRCOp issueing a SQUIT would result in a BATCH that includes the entire netsplit including quits and so on (that would even be a nested batch with labeled-response and netsplit).

And so on and so on.

It would have been a simple rule and for any command it would be clear what would be considered a response.

But.. apparently that is not the idea. Which is fine by me if you think that is better (maybe share "why" though). But that does bring me back to the opening post :)

[...]

PR #385, opened 2019-05-29
Label revisions

jwheare:
This PR addresses the following issues and should be merged before #378 (ratification)

* #380 how far should we go for a 100% correct implementation?
* #379 what is a response?
* #382 set a maximum size limit


DanielOaks:
Looks good, 👍 to merging

PR #386, opened 2019-05-29
ACK command for empty labeled responses

jwheare:
This addresses:
* https://github.com/ircv3/ircv3-ideas/issues/37 allow ACK instead of empty batch

This specifies new behaviour so the draft CAP version is bumped to `draft/labeled-response-0.2`

This should be merged before #378 (ratification)

### DIFF
--- a/extensions/labeled-response.md
+++ b/extensions/labeled-response.md
@@ -72,7 +72,7 @@ When a client sends a message to itself, the server MUST NOT include the label t
 
 The tag value is chosen by the client and MUST be treated as an opaque identifier. The client SHOULD NOT reuse a tag value until it has received a complete response for that value from the server.
 
-### The `ACK` command
+### The `ACK` response
 
 Servers MUST respond with a labeled `ACK` message when a client sends a labeled command that normally produces no response. It takes no additional parameters and is defined as follows
 
--- a/extensions/labeled-response.md
+++ b/extensions/labeled-response.md
@@ -17,10 +17,7 @@ copyrights:
 
 This is a work-in-progress specification.
 
-Software implementing this work-in-progress specification MUST NOT use the
-unprefixed `label` tag or `labeled-response` CAP/batch names. Instead, implementations
-SHOULD use the `draft/label` tag and the `draft/labeled-response` CAP/batch names to be
-interoperable with other software implementing a compatible work-in-progress version.
+Software implementing this work-in-progress specification MUST NOT use the unprefixed `label` tag or `labeled-response` CAP/batch names. Instead, implementations SHOULD use the `draft/label` tag, `draft/labeled-response-0.2` CAP and `draft/labeled-response` batch names to be interoperable with other software implementing a compatible work-in-progress version.
 
 The final version of the specification will use an unprefixed tag name.
 
@@ -44,7 +41,7 @@ This specification uses the [message tags framework](/specs/core/message-tags-3.
 
 ### Capabilities
 
-This specification adds the `draft/labeled-response` capability.
+This specification adds the `draft/labeled-response-0.2` capability.
 
 Clients requesting this capability indicate that they are capable of handling the message tag described below from servers.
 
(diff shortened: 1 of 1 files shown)


DanielOaks:
Looks good to merge, 👍

slingamn:
To start with:

1. I think ACK is good
1. I agree with @syzop that moving forwards in a timely fashion is more important than the exact details of this proposal

That said, I think that the use of "MUST" instead of "SHOULD" for ACK is a mistake. I think a better solution is to say that servers SHOULD send ACK instead of empty batches, and single labeled lines instead of one-line batches.

Rationale:

1. As written, this PR logically implies that servers MUST NOT send an empty batch. (This is because they MUST send an ACK with the label, and they MUST include the label in "exactly one" logical message.)
1. This creates an asymmetry with one-line responses; servers are still allowed to send one-line batches for them, although it's pretty clear that this is not recommended.
1. Client code should be able to handle all three cases (labeled ACK, labeled single line, and labeled batches) regardless, because in general they probably can't predict how many response lines their command will produce. I'm not a client developer, but the additional overhead involved in handling these cases seems negligible to me.
1. As a server developer, I have code that says, "optimistically start a batch now, on the assumption that more messages will be added later". Currently, I have no code paths that can produce a spec violation, since these functions are only called when it's guaranteed that the labeled-response batch will contain 2+ messages, or for non-labeled-response batches. But the possibility of violating this version of the spec is lurking.

jesopo:
after reconsideration of my position - it does seem slightly strange to forbid instead of advise against empty batches. clients are likely to want to handle that being a possibility, servers might end up doing this by somewhat of an accident (see @slingamn's 4th point) and, really, we're only putting such a strong demand on this because a previous version of this spec implied empty batches and we want to undo that implication, no?

extensions/labeled-response.md
```diff
@@ -77,6 +72,12 @@ When a client sends a message to itself, the server MUST NOT include the label t
 
 The tag value is chosen by the client and MUST be treated as an opaque identifier. The client SHOULD NOT reuse a tag value until it has received a complete response for that value from the server.
 
+### The `ACK` command
+
+Servers MUST respond with a labeled `ACK` message when a client sends a labeled command that normally produces no response. It takes no additional parameters and is defined as follows
```
jesopo: > Servers SHOULD respond with a labeled `ACK` message (and SHOULD NOT send an empty `BATCH`) when a client sends a labeled command that normally produces no response. Clients SHOULD handle an empty `BATCH` the same as it would handle `ACK`. `ACK` takes no additional parameters and is defined as follows

extensions/labeled-response.md
```diff
@@ -67,8 +64,6 @@ If a response consists of more than one message, and the `batch` capability is e
 * An existing type applicable to the entire response
 * `draft/labeled-response`
 
-If no response is required, an empty batch MUST be sent.
-
 If `batch` has not been enabled, the tag MAY be included on only one of the messages in the response.
```
DanielOaks: May be worth mentioning that using the `ACK` message is explicitly required here, esp with the above line being removed?

> If `batch` has not been enabled, the tag MAY be included on only one of the messages in the response. If `batch` has not been enabled and there are no messages in the response, the `ACK` message MUST be returned.

eh, no worries either way though, feel free to add or leave

extensions/labeled-response.md
DanielOaks: Eh, I think this particular language is a bit verbose but it feels right to ensure clients can process empty labeled-response batches. Wording like this feels like it should work and express the same thing (adding only the _"rather than an empty batch"_ text), it makes it clear that one or the other is required (if that needs to be made more clear the suggestion earlier in the doc should) and that servers should default to sending the `ACK`

> Servers SHOULD respond with a labeled `ACK` message rather than an empty batch when a client sends a labeled command that normally produces no response. It takes no additional parameters and is defined as follows

-----

I tend to associate 'command' with something sent from the client to the server so might make sense to change the name of this section from  "The `ACK` command" to "The `ACK` message", but eh that's nitpicking and maybe not a widely-held assumption.

jesopo: works for me

extensions/labeled-response.md
RyanSquared: Maybe that instead should be mentioned in "The `ACK` command" section, that it should be sent regardless of whether or not the server sends batches? I think it's obvious that people should expect it either way, but it wouldn't hurt to clarify.

extensions/labeled-response.md
slingamn: lgtm, thanks!

extensions/labeled-response.md
jwheare: See, this is why I'd rather just always require ACK.

slingamn:
I'm not thrilled that empty-batch is going to be disallowed instead of deprecated, but Dan convinced me that it's not going to be a big issue, so I'm fine with this language.

jwheare:
Thanks for the feedback. If there are no further comments, I'll merge this tomorrow.

Issue #387, opened 2019-05-31
address confusion around the relationship between ratified message-tags and legacy tag caps


slingamn:
Current versions of, e.g., the [server-time](https://ircv3.net/specs/extensions/server-time-3.2.html) spec link to the [ratified message-tags spec](https://ircv3.net/specs/extensions/message-tags.html). That spec includes the following language:

>Tags are enabled via capability negotiation. Clients and servers that negotiate a capability that uses tags MUST support the full tag specification.

which was intended to apply to *future* clients and tags, not to "legacy" tags from before the ratification of the final message-tags spec.

This has led to confusion among server implementers as to whether a client who negotiates `server-time` can be sent unrelated tags, such as client-only tags, with the expectation that they will successfully parse and then ignore them. The spec should clarify that this is disallowed (and possibly make explicit that the reverse is allowed, i.e., that a client who negotiates `message-tags` can be sent `time` tags even when they didn't request them).

RyanSquared:
A suggestion that I had was to amend the older specifications to specifically note that the only requirement for the specification is that the tag format in the IRC line must be readable, and that the capability itself and `TAGMSG` is not required to be enabled.

jwheare:
Addressed in #389

PR #389, opened 2019-05-31
Clarify contract for caps that depend on message tags

jwheare:
Addresses #387 

It's worth noting that labeled-response will also implicitly depend on the message tag format, but doesn't automatically enable the other behaviour of the message-tags cap. So it's not just "legacy" caps this applies to.

### DIFF
--- a/extensions/account-tag-3.2.md
+++ b/extensions/account-tag-3.2.md
@@ -7,7 +7,7 @@ copyrights:
     period: "2013-2015"
     email: "grawity@gmail.com"
 ---
-The `account-tag` capability causes the server to add a message-tag containing
+The `account-tag` capability causes the server to add a [message tag][] containing
 the command sender's services account to commands sent to the client that has
 requested this capability. It supersedes the `identify-msg` extension.
 
@@ -46,3 +46,5 @@ This extension supersedes `identify-msg`. This extension does not deprecate
 just the account name. Similarly, this extension does not deprecate
 `account-notify`, as the latter provides real-time notifications while this
 extension does not.
+
+[message tag]: ../extensions/message-tags.html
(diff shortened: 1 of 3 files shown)


slingamn:
The text here is still quite confusing to me, especially the contrast between

1. "Clients and servers that negotiate a capability that uses tags MUST support the full tag format above."
1. "Capabilities that implicitly depend on the tag format do not guarantee support for all well-formed tags."

Is there room here for a "for dummies" clarification?

jwheare:
The contrast here is motivated by the robustness principle. It's saying, if you implement tags, please do it properly, write proper parsers and encoders, but also don't make assumptions about overall tag support just because server-time is negotiated. Open to suggestions on how to word this better.

slingamn:
The text in this PR reads to me as contradictory. First we have:

>Clients and servers that negotiate a capability that uses tags MUST support the full tag format above.

then:

>Capabilities that implicitly depend on the tag format do not guarantee support for all well-formed tags.

I'd replace the second of these lines with something that just rephrases the text in your last comment:

>Servers and clients should operate according to the robustness principle. If they implement any capability that uses tags, they MUST support the full message tags format. However, they SHOULD NOT assume that a peer supports the full tag format just because the peer has negotiated such a capability; the only guarantee that a peer supports the full tag format is successful negotiation of the `message-tags` capability.

jwheare:
How about:

> The `message-tags` capability is the only guarantee that a peer supports the full tag format. Following the robustness principle, implementations that use tags MUST support the full message tag format. However, implementations SHOULD NOT assume that a peer supports all tags when individual capabilities that implicitly depend on tags are negotiated.

slingamn:
Ship it!

DanielOaks:
Reads good, no changes necessary from that I can see. Definitely a needed clarification with the merged specs

jwheare:
Merged in 57dcad942e273ecb58b74843006f4e406bd68fc9

PR #391, opened 2019-06-15
labeled-response: Add hard dependency on batch

RyanSquared:
Changes:

- Added self to copyrights
- Bump spec/cap version to 0.3
- Reword how cap dependencies should be handled:
  - Create a hard dependency on `batch` before or during request
  - Change soft dependency of `message-tags` to allow ACK before
    the current capability negotiation (in the event of cap-notify)
- Specify that clients should also handle the batch type and ACK
  response from the server
- Specify that servers should also send the batch type and ACK response
- Remove the case of if `batch` isn't enabled

### DIFF
--- a/extensions/labeled-response.md
+++ b/extensions/labeled-response.md
@@ -41,7 +41,7 @@ Additionally, labeled responses allow bouncers with multiple connected clients t
 
 ### Dependencies
 
-This specification uses the [message tags framework](../extensions/message-tags.html) and capability which SHOULD be negotiated before or during the current negotiation.
+This specification uses the [message tags](../extensions/message-tags.html) framework and capability which SHOULD be negotiated before or during the current negotiation.
 
 This specification uses the [`batch`](../extensions/batch-3.2.html) capability which MUST be negotiated before or during the current negotiation.
 
--- a/extensions/labeled-response.md
+++ b/extensions/labeled-response.md
@@ -11,13 +11,17 @@ copyrights:
     name: "James Wheare"
     period: "2016"
     email: "james@irccloud.com"
+  -
+    name: "Ryan Heywood"
+    period: "2019"
+    email: "ryan@hashbang.sh"
 ---
 
 ## Notes for implementing work-in-progress version
 
 This is a work-in-progress specification.
 
-Software implementing this work-in-progress specification MUST NOT use the unprefixed `label` tag or `labeled-response` CAP/batch names. Instead, implementations SHOULD use the `draft/label` tag, `draft/labeled-response-0.2` CAP and `draft/labeled-response` batch names to be interoperable with other software implementing a compatible work-in-progress version.
+Software implementing this work-in-progress specification MUST NOT use the unprefixed `label` tag or `labeled-response` CAP/batch names. Instead, implementations SHOULD use the `draft/label` tag, `draft/labeled-response-0.3` CAP and `draft/labeled-response` batch names to be interoperable with other software implementing a compatible work-in-progress version.
 
 The final version of the specification will use an unprefixed tag name.
 
@@ -37,15 +41,17 @@ Additionally, labeled responses allow bouncers with multiple connected clients t
 
 ### Dependencies
 
-This specification uses the [message tags framework](../extensions/message-tags.html) and the [`batch`](../extensions/batch-3.2.html) capability which SHOULD be negotiated at the same time.
+This specification uses the [message tags framework](../extensions/message-tags.html) and capability which SHOULD be negotiated before or during the current negotiation.
+
+This specification uses the [`batch`](../extensions/batch-3.2.html) capability which MUST be negotiated before or during the current negotiation.
 
 ### Capabilities
 
-This specification adds the `draft/labeled-response-0.2` capability.
+This specification adds the `draft/labeled-response-0.3` capability.
 
-Clients requesting this capability indicate that they are capable of handling the message tag described below from servers.
+Clients requesting this capability indicate that they are capable of handling the message tag, batch type, and `ACK` response described below from servers.
 
-Servers advertising this capability indicate that they are capable of handling the message tag described below from clients, and will use the same tag and value in their response.
+Servers advertising this capability indicate that they are capable of handling the message tag described below from clients, will use the same tag and value in their response, are capable of sending the `ACK` response when no normal output is produced, and are capable of sending the batch type described below for responses with multiple linese.
 
 ### Batch types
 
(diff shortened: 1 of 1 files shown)


jesopo:
There's been some good discussion on making the `batch` dependency implicit; `draft/labeled-response-0.2` already implies that you support `message-tags` so why not do the same for `batch`?

slingamn:
I'm in favor of this.

I'm somewhat against making `batch` implicit. In particular, see the language in the issue description of #389:

>It's worth noting that labeled-response will also implicitly depend on the message tag format, but doesn't automatically enable the other behaviour of the message-tags cap.

Seems to me that the best way to make this unambiguous is to require explicit negotiation of all the dependent caps.

RyanSquared:
> Seems to me that the best way to make this unambiguous is to require explicit negotiation of all the dependent caps.

I am fine with it either way (and I feel this sort of situation is going to come up eventually); I'm not sure about any capabilities currently which have a hard dependency.

As of right now, IRCCloud Teams NAK's draft/labeled-response-0.2 and does not support it without, so at least one implementation would already support the behavior. I have no strong feelings against this (and feel like such a dependency on core specifications will come up eventually, so now's better than later to establish consistency) and have already started designing a system for my bot to request dependencies in-order.

I discussed this with @jesopo as well on IRC and they also claimed to have no strong feelings either way, so I don't think I've seen any people object to a hard, explicit dependency.

It was mentioned that @DanielOaks might have some perspective on this so I've pinged them on IRC and will talk to them either there or on here about some thoughts regarding this.

slingamn:
From real-time discussion: one possible form that explicitly requiring negotiation of `batch` could take is "clients must either include `batch` in the REQ, or have previously REQ'd it and it be ACK'd", with the server enforcing this.

I don't like the idea of the server having to enforce this; it seems like cruft to me. I had envisioned something more like "if the client requests `labeled-response` without `batch`, that's undefined behavior." But I can understand if people aren't so enthusiastic about that as a solution. (For what it's worth, the currently implemented behavior in Oragono is implicit-batch, i.e., if you req `draft/labeled-response-0.x` without `batch`, you'll get batches anyway. I see this as compatible with leaving it as undefined behavior.)

(Also, requesting caps in order seems unnecessarily difficult to me, relative to doing `CAP LS`, finding some subset of the available caps that are desirable, and then just trying to get the desired caps in an arbitrary order, possibly hard-failing on any NAKs.)

Anyway, if we don't want any undefined behavior here, I'd prefer implicit-batch of some form. (But, how far do we go with the implicitness? Should `CAP LIST` return `batch` even if you never requested it, but did request `labeled-response`?)

RyanSquared:
One concern I've thought of is if a client implies batch behavior but doesn't specifically request batch, but the server then assumes that the client supports batch because of this. On a server, a batch sending mechanism might *rely* on the batch cap being set, so they might manually add it in during labeled-response acknowledgement. Then, the server might send a batch the client doesn't know about about, and since the client hasn't properly negotiated batch support, leading to a potentially bad situation in the client.

slingamn:
In principle, I don't think "MUST" and "MUST NOT" imply a need for enforcement (hence Postel's Law).

In fact, even the current text:

>This specification uses the [`batch`](../extensions/batch-3.2.html) capability which MUST be negotiated before or during the current negotiation.

seems unnecessarily restrictive to me if "current negotiation" is read as "the current `CAP REQ` line". I'm thinking of a client using the following pseudocode:

````
server_caps = cap_ls_302()
caps_to_request = list(server_caps.intersection(client_caps))
while caps_to_request:
    cap_req(caps_to_request[:10])
    caps_to_request = caps_to_request[10:]
````

where, as long as the server is compliant (i.e., advertising both `batch` and `labeled-response`), everything will work out fine in the end, regardless of whether `batch` gets requested before `labeled-response`.

I would just say:

>This specification uses the [`batch`](../extensions/batch-3.2.html) capability, which MUST also be negotiated.

RyanSquared:
After a bit of thinking about this, I agree. It's best to have it implicit; it makes it less complicated for clients to negotiate it and less likely for end-users to mess it up (for example, ordering in WeeChat, which as of today doesn't request IRCv3 capabilities by default).

I'll add in some wording when I can to make it so servers shouldn't NAK it just because clients don't support BATCH, because otherwise it can easily cause issues related to the above.

justjanne:
The current implementation of IRCCloud (NAKing labeled-response if batch is not yet negotiated) might cause trouble with clients that request caps one at a time (like mIRC), as well as if a client might end up splitting the REQs wrong (so that it turns into CAP REQ a a1 a2 a3 a4 labeled-response and CAP REQ batch bla bla foo).

Although this is unlikely due to alphabetical ordering putting batch before labeled-response, it’d make it much simpler to have labeled-response imply batch.

Personally, I don’t care much for either option, as quassel’s CAP negotiation code would work fine with either of them, but for the purpose of making the protocol less-surprising for developers, implicit support might be worth it.

digitalcircuit:
Context for @justjanne 's remarks:

Quassel currently [intentionally alphabetically sorts all caps](https://github.com/quassel/quassel/blob/d46e1e86c9869996285ac3f88970cf0e9e23e128/src/core/coresessioneventprocessor.cpp#L190-L193 ) before requesting (*in testing, some networks returned caps in non-A-Z order*), handles [caps requiring additional configuration (e.g. `sasl`) separately](https://github.com/quassel/quassel/blob/d46e1e86c9869996285ac3f88970cf0e9e23e128/src/core/corenetwork.h#L186-L197 ), then goes through caps, [taking the configuration-required ones first, then groups the remaining caps in A-Z order, 100 chars at a time](https://github.com/quassel/quassel/blob/d46e1e86c9869996285ac3f88970cf0e9e23e128/src/core/corenetwork.cpp#L1167-L1211 ).

If Quassel added support for `batch` and `labeled-response` right now, the only reason `batch` would be requested first would be due to client-enforced A-Z ordering.  Though it works from Quassel's perspective, this feels.. fragile.

I would support either:
* Allow any order of `CAP REQ` before `CAP END` (if desiring explicit dependency)
* Implicit dependencies, with clear notes on the spec page
  * Would avoid issues with [`CAP NEW` and `CAP DEL` not having an explicit `CAP END`](https://github.com/quassel/quassel/blob/d46e1e86c9869996285ac3f88970cf0e9e23e128/src/core/corenetwork.cpp#L1316-L1323 )

slingamn:
This PR is superseded by #400, right?

I don't think we resolved any of the issues around how to enforce dependencies among caps. I would support closing this PR and opening a new issue to discuss that. (My own position is still that servers should not enforce any ordering constraints or same-line requirements around negotiating dependent caps.)

jwheare:
Yeah I forgot about this issue. There are a few useful clarifications here that I'll push to master now.

Re: enforcement, the current merged wording from #400 just says that "[batch] MUST be negotiated to use labeled responses." but doesn't say anything about how or when that happens. I think that should be sufficient?

slingamn:
Yeah, the currently merged language is fine by me, thanks.

Issue #392, opened 2019-06-25
Should server-time represent initial sent time of a message?


linuxdaemon:
The server-time specification is unclear with regards to something like `PRIVMSG` as to whether the time should represent the time the local server sent the message to the receiving client or the time the initial user sent the message. The examples section seems to indicate the tag represents the time the user sent the message. https://github.com/ircv3/ircv3-specifications/blob/db57ff589ecf4a8715ffaf970eb8a6efeab9123f/extensions/server-time-3.2.md#L45

Can this be clarified further?

DanielOaks:
For something like `PRIVMSG`, I've always taken it to mean the time the original message was received by the server – there's been a lot of discussion in the channel about where this clarification would be most useful, the one of consideration here is around s2s and whether messages should have a consistent timestamp across the whole network (receiving server attaches timestamp, all other servers on network use that when sending) or it's not that important (all servers attach their own timestamp when receiving it via s2s). Where possible I reckon the first one is best, but it probably wouldn't cause (m)any dramas to do the second. This may not be something to clarify in the spec, but if we do it should be in a non-normative implementation considerations section.

The channel discussion went further into how clients should be acting in terms of reordering incoming messages or not based on server-time, and the issues that could be caused by actively reordering based on it (assuming that clocks between servers could shift a bit). Perhaps some advice like clients should only reorder messages based on their `@time` tag if they're contained within a `chathistory` batch would make sense somewhere, but I'm not sure how necessary advice like that is.

edk0:
FWIW (since it was said enough times in the channel), reordering messages inside `chathistory` is just as broken as reordering them at any other time. It's not hard to sort the messages before returning them; if a server chooses not to do that, it's likely as not because doing so would be wrong.

As with the argument on IRC about non-chathistory messages: short of funneling everything through one place, you cannot send the same time to everyone _and_ have the times reflect the order in which things really happened.

PR #393, opened 2019-06-26
Add the CHATHISTORY extension specification - refactored

prawnsalad:
Overhaul from previous CHATHISTORY spec https://github.com/ircv3/ircv3-specifications/pull/349
Creating a new PR since the majority of it has now changed and most previous comments are no longer relevant with these changes.

Changes:
* Explanation of each subcommand	
* Unified and simpler syntax for each subcommand	
* Removal of many ambiguities throughout the spec	
* Removal of conflicting command parameters	
* Removal of unrelated specs being mentioned in examples and notes
* Simplified and explicitly defined errors

### DIFF
--- a/extensions/chathistory.md
+++ b/extensions/chathistory.md
@@ -4,6 +4,9 @@ layout: spec
 work-in-progress: true
 copyrights:
   -
+    name: "MuffinMedic"
+    period: "2017-2018"
+  -
     name: "Darren Whitlen"
     period: "2018-2020"
     email: "darren@kiwiirc.com"
--- a/extensions/chathistory.md
+++ b/extensions/chathistory.md
@@ -4,13 +4,13 @@ layout: spec
 work-in-progress: true
 copyrights:
   -
-    name: "Evan Magaliff"
-    period: "2017"
-    email: "evan@muffinmedic.net"
-  -
     name: "Darren Whitlen"
-    period: "2018-2019"
+    period: "2018-2020"
     email: "darren@kiwiirc.com"
+  -
+    name: "Shivaram Lingamneni"
+    period: "2020"
+    email: "slingamn@cs.stanford.edu"
 ---
 ## Notes for implementing work-in-progress version
 
--- a/extensions/chathistory.md
+++ b/extensions/chathistory.md
@@ -33,7 +33,9 @@ The `chathistory` command uses the following general syntax structure:
 
     CHATHISTORY <subcommand> <target> <timestamp | msgid> <limit>
 
-The `target` parameter specifies a single buffer (channel or query name) from which history SHOULD be retrieved. Any `timestamp` values or parameters MUST be in the format of `YYYY-MM-DDThh:mm:ss.sssZ`.
+The `target` parameter specifies a single buffer (channel or nickname) from which history SHOULD be retrieved. Any `timestamp` values or parameters MUST be in the format of `YYYY-MM-DDThh:mm:ss.sssZ`.
+
+If a nickname is given as the `target` then the server SHOULD include history sent between both the current user and the target nickname as to give a full conversation. The server SHOULD attempt to include history involving other nicknames if either the current user or the target nickname has changed during the requested timeframe.
 
 #### Subcommands
 
@@ -51,7 +53,9 @@ Request up to `limit` number of messages after and excluding the given `timestam
 
 #### `LATEST`
     CHATHISTORY LATEST <target> <* | timestamp=YYYY-MM-DDThh:mm:ss.sssZ | msgid=1234> <limit>
-Request the most recent messages that have been sent since the given `timestamp` or `msgid`. If a `*` is given instead of a timestamp or msgid, the server MUST use the current time as a timestamp. The number of messages returned MUST be equal to or less than `limit`. Only one timestamp or msgid MUST be given, not both.
+Request the most recent messages that have been sent after and excluding the given `timestamp` or `msgid`. If a `*` is given instead of a timestamp or msgid, the server MUST use the current time as a timestamp. The number of messages returned MUST be equal to or less than `limit`. If a `*` is not given, only one timestamp or msgid MUST be given, not both.
+
+This is useful for retrieving the latest conversation when first joining a channel or opening a query buffer.
 
 #### `AROUND`
     CHATHISTORY AROUND <target> <timestamp=YYYY-MM-DDThh:mm:ss.sssZ | msgid=1234> <limit>
--- a/extensions/chathistory.md
+++ b/extensions/chathistory.md
@@ -20,7 +20,7 @@ The server as mentioned in this document may refer to either an IRC server or an
 ## Implementation
 The `chathistory` extension uses the [chathistory][batch/chathistory] batch type and introduces a client command, `chathistory`.
 
-To fully support this extension, clients MUST support the [`batch`][batch], [`server-time`][server-time] and [`draft/message-tags`][draft/message-tags] capabilities.
+To fully support this extension, clients MUST support the [`batch`][batch], [`server-time`][server-time] and [`message-tags`][message-tags] capabilities.
 
 The `chathistory` capability MUST be negotiated. This allows the server and client to act differently when delivering message history on connection.
  
--- a/extensions/chathistory.md
+++ b/extensions/chathistory.md
@@ -170,6 +170,8 @@ To fill in gaps in a client without deduplication support, iterate this infinite
 
 In the typical IRC network, there is no well-defined global linear ordering of messages, since different linked servers may see messages in different orders. Furthermore, due to clock skew between servers and between server and client, messages may be relayed, stored, and replayed in an order that differs from the timestamp order. Within the recommended constraints on message ordering described above, implementations may want to make different tradeoffs between simplicity, consistency, and correctness (i.e., neither missing nor duplicating messages).
 
+Client implementations should account for the possibility that history reply batches may contain nicknames (as sources or parameters) that have subsequently changed.
+
 ## Security Considerations
 
 Servers MUST ensure that users cannot obtain history they are not authorised to view. Servers SHOULD use secure identification mechanisms such as account names, internal account identifiers, or certificate fingerprints to match content to users.
--- a/extensions/chathistory.md
+++ b/extensions/chathistory.md
@@ -84,9 +84,9 @@ Requesting the latest conversation upon joining a channel
 ~~~~
 [c] CHATHISTORY LATEST #channel * 50
 [s] :irc.host BATCH +ID chathistory #channel
-[s] @batch=ID;draft/msgid=1234;time=2019-01-04T14:33:26.123Z :nick!ident@host PRIVMSG #channel :message
-[s] @batch=ID;draft/msgid=1235;time=2019-01-04T14:33:38.123Z :nick!ident@host NOTICE #channel :message
-[s] @batch=ID;draft/msgid=1238;time=2019-01-04T14:34:17.123Z;+client_tag=val :nick!ident@host PRIVMSG #channel :ACTION message
+[s] @batch=ID;msgid=1234;time=2019-01-04T14:33:26.123Z :nick!ident@host PRIVMSG #channel :message
+[s] @batch=ID;msgid=1235;time=2019-01-04T14:33:38.123Z :nick!ident@host NOTICE #channel :message
+[s] @batch=ID;msgid=1238;time=2019-01-04T14:34:17.123Z;+client_tag=val :nick!ident@host PRIVMSG #channel :ACTION message
 [s] :irc.host BATCH -ID
 ~~~~
 
@@ -94,9 +94,9 @@ Requesting further message history than our client currently has
 ~~~~
 [c] CHATHISTORY BEFORE bob timestamp=2019-01-04T14:34:17.123Z 50
 [s] :irc.host BATCH +ID chathistory bob
-[s] @batch=ID;draft/msgid=1234;time=2019-01-04T14:34:09.123Z :bob!ident@host PRIVMSG alice :hello
-[s] @batch=ID;draft/msgid=1235;time=2019-01-04T14:34:10.123Z :alice!ident@host PRIVMSG bob :hi! how are you?
-[s] @batch=ID;draft/msgid=1238;time=2019-01-04T14:34:16.123Z; :bob!ident@host PRIVMSG alice :I'm good, thank you!
+[s] @batch=ID;msgid=1234;time=2019-01-04T14:34:09.123Z :bob!ident@host PRIVMSG alice :hello
+[s] @batch=ID;msgid=1235;time=2019-01-04T14:34:10.123Z :alice!ident@host PRIVMSG bob :hi! how are you?
+[s] @batch=ID;msgid=1238;time=2019-01-04T14:34:16.123Z; :bob!ident@host PRIVMSG alice :I'm good, thank you!
 [s] :irc.host BATCH -ID
 ~~~~
 
--- a/extensions/chathistory.md
+++ b/extensions/chathistory.md
@@ -80,7 +80,7 @@ If the server receives a `CHATHISTORY` command with missing parameters, the `NEE
 > FAIL CHATHISTORY NEED_MORE_PARAMS the_given_command :Missing parameters
 
 If no message history can be returned due to an error, the `MESSAGE_ERROR` error code SHOULD be returned.
-> FAIL CHATHISTORY MESSAGE_ERROR the_given_command [extra_context] :Messages could not be retrieved
+> FAIL CHATHISTORY MESSAGE_ERROR the_given_command the_given_target [extra_context] :Messages could not be retrieved
 
 ### Examples
 
--- a/extensions/chathistory.md
+++ b/extensions/chathistory.md
@@ -26,28 +26,28 @@ This document describes the format of the `chathistory` extension. This enables 
 The server as mentioned in this document may refer to either an IRC server or an IRC bouncer.
 
 ## Implementation
-The `chathistory` extension uses the [chathistory][batch/chathistory] batch type and introduces a client command, `chathistory`.
+The `chathistory` extension uses the [chathistory][batch/chathistory] batch type and introduces a new client command, `CHATHISTORY`.
 
 Full support for this extension requires support for the [`batch`][batch], [`server-time`][server-time] and [`message-tags`][message-tags] capabilities. However, limited functionality is available to clients without support for these CAPs. Servers SHOULD NOT enforce that clients support all related capabilities before using the `chathistory` extension. Meanwhile, bouncers implementing the server side of this specification may not be able to provide message IDs (when they are mediating for a server that does not support the `message-tags` capability). Therefore, clients with full support for `message-tags` MAY wish to implement fallback logic that relies only on `server-time`.
 
-The `draft/chathistory` capability MUST be negotiated. This allows the server and client to act differently when delivering message history on connection.
+The `draft/chathistory` capability MUST be negotiated. This allows the server and client to behave differently when message history is available.
 
-An ISUPPORT token MUST be sent to the client to state the maximum number of messages a client can request in a single command, represented by an integer. `CHATHISTORY=50`. If `0`, the client SHOULD assume that there is no maximum number of messages.
+An ISUPPORT token MUST be sent to the client to state the maximum number of messages a client can request in a single command, e.g., `CHATHISTORY=50`. If `0`, the client SHOULD assume that there is no maximum number of messages.
 
 The `draft/event-playback` capability MAY be negotiated. This allows the client to signal that it is capable of receiving and correctly processing lines that would normally produce a local state change (such as `JOIN` or `MODE`) in its history batches.
 
 ### `CHATHISTORY` Command
-`CHATHISTORY` content can be requested by the client by sending the `CHATHISTORY` command to the server. A `batch` MUST be returned by the server. If no content exists to return, an empty batch SHOULD be returned to avoid the client waiting for a reply and to indicate that no content is available.
+Message history content can be requested by the client by sending the `CHATHISTORY` command to the server. A [`batch`][batch] MUST be returned by the server; its type SHOULD be `chathistory`. If no content exists to return, an empty batch SHOULD be returned, in order to avoid the client waiting for a reply.
 
-The `chathistory` command uses the following general syntax structure:
+The `CHATHISTORY` command has the following general syntax:
 
     CHATHISTORY <subcommand> <target> <timestamp | msgid> <limit>
 
-The `target` parameter specifies a single buffer (channel or nickname) from which history SHOULD be retrieved. Any `timestamp` values or parameters MUST be in the format of `YYYY-MM-DDThh:mm:ss.sssZ`.
+The `target` parameter specifies a single buffer (channel or nickname) from which history is to be retrieved. If a nickname is given as the `target` then the server SHOULD include history sent between the current user and the target nickname, including outgoing messages ("self messages"). The server SHOULD attempt to include history involving other nicknames if either the current user or the target nickname has changed during the requested timeframe.
 
 The special target `*` refers to all direct messages sent to or from the current user, regardless of the other party. This allows the client to retrieve conversations with users it is not yet aware of.
 
-If a nickname is given as the `target` then the server SHOULD include history sent between the current user and the target nickname, including outgoing messages ("self messages"), so as to give a full conversation. The server SHOULD attempt to include history involving other nicknames if either the current user or the target nickname has changed during the requested timeframe.
+A `timestamp` parameter MUST have the format `timestamp=YYYY-MM-DDThh:mm:ss.sssZ`, as in the [server-time][server-time] extension. A `msgid` parameter MUST have the format `msgid=foobar`, as in the [message-ids][message-ids] extension.
 
 #### Subcommands
 
@@ -75,10 +75,10 @@ This is useful for retrieving conversation context around a single message.
 
 #### `BETWEEN`
     CHATHISTORY BETWEEN <target> <timestamp=YYYY-MM-DDThh:mm:ss.sssZ | msgid=1234> <timestamp=YYYY-MM-DDThh:mm:ss.sssZ | msgid=1234> <limit>
-Request up to `limit` number of messages between the given `timestamp` or `msgid` values. The returned messages MUST start from and exclude the first message selector, while finishing on and excluding the second - this may be forwards or backwards in time.
+Request up to `limit` number of messages between the given `timestamp` or `msgid` values. With respect to the limit, the returned messages MUST be counted starting from and excluding the first message selector, while finishing on and excluding the second. This may be forwards or backwards in time.
 
 #### Returned message notes
-The order of returned messages within the batch is implementation-defined, but SHOULD be ascending time order or some approximation thereof. The `server-time` tag on each message SHOULD be the time at which the message was received by the IRC server. The `msgid` tag that identifies each individual message in a response MUST be the `msgid` tag as originally sent by the IRC server.
+The order of returned messages within the batch is implementation-defined, but SHOULD be ascending time order or some approximation thereof, regardless of the subcommand used. The `server-time` tag on each message SHOULD be the time at which the message was received by the IRC server. The `msgid` tag that identifies each individual message in a response MUST be the `msgid` tag as originally sent by the IRC server.
 
 Servers SHOULD provide clients with a consistent message order that is valid across the lifetime of a single connection, and which determinately orders any two messages (even if they share a timestamp); this will allow BEFORE, AFTER, and BETWEEN queries that use msgids for pagination to function as expected. This order SHOULD coincide with the order in which messages are returned within a response batch. It need not coincide with the delivery order of messages when they were relayed on any particular server.
 
@@ -95,6 +95,9 @@ If the server receives a `CHATHISTORY` command with missing parameters, the `NEE
 
 If no message history can be returned due to an error, the `MESSAGE_ERROR` error code SHOULD be returned.
 > FAIL CHATHISTORY MESSAGE_ERROR the_given_command the_given_target [extra_context] :Messages could not be retrieved
+
+If the selectors or limit supplied were invalid, the `INVALID_PARAMS` error code SHOULD be returned.
+> FAIL CHATHISTORY INVALID_PARAMS the_given_command [the_invalid_parameters] :Invalid parameters
 
 ### Examples
 
@@ -119,9 +122,9 @@ Requesting further message history than our client currently has
 ~~~~
 
 ## Use Cases
-The batch type and supporting command are useful for allowing an "infinite scroll" type capability within the client. A client will, upon scrolling to the top of the active window or a manual trigger, may request `chathistory` from the server and, after receiving returned content, append it to the top of the window. Users can repeat this historic scrolling to retrieve prior history until limitations are met (see below).
+Upon joining a channel, a client may request the latest messages for the channel, to retrieve the immediate context of the active conversation.
 
-Upon joining a channel, a client may request the latest messages for the buffer so that the active conversation context may be retrieved.
+Clients can use `CHATHISTORY` to implement "infinite scroll". When the user scrolls to the top of the active window or engages a manual trigger, the client can request `CHATHISTORY` from the server and then insert the results at the top of the window. The user can repeat this action to retrieve more history, possibly until some limit is met.
 
 ### Client pseudocode
 A client with full support for BATCH, message IDs, and deduplication can fill in gaps in its history using the following pseudocode. `FUZZ_INTERVAL` is a constant that compensates for clock skew across the IRC network (perhaps 1 to 10 seconds):
@@ -162,10 +165,16 @@ To fill in gaps in a client without deduplication support, iterate this infinite
 
 ## Implementation Considerations
 
-In the typical IRC network, there is no well-defined global linear ordering of messages, since different linked servers may see messages in different orders. Furthermore, due to clock skew between servers and between server and client, messages may be relayed, stored, and replayed in an order that differs from the timestamp order. Within the recommended constraints on message ordering described above, implementations may want to make different tradeoffs between simplicity, consistency, and being sure of receiving all relevant history messages.
+In the typical IRC network, there is no well-defined global linear ordering of messages, since different linked servers may see messages in different orders. Furthermore, due to clock skew between servers and between server and client, messages may be relayed, stored, and replayed in an order that differs from the timestamp order. Within the recommended constraints on message ordering described above, implementations may want to make different tradeoffs between simplicity, consistency, and correctness (i.e., neither missing nor duplicating messages).
 
 ## Security Considerations
 
-Secure identification of users and clients MUST exist in order to ensure that users cannot obtain history they are not authorised to view. Use of account names, internal account identifiers, or certificate fingerprints SHOULD be strongly considered when matching content to users. If a client requests content for a target that they do not have permission for, eg. a channel they are banned from, an empty batch SHOULD be returned as if no content exists.
+Secure identification of users and clients MUST exist in order to ensure that users cannot obtain history they are not authorised to view. Use of account names, internal account identifiers, or certificate fingerprints SHOULD be strongly considered when matching content to users. If a client requests content for a target they are not authorised to view, e.g., a channel they are banned from, an empty batch SHOULD be returned as if no content exists.
 
 While an ISUPPORT token value of `0` may be used to indicate no message limit exists, servers SHOULD set and enforce a reasonable maximum and properly throttle `CHATHISTORY` commands to prevent abuse.
+
+[batch/chathistory]: ../extensions/batch/chathistory-3.3
+[batch]: ../extensions/batch-3.2
+[server-time]: ../extensions/server-time-3.2
+[message-tags]: ../extensions/message-tags
+[message-ids]: ../extensions/message-ids
--- a/extensions/chathistory.md
+++ b/extensions/chathistory.md
@@ -40,9 +40,7 @@ An ISUPPORT token MUST be sent to the client to state the maximum number of mess
 The `draft/event-playback` capability MAY be negotiated. This allows the client to signal that it is capable of receiving and correctly processing lines that would normally produce a local state change (such as `JOIN` or `MODE`) in its history batches.
 
 ### `CHATHISTORY` Command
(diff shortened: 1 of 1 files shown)


extensions/chathistory.md
```diff
@@ -0,0 +1,111 @@
...
+---
+## Description
+This document describes the format of the `chathistory` extension. This enables clients to request messages that were previously sent if they are still available on the server.
+
+The server as mentioned in this document may refer to either an IRC server or an IRC bouncer.
+
+## Implementation
+The `chathistory` extension uses the [chathistory][batch/chathistory] batch type and introduces a client command, `chathistory`.
+
+To fully support this extension, clients MUST support the [`batch`][batch], [`server-time`][server-time] and [`draft/message-tags`][draft/message-tags] capabilities.
```
SaberUK:
```suggestion
To fully support this extension, clients MUST support the [`batch`][batch], [`server-time`][server-time] and [`message-tags`][message-tags] capabilities.
```

extensions/chathistory.md
```diff
@@ -0,0 +1,111 @@
...
+If no message history can be returned due to an error, the `MESSAGE_ERROR` error code SHOULD be returned.
+> FAIL CHATHISTORY MESSAGE_ERROR the_given_command [extra_context] :Messages could not be retrieved
+
+### Examples
+
+Requesting the latest conversation upon joining a channel
+~~~~
+[c] CHATHISTORY LATEST #channel * 50
+[s] :irc.host BATCH +ID chathistory #channel
+[s] @batch=ID;draft/msgid=1234;time=2019-01-04T14:33:26.123Z :nick!ident@host PRIVMSG #channel :message
```
SaberUK: `msgid` should be used instead of `draft/msgid` here.

extensions/chathistory.md
```diff
@@ -0,0 +1,111 @@
...
+Request a number of messages before and after the `timestamp` or `msgid` with the total number of returned messages not exceeding `limit`. The implementation may decide how many messages to include before and after the selected message. Only one timestamp or msgid MUST be given, not both.
+
+This is useful for retrieving conversation context around a single message.
+
+#### `BETWEEN`
+    CHATHISTORY BETWEEN <target> <timestamp=YYYY-MM-DDThh:mm:ss.sssZ | msgid=1234> <timestamp=YYYY-MM-DDThh:mm:ss.sssZ | msgid=1234> <limit>
+Request up to `limit` number of messages between the given `timestamp` or `msgid` values. The returned messages MUST start from the inclusive first message selector, while excluding and finishing on the second - this may be forwards or backwards in time.
+
+#### Returned message notes
+The returned messages MUST be in ascending time order and the `server-time` SHOULD be the time at which the message was received from the IRC server. The `msgid` that identifies each individual message in a response MUST be the `msgid` as when  received from the IRC server.
```
SaberUK: Received from?

I think this should be received by.

edk0:
I think "The returned messages MUST be in ascending time order" (especially, but not only, when combined with the following SHOULD) is wrong for the same line of reasoning expressed [here](https://github.com/ircv3/ircv3-specifications/issues/392#issuecomment-505784581). The spec should not require servers to break the causal order of messages.

It also feels like this section is specifying the batch type, not the command.

slingamn:
Reraising the two issues I brought up on #349:

* It's still unclear how targets work with direct messages. (What exactly is a "query name"? If it's a nickname, does it include messages sent to the client under a different nickname? Does it include messages sent *from* the client?) I think we should specify:
  * that the user's nickname, as a target, includes messages sent by the user
  * that it includes messages sent to and by the user under different nicknames
  * possibly add a special-purpose target like `*self` for retrieving these messages
* I think we should either deprecate or remove the use of msgid in the order queries (BEFORE, AFTER, LATEST, and BETWEEN); given the weakness of the guarantees on the relationship between msgid and timestamp, I don't see a good way to use msgid correctly in those contexts

slingamn:
I got confused about the distinction between AFTER and LATEST despite having actually implemented the previous version of the spec. I think a minor language tweak would help: describing LATEST as "Request the most recent messages that have been sent, excluding anything sent before the given `timestamp` or `msgid`".

prawnsalad:
@slingamn 
> It's still unclear how targets work with direct messages

Now specified. Though I'm not sure in what case your `*self` suggestion would be used?

prawnsalad:
> I think "The returned messages MUST be in ascending time order" (especially, but not only, when combined with the following SHOULD) is wrong for the same line of reasoning expressed [here](https://github.com/ircv3/ircv3-specifications/issues/392#issuecomment-505784581). The spec should not require servers to break the causal order of messages.
> 
> It also feels like this section is specifying the batch type, not the command.

After some discussion re. this comment, the working phrase at the moment is:
> The returned messages MUST be in chronological order

If no other input comes up I'll update the spec to use this.

slingamn:
Questions about the BETWEEN semantics with msgids as the query parameters:

1. BETWEEN says "The returned messages MUST start from the inclusive first message selector", which seems to potentially contradict the general instruction that "The returned messages MUST be in ascending time order", given that the msgid order need not agree with the timestamp order
1. What should happen if one of the msgids is not found in the message buffer / datastore?
1. Is it illegal to send a timestamp as one endpoint of the BETWEEN interval and a msgid as the other?

prawnsalad:
> 1. BETWEEN says "The returned messages MUST start from the inclusive first message selector", which seems to potentially contradict the general instruction that "The returned messages MUST be in ascending time order", given that the msgid order need not agree with the timestamp order

Message IDs are not reliable for any type of ordering as they are simply an identifier, so I'm not sure in what case you would attempt to order with them?

> 2. What should happen if one of the msgids is not found in the message buffer / datastore?

Good point... I've just been returning the empty batch and leaving it at that. I'll add this as a note.

> 3. Is it illegal to send a timestamp as one endpoint of the BETWEEN interval and a msgid as the other?

I don't see why it should be limited in this way. If I have a msgid and want to find the next 5minutes worth of conversation this would allow that.

slingamn:
>Message IDs are not reliable for any type of ordering as they are simply an identifier, so I'm not sure in what case you would attempt to order with them?

I actually agree, but if this is so, what are the semantics of `BETWEEN msgid=x msgid=y` anyway? Basically it seems like we have to decide that the messages are in some linear order, then find where `x` and `y` appear in the linear order, then decide which one is first, then return messages from the first (inclusive) to the last (exclusive). But there are potential problems with this:

1. There's no single linear order of all the messages across the entire network (even if there is on each individual server)
1. In a scenario where multiple servers are inserting history items into a distributed datastore, the datastore's view of the ordering may not coincide with the order of the messages seen by any one server, or with the delivery order seen by any one client; it may even violate "causal order" constraints

I'm still bearish about the use of msgids in AFTER/BEFORE/BETWEEN/LATEST queries. I think time-based queries have more predictable behavior, since everyone (the client, each server, and a shared datastore) can agree on whether timestamp A precedes timestamp B.

slingamn:
I'd like to propose removing BETWEEN, since we haven't really been able to determine what it's useful for. (Infinite scroll can use LATEST and then iterated BEFORE; as a sidenote, it might actually be helpful to provide pseudocode for this. If a client really wants the functionality of BETWEEN, they can implement it using AFTER.)

slingamn:
What's the rationale for BEFORE being inclusive of the parameter, but AFTER being exclusive? Seems like it would make more sense for them both to be exclusive. (If we also remove BETWEEN, then BEFORE, AFTER, and LATEST all exclude the parameter, while AROUND includes it for obvious reasons --- this seems nice and neat.)

slingamn:
I think input from client developers would be helpful on both of these questions (removing BETWEEN and inclusive vs. exclusive). The impression I'm getting is that both BETWEEN and exclusive queries are more useful if clients plan to use the `CHATHISTORY` command without support for message deduplication. But even so, the alternative of inclusive BEFORE and AFTER queries is viable without full support for deduplication, so I continue to think that we should remove BETWEEN and make all the queries inclusive.

extensions/chathistory.md
```diff
@@ -0,0 +1,115 @@
...
+This is useful for retrieving conversation context around a single message.
+
+#### `BETWEEN`
+    CHATHISTORY BETWEEN <target> <timestamp=YYYY-MM-DDThh:mm:ss.sssZ | msgid=1234> <timestamp=YYYY-MM-DDThh:mm:ss.sssZ | msgid=1234> <limit>
+Request up to `limit` number of messages between the given `timestamp` or `msgid` values. The returned messages MUST start from the inclusive first message selector, while excluding and finishing on the second - this may be forwards or backwards in time.
+
+#### Returned message notes
+The returned messages MUST be in ascending time order and the `server-time` tag SHOULD be the time at which the message was received by the IRC server. The `msgid` tag that identifies each individual message in a response MUST be the `msgid` tag as originally sent by the IRC server.
+
+#### Errors and Warnings
```
SaberUK: Suggestions for more FAIL messages:

* `UNKNOWN_CRITERIA <criteria>` &mdash; for if someone specifies an invalid criteria like wibble=foo instead of e.g.  msgid=abc123. 

* `INVALID_CRITERIA <criteria>` &mdash; for if someone specifies a criteria thats not valid for a command e.g. the wildcard (`*`) criteria with a non-LATEST command.

* `INVALID_LIMIT <limit>` &mdash; for if someone specifies a limit that is out of the allowed range.

extensions/chathistory.md
```diff
@@ -0,0 +1,115 @@
...
+Request a number of messages before and after the `timestamp` or `msgid` with the total number of returned messages not exceeding `limit`. The implementation may decide how many messages to include before and after the selected message. Only one timestamp or msgid MUST be given, not both.
+
+This is useful for retrieving conversation context around a single message.
+
+#### `BETWEEN`
+    CHATHISTORY BETWEEN <target> <timestamp=YYYY-MM-DDThh:mm:ss.sssZ | msgid=1234> <timestamp=YYYY-MM-DDThh:mm:ss.sssZ | msgid=1234> <limit>
+Request up to `limit` number of messages between the given `timestamp` or `msgid` values. The returned messages MUST start from the inclusive first message selector, while excluding and finishing on the second - this may be forwards or backwards in time.
+
+#### Returned message notes
+The returned messages MUST be in ascending time order and the `server-time` tag SHOULD be the time at which the message was received by the IRC server. The `msgid` tag that identifies each individual message in a response MUST be the `msgid` tag as originally sent by the IRC server.
```
kylef: > the `server-time` tag SHOULD be the time at which the message was received by the IRC server

When it comes to linked servers; does this state that the time will be the one from which I am connected to, or the server that processed the message?

If the former, then a message's "server-time" could change depending on which server I've connected to?

extensions/chathistory.md
```diff
@@ -0,0 +1,115 @@
...
+    
+Request up to `limit` number of messages before and including the given `timestamp` or `msgid`. Only one timestamp or msgid MUST be given, not both.
+
+#### `AFTER`
+    CHATHISTORY AFTER <target> <timestamp=YYYY-MM-DDThh:mm:ss.sssZ | msgid=1234> <limit>
+Request up to `limit` number of messages after and excluding the given `timestamp` or `msgid`. Only one timestamp or msgid MUST be given, not both.
+
+#### `LATEST`
+    CHATHISTORY LATEST <target> <* | timestamp=YYYY-MM-DDThh:mm:ss.sssZ | msgid=1234> <limit>
+Request the most recent messages that have been sent after and excluding the given `timestamp` or `msgid`. If a `*` is given instead of a timestamp or msgid, the server MUST use the current time as a timestamp. The number of messages returned MUST be equal to or less than `limit`. If a `*` is not given, only one timestamp or msgid MUST be given, not both.
```
kylef: Multiple messages might have the exact same server-time; as a client I'd likely want to provide an inclusive timestamp instead of exclusive to the result and do de-duplication based on msgid locally. Otherwise there can be cases where I disconnected while receiving a series of messages at the exact same time and loose one of them.

It might even be better to ask for an older timestamp from a few minutes before and handle msgid based deduplication. If there are messages I didn't receive but have an older timestamp due to originating from a different linked server which can have latency between the server I am connected.

msgid lookup can face similiar problems.

extensions/chathistory.md
slingamn: I interpreted this as "the server that initially processed the message", i.e., the time tag should be the same across all relays and replays of the message.

Intriguingly, the ambiguity you're identifying exists even outside the context of history, and AFAICT is not clarified by the [latest server-time spec](https://ircv3.net/specs/extensions/server-time-3.2).

kylef: I was thinking “the server that initially processed the message” too, although that likely means you might not see messages in time order (ascending) when they come from other linked servers live which makes some of the below commands tricky to use.

For example I may receive the following messages, the various nick are connected to different linked servers. doe@aus is notable as they are connected to an Australian IRCd which has the higest lattency:

```
@time=2020-01-23T08:39:42.102Z :bot@network.com PRIVMSG #example :Radio is online
@time=2020-01-23T08:39:42.102Z :bot@network.com PRIVMSG #example :Now playing XYZ
@time=2020-01-23T08:39:42.042Z :dennis@eu PRIVMSG #example :Hello World
@time=2020-01-23T08:39:42.021Z :kyle@us PRIVMSG #example :Hello World
@time=2020-01-23T08:39:41.944Z :doe@aus PRIVMSG #example :Hello World
```

If I disconnected after the first message was received or processed, then reconnect and ask for `CHATHISTORY LATEST` giving the last message timestamp that I’ve seen as `2020-01-23T08:39:42.102Z`. Then I SHOULD receive no additional messages despite never having seen the additional 4 messages. `msgid` lookup has similar problems because the server I connect to may not know the order that I’ve received the messages.

extensions/chathistory.md
```diff
@@ -0,0 +1,115 @@
...
+An ISUPPORT token MUST be sent to the client to state the maximum number of messages a client can request in a single command, represented by an integer. `CHATHISTORY=50`. If `0`, the client SHOULD assume that there is no maximum number of messages.
+
+### `CHATHISTORY` Command
+`CHATHISTORY` content can be requested by the client by sending the `CHATHISTORY` command to the server. A `batch` MUST be returned by the server. If no content exists to return, an empty batch SHOULD be returned to avoid the client waiting for a reply and to indicate that no content is available.
+
+The `chathistory` command uses the following general syntax structure:
+
+    CHATHISTORY <subcommand> <target> <timestamp | msgid> <limit>
+
+The `target` parameter specifies a single buffer (channel or nickname) from which history SHOULD be retrieved. Any `timestamp` values or parameters MUST be in the format of `YYYY-MM-DDThh:mm:ss.sssZ`.
```
kylef: In the situation where a client is disconnected to a bouncer offering chathistory. How can I know which nickname targets may have messaged me while I am disconnected? Someone I have never spoken to with the client may message me  while I am disconnected and the API offers no way to know which targets are available so the client can reconnect but it doesn’t know there are messages for a specific target.

This is less of a problem for channels because you likely get a JOIN event so you can retrieve the history afterwards. ZNC-Playback solves this problem by allowing wildcards here (although that introduces other problems of its own).

extensions/chathistory.md
slingamn: Over on https://github.com/prawnsalad/ircv3-specifications/pull/1 we're planning to make most of the commands use timestamps only as parameters. That will allow clients to compensate for skew/latency across the network by adding or subtracting a fuzz duration. (10 seconds seems like a sane value: [Unreal disables linking entirely when it detects 30 seconds of skew](https://www.unrealircd.org/docs/FAQ#Why_correct_time_is_important_and_how_to_fix_your_clock).)

edk0: In what sense does removing the ability to specify msgids allow clients to do anything?

The situation where it makes sense to use times and a fudge factor is when you're starting to update your existing partial history which came from another server. The rest of the time it's pointless waste.

Either way, with the time order requirement, a distributed implementation that's not strictly serialising has to say inconsistent things, either by returning history in an order it didn't really happen or by returning messages with timestamps that they didn't really have. I could imagine either causing problems for a client that tries to construct a canonical history, but I also just don't want to be required to do extra work to provide less accurate information.

slingamn: I just want to get clear on the terms of this debate. If I understand you correctly, your primary objection is to the requirement that results be returned in time order, *not* to the requirement that queries support timestamps as parameters. As a server vendor, you're going to have to do the work required to answer queries of the form "what messages were sent between 2020-01-27T17:22:09.270Z and 2020-01-27T17:40:09.282Z?" regardless.

Here's my position on this issue:

1. Since there is no global total order on messages, it is unnecessarily difficult for client developers to correctly use msgids in ordering queries. With timestamps, the client and all servers in the network can agree on the intended semantics of the query. (If clients need to support timestamp fudging regardless to handle the case where they reconnect to a different server, in what sense is "waste" occurring? What is being wasted?)
1. The preferred compromise is "returning history in an order it didn't really happen", not sending fake timestamps. The only thing that can cause the timestamp order to contradict the "causal order" of messages is clock skew among servers (not latency). Clock skew among well-managed servers should be on the order of a few milliseconds, so this is a rare edge case.
1. The server-time spec already says "a client SHOULD treat the message as having occurred at the given time instead of its current time": it is implicit in this that the timestamp order overrides the order of message delivery (and I think this is the behavior implemented in many clients that support server-time).

slingamn:
We had a pretty productive conversation about this. Here's what I learned:

1. Client developers want a linear pagination order that is addressable by message IDs, not just timestamps
1. It is not critical from the client's perspective that this linear pagination order coincide with the delivery order on any particular server in the network; it just has to be stable enough to do pagination in the context of a single connection
1. Server developers do not want to provide a linear pagination order that is consistent across multiple servers in the network
1. It's important to some people that the "standard" way of paging not involve any duplicated messages
1. There is no reason to specify that CHATHISTORY return messages in timestamp order; this just throws away information (if clients want to sort the batch, they can sort it)

It sounds like we can probably meet in the middle and specify that CHATHISTORY provides a linear pagination order that is consistent across a single client connection, but is not guaranteed to coincide with any real delivery order (or "causal order") or be stable across reconnections to the network. Examples of how this might be implemented on the server side:

1. A ring buffer of messages in the order they were delivered to the server (this respects the causal order, and is unstable across reconnections to different servers).
1. A MySQL table with an autoincrementing integer ID --- not necessarily exposed directly as the msgid, but the semantics of "messages before msgid XYZ" are defined as "look up the autoincrementing integer ID corresponding to XYZ and return the messages with integer IDs immediately below it". This may violate the causal order, but is stable across reconnections.
1. A table in a distributed datastore keyed on (timestamp + tiebreaker_bits), where the timestamp is a sufficiently high-resolution timestamp and tiebreaker_bits are random bits that provide uniqueness and a determinate linear ordering even if two messages have the same timestamp. This may also violate the causal order, but is stable across the network.

Clients backfilling history after a disconnect who don't want to miss messages will still need to take latency/skew into account to some degree.

What are the implications for the spec? I propose:

1. Keep msgids in queries
1. Make all ordering-based queries exclusive. This facilitates the use of msgids in pagination without requiring deduplication in the typical case. For timestamps, an exclusive query can always be used to emulate an inclusive query (just add or subtract a millisecond to or from the timestamp, as desired)

extensions/chathistory.md
slingamn: I was thinking a special-purpose target like `*self` that refers to your own direct messages with all other clients, including your self-messages.

edk0:
I agree with both concrete points. We probably need some non-normative discussion about history not being stable across reconnects and how to do pagination.

extensions/chathistory.md
```diff
@@ -0,0 +1,115 @@
...
+
+#### `AROUND`
+    CHATHISTORY AROUND <target> <timestamp=YYYY-MM-DDThh:mm:ss.sssZ | msgid=1234> <limit>
+Request a number of messages before and after the `timestamp` or `msgid` with the total number of returned messages not exceeding `limit`. The implementation may decide how many messages to include before and after the selected message. Only one timestamp or msgid MUST be given, not both.
+
+This is useful for retrieving conversation context around a single message.
+
+#### `BETWEEN`
+    CHATHISTORY BETWEEN <target> <timestamp=YYYY-MM-DDThh:mm:ss.sssZ | msgid=1234> <timestamp=YYYY-MM-DDThh:mm:ss.sssZ | msgid=1234> <limit>
+Request up to `limit` number of messages between the given `timestamp` or `msgid` values. The returned messages MUST start from the inclusive first message selector, while excluding and finishing on the second - this may be forwards or backwards in time.
```
slingamn: Can you use a timestamp for one selector and a msgid for the other?

slingamn:
Should we bring back the `NO_SUCH_CHANNEL` fail code? #349 used it both for the case where there is no such channel, and for the case where the user doesn't have permissions to view the channel. AFAICT these cases are not explicitly handled in the current draft.

extensions/chathistory.md
```diff
@@ -0,0 +1,158 @@
...
+---
+## Description
+This document describes the format of the `chathistory` extension. This enables clients to request messages that were previously sent if they are still available on the server.
+
+The server as mentioned in this document may refer to either an IRC server or an IRC bouncer.
+
+## Implementation
+The `chathistory` extension uses the [chathistory][batch/chathistory] batch type and introduces a client command, `chathistory`.
+
+To fully support this extension, clients MUST support the [`batch`][batch], [`server-time`][server-time] and [`message-tags`][message-tags] capabilities.
```
slingamn: Someone interpreted this language as meaning that chathistory has a hard dependency on batch, server-time, and message-tags, and therefore that chathistory could not necessarily be used to replace `znc.in/playback`.

I think it's worth clarifying that this is not the case: degraded functionality (in particular, functionality equivalent to `znc.in/playback`) is still available even without support for any of these caps.

extensions/chathistory.md
```diff
@@ -0,0 +1,158 @@
...
+### `CHATHISTORY` Command
+`CHATHISTORY` content can be requested by the client by sending the `CHATHISTORY` command to the server. A `batch` MUST be returned by the server. If no content exists to return, an empty batch SHOULD be returned to avoid the client waiting for a reply and to indicate that no content is available.
+
+The `chathistory` command uses the following general syntax structure:
+
+    CHATHISTORY <subcommand> <target> <timestamp | msgid> <limit>
+
+The `target` parameter specifies a single buffer (channel or nickname) from which history SHOULD be retrieved. Any `timestamp` values or parameters MUST be in the format of `YYYY-MM-DDThh:mm:ss.sssZ`.
+
+The special target `*` refers to all direct messages sent to or from the current user, regardless of the other party. This allows the client to retrieve conversations with users it is not yet aware of.
```
emersion, line 40: Could we allow servers to send highlights as part of `*` as well, so that notifications could still work without needing clients to fetch history for all channels?

extensions/chathistory.md
```diff
@@ -0,0 +1,166 @@
...
+
+The `draft/chathistory` capability MUST be negotiated. This allows the server and client to act differently when delivering message history on connection.
+
+An ISUPPORT token MUST be sent to the client to state the maximum number of messages a client can request in a single command, represented by an integer. `CHATHISTORY=50`. If `0`, the client SHOULD assume that there is no maximum number of messages.
+
+The `draft/event-playback` capability MAY be negotiated. This allows the client to signal that it is capable of receiving and correctly processing lines that would normally produce a local state change (such as `JOIN` or `MODE`) in its history batches.
+
+### `CHATHISTORY` Command
+`CHATHISTORY` content can be requested by the client by sending the `CHATHISTORY` command to the server. A `batch` MUST be returned by the server. If no content exists to return, an empty batch SHOULD be returned to avoid the client waiting for a reply and to indicate that no content is available.
```
hhirtz, line 40: Thoughts on requiring the batch contents to be sorted by increasing time?  This should already be done by the server (or else their chathistory impl would be very complex...).

hhirtz (commented):
Feedback of the dev of a user-facing client (ie. not a bot) that implemented some of this spec:

## don't use timestamps

Ever talked to BitBot?  Ever sent multi-line messages?  Multiple messages may have the same timestamp. You'll end up with either duplicate or missing ones, depending on whether you put a `>=` or a `>` in your code.  Only use the `msgid=` criteria to fill your timeline, and ask for your server admin to use a good message-id impl (you don't want duplicates, and you want messages to have the same ID when you fetch them with CHATHISTORY).

## BEFORE and LATEST are the only useful subcommands

Fetch history from future to past (ie. don't use AFTER).  Imagine you return from a week-long trip and have to wait for your client to fetch history from the last read message (that is, [if it does it properly][0] and if your server admins are as nice to you as root dns operators are to google).

When I start my client, I expect the latest messages to be shown instantly (with CHATHISTORY LATEST), so that I can talk with others with no delay and the conversation context in mind. I also expect the timeline to fill itself when needed and provide infinite scroll (with CHATHISTORY BEFORE).

AROUND may be useful with a "search" feature, but chathistory doesn't provide one.

BETWEEN and AFTER would be useful to fill a local cache, but that'd be taking it too far on complexity for little gain.

[0]: https://todo.sr.ht/~emersion/gamja/7

rcr:
This extension has the same issue for clients as synchronising MODE messages around an initial numeric 324 ...



Consider the order of events when I join #channel


 - server receives my JOIN #chan
 - server sends me JOIN reply
 - server receives some PRIVMSG from someuser to #chan
 - server sends me user's PRIVMSG to #chan
 - I receive JOIN reply
 - I send CHATHISTORY for #chan
 - I receive PRIVMSG from someuser
 - I receive CHATHISTORY reply



Not only does a client now have to synchronize around receiving the initial chat history to get the order of messages displaying correctly, but in that scenario above the client is going to receive the PRIVMSG directly and in the chat history reply because the timestamp I supply is when I receive the JOIN which is after the server receives the PRIVMSG



The way to fix this is just to send the chat history implicitly after the JOIN exactly like how a RPL_NAMREPLY is sent and NOT have the client need to request it after joining

edk0:
> Not only does a client now have to synchronize around receiving the initial chat history to get the order of messages displaying correctly, but in that scenario above the client is going to receive the PRIVMSG directly and in the chat history reply because the timestamp I supply is when I receive the JOIN which is after the server receives the PRIVMSG
> 
> The way to fix this is just to send the chat history implicitly after the JOIN exactly like how a RPL_NAMREPLY is sent and NOT have the client need to request it after joining

querying for before the msgid of your own join avoids all thse problems without needing to forcefeed the client data

rcr:
Well it still forces a client to have mutable buffers and handle the logic of receiving older messages AFTER newer messages

edk0:
Sure. I think avoiding that one would involve making clients eat all old data (starting from when?) before they get anything new, which would be impractical with current server architectures and undesirable in any event for UX reasons.

jwheare:
Could clients perhaps indicate the chathistory they want by sending tags alongside the JOIN command?

edk0:
That's a good idea, but realistically what they get in one burst will have to be limited, so they should expect to have to ask for more.

rcr:
Sending the desired chathistory along with the JOIN seems like a good compromise here, yeah

hhirtz:
> Well it still forces a client to have mutable buffers and handle the logic of receiving older messages AFTER newer messages

You need that logic for infinite scrolling anyway.

prawnsalad:
If a client requests chathistory for different points in time then it would be safe to assume that the client either has a mutable buffer or some other way to display those messages. If a client has no sensible way to display out-of-order message then it knows itself not to request those messages.

A message tag on the JOIN command for the history is a neat idea that would suit both im/mutabe types. Though I think servers that currently have a channel history option sends a chathistory batch on JOIN already? What would the difference here be?

prawnsalad:
@hhirtz 
> don't use timestamps

@slingamn and I had come to the same conclusion too. But this is only relevant when implementing buffer scrollback, since timestamps are useful for casually jumping around history when you don't have a msgid reference to go from. Eg. you have a general message stored somewhere or perhaps from a pasted log, and you're searching for context. Same applies for the not-so-useful commands you mentioned. Perhaps this is noteworthy enough to mention in the spec

slingamn:
First of all, I'm grateful to have additional feedback on this specification from client developers: server developers have been overrepresented in recent discussion.

>don't use timestamps

The problem being addressed here is that there is no global linear order of messages sent to a channel: different servers on a network may see the messages in different orders, and the history replay order may disagree with the order in which messages were originally relayed.

The consequence of these limitations is that the most robust way to fetch history will involve both timestamps and msgids --- timestamps can be modified to compensate for lag and clock skew when reconciling relayed and replayed messages, msgids can be used for pagination within the locally consistent history replay order available on any given connection.

That said, I think the current pseudocode examples could be adjusted so that all queries use msgids as parameters (or use LATEST with no parameters); I'll consider doing this.

>BEFORE and LATEST are the only useful subcommands

The pseudocode includes an example use of BETWEEN (backfilling a gap when the client lacks deduplication support). I can imagine use cases for AFTER as well (backfilling short gaps in history in cases where the client only wants to handle messages in ascending order). AROUND is for future-proofing, essentially; I agree that in its current form there are no good use cases for it.

>querying for before the msgid of your own join avoids all these problems without needing to forcefeed the client data

Here's an edge case that I expect will make everyone unhappy: JOIN lines that don't correspond to an actual JOIN event, for example, bouncer-style reattach in Oragono, may not receive msgids. (The pseudocode section of the spec currently recommends `CHATHISTORY(BEFORE, join_message.msgid)`; I will change this to recommend `CHATHISTORY LATEST`.)

I feel like the way to handle @rcr's case (without increasing client implementation complexity too much) is just to discard messages relayed from the channel until the response to a `CHATHISTORY LATEST` query has been received, on the expectation that the results of this query will include the messages that were discarded. (There will still probably be minor race conditions here in most server-side implementations --- but it seems to me that the only path to removing all such race conditions is to add deduplication and/or reordering support, which increases complexity.)

At a "meta" level, I think a lot of the discussion so far has tended towards a relatively weak / flexible specification, one that gives implementers a lot of freedom but which makes it more complex (but still possible!) to write fully correct client code. If it seems hard to use this API in a way that's completely free of race conditions, then that reflects a conscious design tradeoff we made. That doesn't mean that the issues can't be reexamined, but my point is that they were examined previously (perhaps too many times).

hhirtz:
> The pseudocode includes an example use of BETWEEN (backfilling a gap
> when the client lacks deduplication support
The client pseudocode part needs clarification:



It seems the client is assumed to keep logs locally.  This is not always

the case however:  logs might be on a bouncer and fetched via

CHATHISTORY.  Although no "real-life" user-facing client does that right

now because people need them to act as bouncers (irssi+ssh, weechat+its

custom protocol, kiwiirc+http, etc.), CHATHISTORY may (will?) change

this fact since it allows a client to fetch their logs dynamically from

an IRC bouncer (bouncer + user-facing client and CHATHISTORY).



The pseudocodes are indeed useful for bouncers, since they'll want to

fill the gaps in their logs.



For user-facing clients that don't keep logs, the first and third

pseudocodes don't make sense because there is no "last message relayed

to previous session":  the user is presented with the latest messages

via LATEST or BEFORE, and if needed previous messages are fetched with

BEFORE.  Using BETWEEN or AFTER with a "last-seen message ID/timestamp"

instead will introduce large delays when scrolling.



TL;DR:  separate between the two use-cases (log filling vs infinite

scroll), and the first and third listings do the same thing.

emersion:
AROUND may be useful to jump to a specific message in the history. For instance the client receives a reply to a message and may want to provide a link to jump to the original message in the buffer.

slingamn:
@hhirtz: I'm open to revising the pseudocode examples, but I think you're taking too narrow a view of what it means to "keep logs locally". An example of a case where I think full backfill would be warranted is a mobile client (connected directly to an IRC server that supports CHATHISTORY) that loses connectivity for a macroscopic amount of time, say 30 minutes to an hour, and then wants to do a complete backfill covering the time away to look for missed mentions or direct messages (since direct messages from possibly unknown sources are queried via the `*` target). (You could say that the proper solution for this use case is push notifications, but that opens a different can of infrastructural worms.)

extensions/chathistory.md
slingamn: This is resolved now by the `*` target (merged into the PR).

extensions/chathistory.md
RyanSquared: I think this should instead be done by some other mechanism. If I understand correctly, the "*" target denotes targets that the client might not know about. However, the client _should_ know about channels it is in. This may be more applicable to some other subcommand.

slingamn:
I made revisions to the pseudocode examples here: https://github.com/prawnsalad/ircv3-specifications/pull/4

slingamn:
In terms of getting this merged as a draft, it would help to take a census of implementations. The server implementations I'm aware of are oragono (fully compliant as of 2.3.0, to the best of my knowledge) and kiwibnc (which may be missing some edge cases). The only client implementation I'm aware of is kiwiirc (modulo https://github.com/kiwiirc/kiwiirc/issues/1397 and https://github.com/kiwiirc/kiwiirc/issues/1250).

I've heard that there are other client implementations that have been released privately but not published. Maybe there are other server implementations? I think we're fairly close to the "two servers, two clients" threshold for merging this.

hhirtz:
Here are other implementations (not fully compliant though):



- soju (as a server): <https://sr.ht/~emersion/soju/>
 see #12 [0] for what's missing,

- senpai: <https://sr.ht/~taiite/senpai/>
 which currently only uses timestamps and doesn't handle disconnections.



[0] https://todo.sr.ht/~emersion/soju/12

slingamn:
I think a client implementation counts as compliant even if it doesn't exercise all the functionality of the spec --- just as long as the commands it does exercise are specified, and it uses them in accordance with the specified semantics.

slingamn:
One more client implementation: https://sr.ht/~emersion/gamja/

It seems to me like this might actually be ready for merge as a draft?

jwheare:
Yup, if the main participants here are all happy there are no TODOs left, let's do it.

slingamn:
My bad: @emersion brought up an issue that we should probably address. As written, this specification assumes that the server supports `message-tags`. This is problematic for bouncers that do support `message-tags` sitting in front of servers that don't: such a bouncer shouldn't publish its own msgids (since they would be invalid if the client tried to use them in a reply tag or similar). It therefore seems necessary to explicitly allow for fully compliant server implementations that don't send or accept msgids.

I think what this needs is non-normative language suggesting that clients implement a fallback to timestamp-based queries?

I can't think of a reason servers wouldn't support timestamp-based queries, but I also don't immediately see a compelling reason to promise that servers MUST support them.

slingamn:
One other thing I wanted to ask about: the current spec language doesn't recommend explicitly that clients have to be joined to a channel in order to query its history (some non-normative language suggests that this should be allowed unless the client is banned).

My sense is that given prevailing expectations around IRC channel membership and privacy, this should be a MAY, e.g.,

>Servers MAY wish to disallow clients from querying the history of channels they are not joined to. If they do not, they SHOULD disallow clients from querying channels that they are banned from, or which are private.

slingamn:
I've addressed all the outstanding issues I'm aware of. Let's take comments for a week or so and then merge as a draft?

slingamn:
Last call for comments!

slingamn:
Let's merge this on Monday.

slingamn:
Concerns were raised in #ircv3 about replay of self-messages across nickname changes:

>As it currently stands, it seems there is no reliable way of determining whether a message in the chat history was sent by the user itself.

The person raising those concerns suggested adding a `self` tag to these messages. My own preference is for a split approach:

1. For channel message history, do nothing
2. For direct message history, servers MAY (SHOULD?) rewrite self-messages such that the source is the client's current nickmask (i.e., nickmask at the time of the replay)

the motivation being that rewriting is at once less important in the channel setting, and more disruptive (since nickname changes are often "part of the conversation" there). (In the channel setting, this is also potentially mitigated by `account-tag`, although I don't know of anyone who is currently using `account-tag` to identify the sources of messages to users.)

slingamn:
Added a very hesitant line about rewriting:

>With direct message history, servers may wish to rewrite the sources or targets of messages to correspond to the current nicknames of the two users.

slingamn:
I think we should merge this.

jwheare:
Happy to merge this if all contributors are happy. A few non-blocking thoughts:

* Is there a reasons TAGMSG is only included with event-playback?
* Would e.g. `chathistory-events` be a better name than `event-playback`?
* Is there scope for a way to indicate that you've reached the end of history either forward or backwards, or are we content for clients to just keep requesting more until they stop seeing new messages?

slingamn:
>Is there a reason TAGMSG is only included with event-playback?

I'm not sure. #362 and #293 assume that by default, only PRIVMSG and NOTICE are suitable for inclusion in a chathistory batch.

It's possible that the default/baseline assumption about TAGMSG is that they may produce a state change in the client, hence this caution?

>Is there scope for a way to indicate that you've reached the end of history either forward or backwards, or are we content for clients to just keep requesting more until they stop seeing new messages?

Right now you get an empty batch. This seems fine to me on its own terms, but it's the same response you get for a permissions error or a nonexistent target, which might be the result of an excess of caution on security/privacy issues.

jwheare:
There's no risk of TAGMSG causing a state related desync, they are by design lossy, so can't communicate important consistent state. #293 looks too old to consider TAGMSG, and no one in #362 never provided a rationale.

Re: empty batch, I'm thinking more if the end comes half way through a reply, you have to make another request to confirm it's the end (with an empty batch). Seems minorly wasteful, but might not be a significant issue.

emersion:
> Re: empty batch, I'm thinking more if the end comes half way through a reply, you have to make another request to confirm it's the end (with an empty batch). Seems minorly wasteful, but might not be a significant issue.

You can check if the number of returned messages == the limit the client sent, and only request more if that's the case. So the empty batch should only happen when there are as many remaining messages as the limit.

jwheare:
Yes, it's just that's a heuristic rather than something explicit. I think I'd prefer the latter, but it's not a blocker for merge.

jwheare:
To give a concrete hypothetical of what something explicit might be, a tag on the BATCH might work:

`@chathistory-end :irc.host BATCH +ID chathistory #channel`

I prefer to avoid positional params, for extensibility purposes, but a param is another option that would support clients without tags.

e.g.

`:irc.host BATCH +ID chathistory #channel end`

jwheare:
An additional comment – probably more of a blocker in the absence of any rationale – I'd upgrade these SHOULDs to MUSTs:

> The type of the batch SHOULD be `chathistory`, and it SHOULD take a single additional parameter, the canonical name of the target being queried

slingamn:
I'm not that enthusiastic about adding a "no more messages" indicator.

I think I might like to add a new `FAIL` code (perhaps `INVALID_TARGET`) covering all the cases where either the target does not exist or the client doesn't have permissions to query it. I don't think this would add any security concerns?

emersion:
>I think I might like to add a new FAIL code (perhaps INVALID_TARGET) covering all the cases where either the target does not exist or the client doesn't have permissions to query it. I don't think this would add any security concerns?

+1

slingamn:
Let's merge this?

jwheare:
Sure, can do tomorrow.

PR #394, opened 2019-06-26
[message-ids] Forbid SPACE/NUL/CRLF/leading-colon in message IDs

jwheare:
This makes using them in e.g. CHATHISTORY more robust.

### DIFF
--- a/extensions/message-ids.md
+++ b/extensions/message-ids.md
@@ -34,7 +34,7 @@ Servers MAY attach this tag on any event. If this tag is being used, it SHOULD b
 
 The tag value MUST be a unique ID chosen by the originating server. Uniqueness in this context means that any other message transmitted on the entire network at any time MUST NOT share the same value.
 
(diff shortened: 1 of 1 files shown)


SaberUK:
Is there any need to blacklist NUL, CR, and LF considering that those are banned by the message-tags spec?

https://ircv3.net/specs/extensions/message-tags.html#format

><escaped_value> ::= <sequence of zero or more bytes except NUL, CR, LF, semicolon (`;`) and SPACE>

jwheare:
They’re banned in escaped form, not unescaped.

prawnsalad:
It seems that white listing a range of chars would be easier and more precise? It would also ensure we know what servers are using in case we need to use the msgid somewhere else in future too.

jwheare:
My preference is to keep this a blacklist, but I'd be open to a second choice of limiting message ids to a whitelist of just ascii printables exluding space and leading-colon, but no further, if there's consensus for that.

Have to be careful about adding overly restrictive errata to ratified specs.

Spaces and leading-colon will never be allowed, but please upvote this comment if you'd prefer an ascii-printable whitelist or down vote for just a blacklist of CR and LF.

jwheare:
There doesn't seem to be any support for an ascii printable only whitelist, so I'll merge this as is tomorrow.

Issue #208, opened 2016-02-04
Multiline messages


highflyer77:
CAP multi-line
This is only an incomplete example of how it could work, syntax or method to impliment would vary.

<lf>   would mark the new line (or whatever char. to be determined).
<crlf> would mark end of message, same as irc now uses.
- Raw message sent by user with multi-line support enabled:

```
<myNick>    This is my message I want to send. <lf>
            This would be line 2, I want to keep it intact with the first line. <lf>
            line 3 and end of the message. <crlf>
```
- How message "is" shown with "no" multi-line support:

`<myNick>   This is my message I want to send. This would be line 2, I want to keep it intact with the first line. line 3 and end of the message.`
- How message should be shown with multi-line support:

```
<myNick>    This is my message I want to send.
            This would be line 2, I want to keep it intact with the first line.
            line 3 and end of the message.
<OtherNick> Yeah cool.
```
- multi-line support would be usefull for pretty printing short code examples too.

```
<myNick>    #! /bin/bash
            echo "my cool script"
            if [ -n "$result" ] ; then
                echo "$result"
                exit 0
            else
                echo "File pattern not found"
                exit 1
            fi
<OtherNick> Nice code dude!
```
- The benefit is:  new lines are kept together without splitting up the message format.

I know some channels would have a limit or ban, but others could enjoy the multi-line support.
The 512 character limit would still apply to the message with multi-line support including the <lf> and <cr-lf> characters

Please comment and let me know if this is feasible,
Thank you.

ChickenNuggers:
As I said on IRC, I think this would be better implemented via BATCH.

dwfreed:
This wouldn't work in a lot of IRCds as they're currently implemented, because they accept LF from clients to mean the same thing as CRLF.

ChickenNuggers:
> This wouldn't work in a lot of IRCds as they're currently implemented, because they accept LF from clients to mean the same thing as CRLF.

I know that, using [cqueues](https://github.com/wahern/cqueues/) for example, would not allow this; LF is automatically translated whether it be CRLF or LF coming in from the server.

highflyer77:
> This wouldn't work in a lot of IRCds as they're currently implemented, because they accept LF from clients to mean the same thing as CRLF.

I sorta already knew that, that's why it says at the top:  " (or whatever char. to be determined). "

1. This was not meant to be used "as currently implemented" that's why it's being proposed.
[...]

PR #398, opened 2019-09-23
[multiline] Multiline message batches

jwheare:
A solution to multiline messages (#208) using client to server batches based on https://gist.github.com/jesopo/092866e52b40cda6f6205263d04dbcf9

---

Currently implemented and testable with the IRCCloud client and servers (prattle: teams, bnc, slack) and bitbot.

---

Client sending a mutliline batch

    Client: BATCH +123 draft/multiline #channel
    Client: @batch=123 PRIVMSG #channel hello
    Client: @batch=123 privmsg #channel :how is<SPACE>
    Client: @batch=123;draft/multiline-concat PRIVMSG #channel :everyone?
    Client: BATCH -123

Server sending the same batch

    Server: @msgid=xxx;account=account :n!u@h BATCH +123 draft/multiline #channel
    Server: @batch=123 :n!u@h PRIVMSG #channel hello
    Server: @batch=123 :n!u@h PRIVMSG #channel :how is<SPACE>
    Server: @batch=123;draft/multiline-concat :n!u@h PRIVMSG #channel :everyone?
    Server: BATCH -123

Server sending messages to clients without multiline support

    Server: @msgid=xxx;account=account :n!u@h PRIVMSG #channel hello
    Server: @account=account :n!u@h PRIVMSG #channel :how is<SPACE>
    Server: @account=account :n!u@h PRIVMSG #channel :everyone?

Final concatenated message

    hello
    how is everyone?

### DIFF
--- a/extensions/multiline.md
+++ b/extensions/multiline.md
@@ -69,7 +69,7 @@ When receiving a well-formed mulitiline message batch, implementations MUST coll
 * Servers: delivering the batch to the intended recipients
 * Clients: displaying the batched message to the user
 
-Messages in a multiline batch MUST be concatenated with a line break unless the `draft/multiline-concat` message tag is sent, in which case the message is directly concatenated with the previous message without any separation.
+Messages in a multiline batch MUST be concatenated with a single line feed (`\n`) byte unless the `draft/multiline-concat` message tag is sent, in which case the message is directly concatenated with the previous message without any separation.
 
 Clients MUST NOT send messages other than PRIVMSG while a multiline batch is open.
 
--- a/extensions/multiline.md
+++ b/extensions/multiline.md
@@ -214,9 +214,9 @@ Invalid multiline batch target
 
     Server: :irc.example.com FAIl BATCH MULTILINE_MAX_BYTES 40000 :Multiline batch max-bytes exceeded
 
-`max-bytes` exceeded multiline batch
+`max-lines` exceeded multiline batch
     
-    Server: CAP * LS :draft/multiline=max-lines=10
+    Server: CAP * LS :draft/multiline=max-bytes=40000,max-lines=10
 
     Client: BATCH +abc draft/multiline #channel
     Client: @batch=abc PRIVMSG #channel hello
--- a/extensions/multiline.md
+++ b/extensions/multiline.md
@@ -112,6 +112,8 @@ When using `draft/multiline-concat` to split longer lines, care is required to e
 
 The recommended splitting method is to split long lines between words where possible and leave the space character at the end of the line. This ensures that pairs of words spanning a line concatenation are still space separated, without adding a space to the start of any fallback lines.
 
+Take care not to split in the middle of multi-byte character sequences.
+
 In order to calculate the maximum allowed space before splitting a line, consider the following line:
 
 ```
--- a/extensions/multiline.md
+++ b/extensions/multiline.md
@@ -35,7 +35,9 @@ Mulitline messages allow for a more coherent experience that avoids the disconne
 
 ### Dependencies
 
-This specification uses the [`batch`][] capability which MUST be negotiated to use multiline messages. It also uses the [message tags][] and [standard replies][] frameworks.
+This specification depends on the [`batch`][] capability which MUST be negotiated to use multiline messages. The order of capability negotiation is not significant and MUST not be enforced.
+
+This specification also uses the [message tags][] and [standard replies][] frameworks.
 
 ### Capabilities
 
--- a/extensions/multiline.md
+++ b/extensions/multiline.md
@@ -115,7 +115,7 @@ The recommended splitting method is to split long lines between words where poss
 In order to calculate the maximum allowed space before splitting a line, consider the following line:
 
 ```
-:nick!~user@host PRIVMSG #channel :hello \n\r
+:nick!~user@host PRIVMSG #channel :hello \r\n
 ```
 
 The parts of a message that must fit within the 512 byte limit are as follows
@@ -131,7 +131,7 @@ The parts of a message that must fit within the 512 byte limit are as follows
 | command | `PRIVMSG` | 7 |
 | channel | `#channel` | variable |
 | message separator | ` :` | 2 |
-| crlf | `\n\r` | 2 |
+| crlf | `\r\n` | 2 |
 | message | `hello ` | variable |
 
 Not all parts may appear in a message, but a safe limit should take them into account and include an extra safety buffer in case of any unexpected syntax added by a server, e.g. a statusmsg prefix.
--- a/extensions/multiline.md
+++ b/extensions/multiline.md
@@ -92,7 +92,7 @@ Servers MUST use `FAIL` messages from the [standard replies][] framework to noti
 
 ### Message tags ([spec][message tags])
 
-Clients MUST only send non-`batch` tags on the opening BATCH command when sending multiline batches.
+Clients MUST NOT send tags other than `draft/multiline-concat` and `batch` on messages within the batch. In particular, all client-only tags associated with the message must be sent attached to the initial BATCH command.
 
 ### Message ids ([spec][message ids])
 
--- a/extensions/multiline.md
+++ b/extensions/multiline.md
@@ -50,7 +50,7 @@ Clients MUST ignore every token with a key that they don't understand.
 The only defined capability key so far is:
 
 * `max-bytes` - This defines the maximum allowed total byte length of multiline batched content *REQUIRED*
-* `max-lines` - This defines the maximum allowed number of lines in a multiline batch content *OPTIONAL*
+* `max-lines` - This defines the maximum allowed number of lines in a multiline batch content *RECOMMENDED*
 
 ### Tags
 
--- a/extensions/multiline.md
+++ b/extensions/multiline.md
@@ -35,7 +35,7 @@ Mulitline messages allow for a more coherent experience that avoids the disconne
 
 ### Dependencies
 
-This specification uses the [batch][] capability which MUST be negotiated at the same time. It also uses the [message tags][] and [standard replies][] frameworks.
+This specification uses the [`batch`][] capability which MUST be negotiated to use multiline messages. It also uses the [message tags][] and [standard replies][] frameworks.
 
 ### Capabilities
 
--- a/extensions/multiline.md
+++ b/extensions/multiline.md
@@ -1,0 +1,237 @@
+---
+title: Multiline messages
+layout: spec
+work-in-progress: true
+copyrights:
+  -
+    name: "James Wheare"
+    period: "2019"
+    email: "james@irccloud.com"
+---
+
+## Notes for implementing work-in-progress version
+
+This is a work-in-progress specification.
+
+Software implementing this work-in-progress specification MUST NOT use the unprefixed `multiline-concat` tag or `multiline` CAP/batch names. Instead, implementations SHOULD use the `draft/multiline-concat` tag, `draft/multiline` CAP and `draft/multiline` batch names to be interoperable with other software implementing a compatible work-in-progress version.
+
+The final version of the specification will use an unprefixed tag name.
+
+## Introduction
+
+This specification adds a new batch type and tag sent by clients and servers to send messages that can exceed the usual byte length limit and that can contain line breaks.
+
+## Motivation
+
+IRC messages have been traditionally limited by the line-based nature of the protocol and the line length limit of 512. To work around these limitations, client implementations split longer messages into multiple lines, and users need to rely on snippet hosting services to send multiple lines without flooding the chat or having their messages interrupted by other users.
+
+Mulitline messages allow for a more coherent experience that avoids the disconnected messages that result from these workarounds.
+
+## Architecture
+
+### Dependencies
+
+This specification uses the [batch][] capability which MUST be negotiated at the same time. It also uses the [message tags][] and [standard replies][] frameworks.
+
+### Capabilities
+
+This specification adds the `draft/multiline` capability.
+
+Implementations requesting this capability indicate that they are capable of handling the batch type and message tag described below.
+
+The capability has a REQUIRED value: a comma (`,`) (0x2C) separated list of tokens. Each token consists of a key which might have a value attached. If there is a value attached, the value is separated from the key by an equals sign (`=`) (0x3D). That is, `<key>[=<value>][,<key2>[=<value2>][,<keyN>[=<valueN>]]]`. Keys specified in this document MUST only occur at most once.
+
+Clients MUST ignore every token with a key that they don't understand.
+
+The only defined capability key so far is:
+
+* `max-bytes` - This defines the maximum allowed total byte length of multiline batched content *REQUIRED*
+* `max-lines` - This defines the maximum allowed number of lines in a multiline batch content *OPTIONAL*
+
+### Tags
+
+This specification adds the `draft/multiline-concat` message tag, which has no value and is used to send message lines that extend beyond the usual byte length limit. Its usage is described below.
+
+### Batch types
+
+This specification adds the `draft/multiline` batch type, and introduces client initiated batches. These use the same syntax as server initiated batches.
+
+In addition to the base batch parameters (reference-tag and type) a multiline batch has one additional parameter, the target recipient.
+
+Multiline batches MUST contain one or more PRIVMSG lines whose target MUST match the batch target.
+
+When receiving a well-formed mulitiline message batch, implementations MUST collect each PRIVMSG message content and wait until the full batch has been received before processing the message. Processing in this context refers to:
+
+* Servers: delivering the batch to the intended recipients
+* Clients: displaying the batched message to the user
+
+Messages in a multiline batch MUST be concatenated with a line break unless the `draft/multiline-concat` message tag is sent, in which case the message is directly concatenated with the previous message without any separation.
+
+Clients MUST NOT send messages other than PRIVMSG while a multiline batch is open.
+
+### Fallback
+
+When delivering multiline batches to clients that have not negotiated the multiline capability, servers MUST deliver the component messages without using a multiline BATCH.
+
+Any tags that would have been added to the batch, e.g. message IDs, account tags etc MUST be included on the first message line. Tags MAY also be included on subsequent lines where it makes sense to do so.
+
+### Errors
+
+Servers MUST use `FAIL` messages from the [standard replies][] framework to notify clients of errors with multiline batches. The command is `BATCH` and the following codes are defined:
+
+* `MULTILINE_MAX_BYTES <limit>`: `max-bytes` limit exceeded
+* `MULTILINE_MAX_LINES <limit>`: `max-lines` limit exceeded
+* `MULTILINE_INVALID_TARGET <batch-target> <provided-target>`: mismatched PRIVMSG target within a batch
+* `MULTILINE_INVALID`: any other error
+
+## Interaction with other specifications
+
+### Message tags ([spec][message tags])
+
+Clients MUST only send non-`batch` tags on the opening BATCH command when sending multiline batches.
+
+### Message ids ([spec][message ids])
+
+Servers MUST only include a message ID on the first message of a batch when sending a fallback to non-supporting clients.
+
+### Labeled response ([draft][labeled response]) with echo-message ([spec][echo message])
+
+Servers MUST only include the label tag on the opening BATCH command when replying to a client using `echo-message`.
+
+## Client implementation considerations
+
+This section is non-normative.
+
+### Splitting long lines
+
+When splitting long lines, care is required to ensure messages are displayed appropriately when combined or in the fallback case.
+
+The recommended splitting method is to split long lines between words where possible and leave the space character at the end of the line. This ensures that concatenated lines don't lose the space, while fallback lines don't start with a space.
+
+In order to calculate the maximum allowed space before splitting a line, consider the following line:
+
+```
+:nick!~user@host PRIVMSG #channel :hello \n\r
+```
+
+The parts of a message that must fit within the 512 byte limit are as follows
+
+| part | syntax | length |
+|-|-|-|
+| leading | `:` | 1 |
+| nick | `nick` | variable |
+| nick/mask separator | `!` | 1 |
+| user | `~user` | variable |
+| user/host separator | `@` | 1 |
+| host | `host` | variable |
+| command | `PRIVMSG` | 7 |
+| channel | `#channel` | variable |
+| message separator | ` :` | 2 |
+| crlf | `\n\r` | 2 |
+| message | `hello ` | variable |
+
+Not all parts may appear in a message, but a safe limit should take them into account and include an extra safety buffer in case of any unexpected syntax added by a server, e.g. a statusmsg prefix.
+
+With a safety buffer of 10, the total message size available before a split is needed can be calculated as follows
+
+```
+512 - 14 - 10 - <nick> - <user> - <host> - <channel>
+```
+
+Clients might wish to perform a `WHO` query on themselves when connecting to a server, and keep track of numeric `396 RPL_VISIBLEHOST` to keep their `user@host` mask in sync.
+
+Clients might alternatively prefer to use a hard coded limit. A possible worst case scenario for the variable parts might be:
+
+* nick: 20
+* user: 20
+* host: 63
+* channel: 32
+
+Which would leave 353 bytes for each message.
+
+## Server implementation considerations
+
+This section is non-normative.
+
+In the context of flood protection, keep in mind that clients might send an entire batched multiline message all at once.
+
+## Examples
+
+This section is non-normative.
+
+NOTE: In these examples, `<SPACE>` indicates a space character which would otherwise not be clearly visible.
+
+Client sending a mutliline batch
+
+    Client: BATCH +123 draft/multiline #channel
+    Client: @batch=123 PRIVMSG #channel hello
+    Client: @batch=123 privmsg #channel :how is<SPACE>
+    Client: @batch=123;draft/multiline-concat PRIVMSG #channel :everyone?
+    Client: BATCH -123
+
+Server sending the same batch
+
+    Server: @msgid=xxx;account=account :n!u@h BATCH +123 draft/multiline #channel
+    Server: @batch=123 :n!u@h PRIVMSG #channel hello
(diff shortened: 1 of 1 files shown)


extensions/multiline.md
```diff
@@ -0,0 +1,241 @@
...
+
+## Interaction with other specifications
+
+### Message tags ([spec][message tags])
+
+Clients MUST only send non-`batch` tags on the opening BATCH command when sending multiline batches.
+
+### Message ids ([spec][message ids])
+
+Servers MUST only include a message ID on the first message of a batch when sending a fallback to non-supporting clients.
```
ProgVal: Why the first one? If the message ID is used eg for a reply or a reaction emoji, it visually makes more sense to have it attached to the last message.

slingamn: If I understand correctly, this is saying that (some of) the fallback `PRIVMSG` should be sent without any `msgid` tag at all? This seems to upend some expectations (e.g., msgids being available for deduplication purposes).

slingamn: For comparison, [oragono.io/maxline-2](https://oragono.io/maxline-2) generates n+1 distinct message IDs for the original message and all n lines of the split message. This isn't perfect, but it maintains the expectation that `PRIVMSG` should have an ID.

The msgid spec says: "If this tag is being used, it SHOULD be attached to all PRIVMSG and NOTICE events."

jwheare: msgids can't appear more than once. Generating unique n+1 msgids and only sending them on the fallback creates undesirable situations where people might reply to a fallback message and the context would then be missing for people who fully support maxline (i.e. discard fallback message tags)

If you're concerned about missing msgids, the solution would be to properly support the maxline batch spec.

edk0: I don't think that's true. It'd be easy to generate unique fallback msgids that the server can recognise as such and translate back or refuse replies for or something.

slingamn: @jwheare, just to be clear, when you say "maxline" you're referring to the present multiline spec, not to oragono.io/maxline-2?

Some degradation of experience is inevitable under any proposal to support PRIVMSG over 512 bytes. I'm much less concerned about the case you're describing than I am about sending PRIVMSG without IDs, which seems to complicate the handling of chat history at a more fundamental level.

I'm not 100% sold on @edk0's suggestion because it violates an expectation I have that servers shouldn't need to care about client-only tags, but I could be convinced.

Hypothetically, we could define a new tag like `draft/fmsgid` ("fallback msgid") for transmitting the fallback msgids to clients that support the multiline spec.

edk0: I'm going to care about client-only tags anyway, so it's not a bother for me. You don't have to implement my suggestion; I'm just saying a solution is possible.

slingamn:
Are there any constraints on which messages in the client batch can carry client-only tags, and what those tags can be?

jwheare:
The way I've implemented this in the client, the only tags within the batch that I use are the @draft/multiline-concat, every other tag must appear on the opening BATCH message to be recognised.

`Any tags that would have been added to the batch, e.g. message IDs, account tags etc MUST be included on the first message line. Tags MAY also be included on subsequent lines where it makes sense to do so.`

jwheare:
Actually that wording is from the fallback handling and refers to unbatched messages. Could use some clearer guidance on non-fallback BATCHed messages and where to include the tag, though I feel it is implied by the fact that the BATCH indicates a single logical message.

jwheare:
Oh it is mentioned actually:

`Clients MUST only send non-`batch` tags on the opening BATCH command when sending multiline batches.`

That means only `@batch` and `@multiline-concat` are allowed on messages within the batch.

extensions/multiline.md
```diff
@@ -0,0 +1,241 @@
...
+
+    Client: BATCH +abc draft/multiline #channel
+    Client: @batch=abc PRIVMSG #channel hello
+    ... lines that exceeed the max-bytes=40000 limit ...
+
+    Server: :irc.example.com FAIl BATCH MULTILINE_MAX_BYTES 40000 :Multiline batch max-bytes exceeded
+
+`max-bytes` exceeded multiline batch
+    
+    Server: CAP * LS :draft/multiline=max-lines=10
```
slingamn: This cap value is technically invalid, since `max-bytes` is mandatory.

Also, the title of this section still says `max-bytes`.

extensions/multiline.md
```diff
@@ -0,0 +1,241 @@
...
+* host: 63
+* channel: 32
+
+Which would leave 353 bytes for each message.
+
+## Server implementation considerations
+
+This section is non-normative.
+
+In the context of flood protection, keep in mind that clients might send an entire batched multiline message all at once.
```
slingamn: I haven't thought this through fully, but there's an issue here with sendqs. If the sendq is small relative to the value of max-bytes, you can DoS clients by completing multiple multiline batches at once and filling up their sendqs.

extensions/multiline.md
```diff
@@ -0,0 +1,241 @@
...
+* `MULTILINE_MAX_BYTES <limit>`: `max-bytes` limit exceeded
+* `MULTILINE_MAX_LINES <limit>`: `max-lines` limit exceeded
+* `MULTILINE_INVALID_TARGET <batch-target> <provided-target>`: mismatched PRIVMSG target within a batch
+* `MULTILINE_INVALID`: any other error
+
+## Interaction with other specifications
+
+### Message tags ([spec][message tags])
+
+Clients MUST only send non-`batch` tags on the opening BATCH command when sending multiline batches.
```
slingamn: Suggested language from discussion in #ircv3: Clients MUST NOT send tags other than draft/multiline-concat on messages within the batch. In particular, all client-only tags associated with the message must be sent attached to the initial BATCH message.

extensions/multiline.md
slingamn: Following up on a clarifying question from #ircv3: yeah, client A can try to disconnect client B from a server by filling up its sendq. Normally, flood control is a check on this. It's not clear what the best practices are for applying flood control to multiline messages --- the way to get parity with pre-multiline behavior w.r.t. the sendq would be to apply artificial delays to the *outgoing* batch.

slingamn: If `max-lines` is not set, the worst case for a DoS on the sendq is to send `max-bytes` lines, with a single byte on each line, and `multiline-concat` on every line after the first. Assuming the longest possible nickmask, the total size of the relayed message will be about 100 times `max-bytes`. Even for a conservative value of `max-bytes`, this should be large enough to overwhelm both the sendq and the TCP send buffer (I think it's typically around 85kB on modern Linux machines).

slingamn:
re. interactions between multiline and labeled-response: on which message does the *client* send the label tag, the `BATCH +` or the `BATCH -` line?

jwheare:
BATCH +

slingamn:
I have a (likely buggy) implementation of this up on testnet.oragono.io. It uses fallback msgids (each individual message in the multiline batch has an alternative msgid, which is sent to multiline-capable clients under the `draft/fmsgid` tag, and to legacy clients under the `msgid` tag).

Please report any bugs at https://github.com/oragono/oragono :-)

xPaw:
Do I understand correctly that the `<SPACE>` at the end of the line in these examples means there shouldn't be a new line inserted in the concatenated message?

```
foreach msg in messages
 concat_msg += msg
 if msg does not end with space
  concat_msg += "\n"
```

This is *almost* explained in splitting long lines, but for the sake of verbosity, it probably warrants an explicit explanation.

jwheare:
No, the `draft/multiline-concat` tag indicates that. The `<SPACE>` is there because without a newline, the words are separated with a space.

xPaw:
Oh duh, I completely missing the tag there.

The "leave the space character at the end of the line. This ensures that concatenated lines don't lose the space" paragraph mislead me. I do see why it's this way for fallback messages though.

jwheare:
Any suggestions for clarifying that paragraph to make it less misleading?

xPaw:
Probably mentioning the tag there would help clarify it.

I can't think of an example where "This ensures that concatenated lines don't lose the space" is relevant besides buggy code that trims the messages?

jwheare:
Any outstanding comments before we merge this as a draft?

slingamn:
Two issues I raised haven't been addressed:

1. The question of "fallback msgids"
1. Denial-of-service attacks that exhaust the sendq

jwheare:
1. I'm not in favour of them.
2. Any suggested wording that should be in the spec to address that?

jwheare:
To expand on 1) I feel the benefit to complexity ratio is skewed. It's aimed at the small subset of people who do support message tags, but don't support multiline, which will likely only become smaller as time goes on.

slingamn:
On the one hand, I'm concerned about a spec that mandates sending any PRIVMSG without a message ID tag, because it seems like this compromises the use of message IDs as "infrastructure" (e.g., for deduplication). On the other hand, the possibility of skipping directly to the world where everyone supports both message IDs and multiline batches is tempting. So I'm not sure what I think right now.

re. DoS, I think at the very least the spec should require that max-lines be set. Beyond that, I think this is a serious problem to which I do not have a complete solution. A possibility that occurred to me is suggesting that servers apply fakelag to the *outgoing* messages as necessary. This would add a lot of implementation complexity, though.

extensions/multiline.md
```diff
@@ -0,0 +1,241 @@
...
+| leading | `:` | 1 |
+| nick | `nick` | variable |
+| nick/mask separator | `!` | 1 |
+| user | `~user` | variable |
+| user/host separator | `@` | 1 |
+| host | `host` | variable |
+| command | `PRIVMSG` | 7 |
+| channel | `#channel` | variable |
+| message separator | ` :` | 2 |
+| crlf | `\n\r` | 2 |
```
xPaw: crlf is \r\n

slingamn:
Two questions:

1. The stated rationale for sending PRIVMSG without msgids to legacy clients is that this is a transitional period, and eventually every client that supports tags will also support multiline, rendering the question moot. Does this imply that after some point in the future, we will no longer allow new specifications that require sending a PRIVMSG without a msgid, at which point client developers will be able to rely on the presence of msgids?
1. Suppose a client wants to implement support for draft/chathistory (possibly at a very rudimentary level, like just sending a `LATEST` query after every JOIN) during this transitional period, but doesn't want to implement draft/multiline (e.g., because it breaks too many assumptions). Is there any way for them to correctly deduplicate the history? There's seemingly no way to distinguish between duplicate playback of the intermediate lines of a multiline message, and intermediate lines that are genuinely repeated in the message, because the time tag will be the same on each line.

jwheare:
1. There is no hard requirement that all PRIVMSGs have a msgid. It's only a SHOULD in the spec.
2. Are you aware of real world implementors this would affect or is it a hypothetical? Some concrete spelled out examples of how this might manifest might help.

slingamn:
1. Yes, but apart from this, there's no good reason to send a PRIVMSG without a msgid, so it would be reasonable for clients to degrade in that case (e.g., by assuming that anything sent without a msgid is always a duplicate, or is never a duplicate, depending which side of the line they prefer to err on).
1. It's mostly a hypothetical. I'm thinking of adding this kind of support to either Hexchat, irssi, or https://github.com/MCMrARM/revolution-irc, but I haven't looked too deeply at what would be required. It's true that implementing CHATHISTORY sensibly already requires support for BATCH, so the added burden is just to support nested batches.

slingamn:
I tentatively withdraw my objection: I've come around to thinking that instead CHATHISTORY should be designed so that rudimentary use is possible without any awareness of BATCH, msgids, or deduplication.

jwheare:
OK, so without further comments, this will be merged as draft on Monday.

slingamn:
My security concerns still haven't been addressed. I would propose:

1. Make `max-lines` mandatory
2. Add a "security considerations" section, with language along the following lines: "Server implementations and operators should consider the potential effects of relaying large amounts of message data at once. In particular, large multiline messages have the potential to exhaust client send queues, which will either result in client disconnection (a denial-of-service attack) or degraded client performance. Size on the wire will typically be constrained by `max-lines`, not `max-bytes`, because each additional protocol line can carry a large amount of metadata (in particular the sender's full nickmask and the channel name). Operators should set `max-bytes`, `max-lines`, and the send queue length in tandem to mitigate these attacks."

jwheare:
Strongly disagree that max lines needs to be mandatory. *if* a server uses a line based sendq then it might make sense but not all servers do and implementation details shouldn’t leak into the spec. Bytes are a less arbitrary limit on data that more directly maps to universal resource constraints.

jwheare:
Happy to include some server considerations based on the rest of your suggestion though.

jwheare:
Would be happier still to get feedback from other server implementers as well!

slingamn:
>if a server uses a line based sendq then it might make sense but not all servers do and implementation details shouldn’t leak into the spec.

The motivation is not line-based sendq's, but byte-based sendq's. The problem is that if max-lines is unset, the total size of the relayed message can be approximately 100 times max-bytes (after including the full nickmask, the channel target, and the concat tag on every line).

jwheare:
I want to avoid a situation where servers apply arbitrarily low limits on lines because they’re forced to pick a limit. Bearing in mind people use this feature to post code snippets.

If we want servers to consider the byte budget of repeated line envelopes, we should probably talk about it in terms of bytes. e.g. if you’re trying to account for 100 extra bytes per line, and your byte limit is 4000, the max possible lines is 4000 if each new line counts for a 1B \n That would require a sendq of at most 400KB. If you want to reduce that to 40KB, you’d need a max-lines 400.

What are typical sendq values in default or common configurations or on popular servers in the real world? Would we consider them overly conservative on the whole?

It’s not for the spec to decide resource constraints for servers, but we should help them come to a non arbitrary decision by thinking about them appropriately.

jwheare:
Side note: we should probably clarify how many bytes of *content* a line break takes up in terms of max-bytes. Internally on the server, in our implementation, it’s a single \n byte, rather than the CRLF of the protocol separator.

slingamn:
Unreal's default is [200 KB](https://github.com/unrealircd/unrealircd/blob/ead0ad0aa69016e04fc4f4959ca2efabb3cad2c1/doc/conf/examples/example.conf#L88). Inspircd's default is [1MB](https://github.com/inspircd/inspircd/blob/b7382a91384e6e9463ba85c6effec32f071dd4ca/docs/conf/inspircd.conf.example#L408). Oragono's is 16 MB (probably too low, we should fix that).

jwheare:
The orders of magnitude there suggest that there often isn’t really a need for a separate max-lines, unless you’re unreal and don’t want to raise the sendq but still allow 4000 max bytes. So yeah, I’m still against it being mandatory, but there should certainly be some discussion, guidance around it.

slingamn:
You can exceed the default inspircd sendq using three coordinating clients who terminate their batches at the same time.

It's also worth thinking about this in terms of raw bandwidth. If we accept that a single multiline message can take up 400 KB on the wire, relaying it to a 1000-client channel is 400 MB, which will exhaust a 100 mbps connection for approximately 4 seconds.

slingamn:
This doesn't eliminate the problem (since it does nothing for legacy clients), but a potential amelioration is to say that the server MAY send `*` in place of both the nickmask and the target to fully capable clients, on the PRIVMSGs inside the batch. This could be about a 3x reduction in wire size?

To be clear, I still think max-lines should be required; values between 128 and 512 seem sane to me.

jwheare:
I’d be up for such an optimisation or similar even if it doesn’t help legacy.

One reason i don’t want to mandate max-lines: we’ve implemented this in a Slack gateway, and Slack has no such limitation.

Definitely worth mentioning the case for having such a limit though.

slingamn:
>One reason i don’t want to mandate max-lines: we’ve implemented this in a Slack gateway, and Slack has no such limitation.

Totally fair, it's much less of a concern in any setting where you're not worried about abuse. I'd be fine with recommending it instead of mandating it.

Suggested language: "To save space on the wire, when relaying multiline batches to fully compatible clients, servers MAY omit the source nickmask and send a `*` in place of the channel name on the PRIVMSG lines inside the batch (since this information is redundant with the initial BATCH line)."

jwheare:
I think we decided in IRC that this optimisation is probably not worth it. But we still need some more advice on setting a sensible max-lines value.

slingamn:
As I remember it, we decided that the most aggressive optimization possible (sending `*` for the command and sending the message as a single parameter) wasn't worth it. I'm still very interested in doing a more modest optimization (omitting the prefix, keeping PRIVMSG as the command, and sending `*` as the first parameter and the message as the second). Suggested language for this is in my previous comment.

The main "upwards" constraint you identified earlier on max-lines is the need to hold code snippets. I think 128 lines is a reasonable limit for this.

SadieCat:
Messages without a prefix are from the locally connected server. Also, this "optimisation" seems like it would complicate both client and server code for very little benefit.

slingamn:
These messages would only be sent to fully compliant clients, which in any case need to take the batch tag into account when interpreting the message. So there should be no increase in client complexity over the baseline required to implement multiline. As for server complexity, it's a MAY, so servers don't have to implement it if they don't want it.

The potential benefit is a 60% reduction in the size of relayed messages, which seems worthwhile to me.

kythyria:
Compromise: Put only the nick as the sender. Could even put `*` as the sender, though that's perhaps inconsistent with other extensions using `*` to mean "yourself".

SadieCat:
I don't see the need for this optimisation and I don't think it should be mandated by the spec.

slingamn:
Implementations might want to use the target for routing messages (in addition to the batch tag), so I withdraw my proposal to send `*` as the target.

I still think we should allow servers to omit the source nickmask: this is syntactically unproblematic, and should be semantically unproblematic for clients as well (at least one implementation supports this "by default", since sources for messages within the relayed S2C batch are ignored).

A lot of numbers have been bandied back and forth, so here's what I think is a realistic estimate: for a 100-line code snippet with an average line length of 60 lines, relayed to a channel with an 8-character name by a user with a 40-character nickmask, I compute a 30% reduction in wire size.

jwheare:
Did we decide if this needs edits before merging as draft?

slingamn:
1. CRLF is still incorrectly specified as `\n\r`
2. No action has been taken on my suggested "security considerations" section ([proposed language](https://github.com/ircv3/ircv3-specifications/pull/398#issuecomment-578458879))
3. I am still holding out hope for including "servers MAY omit the source nickmask on PRIVMSG within the batch" --- is this proposal definitely vetoed? I haven't really heard compelling arguments against it, other than the convention that lines without a source are sent by the local server. Since this change would only affect clients with full support for multiline, this doesn't seem that persuasive to me.

jwheare:
I’ll take a look at 1 and 2.

3: I think it comes under the same “not worth the micro optimisation” justification as the other stuff on this theme.

slingamn:
The other proposed optimizations had stronger arguments against them. I think this one is safer than the other proposals, and achieves most of their possible gains.

jesopo:
I find the inclusion of entirely redundant information (`:source`) to make this look and feel strange. I'd much prefer we totally omit this, even if I ignore the fact that it is an optimisation.

jwheare:
@slingamn those changes should cover your 1 and 2 suggestions.

I still think we need some sort of guidelines for configuring limits, but I'm struggling to arrive at something that's sensible, future-proof, not overly restrictive etc.

I tried plotting some numbers but it didn't really help.

Any ideas?

slingamn:
The proposed language in [my earlier comment](https://github.com/ircv3/ircv3-specifications/pull/398#issuecomment-578458879) is aimed more at drawing attention to the issue, rather than suggesting concrete numbers (which I agree is a difficult problem).

jwheare:
Yes I added a version of your proposal in https://github.com/ircv3/ircv3-specifications/pull/398/commits/ff7f73bbb510e3c2caf504a84cb21fabebae9b48 but I still feel like something more is needed.

slingamn:
Oh, sorry I missed that. I don't think I have much insight into this question, other than the suggestion of "envelope math", which you already tried.

It sounds like there is an upwards constraint on these limits (supporting all realistic use cases for the feature) and a downwards constraint (blocking abuse), and probably the best way to get more information about both of these constraints is simply to move forward with the specifications process (i.e., make it an official draft, if that's the correct next step) and see what happens in the wild. This is because:

a. Any guidance about limits we add later will be in a "non-normative" section of the spec
b. These limits will typically be configurable, not hardcoded, so operators will be able to adopt new recommendations without requiring any code changes from developers

p.s. In terms of moving forwards on the specifications process, I haven't given up on item 3, but I don't really have any further arguments for it beyond what I've already stated.

jwheare:
OK, makes sense re: further limits guidance.

I think in the absence of consensus over how to optimise these messages, again we should just proceed to merged draft and see if anything comes out of implementations.

At the end of the day, the unbatched fallback messages will never be able to be optimised, so implementations being able to cope without it is effectively a requirement.

slingamn:
That sounds fair.

jwheare:
OK will merge if no further comments are forthcoming.

extensions/multiline.md
```diff
@@ -0,0 +1,245 @@
...
+
+IRC messages have been traditionally limited by the line-based nature of the protocol and the line length limit of 512. To work around these limitations, client implementations split longer messages into multiple lines, and users need to rely on snippet hosting services to send multiple lines without flooding the chat or having their messages interrupted by other users.
+
+Mulitline messages allow for a more coherent experience that avoids the disconnected messages that result from these workarounds.
+
+## Architecture
+
+### Dependencies
+
+This specification uses the [batch][] capability which MUST be negotiated at the same time. It also uses the [message tags][] and [standard replies][] frameworks.
```
DarthGandalf, line 38: What if client enabled `batch` before? E.g. `CAP REQ batch` followed by `CAP REQ draft/multiline`

Or after, but still during registration time

extensions/multiline.md
```diff
@@ -0,0 +1,245 @@
...
+
+Which would leave 353 bytes for each message.
+
+## Server implementation considerations
+
+This section is non-normative.
+
+For server implementations and operators, consider the impact of relaying a large multiline batch. In particular, the potential to exhaust client send queue budgets, which will either result in client disconnection (a denial-of-service attack) or degraded client performance.
+
+Each additional protocol line carries extra metadata beyond the message content, such as the sender's full source mask, the command, and the target. This means that the byte size of a multiline batch sent to clients will typically be a function of max-lines as well as max-bytes. Configuring max-bytes, max-lines, and the send queue length together is reommended to mitigate these issues.
```
DarthGandalf, line 162: typo: reCommended

extensions/multiline.md
```diff
@@ -0,0 +1,245 @@
...
+
+This specification adds the `draft/multiline-concat` message tag, which has no value and is used to send message lines that extend beyond the usual byte length limit. Its usage is described below.
+
+### Batch types
+
+This specification adds the `draft/multiline` batch type, and introduces client initiated batches. These use the same syntax as server initiated batches.
+
+In addition to the base batch parameters (reference-tag and type) a multiline batch has one additional parameter, the target recipient.
+
+Multiline batches MUST contain one or more PRIVMSG lines whose target MUST match the batch target.
```
DarthGandalf, line 65: Can a batch contain 2 PRIVMSGs with different targets?

extensions/multiline.md
```diff
@@ -0,0 +1,245 @@
...
+Clients might alternatively prefer to use a hard coded limit. A possible worst case scenario for the variable parts might be:
+
+* nick: 20
+* user: 20
+* host: 63
+* channel: 32
+
+Which would leave 353 bytes for each message.
+
+## Server implementation considerations
```
DarthGandalf, line 156: I think server should be prepared to handle a case if an utf-8 character is split mid-way

extensions/multiline.md
```diff
@@ -0,0 +1,245 @@
...
+
+This specification uses the [batch][] capability which MUST be negotiated at the same time. It also uses the [message tags][] and [standard replies][] frameworks.
+
+### Capabilities
+
+This specification adds the `draft/multiline` capability.
+
+Implementations requesting this capability indicate that they are capable of handling the batch type and message tag described below.
+
+The capability has a REQUIRED value: a comma (`,`) (0x2C) separated list of tokens. Each token consists of a key which might have a value attached. If there is a value attached, the value is separated from the key by an equals sign (`=`) (0x3D). That is, `<key>[=<value>][,<key2>[=<value2>][,<keyN>[=<valueN>]]]`. Keys specified in this document MUST only occur at most once.
```
DarthGandalf, line 46: The value is sent by server, not by client?

extensions/multiline.md
jwheare: No

extensions/multiline.md
jwheare: What do you mean "handle"? What would the expected behaviour be?

extensions/multiline.md
DarthGandalf: "Multiline batches MUST only contain one or more PRIVMSG lines, all these lines MUST have a target which matches the batch target" would be clearer.

extensions/multiline.md
jwheare: Yes, is that relevant?

extensions/multiline.md
DarthGandalf: Somehow handle. Even if by dropping the message on the floor and sending FAIL to client. Or by passing it to clients as is.

I think this shouldn't be left to implementation details to every implementation to decide for temselves.

extensions/multiline.md
DarthGandalf: It's just not clear from this paragraph. The text reads like the client MUST send the value as well, because it's REQUIRED.

extensions/multiline.md
slingamn: Dependent caps are a Whole Thing. Here's my take on how they should work:

https://github.com/ircv3/ircv3-specifications/pull/400#issuecomment-579063998

jwheare: I made this match how we phrase it in labeled-response.

extensions/multiline.md
jwheare: I expect servers don't want to inspect the content at this level, so this will probably just involve forwarding onto clients as is. We should probably tell clients not to split within multibyte characters though.

jwheare: e8efcaecb96912df1d73b2f5b9d9b51b9a09f5ed

jwheare:
@DarthGandalf all good?

DarthGandalf:
> @DarthGandalf all good?

For draft, no further comments.

PR #400, opened 2020-01-17
[labeled-response] Make batch a hard dependency

jwheare:
This is a blocker for #378

### DIFF
--- a/extensions/labeled-response.md
+++ b/extensions/labeled-response.md
@@ -37,7 +37,7 @@ Additionally, labeled responses allow bouncers with multiple connected clients t
 
 ### Dependencies
 
-This specification uses the [message tags framework](../extensions/message-tags.html) and the [`batch`](../extensions/batch-3.2.html) capability which SHOULD be negotiated at the same time.
+This specification uses the [message tags framework](../extensions/message-tags.html) and depends on the [`batch`](../extensions/batch-3.2.html) capability which MUST be negotiated to use labeled responses.
 
 ### Capabilities
 
@@ -59,12 +59,10 @@ This tag MAY be sent by a client for any messages that need to be correlated wit
 
 For any message received from a client that includes this tag, the server MUST include the same tag and value in any response required from this message where it is feasible to do so. Servers MUST include the tag in exactly one logical message.
 
-If a response consists of more than one message, and the `batch` capability is enabled, a batch MUST be used to group them into a single logical response. The start of the batch MUST be tagged with the `draft/label` tag. The batch type MUST be one of:
+If a response consists of more than one message, a batch MUST be used to group them into a single logical response. The start of the batch MUST be tagged with the `draft/label` tag. The batch type MUST be one of:
 
 * An existing type applicable to the entire response
 * `draft/labeled-response`
-
-If `batch` has not been enabled, the tag MAY be included on only one of the messages in the response.
 
 When a client sends a message to itself, the server MUST NOT include the label tag, except for any acknowledgment sent with the [`echo-message`](/specs/extensions/echo-message-3.2.html) mechanism. This allows clients to differentiate between the echoed message response, and the delivered message.
 


SadieCat:
LGTM. :shipit:

RyanSquared:
> Re: enforcement, the current merged wording from #400 just says that "[batch] MUST be negotiated to use labeled responses." but doesn't say anything about how or when that happens. I think that should be sufficient?

I hate to have to bring this up now that this is merged but I've been out of it the past few days while discussion of this is active. The main reason behind my PR as a whole was because some implementations would not allow usage of labeled-responses if batch was not negotiated *beforehand*. I think there was also a discussion that it had to be negotiated and ACK'ed (as in, a separate REQ entirely) before labeled-response was able to be REQ'd. I think it's worth mentioning whether or not it's a "soft" or a "hard" dependency, and whether or not `batch` must be requested first.

slingamn:
>The main reason behind my PR as a whole was because some implementations would not allow usage of labeled-responses if batch was not negotiated beforehand. I think there was also a discussion that it had to be negotiated and ACK'ed (as in, a separate REQ entirely) before labeled-response was able to be REQ'd.

From my perspective, this is incorrect behavior. Which implementations do this? irccloud was mentioned but I'm not sure that's accurate.

>I think it's worth mentioning whether or not it's a "soft" or a "hard" dependency, and whether or not batch must be requested first.

What do "soft" and "hard" mean here?

Here's my position:

1. Servers MUST advertise either (a) both batch and labeled-response, or (b) only batch, or (c) neither.
1. Clients MUST request either (a) both batch and labeled-response, or (b) only batch, or (c) neither.
1. There are no ordering or same-line constraints on how the client requests both caps
1. If a client requests labeled-response without batch, the result is undefined behavior

I think all of this is implicit in the currently accepted language.

RyanSquared:
>> The main reason behind my PR as a whole was because some implementations would not allow usage of labeled-responses if batch was not negotiated beforehand. I think there was also a discussion that it had to be negotiated and ACK'ed (as in, a separate REQ entirely) before labeled-response was able to be REQ'd.

> From my perspective, [not allowing usage of labeled-responses if batch was not negotiated beforehand] is incorrect behavior. Which implementations do this? irccloud was mentioned but I'm not sure that's accurate.

It might no longer be accurate but I think it was the case at one point. It may have been some testing branch of an IRCd specification, but I think defining that this behavior is incorrect is important, which coincides with your position #4, "If a client requests labeled-response without batch, the result is undefined behavior"

I think that given position #3, the no-ordering constraint, this directly conflicts with #4 because it both defines a behavior where labeled-response can be requested before batch (thus fulfilling the "no ordering or same-line constraints", but then establishes in the next point that doing so is undefined behavior.

I'm unsure what should happen if a client requests labeled-response without having batch requested beforehand. I think that if labeled-response is requested first, but batch is not requested, that servers should not send the message but still send the `CAP ACK` to imply that they do support the behavior if clients should later request batch.

If I understand correctly, capabilities can be requested after "capability negotiation" at any time using a `CAP REQ` command (for the case of a client or a server being upgraded in-place and being able to support new capabilities, for example). This means that it's not easily possible to just say "wait until CAP END to support it", even though servers *really shouldn't* do this as it can nullify the "agreement" of sorts given by a `CAP ACK`.

To further elaborate on your positions, I would accept them if #4 was changed to the following:

4. If a client requests labeled-response without batch, the result of messages requesting labeled-response is undefined behavior

I believe this is the point that you wanted to get across, as this way labeled-response can be requested without batch, it can still be supported if batch is later requested.

RyanSquared:
> [slingamn] yeah...i would say this language doesn't need to be in the labeled-response spec before ratification though, it could be added to the cap spec eventually

:+1:

slingamn:
>If a client requests labeled-response without batch, the result of messages requesting labeled-response is undefined behavior

The problem we're faced with here is a little easier than the general case, because labeled-response only comes into effect when the client "triggers" it by attaching a label tag to a command. The more general problem is something like `draft/multiline`, where you could asynchronously receive a relayed multiline batch at any time (post-registration), including between requesting `draft/multiline` and `batch` if you were requesting them in that order on separate lines.

Let's say cap Y depends on cap X. Here's how I'd put it (just to be clear, "register" refers to [connection registration](https://modern.ircdocs.horse/#connection-registration), not account registration):

1. Servers MUST advertise either (a) both X and Y, or (b) only X, or (c) neither
1. Clients MUST ensure that when registration is complete, they have received either (a) both X and Y, or (b) only X, or (c) neither
1. During registration, there are no ordering or same-line constraints on how X and Y are requested
1. If a registered client has received Y but not X, the server's behavior is undefined. Possibilities include: disconnecting the client, behaving as though Y were disabled, behaving as though both X and Y were enabled, and sending invalid data.

This implies that post-registration, there *is* an ordering constraint on how X and Y are requested: the client MUST request X before Y, or request them on the same line.

Pros of this approach:

1. Servers don't have to police anything
1. Clients that do all their cap negotiation during registration don't have to worry about anything (in particular, the ["naive intersection pseudocode"](https://github.com/ircv3/ircv3-specifications/pull/391#issuecomment-502438801) is correct during registration)

Cons of this approach:

1. Clients have to worry about dependencies if they want to handle `CAP NEW`
1. Client bugs will not necessarily be caught up front

PR #401, opened 2020-01-18
labeled-response: Make the non-normative section more explicit

DarthGandalf:


DarthGandalf:
I can make it the reverse (the server MUST reply at least with empty ACK if it understood that there will be no response for a certain request), but I want one of such options to be mentioned explicitly.

jwheare:
I don't see how this change makes anything more explicit. For me it raises more questions: why is the server not responding to this WHOIS? We can't enumerate all cases, but any case that we do mention, it should be common or obvious why it's being called out.

It might be better to just say "or for any other server-defined reason" if you want servers to feel able to not reply for any reason in future.

I don't think reversing this to requiring an ACK makes sense, the whole point of this caveat is to say that it's sometimes not feasible to reply with anything.

DarthGandalf:
> I don't see how this change makes anything more explicit.

Even knowing what to look for, I have a hard time finding it. And to find it in the current text, I have to squeeze the meaning of it somewhat.

The point of this change is to emphasize that it's not only about support itself, but also that the same message can be replied to, or not. To make it clearer for the reader what to expect from server, and what not to expect.

> why is the server not responding to this WHOIS?

I can provide the bouncer use case as an example of when that can happen, if you wish. I didn't think it's very valuable.

> "or for any other server-defined reason"

While it's technically true, I don't think that makes the text any clearer, at least by itself. Maybe something like `it might not be feasible for servers to attach a label to a message, for instance on certain remote or asynchronous actions, or for any other server-defined reason. For example, bouncer blah blah, in that case there might be no response blah blah`? (I'll try to improve the wording)

jwheare:
I think for your case, if the server *knows* there won't be a response, the spec already says you MUST send an ACK. This section is more about cases where the server doesn't know whether there will or won't be a response, so clients should be prepared to handle that. Does that need to be clarified?

DarthGandalf:
Well, it wasn't clear to me. Will try to get to this update tomorrow.

jwheare:
How does this sound?

---

There are some cases where a server might not produce a labeled response, or even an ACK. Consider the example of an asynchronous command such as a remote `WHOIS` query that's forwarded to one or more other servers. If the command fails, for instance due to a netsplit, there are several potential outcomes:

* the local server notices that the remote server is unavailable and sends a labeled `ACK` instead of a `WHOIS` response.
* the local server is unable to tell whether the remote server will respond or not and sends neither an `ACK` nor a `WHOIS` response.
* the local server has already begun responding when the netsplit occurs, so the batched response ends early.
* the local server receives a response from the remote server after the netsplit resolves, and sends a `WHOIS` response to the client, potentially without any label.

Clients should handle these cases as they would normally for a server without support for labeled responses.

DarthGandalf:
Yes, that reads easier than what I had written.

PR #406, opened 2020-03-18
webirc: Fix bad handling of IPv6 addresses

RyanSquared:
There could potentially be a case when the IPv6 address of a client
begins with a colon (`:`), such as when testing. This could pose an
issue, as the colon symbol is used to denote that the next argument is
the last argument, and should treat all spaces after it as included in
that argument, which means that no further arguments can be passed. This
creates a problem because there is still another argument, `options`,
which needs to be parsed.

### DIFF
--- a/extensions/webirc.md
+++ b/extensions/webirc.md
@@ -82,5 +82,5 @@ WebIRC allows anyone with a valid password to successfully connect to the IRC se
 
 Because the possibility for hostname spoofing exists, IRC servers MAY attempt to further validate or resolve hostnames and match them to an IP address. It is the responsibility of IRC servers to verify the authenticity of connecting users and perform additional security checks as they see fit. To assist network operators and prevent abuse, IRC servers SHOULD show when a WebIRC connection is in use and SHOULD provide the original host when possible. This behavior is non-normative and implementation defined.
 
-##Errata
+## Errata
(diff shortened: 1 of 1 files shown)


extensions/webirc.md
```diff
@@ -80,4 +80,7 @@ WebIRC requires that the both WebIRC gateway and IRC server be configured to acc
...
 
-Because the possibility for hostname spoofing exists, IRC servers MAY attempt to further validate or resolve hostnames and match them to an IP address. It is the responsibility of IRC servers to verify the authenticity of connecting users and perform additional security checks as they see fit. To assist network operators and prevent abuse, IRC servers SHOULD show when a WebIRC connection is in use and SHOULD provide the original host when possible. This behavior is non-normative and implementation defined.
\ No newline at end of file
+Because the possibility for hostname spoofing exists, IRC servers MAY attempt to further validate or resolve hostnames and match them to an IP address. It is the responsibility of IRC servers to verify the authenticity of connecting users and perform additional security checks as they see fit. To assist network operators and prevent abuse, IRC servers SHOULD show when a WebIRC connection is in use and SHOULD provide the original host when possible. This behavior is non-normative and implementation defined.
+
+##Errata
```
dequis:
```suggestion
## Errata
```

jwheare:
This seems fine to merge. Any further comments?

SadieCat:
LGTM.

PR #407, opened 2020-04-09
[account-notify] Send ACCOUNT to self as well

jwheare:
This clarifies that account-notify should be sent to your own client when changing your own account name (via /nickserv set accountname or otherwise)

I don't see a reason to exclude yourself from this notification or any way this might cause issues for clients.

Note that the SASL 900 RPL_LOGGEDIN numeric might also be sent in this case, but isn't always in practice.

### DIFF
--- a/extensions/account-notify-3.1.md
+++ b/extensions/account-notify-3.1.md
@@ -8,11 +8,12 @@ copyrights:
     email: "nenolod@dereferenced.org"
 ---
 The account-notify client capability allows a client to be notified
-when another client's accountname changes.  This capability MUST be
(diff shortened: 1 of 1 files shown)


jwheare:
Oh I also changed `accountname` to `account name` in the prose sections. accountname isn't a word.

jesopo:
some things do not currently do this and making this part of the spec implies reliability on this being the case. is that ok?

jwheare:
@jesopo yes probably OK, better than the current status quo anyway. Will merge this in a few days if no further comments.

PR #408, opened 2020-04-21
extensions/sasl-3.1: Fix when to send 904

RyanSquared:
It's been pointed out to me that servers often send an 908 numeric to
the client, but then also send an 904. When looking at this behavior, I
think this is so is because servers are generally built by default with
904 no matter what failed and then later, the 908 functionality is
implemented. In order to keep compatibility with clients that may not
understand the 908 numeric but do understand the 904 numeric as a
generic "retry". However, some clients do understand a 908 and use that
as a way to say "downgrade the mechanism and try again". So, they can
begin reconnecting as of the 908 numeric, but then get sent a 904
numeric, which some clients also take as a "try again" and do not
blindly consume the message.

The fix for the client would be to keep a state of which related
numerics are sent to the server, then begin the next attempt once the
904 numeric is received. This would *still work* with this proposed
change.

This also means that any numerics added in the future are compatible
with this specification, as the numerics will need to be sent before the
904 numeric which the client can use to restart the authentication with
any context sent from the other numerics.

---

I also noticed that there was only example of a "failed" case. I have
manually added one. I will be reattempting it sometime in the future to
get an actual response, I was just being ~~really lazy~~ busy writing
the example. I will also go attempt to get further examples of numerics
that are not currently covered by examples.

I do know the branch is called 804. It is a typo.

### DIFF
--- a/extensions/sasl-3.1.md
+++ b/extensions/sasl-3.1.md
@@ -157,6 +157,33 @@ an additional capability.
     S: :jaguar.test 001 jilles :Welcome to the jillestest Internet Relay Chat Network jilles
     (usual welcome messages)
 
+If the client requests a capability that is not supported, the server should
+send both a 908 and a 904 message, allowing the client to reattempt
+authentication with a different method.
+
+    C: CAP LS
+    C: NICK ryan
+    C: user ryan * * :RyanSquared
+    S: NOTICE AUTH :*** Processing connection to irc.test.hashbang.sh
+    S: NOTICE AUTH :*** Looking up your hostname...
+    S: NOTICE AUTH :*** Checking Ident
+    S: NOTICE AUTH :*** No Ident response
+    S: NOTICE AUTH :*** Found your hostname
+    S: :irc.test.hashbang.sh CAP * LS :multi-prefix sasl
+    C: CAP REQ :multi-prefix sasl
+    S: :irc.test.hashbang.sh CAP ryan ACK :multi-prefix sasl
+    C: AUTHENTICATE SCRAM-SHA-1
+    S: :irc.test.hashbang.sh 908 ryan EXTERNAL,PLAIN :are available SASL mechanisms
+    S: :irc.test.hashbang.sh 904 ryan :SASL authentication failed
+    C: AUTHENTICATE PLAIN
+    S: AUTHENTICATE +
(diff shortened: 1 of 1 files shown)


extensions/sasl-3.1.md
```diff
@@ -175,7 +202,7 @@ an additional capability.
 
     :server 903 <nick> :SASL authentication successful
 
-`904` aka `ERR_SASLFAIL` is sent when the SASL authentication fails because of invalid credentials or other errors not explicitly mentioned by other numerics.
+`904` aka `ERR_SASLFAIL` is sent when the SASL authentication fails because of invalid credentials or other errors, even if they're additionally sent by other numerics, and must be sent after all additional numerics.
```
grawity, line 205: I would nitpick that because 908 isn't an ERR numeric, it _doesn't_ technically indicate failure, just provide information... On the other hand, how does this change interact with failures which currently _do_ result in an ERR numeric? Should e.g. ERR_TOOLONG or ERR_ABORTED be also followed by ERR_SASLFAIL (they currently aren't)?

Maybe it's better to say "other **error** numerics"?

RyanSquared: Wow. That is a point that I had COMPLETELY missed. Thank you.

I think that there should then be two RPL and ERR sections to avoid confusion that the RPL_SASLMECHS is not the "ERR" numeric, or at the very least very heavily mention it.

I will have to do some additional research to see if this happens with ERR_TOOLONG and ERR_ABORTED. If that is not the case, then I may revert the change and say that it'll be sent *after* RPL numerics.

jwheare:
First bit and example looks good. Actually had a bug around this in IRCCloud that I worked through with amdj/atheme a few months ago.

RyanSquared:
TODO split RPL_ and ERR_ into sections, rewrite to say that 904 should be sent in addition to RPL_ messages, NOT other ERR_ messages.

extensions/sasl-3.1.md
RyanSquared: Resolved in TODO comment (Apr 22 4:40 AM CST)

jwheare:
@RyanSquared are you able to finish this up? If not we can try get someone else to take it over.

Issue #410, opened 2020-05-05
Communicating client tag deny/allow lists to clients


jwheare:
Following on from https://github.com/ircv3/ircv3-specifications/issues/333#issuecomment-491496090

Since unreal (@syzop) and now inspircd (@SadieCat - https://github.com/inspircd/inspircd/commit/7488b2de62811d7f831e84fb0aeac2345a7f9ee9) both allow or enforce client tag disabling, we need a solution for clients that need to know which client tags can be used, so they can show e.g. a reply ui to users.

The use cases are, in order of how likely they are:

1. All allowed (default)
2. All blocked (catch-all deny)
3. Some allowed (catch-all deny + allow)
4. Some blocked (deny)

A few decisions:

* Where should this information be published? In ISUPPORT or the tag value or via a negotiated command?
* What format should be used?
    * What to call the key?
    * Do we need to include the + when listing tags?
    * Can we use * for a catch-all?
    * Should we allow * as a wildcard, e.g. foo*?
    * Can we use ! for negation?
    * Can we use !* for both?
    * Can we use , or ; for separating rules?
    * Should allow and deny rules be listed in separate keys?
* Which order are deny/allow/defaults applied in?
* How are empty or omitted deny/allow rules interpreted?

Here are some examples of how this might look, using isupport. If no key is present, the default behaviour is always 1) All allowed.

# Single allow key `CLIENTTAGALLOW`
If omitted *or empty* defaults to *, all allowed

1. `CLIENTTAGALLOW=*`
2. `CLIENTTAGALLOW=!*` (negative catch-all)
3. `CLIENTTAGALLOW=!*,foo` (negative catch-all)
4. `CLIENTTAGALLOW=!foo`

# Single deny key `CLIENTTAGDENY`
If omitted, defaults to empty, nothing blocked

1. `CLIENTTAGDENY=`
2. `CLIENTTAGDENY=*`
3. `CLIENTTAGDENY=*,!foo` (double-negative)
4. `CLIENTTAGDENY=foo`

# Separate keys `CLIENTTAGALLOW` `CLIENTTAGDENY`
Same default rules apply as above: `CLIENTTAGALLOW` defaults to *, including when empty and `CLIENTTAGDENY` defaults to empty

Allow applied first:

1. `CLIENTTAGALLOW=* CLIENTTAGDENY=`
2. `CLIENTTAGALLOW= CLIENTTAGDENY=*`
[...]

PR #412, opened 2020-05-06
[message-tags] CLIENTTAGDENY isupport token

jwheare:
This implements the consensus (afaict) from #410

It also makes the allowance for moderating tags less specific to TAGMSG.

### DIFF
--- a/extensions/message-tags.md
+++ b/extensions/message-tags.md
@@ -212,9 +212,9 @@ If a server sends a message with more tag data than the allowed limit, clients M
 
 This specification defines the optional `CLIENTTAGDENY` token for use in `RPL_ISUPPORT` (005) responses.
 
-This token allows servers to communicate to clients that certain client-only tags are blocked and will be silently ignored. This scenario is not a recommended default, but server administrators may wish to enable it for moderation reasons.
+Servers SHOULD use this token to communicate to clients that certain client-only tags are blocked and will be silently ignored. Blocking client-only tags is not a recommended default behaviour, but server administrators might wish to do so for moderation reasons.
 
-This allows clients to selectively remove features from their user interface that rely on any client-only tag that the server has blocked.
+This token allows clients to selectively remove features from their user interface that rely on any client-only tag that the server has blocked.
 
 Note that blocking tags will not necessarily block the message that the tag is attached to.
 
@@ -222,15 +222,15 @@ Clients MAY still send blocked tags to the server.
 
 The `CLIENTTAGDENY` token value is a comma `,` (0x2C) separated list of blocked client-only tags. The client-only prefix (`+`) is omitted when a tag appears in this list.
 
-An asterisk `*` (0x2A) indicates that **all** client-only tags are blocked.
+An asterisk `*` (0x2A) indicates that **all** client-only tags are blocked. When used, this MUST be the first item in the list.
 
 A hyphen `-` (0x2D) indicates that a block is negated, i.e. when certain client-only tags are exempt from a catch-all block.
 
-An empty or missing `CLIENTTAGDENY` indicates the default case that all client-only tags are allowed.
+An empty or missing `CLIENTTAGDENY` matches the default case and indicates that all client-only tags are allowed.
 
 #### `CLIENTTAGDENY` examples
 
-An example where all client-only tags are allowed (ommiting the token entirely is recommended)
+An example where all client-only tags are allowed (ommiting the token entirely is recommended instead)
 
     CLIENTTAGDENY=
 
--- a/extensions/message-tags.md
+++ b/extensions/message-tags.md
@@ -129,6 +129,8 @@ if they are appropriate for more widespread adoption. See [Rules for naming mess
 
 Client-only tags are intended to replace the use of future [CTCP commands][ctcp].
 
+Servers MAY apply moderation to client-only tags using existing or newly specified modes or configuration. See the [RPL_ISUPPORT Tokens](#rpl_isupport-tokens) section for further information.
+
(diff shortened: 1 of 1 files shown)


ProgVal:
LGTM!

Should it be mentioned explicitly that clients can ignore CLIENTTAGDENY entirely, and that their PRIVMSGs shouldn't be dropped entirely if they use a denied tag?

jwheare:
Yeah I can try to make that more explicit.

jwheare:
@ProgVal better?

ProgVal:
Great!

prawnsalad:
As someone that this will directly effect and will impact users using my client, I still don't fully follow the logic for this. I've asked before but was dismissed which means I can't even get to understand the logic. If someone could explain the following to me that would be great.

(I've read the previous threads, most of the points made are already solved problems as far as I can tell. Below in more detail.)

I understand that there is an optional implementation of tag whitelisting already, however the conversation seems to have jumped from some concerns to standardising the practise without any discussion of the impact it actually has or any better alternatives.

1) There were security concerns with allowing arbitrary data in tags. The explanation for the security concerns was "I don't like that at all". What security concern does this spec fix exactly? If it's security related then there should be the specific attack and also the specific mitigation. Otherwise, literally nothing has been accomplished.

2) One of the main points towards c2c tags is that clients can experiment and implement new features without having to wait for servers to notice, then acknowledge, then config or implement the features. If an IRC server blocks client tags then this innovation is entirely blocked.
  This has already been an issue with some IRC servers that didn't implement c2c tags fully despite advertising message-tags capabilities, and some that have a whitelist for tags, causing extra burden for client developers to know how to enable tags on each IRC server software for when users request support. That is assuming the users of the IRC server and/or general network users has gone out of their way to come to complain to the client developer of a broken client and actually stayed around to hear that it was a server blocking the client features. This actively happens already even before this is widespread. Unfortunately most of the users won't even do that, but only assume the client is broken.

3) If a client has to determine which features are possible with their client tags, then mainstream (ie. not technical, not IRC literate) users will become confused with their client if "A (feature) may work but B wont, oh but C does work, but D wont work either", simply because the IRC network they connected to are arbitrarily limiting client tags. A mainstream user will however understand more clearly "this network will only support sending messages, nothing more". Why do we want to confuse IRC users? We should be simplifying the experience.

4) As a user, if I have a client that supports (as an example, it could be anything a client wants to do) e2e encryption that uses some named c2c tags, why should a network admin refuse to allow me to use that feature of my client? It is using supported features of the IRC server and protocol, not using any abused tags, but I can't use it because some network admin decided it was not in my best interest.

5) There was an issue raised regarding flooding or abuse of arbitrary tags towards a channel. How does this differ to normal abuse? If a user notices that they are getting abused by a c2c tag then this means that they are able to see the tag value or the side effect that the tag is having on their client. In this case, reporting the user to channel or network staff should be sufficient - exactly the same as it currently is for floods or other types of abuse towards a channel. If the abused user does not feel that this is sufficient then they also have the option to ignore the source of abuse or entirely disable c2c tags within their client without impacting everybody else on the network. So far there hasn't been any examples of abuse that this spec is trying to prevent specific to c2c tags, that I’ve seen. Any examples for this?

As it stands, if a network admin decides to limit a clients functionality that makes use of c2c tags then the obvious option for me as a client dev is to display a warning that client functionality has been restricted by the network and that only basic messaging is available. Listing specific disabled/enabled features is mental overhead and will be overlooked by users just making the client look broken in their eyes. Therefore a "basic messaging" or a "fully featured" client is what makes more sense here.

ProgVal:
> causing extra burden for client developers to know how to enable tags on each IRC server software for when users request support

This is what this PR aims to fix.

> In this case, reporting the user to channel or network staff should be sufficient - exactly the same as it currently is for floods or other types of abuse towards a channel

Only if the staff has a client that can see these tags. They may not even be aware their client doesn't.

jwheare:
To avoid a long heated back and forth over this comment I’ll take this to IRC privately and we’ll see if we can resolve this.

jwheare:
I believe this has broad consensus. There are obviously still some unshared viewpoints around whether client tags should be blocked at all, but I think given that is something servers will be doing anyway, for various reasons, we should go ahead with this errata.

I've chatted to the people involved and there are disagreements concerning the *reasons* server authors/operators have given, but I feel they are still valid reasons. And I don't think anyone is outright vetoing this proposal.

With all that in mind, I will be merging this by the middle of the week.

jwheare:
I am actually going to delay this until I've tested an implementation. If anyone has implementations on e.g. testnets, please post here.

jwheare:
Thoughts on requiring that * appear before any negated tags? e.g. this is allowed:

`CLIENTTAGDENY=*,-typing`

This is not:

`CLIENTTAGDENY=-typing,*`

Means clients can iterate the list in order and set a "denied" flag on or off if there's a match.

Otherwise we need to specify that arguments are interpreted in a certain order.

This is only ambiguous for the deny-all-except-some case.

k4bek4be:
I have an implementation for this avaliable for testing on a public server (in fact for 8 days now). It is currently accessible via irc://aga.k4be.pl:6667/

jwheare:
I have implemented this in the IRCCloud web client to toggle the reply, react, edit and delete UI, and typing notifications when the respective tags are being blocked.

k4bek4be:
It is now merged into unrealircd-5.0.5-dev.
https://github.com/unrealircd/unrealircd/commit/0aa5fb6e7ca8facb02382dab3fa2e42552961e6e

jwheare:
This is probably OK to merge now. Will do so tomorrow if no further comments.

PR #413, opened 2020-05-14
[multiline] Clarify blank line and line feed handling

jwheare:
This addresses issues raised in https://github.com/oragono/oragono/issues/1005#issuecomment-628000227

* Clarify line feed joiner takes up one byte
* Servers MUST NOT reject blank lines (they can be used to create extra new lines)
* Clients MUST NOT send concat only blank lines or blank-only messages
* Servers MUST NOT send blank lines in legacy fallback (clients may well handle these but they're normally not allowed. An alternative may be to turn them into lines containing a single space, but not sure about that)

I don't think the suggestion to disallow a leading blank line is correct.

### DIFF
--- a/extensions/multiline.md
+++ b/extensions/multiline.md
@@ -71,9 +71,9 @@ When receiving a well-formed mulitiline message batch, implementations MUST coll
 * Servers: delivering the batch to the intended recipients
 * Clients: displaying the batched message to the user
 
-Messages in a multiline batch MUST be joined with a single line feed (`\n`) byte unless the `draft/multiline-concat` message tag is sent, in which case the message is directly joined with the previous message with no separation.
+The combined message value of a multiline batch is defined as the concatenation of the messages from each individual line within the batch. Line messages are joined by a single line feed (`\n`) byte unless the `draft/multiline-concat` message tag is sent, in which case that line's message is directly joined with the previous line's message with no separation.
 
-The line feed joiner uses one byte from the `max-bytes` limit. No line feed is appended to the final line of a batch.
+Each line feed used to join line messages contributes one byte towards the `max-bytes` limit. No line feed is appended to the final line message of a batch.
 
 Servers MUST NOT reject blank lines.
 
--- a/extensions/multiline.md
+++ b/extensions/multiline.md
@@ -87,6 +87,8 @@ When delivering multiline batches to clients that have not negotiated the multil
 
 Servers MUST NOT send blank lines to clients that have not negotiated the multiline capability.
 
+Servers SHOULD maintain the line composition sent by the client instead of combining to a normalised form before re-splitting. This ensures that steps taken to [split long lines](#splitting-long-lines) appropriately are preserved.
+
 Any tags that would have been added to the batch, e.g. message IDs, account tags etc MUST be included on the first message line to be sent. Tags MAY also be included on subsequent lines where it makes sense to do so.
 
 ### Errors
--- a/extensions/multiline.md
+++ b/extensions/multiline.md
@@ -75,9 +75,10 @@ The combined message value of a multiline batch is defined as the concatenation 
 
 Each line feed used to join line messages contributes one byte towards the `max-bytes` limit. No line feed is appended to the final line message of a batch.
 
-Servers MUST NOT reject blank lines.
+Servers MUST NOT reject blank lines other than in the following cases:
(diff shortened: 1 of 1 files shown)


slingamn:
I agree with the substantive changes here. (In particular: my legacy users were displeased with the idea of blank lines in chat; I think leaving them out is preferable to sending a space.) Allowing a leading blank line (for non-legacy clients) is fine too.

This language is weird to me:

>Messages in a multiline batch MUST be joined with a single line feed (`\n`) byte unless the `draft/multiline-concat` message tag is sent, in which case the message is directly joined with the previous message with no separation.

I have the same questions about what it means to say "MUST be joined" as I did with the previous "MUST be concatenated" language --- who is doing this joining or concatenation? If anyone, it's the client, but what does it mean to say that this is a MUST? It seems like the intent here is to specify the "value" of a multiline message, in a sense that the sending and the receiving clients agree upon, but I don't think the current language communicates that clearly.

All the other language looks good.

slingamn:
Suggestion:

>The value of a multiline batch (i.e., the single underlying message that it represents) is the concatenation of all the individual messages, with a single line feed (`\n`) byte interposed between each pair of consecutive messages, except in the case where the second message in the pair has the `draft/multiline-concat` message tag, in which case no separator is interposed. Note that no line feed is appended to the final message of the batch. The `max-bytes` limit is enforced against the length of the batch's value.

slingamn:
This never occurred to me until now, but `max-lines` refers to messages within the batch, not line feeds, right? You get billed for a line against `max-lines` even when sending the concat tag?

jwheare:
> This never occurred to me until now, but `max-lines` refers to messages within the batch, not line feeds, right? You get billed for a line against `max-lines` even when sending the concat tag?

Yes. max-lines is about taking into account all the prefix/cmd/target boilerplate that accompanies each protocol line.

jwheare:
How's that? No more confusing MUST and a bit more explicit.

slingamn:
Awesome.

slingamn:
Oh, for what it's worth, the reason I intend to strictly enforce that clients MUST NOT send blank lines with the concat tag is that I'm disabling fakelag during the sending of the multiline. So if I normalize out these lines, and max-lines isn't configured, a client can send an arbitrary number of unthrottled empty concat lines. I could work around this but it seems cleaner just to reject.

slingamn:
The leftover item of business here is to decide what kind of re-combining and re-splitting the server is allowed to do. On my reading of the current language:

>When delivering multiline batches to clients that have not negotiated the multiline capability, servers MUST deliver the component messages without using a multiline BATCH.

the server is not allowed to do any of this and must retransmit the original component messages. I think this is correct, given that one of the goals of this proposal was to give the client more control over word wrap behavior. (Besides optimization, what purpose is served by allowing the server to recombine or resplit?)

Proposed language that makes this more explicit:

>When delivering multiline batches to clients that have not negotiated the multiline capability, servers MUST deliver the individual component messages without using a multiline BATCH, with the exception of blank lines, which MUST NOT be sent.

jwheare:
@slingamn Yeah, fair enough that server SHOULD probably keep the original line breaking sent by the client. I don't follow how the language change you suggest relates to this at all though.

slingamn:
I understood "individual component messages" to refer (both before and after the proposed language change) to the original line breaks sent by the client. However, the original language appears to require that blank lines be sent to legacy clients. The language change therefore indicates that blank lines MUST NOT be sent; they are the one exception to the requirement that the server retransmit the original component messages.

jwheare:
I think it should say what we want it to say, clearly :) I'll have a go at some better wording.

jwheare:
@slingamn that sound ok?

slingamn:
Looks good to me, thanks!

jwheare:
This should now be good to merge. Will aim to do so on Friday, barring any further issues,

jwheare:
Cleared up when servers can and can't reject blank lines with examples. Hope it's better.

I made the example response to an entirely blank batch a `FAIL BATCH MULTILINE_INVALID`. The single line equivalent would be to use a 412 ERR_NOTEXTTOSEND numeric. I think the FAIL code is probably better and more explicit but I could be swayed either way.

jwheare:
Alright, merging this tomorrow, barring any further comments.

PR #414, opened 2020-05-21
[typing] Fix trivial typo

delthas:
`paused` is a value for +typing, like `active` and `done`. The value is *not* `typing-paused`.


DanielOaks:
Thanks much :D

PR #416, opened 2020-06-05
draft specification for the persist tag

slingamn:
This is "option 3" from https://github.com/ircv3/ircv3-ideas/issues/68.

### DIFF
--- a/client-tags/persist.md
+++ b/client-tags/persist.md
@@ -37,7 +37,9 @@ These messages have some salient common characteristics:
 
 It is desirable to be able to advise the server not to store them and replay them to offline clients --- either because they would consume server resources unnecessarily, or because replaying them would actually risk degrading the client user experience.
 
-In contrast, some extensions involve TAGMSG that have more enduring relevance, in particular, [reactions](./react). The persist tag identifies these messages, with the intent that servers will store only those TAGMSG that carry the persist tag.
+In contrast, some extensions involve TAGMSG that have more enduring relevance, in particular, [reactions](./react). If a server offers history storage and replay functionality, then it should store and replay these messages.
+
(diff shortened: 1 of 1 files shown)


client-tags/persist.md
```diff
@@ -0,0 +1,80 @@
...
+
+### Format
+
+The persist tag is sent by a client with the client-only prefix `+`. It has no value:
+
+    +draft/persist
+
+## Server implementation considerations
+
+Servers SHOULD NOT store TAGMSG that do not contain the `+draft/persist` tag. If a server implements history storage and replay, it SHOULD store TAGMSG that contain the `+draft/persist` tag. Servers SHOULD ignore the presence of `+draft/persist` on all other message types.
```
ProgVal, line 58: I don't think these should be requirement on the servers, but hints.

client-tags/persist.md
```diff
@@ -0,0 +1,80 @@
...
+### Dependencies
+
+Clients wishing to use this tag MUST negotiate the [`message-tags`](../extensions/message-tags.html) capability with the server.
+
+### Format
+
+The persist tag is sent by a client with the client-only prefix `+`. It has no value:
+
+    +draft/persist
+
```
ProgVal, line 55: and servers should ignore the value if any (to give some wiggle room for future extensions)

client-tags/persist.md
slingamn: As in, something weaker than a SHOULD? I'm not sure what alternative language would make sense.

ProgVal: `should` instead of `SHOULD` would be fine

SadieCat:
As I mentioned previously this has a poor fallback for existing client implementations (i.e. nothing being stored) and I can't really see it being useful for server implementations.

PR #423, opened 2020-07-30
Clarify chghost

jwheare:
This PR replaces and builds on #364, incorporating the latest suggestions around showing unidented tildes, and also moving the fallback description to its own subheading.

### DIFF
--- a/extensions/chghost-3.2.md
+++ b/extensions/chghost-3.2.md
@@ -26,11 +26,23 @@ message to clients who have enabled the `chghost` capability and either share
 the same channel as the target client or have the client on a `MONITOR` list.
 Servers SHOULD additionally send the `CHGHOST` message to the client whose
 user and/or host has changed if the client supports the `chghost` capability.
+The server MUST defer sending CHGHOST messages to the client about successfully
+changing it's own user or host (for example, if using SASL and a virtual host
+is set) until both a successful `NICK` command and a successful `USER` command
+have been received.
 
-When the capability is not enabled for clients who share the same channel,
-servers should fall back to having the client `QUIT` the server, rejoin all
-channels, and re-establish the channel and user modes the client had before, as
-though the client had reconnected.
+When the capability is not enabled for other clients who share channels with or
+monitor the changed client, servers SHOULD send messages to simulate the client
+reconnecting. This allows clients to keep their user state up to date. For
+shared channels, the simulated events SHOULD include appropriate QUIT, JOIN and
+MODE commands, to restore membership and user channel modes. For monitored
+clients, the events SHOULD include appropriate RPL_MONOFFLINE and RPL_MONONLINE
+numerics.
+
+The server MUST send a `CHGHOST` message to a client, but MUST defer doing so
+until both a successful `NICK` command a successful `USER` command are received
+by the server. The server CAN choose to defer it until after registration is
+completed.
 
 ## The `CHGHOST` message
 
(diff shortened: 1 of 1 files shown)


jwheare:
I still need to do a fresh reading of these changes. Review comments from anyone else welcome.

extensions/chghost-3.2.md
```diff
@@ -5,59 +5,87 @@ copyrights:
...
+When a client changes their user or host, servers MUST send the `CHGHOST`
+message to clients who have enabled the `chghost` capability and either share
+the same channel as the target client or have the client on a `MONITOR` list.
+Servers SHOULD additionally send the `CHGHOST` message to the client whose
+user or host has changed if the client supports the `chghost` capability.
 
-The CHGHOST message is one of the following:
+When the server sends a `CHGHOST` message to a client, it MUST defer doing so
+until both successful `NICK` and `USER` commands have been received by the
+server.  The server MAY choose to defer it until after registration is
```
edk0, line 32: I find this form (when you do X, you must defer doing X) pretty confusing.

extensions/chghost-3.2.md
```diff
@@ -5,59 +5,87 @@ copyrights:
...
 
-1) Enable the chghost capability at capability negotiation time during the
-   login handshake.
+When the capability is not enabled for other clients who share channels with or
+monitor the changed client, servers SHOULD send fallback messages to simulate the client
+reconnecting. This allows clients to keep their user state up to date. For
+shared channels, the simulated events SHOULD include appropriate `QUIT`, `JOIN`
+and `MODE` commands, to restore membership and user channel modes. For monitored
+clients, the events SHOULD include appropriate `RPL_MONOFFLINE` and
+`RPL_MONONLINE` numerics.
```
edk0, line 62: This seems questionable since it'll pointlessly highlight people.

extensions/chghost-3.2.md
```diff
@@ -5,59 +5,87 @@ copyrights:
...
+    :tim!~toolshed@backyard CHGHOST ~b ckyard
 
-In this example, `tim!b@ckyard` gets their username changed to `~toolshed` and
+In this example, `tim!b@ckyard` gets their username changed to `toolshed` and
 their hostname changed to `backyard`:
 
-    :tim!b@ckyard CHGHOST ~toolshed backyard
+    :tim!b@ckyard CHGHOST toolshed backyard
 
 ## Errata
```
edk0, line 81: The requirement for monitored clients seems to be new, but isn't mentioned here.

extensions/chghost-3.2.md
jwheare: Yeah I want to rewrite a bunch of this.

extensions/chghost-3.2.md
jwheare: Well, highlighting depends on the client. This allows clients that are monitoring users but aren't in shared channels to have an up to date mask, which is a requirement for various things.

jwheare: Can't find much discussion of it other than these two comments:

https://github.com/ircv3/ircv3-specifications/pull/364#issuecomment-464385022
https://github.com/ircv3/ircv3-specifications/pull/364#pullrequestreview-204541454

Not sure which implementations if any do this. As a SHOULD it's not a breaking change, but it should definitely be errataed if kept, as you noted elsewhere.

jwheare: From a glance at inspircd, unreal, charybdis, hybrid, none of these servers send monitor numerics in the fallback case, nor send the cap-enabled message to monitoring clients.

Given we aren't routinely extending MONITOR to support other forms of change notification (e.g. SETNAME) there's probably an argument for dropping it from this.

There's probably also an argument for extending MONITOR more thoroughly, but given lack of implementation, I reckon we can defer that.

Thoughts @SadieCat @syzop @slingamn?

jwheare:
I rewrote the intro and dropped the stuff about waiting till after NICK/USER. I think it's an edge case and don't think it's worth documenting in the spec. There are no known server implementations that still have an issue with this. Ref: https://github.com/ircv3/ircv3-specifications/pull/364#issuecomment-499463530

jwheare:
Will merge next week if no further comments.

PR #424, opened 2020-08-09
Add +draft/language client tag.

ProgVal:
A small specification for once, hopefully without too many implications. :)

Some use cases I can think of:

* Using specific fonts for different languages, like browsers do. (Useful when the default font is partial, or [when the same unicode codepoint has variants depending on language](https://chinese.stackexchange.com/questions/792/characters-which-have-several-different-shapes))
* Helping text to speech software
* Bots which provide automatic translation don't need to guess the language (which usually involves an HTTP API call); Limnoria has a plugin doing that.
* You can add a button next to languages in a foreign language, like "Translate with Bing" / "Translate with Google" on some microblogging websites
* Hiding/folding messages in some languages if the user so desires, eg. for multilingual channels

### DIFF
--- a/client-tags/language.md
+++ b/client-tags/language.md
@@ -35,11 +35,11 @@ Clients wishing to use this tag MUST negotiate the
 ### Format
 
 The language tag is sent by a client with the client-only prefix `+` and
-its value must be a a ISO 639-3 code.
+its value must be a [BCP 47 language tag](https://tools.ietf.org/html/bcp47).
 
-    +draft/language=<language-code>
+    +draft/language=<language-tag>
 
-In case the value is not a valid language code, clients SHOULD ignore it.
+In case the value is not a valid language tag, clients SHOULD ignore it.
 
 ### Guessing languages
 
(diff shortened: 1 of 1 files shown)


client-tags/language.md
```diff
@@ -0,0 +1,87 @@
...
+    C: @+draft/language=fra PRIVMSG #channel :Je parle aussi français.
+
+### Multiline message
+
+NOTE: In this examples, `<SPACE>` indicates a space character which would
+otherwise not be clearly visible.
+
+    C: BATCH +123 draft/multiline #channel
+    C: @batch=123 +draft/language=eng PRIVMSG #channel :The French word for "cat" is<SPACE>
+    C: @batch=123;draft/multiline-concat;+draft/language=fra +draft/language=fra PRIVMSG #channel :"chat"
```
DarthGandalf, line 86: Why language is repeated, and second one as a separate parameter?

client-tags/language.md
```diff
@@ -0,0 +1,87 @@
...
+
+### Dependencies
+
+Clients wishing to use this tag MUST negotiate the
+[`message-tags`](../extensions/message-tags.html) capability with the server.
+
+### Format
+
+The language tag is sent by a client with the client-only prefix `+` and
+its value must be a a ISO 639-3 code.
```
DarthGandalf, line 38: No difference between e.g. British and American English?

client-tags/language.md
ProgVal: fixed

client-tags/language.md
ProgVal: No. Even variants of arabics are considered the same macrolanguage.

prawnsalad: There's https://tools.ietf.org/html/bcp47 which is the commonly used `en-US` `en-GB` `en` `fr` etc

ProgVal:
There doesn't seem to be much interest in this, so I'm closing the proposal. Ping me if you want to reopen it.

Issue #405, opened 2020-03-09
Can a server reply with several CAP LS in 302-less implementation ?


lel-amri:
From my understanding of the spec, a server that doesn't supports 302 can reply with several CAP LS to a CAP LS request.

Let's say a client that does supports 302 sends CAP LS 302 to a server that doesn't supports CAP LS 302, and that the said server has so much capabilities that it needs to reply with two CAP LS. In this case, the exchange would be:

    Client: CAP LS 302
    Server: CAP * LS :cap1 cap2 cap3
    Server: CAP * LS :cap4 cap5

From my understanding of the spec, I can raise these statements:

1. A server isn't required to send multiline CAP LS replies
2. A client can't know whether the server does supports CAP LS 302 or doesn't
3. It is unspecified whether upon receiving CAP LS (non-302, and 302 altogether) the client should _replace_ or _make and addition to_ its set of known server's supported capabilities
4. A client isn't required to wait for several CAP LS replies

Given these statements, for a client making a CAP LS 302 request to a server that doesn't supports 302, all these behavior are valid:

1. The client gets `CAP * LS :cap1 cap2 cap3`, replaces its internal set so it becomes `(cap1, cap2, cap3)`, then drops `CAP * LS :cap4 cap5` because it appear to be unsolicited
2. The client gets `CAP * LS :cap1 cap2 cap3`, replaces its internal set so it becomes `(cap1, cap2, cap3)`, then upon receiving `CAP * LS :cap4 cap5` replaces its internal set so it becomes `(cap4, cap5)`
3. The client gets `CAP * LS :cap1 cap2 cap3`, updates (or initialize, if it's a fresh connection) its internal set so it becomes `(cap1, cap2, cap3)`, then upon receiving `CAP * LS :cap4 cap5` updates its internal set so it becomes `(cap1, cap2, cap3, cap4, cap5)`

[...]

PR #429, opened 2020-11-02
Clarify that multiline LS and LIST replies must only be used for CAP 302

jwheare:
Attempt to address #405

This version doesn't allow multiline LS replies to be sent to 301 clients. An alternative version that does might look like:

> Servers MAY send multiple lines in response to `CAP LS` and `CAP LIST` but SHOULD NOT send them to clients that do not support CAP version `302`. Clients that support CAP version `302` MUST and other clients SHOULD handle the continuation format described as follows:

But this seems odd. Why should 301 clients support this if servers shouldn't be sending it?

### DIFF
--- a/core/capability-negotiation.md
+++ b/core/capability-negotiation.md
@@ -204,7 +204,7 @@ for the future life of the connection. Clients that do not send any version numb
 If a client has not indicated support for `CAP LS 302` features, the server MUST NOT send
 these new features to the client.
 
-When CAP Version `302` is enabled, the client also implicitly indicates support for the
+When CAP version `302` is enabled, the client also implicitly indicates support for the
 `cap-notify` capability listed below, and support for the relevant `NEW` and `DEL`
 subcommands.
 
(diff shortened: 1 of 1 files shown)


grawity:
I have to guess that clients "SHOULD" understand them because they don't know whether they're talking to a 301 server or to a server implementing [the draft spec][1] (which _did_ have multiline responses).

Of course the server also doesn't know whether it's replying to a 301 client or a draft-mitchell client, so I agree it should be conservative and only send a single line.

[1]: http://web.archive.org/web/20140728040332/http://www.leeh.co.uk/draft-mitchell-irc-capabilities-02.html

jwheare:
Yeah it's a case of "gee it would be nice if every client did this then we could go back to the mitchell spec for all". But I suspect the non-mitchell behaviour has been published for long enough now that that ship has sailed. So I think I'd rather just have a current spec that doesn't raise more questions than it answers.

lel-amri:
@jwheare I'd keep the examples for single-line CAP LS/LIST replies while modifying them so they show only one-lined replies.

jwheare:
Not sure that's necessary. Ordinary LS/LIST examples are already shown elsewhere

jwheare:
Dunno why "Version" was being capitalised all over the place, but I fixed that and rewrote the start of the multiline bit to match the language of the cap values bit.

PR #432, opened 2021-01-24
spec for UTF8ONLY

slingamn:
The ISUPPORT token `UTF8ONLY` indicates "this server only accepts or relays UTF-8 content".

This is implemented (configurable, but a recommended default) in Oragono's master and the new [v2.5.0-rc1](https://github.com/oragono/oragono/releases/tag/v2.5.0-rc1) release candidate.

### DIFF
--- a/extensions/utf8-only.md
+++ b/extensions/utf8-only.md
@@ -17,7 +17,7 @@ copyrights:
 IRC predates the Unicode standard. Consequently, although UTF-8 has been widely adopted on IRC, clients cannot assume that all IRC data is UTF-8. This specification defines a way for servers to advertise that they only allow UTF-8 on their network, letting clients change their processing of outgoing and incoming messages accordingly.
 
 ## The `UTF8ONLY` ISUPPORT token
-This specification introduces a new token `UTF8ONLY` that servers can include in their [ISUPPORT](https://modern.ircdocs.horse/#feature-advertisement) (`005`) output. Servers publishing this token MUST NOT relay content (such as `PRIVMSG` or `NOTICE` message data, channel topics, or realnames) containing non-UTF-8 data to clients. Clients implementing this specification MUST NOT send non-UTF-8 data to the server once they have seen this token. Server handling of such messages is implementation-defined, and MAY involve sending the `FAIL` code described below.
+This specification introduces a new token `UTF8ONLY` that servers can include in their [ISUPPORT](https://modern.ircdocs.horse/#feature-advertisement) (`005`) output. Servers publishing this token MUST NOT relay content (such as `PRIVMSG` or `NOTICE` message data, channel topics, or realnames) containing non-UTF-8 data to clients. Clients implementing this specification MUST NOT send non-UTF-8 data to the server once they have seen this token. Server handling of such messages is implementation-defined; for example, they MAY send the `FAIL` code described below, disconnect the client with an `ERROR` message, or respond in some other way.
 
 If a client implementing this specification sees this token, they MUST set their outgoing encoding to UTF-8 without requiring any user intervention. This allows clients to work transparently on networks that only allow UTF-8 traffic.
 
@@ -34,4 +34,10 @@ Server: FAIL PRIVMSG INVALID_UTF8 :Message rejected, your IRC software MUST use 
 ```
 Client: USER u s e :<non-utf8 realname>
 Server: FAIL USER INVALID_UTF8 :Message rejected, your IRC software MUST use UTF-8 encoding on this network
+```
+
+```
+Client: PRIVMSG #ircv3 :<non-utf-8 message>
+Server: ERROR :Your IRC software MUST use UTF-8 encoding on this network
+... server disconnects the client ...
 ```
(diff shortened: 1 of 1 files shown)


slingamn:
Given that it's impossible to recover from this error in software (a specification-compliant client will never send a message that produces the error), the only purpose of the error is to produce a user-visible message. Given this, it's been suggested that the recommended error be an RFC1459 code (probably `400 ERR_UNKNOWNERROR`) rather than a `FAIL`. (Even if the recommendation is changed, it seems better to leave it as implementation-defined.)

slingamn:
Another suggestion: `ERROR` with a descriptive message, then disconnecting the client, might be the best way to get the user's attention.

DanielOaks:
Let's keep the more specific `FAIL` code defined for any implementations that do want to use it, and if implementations want to fail with an `ERROR` or respond in some other way then the spec's flexible enough to allow that. We can definitely add an example showing the server just straight-up sending the `ERROR` and disconnecting the client too, make it explicit that servers can respond in whatever way they see fit.

SadieCat:
Given this, it's been suggested that the recommended error be an RFC1459 code (probably 400 ERR_UNKNOWNERROR) rather than a FAIL
Is there any reason to not use the already existing ERR_CANTSENDTOUSER (404) and ERR_CANTSENDTOUSER (531) numerics?

slingamn:
It's not evident to me that `404 ERR_CANNOTSENDTOCHAN` and/or `531 ERR_CANNOTSENDTOUSER` cover all the cases where this error might be produced, e.g., `SETNAME`, `USER`, and `TOPIC`. I'd also prefer to recommend a single response that is applicable to every command (choosing between two possible responses would be vaguely burdensome in my implementation).

But these are just arguments against making those numerics the official recommendation; the spec would certainly allow sending them.

slingamn:
@SadieCat notes that a recommendation to `ERROR` and disconnect would facilitate DoS attacks against legacy bots / client scripts.

slingamn:
@SadieCat I believe you had a question or concern here?

extensions/utf8-only.md
```diff
@@ -0,0 +1,43 @@
...
+This specification introduces a new token `UTF8ONLY` that servers can include in their [ISUPPORT](https://modern.ircdocs.horse/#feature-advertisement) (`005`) output. Servers publishing this token MUST NOT relay content (such as `PRIVMSG` or `NOTICE` message data, channel topics, or realnames) containing non-UTF-8 data to clients. Clients implementing this specification MUST NOT send non-UTF-8 data to the server once they have seen this token. Server handling of such messages is implementation-defined; for example, they MAY send the `FAIL` code described below, disconnect the client with an `ERROR` message, or respond in some other way.
+
+If a client implementing this specification sees this token, they MUST set their outgoing encoding to UTF-8 without requiring any user intervention. This allows clients to work transparently on networks that only allow UTF-8 traffic.
+
+## The `INVALID_UTF8` `FAIL` code
+This is a code that can be used with the `FAIL` command, as defined by the [standard replies](https://ircv3.net/specs/extensions/standard-replies) specification. This code indicates to the client that their message was dropped or modified because it contained non-UTF-8 bytes.
```
DanielOaks, line 25: This code should be used with `FAIL` when the client's message is dropped, and with `WARN` when their message is modified but still gets processed/sent/etc.

extensions/utf8-only.md
```diff
@@ -46,3 +46,6 @@ Server: ERROR :Your IRC software MUST use UTF-8 encoding on this network
 Client: PRIVMSG #ircv3 :<non-utf-8 message>
 Server: WARN PRIVMSG INVALID_UTF8 :Your message was not correctly encoded as UTF-8 and had to be modified
 ```
+
+## Implementation considerations
+Implementations MUST ensure that if they truncate messages to meet a length limit, they do not do so in the middle of a UTF-8 codepoint.
```
DanielOaks, line 51: This should be 'must' rather than 'MUST' - our implementation consideration sections are non-normative and so the `MUST` and `SHOULD` and official-uppercasey-terms can't be used in 'em (don't you love some spec pedantry ;D ).

slingamn:
@justjanne suggests that we should include non-normative language about how this proposal interacts with text-based websocket messages.

DanielOaks:
If we can somehow come up with a one-sentence way to mention that this _interacts_ with that... cool? From our side the implications seem fairly obvious though, not sure it's worth mentioning, and given the more volatile of the two specs is most certainly websockets I'd leave it for that spec to have the mention. This is small and happy and fine, let's leave it alone for anything but critical fixes.

DanielOaks:
We can merge this and wait for implementations. It's very, overwhelmingly clear that this isn't useful for the existing large IRC nets that want to retain their different encoding stuff, but for what it's designed for it works and it works well.

Spec is 100% fine as-is, whether it lives on the IRCv3 website or on the Oragono website.

slingamn:
Sounds good to me. I think this spec doesn't need to mention websockets, and mentioning this spec in the websockets spec would be unnecessarily controversial.

PR #434, opened 2021-01-29
Add the bouncer-networks extension specification

emersion:
This is based on a previous proposal by Darren Whitlen [1], but with
its scope reduced to networks (buffers are left out). Discussion has
happened in an ircv3-ideas issue [2].

[1]: https://gist.github.com/prawnsalad/d864822d9bb8116626a4370bee6e6fc6
[2]: https://github.com/ircv3/ircv3-ideas/issues/39

cc @prawnsalad

### DIFF
--- a/extensions/bouncer-networks.md
+++ b/extensions/bouncer-networks.md
@@ -171,6 +171,10 @@ If a client sends an `ADDNETWORK` subcommand without a mandatory attribute:
 
     FAIL BOUNCER NEED_ATTRIBUTE ADDNETWORK <attribute> :Missing required attribute
 
+If a client sends an unknown subcommand:
+
+    FAIL BOUNCER UNKNOWN_COMMAND <subcommand> :Unknown subcommand
+
 TODO: more errors
 
 ### Standard network attributes
--- a/extensions/bouncer-networks.md
+++ b/extensions/bouncer-networks.md
@@ -139,41 +139,67 @@ Errors are returned using the standard replies syntax. The general syntax is:
 
     FAIL BOUNCER <code> <subcommand> [context...] <description>
 
-If a client sends a `BIND` subcommand before authentication:
+If a client sends an unknown subcommand, the server MUST reply with:
+
+    FAIL BOUNCER UNKNOWN_COMMAND <subcommand> :Unknown subcommand
+
+#### `ACCOUNT_REQUIRED` error
+
+If a client sends a `BIND` subcommand before authentication, the server MAY
+reply with:
 
     FAIL BOUNCER ACCOUNT_REQUIRED BIND :Authentication required
 
-If a client sends a `BIND` subcommand after registration:
+#### `REGISTRATION_IS_COMPLETED` error
+
+If a client sends a `BIND` subcommand after registration, the server MAY reply
+with:
 
     FAIL BOUNCER REGISTRATION_IS_COMPLETED BIND :Cannot bind to a network after registration
 
-If a client sends a `BIND` subcommand with an invalid network ID:
+#### `INVALID_NETID` error
 
-    FAIL BOUNCER INVALID_NETID BIND <netid> :Network not found
+If a client sends a subcommand with an invalid network ID, the server MUST
+reply with:
 
-If a client sends an `ADDNETWORK` subcommand with an invalid attribute:
+    FAIL BOUNCER INVALID_NETID <subcommand> <netid> :Network not found
 
-    FAIL BOUNCER INVALID_ATTRIBUTE ADDNETWORK * <attribute> :Invalid attribute value
+#### `INVALID_ATTRIBUTE` error
 
-If a client sends a `CHANGENETWORK` subcommand with an invalid attribute:
+If a client sends an `ADDNETWORK` or a `CHANGENETWORK` subcommand with an
+invalid attribute, the server MUST reply with:
 
-    FAIL BOUNCER INVALID_ATTRIBUTE CHANGENETWORK <netid> <attribute> :Invalid attribute value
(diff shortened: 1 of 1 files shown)


extensions/bouncer-networks.md
```diff
@@ -0,0 +1,122 @@
...
+
+    BOUNCER LISTNETWORKS
+
+The server replies with:
+
+    BOUNCER LISTNETWORKS <netid> <attributes>
+
+And then ends the list (even if empty) with:
+
+    BOUNCER LISTNETWORKS RPL_OK
```
emersion, line 74: Not a fan if this `RPL_OK` token. Something like `*` would probably be better.

jwheare:
It's not immediately clear from this issue why the buffers bits were left out of this, care to elaborate @emersion?

@prawnsalad any thoughts? Are you keen to standardise the buffers portion of your original proposal as well, or could that be part of a later or vendor-specific extension?

It's a question of general applicability I suppose. One issue is probably that servers don't really have the same concept of "buffers" as clients, but bouncers often do. See https://wiki.znc.in/Query_buffers

emersion:
My bouncer doesn't have a concept of "buffers", which is why these are left out. I guess the reason why ZNC has them is because they store history in memory and need to know when it's safe to release the memory. Since soju stores history on disk, there's no need for this.

jwheare:
Aside from memory management (which seems a red herring to me) I think this disconnect directly relates to the same questions coming up in chathistory around history for private messages. Servers don't have a concept of them, but I think there's a stronger case for it with bouncers, as they operate as both client and server at the same time and need to model more of what the client does.

IRCCloud has a pm buffers concept, as does Quassel, and not for memory management reasons.

emersion:
> IRCCloud has a pm buffers concept, as does Quassel, and not for memory management reasons.

OK, what are they used for then?

jwheare:
They map to the way clients work, messages are grouped within buffers, and requested by the client as an ordered list within the concept of the buffer, so the bouncer has to be able to send them that way, With a shared buffer for all PMs, you get the issue that spam in one PM pushes out legitimate messages in another.

emersion:
Hm, this still looks like this is still tied to history retrieval? Why is a concept of buffer necessary for this? soju doesn't have a concept of buffer, but still has per-PM history buckets and doesn't run into the spam issue you're describing.

In any case, as said earlier, adding another extension on top of this one for bouncer buffers should be pretty trivial. I don't think we should force bouncers to implement a concept of buffers because of this extension.

jwheare:
> has per-PM history buckets

That's basically all I mean by a buffer. I don't mean a literal memory buffer.

> I don't think we should force bouncers to implement a concept of buffers because of this extension.

That's fair, I just wonder if the resistance to treating PM buckets as a first class concept will continue to lead to issues as servers begin to implement bouncer type features (history, notifications, etc)

I also just think this point needed to made explicit in this issue.

slingamn:
>Servers don't have a concept of them, but I think there's a stronger case for it with bouncers, as they operate as both client and server at the same time and need to model more of what the client does. [...] With a shared buffer for all PMs, you get the issue that spam in one PM pushes out legitimate messages in another.

This is subtle, I think. CHATHISTORY mandates that servers be able to retrieve PMs with a specific correspondent (nick targets), as well as with all correspondents (the `*` target). What CHATHISTORY does not provide is a way to enumerate all the "open queries" (open buffers?), since this concept is time-relative on the server side.

jwheare:
Yeah but it's still an issue because you have to know that a nick pmed you to request the chat from it. The * target is the workaround for that, but it has the issue I described.

slingamn:
That's what I meant by "open queries". I'm interested in potential solutions to this problem.

emersion:
WIP implementations are available in the `bouncer-ext` branches of soju and gamja:

- https://git.sr.ht/~emersion/soju/tree/bouncer-ext
- https://git.sr.ht/~emersion/gamja/tree/bouncer-ext

extensions/bouncer-networks.md
```diff
@@ -0,0 +1,259 @@
...
+
+    BOUNCER LISTNETWORKS
+
+The server replies with:
+
+    BOUNCER LISTNETWORKS <netid> <attributes>
+
+And then ends the list (even if empty) with:
+
+    BOUNCER LISTNETWORKS *
```
emersion, line 74: It may be desirable to use a batch instead of this special end-of-list message.

prawnsalad:
I don't think I'm following why you would want to remove buffer support from the spec? I assume your bouncer has a list of joined channels and/or queries either in memory or on disk? If so, these would be your buffers.

If you don't have a list of joined channels, how does your bouncer handle notifying the client of joined channels?

emersion:
I have a list of joined channels in memory, but I don't have a list of queries. It's not useful for my use-case.

The client is notified of the joined channels via regular JOIN messages. The client can manage channel buffers via regular JOIN/PART commands.

DanielOaks:
I don't see a reason to leave the buffers part from the original spec out.

emersion:
I don't see a reason to implement it.

Alright, I'll just ship my vendored ext. I'll reconsider submitting an IRCv3 proposal once things have settled.

emersion:
For future reference, I'll keep an up-to-date spec for the vendored ext here: https://git.sr.ht/~emersion/soju/tree/bouncer-ext/item/doc/ext/bouncer-networks.md

ilmari-lauhakangas:
> For future reference, I'll keep an up-to-date spec for the vendored ext here: https://git.sr.ht/~emersion/soju/tree/bouncer-ext/item/doc/ext/bouncer-networks.md

New location: https://git.sr.ht/~emersion/soju/tree/master/item/doc/ext/bouncer-networks.md

PR #435, opened 2021-02-13
Add draft account-registration spec

progval:
This draft specification adds **simple** `REGISTER` and `VERIFY` commands to standardize account creation, as a replacement for the withdrawn [Account Management Framework](https://github.com/ircv3/ircv3-specifications/pull/276)

It is based on @edk0's [draft](https://gist.github.com/edk0/bf3b50fc219fd1bed1aa15d98bfb6495), and is implemented by Oragono and Limnoria.

[rendered view](https://github.com/ProgVal/ircv3-specifications/blob/register/extensions/register.md)

### DIFF
--- a/extensions/account-registration.md
+++ b/extensions/account-registration.md
@@ -71,7 +71,7 @@ as if it used SASL.
 Sent by the server as a response to `REGISTER` when further action is required
 to complete registration of the `<account>`, as explain in the human-readable `<message>`.
 
-    FAIL REGISTER USERNAME_EXISTS <account> <message>
+    FAIL REGISTER ACCOUNT_EXISTS <account> <message>
 
 Sent by the server as a response to `REGISTER` when the client tried to register
 while using a nick that is not available as a new account's name
--- a/extensions/account-registration.md
+++ b/extensions/account-registration.md
@@ -28,8 +28,8 @@ for future organic expansion.
 
 ### Capability
 
-This specification adds the `draft/register` capability, whose presence
-signifies that the server accepts the `REGISTER` command.
+This specification adds the `draft/account-registration` capability, whose
+presence signifies that the server accepts the `REGISTER` command.
 
 The capability has an optional value, a comma-separated list of key-value
 pairs; the format is intended to follow the precedent set by
@@ -42,22 +42,28 @@ The defined keys are:
    `FAIL REGISTER COMPLETE_CONNECTION_REQUIRED`
  * `email-required` - if present, registrations require a valid email address
    to process
+ * `custom-accountname` - if present, the account name can be different
+   from the user's current nickname
 
 Clients MUST ignore any value assigned to these keys, and MUST ignore
 any unknown key.
 
 Software implementing this work-in-progress specification MUST NOT use
 the unprefixed `register` or `account-registration` capability names.
-Instead, implementations SHOULD use the `draft/register`
+Instead, implementations SHOULD use the `draft/account-registration`
 capability name to be interoperable with other software implementing
 a compatible work-in-progress version.
 
 ### Commands
 
-    REGISTER {<email> | "*"} <password>
+    REGISTER <accountname> {<email> | "*"} <password>
     
 The `REGISTER` command informs the server of a request to register
 an account named for the current nick of the requestor.
+
+If `<accountname>` is `*`, then this value is the user's current nickname.
+If the server advertises the `custom-accountname` key, then this desired
+account name can be different from the user's current nickname.
 
 The `REGISTER` command MAY be sent at any point during the connection
 that the client has a valid nickname.
@@ -90,13 +96,24 @@ to complete registration of the `<account>`, as explain in the human-readable
 
 Sent by the server as a response to `REGISTER` when the client tried to register
 while using a nick that is not available as a new account's name
-(for example, because someone already registered it).
+because someone already registered it.
 
 The server MUST NOT send this response before the client sent a `NICK` command.
 
 The client MAY try registering again later.
 Clients should not automatically retry immediately without changing
 their nickname or waiting.
+
+    FAIL REGISTER BAD_ACCOUNTNAME <account> <message>
+
+Sent by the server to indicate that the desired accountname is invalid or
+otherwise restricted/disallowed. The message should tell the user how or why
+the desired name has been rejected.
+
+    FAIL REGISTER ACCOUNTNAME_MUST_BE_NICK <account> <message>
+
+Sent by the server to indicate that the desired accountname does not match
+the user's current nick, when it must match.
 
     FAIL REGISTER NEED_NICK * <message>
 
--- a/extensions/account-registration.md
+++ b/extensions/account-registration.md
@@ -152,6 +152,11 @@ like a `privmsg` to nickserv), which are limited in length, servers may want to
 prevent or discourage users from setting passwords so long they may not fit
 in these messages. 300 bytes should be a safe reasonable limit.
 
+Servers should ensure that if they allow REGISTER before connection registration,
+this functionality cannot be used to bypass any authentication restrictions
+defined in the server configuration, e.g. requirements that clients send
+a server password with PASS.
+
 
 
 [multiline]: https://github.com/ircv3/ircv3-specifications/pull/398/
--- a/extensions/account-registration.md
+++ b/extensions/account-registration.md
@@ -176,11 +176,9 @@ unavailable.
 
 This section is non-normative.
 
-Passwords are opaque byte strings.
-It is recommended for them to be valid UTF-8;
+Passwords are opaque byte strings. It is recommended for them to be valid UTF-8;
 or authentication may be impossible later (eg. with SASL PLAIN).
-Servers may also choose to reject any non-UTF-8 password with
-`UNACCEPTABLE_PASSWORD`.
+Servers may also choose to reject any non-UTF-8 password with `UNACCEPTABLE_PASSWORD`.
 
 As passwords may need to be sent in non-`authenticate` messages,
 like a `PRIVMSG` to NickServ), which are limited in length, clients may want to
--- a/extensions/account-registration.md
+++ b/extensions/account-registration.md
@@ -83,6 +83,11 @@ The client MAY try registering again later.
 Clients should not automatically retry immediately without changing their nickname
 or waiting.
 
+    FAIL REGISTER NEED_NICK * <message>
+
+Sent by the server if the client sends `REGISTER` before sending a NICK command.
+This is only possible in setups that advertise the `before-connect` key.
+
     FAIL REGISTER ALREADY_REGISTERED <account> <message>
     FAIL VERIFY ALREADY_REGISTERED <account> <message>
 
--- a/extensions/account-registration.md
+++ b/extensions/account-registration.md
@@ -91,7 +91,7 @@ This is only possible in setups that advertise the `before-connect` key.
     FAIL REGISTER ALREADY_REGISTERED <account> <message>
     FAIL VERIFY ALREADY_REGISTERED <account> <message>
 
-Sent by the server is the client is already authenticated.
+Sent by the server if the client is already authenticated.
 
     FAIL REGISTER WEAK_PASSWORD <account> <message>
 
--- a/extensions/account-registration.md
+++ b/extensions/account-registration.md
@@ -40,6 +40,12 @@ The defined keys are:
 
 Clients MUST ignore any value assigned to these keys, and MUST ignore any unknown key.
 
+Software implementing this work-in-progress specification MUST NOT use
+the unprefixed `register` or `account-registration` capability names.
+Instead, implementations SHOULD use the `draft/register`
+capability name to be interoperable with other software implementing
+a compatible work-in-progress version.
+
 ### Commands
 
     REGISTER {<email> | "*"} <password>
--- a/extensions/account-registration.md
+++ b/extensions/account-registration.md
@@ -104,13 +104,13 @@ The client MAY try registering again later.
 Clients should not automatically retry immediately without changing
 their nickname or waiting.
 
-    FAIL REGISTER BAD_ACCOUNTNAME <account> <message>
+    FAIL REGISTER BAD_ACCOUNT_NAME <account> <message>
 
 Sent by the server to indicate that the desired account name is invalid or
 otherwise restricted/disallowed. The message should tell the user how or why
 the desired name has been rejected.
 
-    FAIL REGISTER ACCOUNTNAME_MUST_BE_NICK <account> <message>
+    FAIL REGISTER ACCOUNT_NAME_MUST_BE_NICK <account> <message>
 
 Sent by the server to indicate that the desired account name does not match
 the user's current nick, when it must match.
--- a/extensions/account-registration.md
+++ b/extensions/account-registration.md
@@ -27,12 +27,18 @@ The final version of the specification will use an unprefixed capability name.
 
 ## Motivation
 
-The lack of a standardized way to register an account has been a source
-of frustration for many participants for several years.
+A standard method of registering a user account has historically been missing
+from IRC. Instead various methods such as sending messages to services bots
+(e.g. NickServ) have been employed.
 
-This specification is intentionally restricted in scope to be appealing
-to implementors by being severely. The authors hope that this solution
-will serve as the basis for future organic expansion.
+This specification aims to provide a standard baseline for account
+registration that allows client developers to build more reliable and
+integrated user experiences around account registration.
+
+It does not aim to cover all registration-related use cases that may currently
+be in use by server implementations to varying degrees. However, future
+extensions to the specification might be proposed given a broad enough
+consensus.
 
 ### Capability
 
--- a/extensions/account-registration.md
+++ b/extensions/account-registration.md
@@ -30,12 +30,9 @@ The final version of the specification will use an unprefixed capability name.
 The lack of a standardized way to register an account has been a source
 of frustration for many participants for several years.
 
-This isn't the first attempt to fix it. Unlike the withdrawn `ACC` proposal,
-this effort aims to be more appealing to implementors by being severely
-restricted in scope.
-The only register-verify flows supported are those already in use in the wild.
-The authors hope that this stripped-down solution will serve as the basis
-for future organic expansion.
+This specification is intentionally restricted in scope to be appealing
+to implementors by being severely. The authors hope that this solution
+will serve as the basis for future organic expansion.
 
 ### Capability
 
--- a/extensions/account-registration.md
+++ b/extensions/account-registration.md
@@ -67,7 +67,8 @@ a second attempt after it receives the welcome message.
 
     VERIFY <account> <code>
     
-The `VERIFY` command completes a registration that required email verification.
+The `VERIFY` command completes a registration that required verification
+(eg. via email or CAPTCHA).
 
 ### Responses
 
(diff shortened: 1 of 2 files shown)


westor7:
Very nice idea and i hope for this, having some quick questions about this:

1. What if the irc services having VERIFY disabled, so a user no need to verify his email?
2. What if a user due the VERIFY sends an incorrect code or incorrect account?
3. What if the irc services will be down?

- Thanks!

ProgVal:
>     1. What if the irc services having VERIFY disabled, so a user no need to verify his email?

Then the server responds with `REGISTER SUCCESS` instad of `REGISTER VERIFICATION_REQUIRED`.

>     2. What if a user due the VERIFY sends an incorrect code or incorrect account?
>     3. What if the irc services will be down?

Good catch. Added response codes `FAIL VERIFY INVALID_CODE` and `FAIL REGISTER/VERIFY UNAVAILABLE`.

westor7:
https://github.com/ircv3/ircv3-specifications/pull/435/commits/339361e863dba410c5bea12ec942c4acdc8a9514#diff-2050f0354e63e54a43299643306ae9c6b94a0a5e24af547c1b50a95fadbfc81aR122

You mean if NOT right?

So: ``Sent by the server if the `REGISTER`/`VERIFY` commands are temporarily available.``
Should be: ``Sent by the server if the `REGISTER`/`VERIFY` commands are temporarily unavailable.``

DanielOaks:
We should reach out to see whether Atheme+Anope have any thoughts on this. Any ircds have thoughts about it? (obviously Ora likes it, and we started using the `before-connect` key because that supports certain kinds of restricted installs that we support but I don't think traditional irc nets would be into that aspect).

aaronmdjones:
Why do the success messages resemble e.g. `REGISTER SUCCESS` but the failure messages resemble `FAIL REGISTER`? Should they not be the other way around?

ProgVal:
@aaronmdjones The error messages follow the new [Standard Replies Extension](https://ircv3.net/specs/extensions/standard-replies)

extensions/register.md
```diff
@@ -0,0 +1,127 @@
...
+
+The defined keys are:
+
+ * `before-connect` - if present, indicates the server supports early registration, so it
+   MUST NOT use `FAIL REGISTER COMPLETE_CONNECTION_REQUIRED`
+ * `email-required` - if present, registrations require a valid email address to process
+
+### Commands
+
+    REGISTER {<email> | "*"} <password>
```
hhirtz, line 46: Thoughts on

```suggestion
The defined keys are:

 * `before-connect` - if present, indicates the server supports early registration, so it
   MUST NOT use `FAIL REGISTER COMPLETE_CONNECTION_REQUIRED`
 * `verify` - if present, registrations require verification to proceed
 * `verify-email` - if present, verification can be done via email
 * `verify-tel` - if present, verification can be done via telephone

### Commands

    REGISTER {verify-mean=<verify-value> | "*"} <password>
```

?

It'd allow something like

```
S: CAP * LS :register=before-connect,verify,verify-email,verify-tel
C: REGISTER tel=+33612345678 :a very secure password
```

Vendor verification means could be added like so:

```
S: CAP * LS :register=before-connect,verify,verify-email,vendor/verify-matrix
C: REGISTER vendor/matrix=@freenode_pinkieval:matrix.org :though this should be strictly checked
```

emersion: I don't think we should try to make this spec as generic as ACC was.

ProgVal: I agree with @emersion. I'll add a requirement that clients should ignore unknown keys, so future specification can extend this one by adding keys.

edk0:
null characters can't be in IRC messages anyway so mentioning them feels a bit redundant

extensions/register.md
```diff
@@ -0,0 +1,147 @@
...
+    FAIL REGISTER INVALID_EMAIL <account> <message>
+
+Sent by the server if it cannot send emails to the provided address.
+
+    FAIL REGISTER UNACCEPTABLE_EMAIL <account> <message>
+
+Sent by the server if the email address is valid, but not available for account registration.
+
+    FAIL REGISTER COMPLETE_CONNECTION_REQUIRED
+    FAIL VERIFY COMPLETE_CONNECTION_REQUIRED
```
DanielOaks, line 112: You'll need the `<message>` param on the end of these to match the std replies format.

extensions/register.md
```diff
@@ -0,0 +1,147 @@
...
+Sent by the server if the `REGISTER`/`VERIFY` commands are temporarily unavailable.
+
+
+# Client considerations
+
+This section is non-normative.
+
+Passwords are opaque byte strings.
+It is recommended for them to be valid UTF-8;
+or authentication may be impossible later (eg. with SASL PLAIN).
```
slingamn, line 134: Pure pedantry: I don't think SASL PLAIN would have any problems here, since everything gets base64-encoded. The issue is more with being able to pass them easily to NickServ (e.g., to change the password).

I think a warning about password length may be in order? I've been imposing a cap of 300 bytes to ensure that passwords fit comfortably into IRC lines (it's theoretically possible that someone could submit a passphrase through this framework that gets truncated in other contexts). And although I haven't been validating this until now, it might also be worth checking that passphrases are acceptable as non-final IRC parameters (nonempty, no spaces, no initial `:`).

ProgVal: Purer pedantry: RFC 4616 requires the encoded bytes to be a UTF-8 string.

IMO warnings about password length should be left to the UI. I could mention it in a non-normative section though.

DanielOaks:
So can we put this on the website and wait for implementations?

slingamn:
I'll compound my embarrassment a bit by requesting a "security considerations" section reflecting https://github.com/oragono/oragono/issues/1634 --- something like:

>Servers MUST ensure that if they allow `REGISTER` before connection registration, this functionality cannot be used to bypass any authentication restrictions defined in the server configuration, e.g. requirements that clients send `PASS` or authenticate with SASL.

SadieCat:
Trivial suggestion (don't block merging on this): Might be a good idea to rename the spec to "account registration" rather than just "register".

extensions/register.md
```diff
@@ -0,0 +1,157 @@
...
+
+ * `before-connect` - if present, indicates the server supports early registration, so it
+   MUST NOT use `FAIL REGISTER COMPLETE_CONNECTION_REQUIRED`
+ * `email-required` - if present, registrations require a valid email address to process
+
+Clients MUST ignore any value assigned to these keys, and MUST ignore any unknown key.
+
+### Commands
+
+    REGISTER {<email> | "*"} <password>
```
emersion, line 45: In case the server supports `before-connect`, what is the "curerent nick"? Should we require that a `NICK` command is sent before `REGISTER`?

extensions/register.md
```diff
@@ -0,0 +1,157 @@
...
+The server MUST NOT send this response before the client sent a `NICK` command.
+
+The client MAY try registering again later.
+Clients should not automatically retry immediately without changing their nickname
+or waiting.
+
+    FAIL REGISTER ALREADY_REGISTERED <account> <message>
+    FAIL VERIFY ALREADY_REGISTERED <account> <message>
+
+Sent by the server is the client is already authenticated.
```
DanielOaks, line 89: s/is/if/

extensions/register.md
DanielOaks: nice catch. something like this in the list of responses should be sufficient? and/or a line on the `before-connect` key description. it's possible in `before-connect` setups but given those are mostly an experimental use case tossed on top I'd rather not describe this in the 'mainline' spec path.

>     FAIL REGISTER NEED_NICK <message>
> 
> Sent by the server if the client sends `REGISTER` before sending a `NICK` command. This is only possible in setups that advertise the `before-connect` key.

SadieCat:
It might be a good idea to have password min/max lengths in the register cap value too?

extensions/register.md
```diff
@@ -0,0 +1,157 @@
...
+further action is needed.
+Afterward, the client is assumed to be authenticated to the `<account>`
+as if it used SASL.
+
+    REGISTER VERIFICATION_REQUIRED <account> <message>
+    
+Sent by the server as a response to `REGISTER` when further action is required
+to complete registration of the `<account>`, as explain in the human-readable `<message>`.
+
+    FAIL REGISTER USERNAME_EXISTS <account> <message>
```
SadieCat, line 74:
```suggestion
    FAIL REGISTER ACCOUNT_EXISTS <account> <message>
```

Would prevent overloaded terminology (username can also refer to what is also known as the ident).

extensions/register.md
```diff
@@ -0,0 +1,157 @@
...
+accepts the `REGISTER` command.
+
+The capability has an optional value, a comma-separated list of key-value pairs; the format is intended
+to follow the precedent set by [the multiline draft][multiline].
+
+The defined keys are:
+
+ * `before-connect` - if present, indicates the server supports early registration, so it
+   MUST NOT use `FAIL REGISTER COMPLETE_CONNECTION_REQUIRED`
+ * `email-required` - if present, registrations require a valid email address to process
```
SadieCat, line 39: Should there be a fail code for when the user sends * instead of an email or is INVALID_EMAIL enough?

DanielOaks: INVALID_EMAIL should be enough, the client software can work it out based on that and the message should tell the user what went wrong

emersion:
A WIP client implementation is available for gamja in this branch: https://git.sr.ht/~emersion/gamja/log/register-account

Should be enough to test basic functionality.

DanielOaks:
Do we need Atheme/Anope to look at this or is this something that ircds will implement all on their own without services involvement? If we do need services to look at this who from those projects can look over this and give us the yay/nay on it?

SadieCat:
Services will need per-ircd support like with SASL probably (I'm implementing the Anope side of it).

extensions/register.md
ProgVal: I added a `*` argument before `<message>`, for consistency with the other errors.

ProgVal:
I applied all the suggestions above. In particular, note that the capability was renamed `draft/account-registration` instead of `draft/register`, for consistency with this spec's name, and other capability names.

extensions/account-registration.md
```diff
@@ -0,0 +1,187 @@
...
+the unprefixed `register` or `account-registration` capability names.
+Instead, implementations SHOULD use the `draft/register`
+capability name to be interoperable with other software implementing
+a compatible work-in-progress version.
+
+### Commands
+
+    REGISTER {<email> | "*"} <password>
+    
+The `REGISTER` command informs the server of a request to register
```
kaniini, line 59: For Ophion, and other environments where accounts exist, but nicknames are not owned (like Undernet for example), this is somewhat awkward.

It would be nice if we could have it be:

    REGISTER {<email> | "*"} [accountname] <password>

This would work for Ophion (and presumably cases like Undernet too).

extensions/account-registration.md
```diff
@@ -0,0 +1,187 @@
...
+
+The `REGISTER` command MAY be sent at any point during the connection
+that the client has a valid nickname.
+If the client sends `REGISTER` before completing connection registration,
+and receives a `FAIL REGISTER COMPLETE_CONNECTION_REQUIRED`, it SHOULD make
+a second attempt after it receives the welcome message.
+
+    VERIFY <account> <code>
+    
+The `VERIFY` command completes a registration that required email verification.
```
kaniini, line 70: Not all environments want to send email, for example, Ophion intends to prompt the user to complete a CAPTCHA it generates at some URI.  Changing "email verification" to "verification" would allow more flexibility.

kaniini (changes requested):

extensions/account-registration.md
```diff
@@ -0,0 +1,188 @@
...
+The client MAY try registering again later.
+Clients should not automatically retry immediately without changing
+their nickname or waiting.
+
+    FAIL REGISTER NEED_NICK * <message>
+
+Sent by the server if the client sends `REGISTER` before sending a NICK command.
+This is only possible in setups that advertise the `before-connect` key.
+
+    FAIL REGISTER ALREADY_REGISTERED <account> <message>
```
kaniini, line 106: The `ALREADY_REGISTERED` fail code is sent if a client is already *authenticated*.  Should this be `ALREADY_AUTHENTICATED`?

DanielOaks: makes sense

extensions/account-registration.md
ProgVal: Done by @DanielOaks through a PR to my repo.

extensions/account-registration.md
```diff
@@ -56,13 +56,13 @@ a compatible work-in-progress version.
 
 ### Commands
 
-    REGISTER <accountname> {<email> | "*"} <password>
+    REGISTER <account> {<email> | "*"} <password>
```
DanielOaks, line 59: this should probably be left as accountname so that it's more clear

ProgVal: The rest of the spec uses `<account>`.

extensions/account-registration.md
```diff
@@ -0,0 +1,205 @@
...
+
+# Server considerations
+
+This section is non-normative.
+
+Server implementations should be careful not to accidentally send `INVALID_EMAIL`
+as a response to a valid address.
+
+As passwords may need to be sent in non-`authenticate` messages,
+like a `privmsg` to nickserv), which are limited in length, servers may want to
```
SadieCat, line 194:
```suggestion
like a `PRIVMSG` to NickServ), which are limited in length, servers may want to
```

extensions/account-registration.md
```diff
@@ -0,0 +1,205 @@
...
+This section is non-normative.
+
+Passwords are opaque byte strings.
+It is recommended for them to be valid UTF-8;
+or authentication may be impossible later (eg. with SASL PLAIN).
+Servers may also choose to reject any non-UTF-8 password with
+`UNACCEPTABLE_PASSWORD`.
+
+As passwords may need to be sent in non-`authenticate` messages,
+like a `privmsg` to nickserv), which are limited in length, clients may want to
```
SadieCat, line 181:
```suggestion
like a `PRIVMSG` to NickServ), which are limited in length, clients may want to
```

extensions/account-registration.md
```diff
@@ -0,0 +1,205 @@
...
+Sent by the server if the email address is valid, but not available for
+account registration.
+
+    FAIL REGISTER COMPLETE_CONNECTION_REQUIRED <message>
+    FAIL VERIFY COMPLETE_CONNECTION_REQUIRED <message>
+
+Sent by the server if the client sent a `REGISTER` command before connection
+registration.
+The server MUST NOT sent these replies if it advertizes the `before-connect`
+key of the `draft/register` capability.
```
DanielOaks, line 156: oop, I forgot to update this to the new cap name `draft/account-registration`

slingamn:
Can we keep the draft name as `draft/register`, then plan to verify it as `account-registration`? (Same strategy as chathistory: we're keeping the draft name as `draft/event-playback` and planning to ratify as `chathistory-events`.)

progval:
Can we merge this?

jwheare:
@SadieCat is this implemented in Anope now?

SadieCat:
I do have a WIP implementation. I'll try to get it onto the InspIRCd Testnet this week.

progval:
@SadieCat Any news?

jwheare:
Probably good to merge this as draft even without an anope implementation. Client devs can test using ergo. Can do so next week unless there's any blockers.

jwheare:
Blocking merge pending resolution on ircv3/ircv3-ideas#78

jwheare:
Dropping draft/ isn't clear cut so unblocking this.

jwheare:
The motivation section probably doesn't need to mention the history of ACC or go into how the authors hope it will be acceptable. It can just be more direct about what the spec is for on its own terms. @progval can you have a stab at fixing that? Otherwise I can merge and we'll fix it before ratifying.

extensions/account-registration.md
```diff
@@ -0,0 +1,205 @@
...
+   from the user's current nickname
+
+Clients MUST ignore any value assigned to these keys, and MUST ignore
+any unknown key.
+
+Software implementing this work-in-progress specification MUST NOT use
+the unprefixed `register` or `account-registration` capability names.
+Instead, implementations SHOULD use the `draft/account-registration`
+capability name to be interoperable with other software implementing
+a compatible work-in-progress version.
```
jwheare, line 55: This needs to be moved into the usual "Notes for implementing work-in-progress version" style at the start. See e.g. #466 instead of burying it in another section.

extensions/account-registration.md
```diff
@@ -0,0 +1,205 @@
...
+for future organic expansion.
+
+### Capability
+
+This specification adds the `draft/account-registration` capability, whose
+presence signifies that the server accepts the `REGISTER` command.
+
+The capability has an optional value, a comma-separated list of key-value
+pairs; the format is intended to follow the precedent set by
+[the multiline draft][multiline].
```
jwheare, line 36: Not sure we need to mention this precedent here. The format should already be described in a standalone way.

extensions/account-registration.md
```diff
@@ -0,0 +1,205 @@
...
+
+# Client considerations
+
+This section is non-normative.
+
+Passwords are opaque byte strings.
+It is recommended for them to be valid UTF-8;
+or authentication may be impossible later (eg. with SASL PLAIN).
+Servers may also choose to reject any non-UTF-8 password with
+`UNACCEPTABLE_PASSWORD`.
```
jwheare, line 178: Fix the line breaking here (and elsewhere). Either hard wrap for line length or (more preferably) soft wrap, hard wrapping on punctuation is odd.

extensions/account-registration.md
```diff
@@ -0,0 +1,205 @@
...
+in these messages. 300 bytes should be a safe reasonable limit.
+
+Servers should ensure that if they allow REGISTER before connection registration,
+this functionality cannot be used to bypass any authentication restrictions
+defined in the server configuration, e.g. requirements that clients send
+a server password with PASS.
+
+
+
+[multiline]: https://github.com/ircv3/ircv3-specifications/pull/398/
```
jwheare, line 205: Can remove this when reference is dropped further up.

jwheare (changes requested):
Actually a few quick things I'd like changing before merge.

extensions/account-registration.md
progval: I don't understand; do you want me to copy-paste it here?

emersion: I think @jwheare wants to drop this part "the format is intended to follow the precedent set by multiline", while keeping the rest of the sentence.

progval: "The capability has an optional value, a comma-separated list of key-value pairs" alone is not enough to describe the format.

jwheare: If you need to copy paste it here to make it standalone then yeah.

progval:
All good now?

extensions/account-registration.md
```diff
@@ -0,0 +1,204 @@
...
+The final version of the specification will use an unprefixed capability name.
+
+## Motivation
+
+The lack of a standardized way to register an account has been a source
+of frustration for many participants for several years.
+
+This specification is intentionally restricted in scope to be appealing
+to implementors by being severely. The authors hope that this solution
+will serve as the basis for future organic expansion.
```
jwheare, line 35: How about this? Historically, other specs lean on the word "historically" in this section.

```suggestion
A standard method of registering a user account has historically been missing
from IRC. Instead various methods such as sending messages to services bots
(e.g. NickServ) have been employed.

This specification aims to provide a standard baseline for account
registration that allows client developers to build more reliable and
integrated user experiences around account registration.

It does not aim to cover all registration-related use cases that may currently
be in use by server implementations to varying degrees. However, future
extensions to the specification might be proposed given a broad enough
consensus.
```

progval: Nice!

jwheare (approved):
Alright looks good to merge now.

PR #436, opened 2021-02-23
allow NOTICE in draft/multiline

slingamn:


slingamn:
I checked and Oragono already allows this.

extensions/multiline.md
```diff
@@ -80,7 +80,7 @@ Servers MUST NOT reject blank lines other than in the following cases:
 * Clients MUST NOT send blank lines with the `draft/multiline-concat` tag.
 * Clients MUST NOT send messages consisting entirely of blank lines.
 
-Clients MUST NOT send messages other than PRIVMSG while a multiline batch is open.
+Clients MUST NOT send messages other than PRIVMSG or NOTICE while a multiline batch is open.
```
ProgVal, line 83: I get the intent, but formally, this probably needs to be reworded to disallow NOTICE in a PRIVMSG batch and vice-versa

slingamn: I believe this is covered by the first change: "Multiline batches MUST only contain one or more PRIVMSG lines, or one or more NOTICE lines."

ProgVal: Hmm, if I understand this line correctly, its purpose is to disallow messages that don't belong to the batch. But it fails to do so (even with PRIVMSG only), because it doesn't disallow this:

```

    Client: BATCH +792da7 draft/multiline #channel
    Client: @batch=792da7 PRIVMSG #channel :message1
    Client: PRIVMSG #channel :message2
    Client: BATCH -792da7
```

Unless it's intentional to allow PRIVMSGs that don't belong to the batch? If so, NOTICE should also be allowed, independently of whether this PR is accepted or not.

extensions/multiline.md
```diff
@@ -80,7 +80,7 @@ Servers MUST NOT reject blank lines other than in the following cases:
 * Clients MUST NOT send blank lines with the `draft/multiline-concat` tag.
 * Clients MUST NOT send messages consisting entirely of blank lines.
 
-Clients MUST NOT send messages other than PRIVMSG while a multiline batch is open.
+Once the client has opened a multiline batch, it MUST NOT send any messages that are not part of the batch, i.e., it may only send PRIVMSG or NOTICE with the appropriate `batch` tag, or the closing `BATCH` command.
```
ProgVal, line 83: `it may only send PRIVMSG or NOTICE` -> `it may send only PRIVMSG or only NOTICE`

slingamn: I really think that's implicit in the previous change?

ProgVal: It is, but it doesn't hurt to say it again. No big deal either way

PR #439, opened 2021-02-25
Bot mode (+B)

DanielOaks:
This specification defines the Bot mode, ISUPPORT token and WHO flags which are already in use by a number of IRC servers. It also defines the `bot` tag, which is in use as a vendor tag by InspIRCd right now (as `inspircd.org/bot`).

Short, sweet, simple. Mostly explains existing behaviour apart from that `bot` tag, and I don't really feel like it's necessary to go through the whole `draft/` prefix for six months stage for something which is a simple tag like this.

Discussion happened in https://github.com/ircv3/ircv3-ideas/issues/43 and servers implemented and converged on this behaviour (apart from the tag) during that discussion.

### DIFF
--- a/extensions/bot-mode.md
+++ b/extensions/bot-mode.md
@@ -35,7 +35,7 @@ This numeric is returned as part of a bot's `WHOIS` reply.
 When a `RPL_WHOREPLY` `(352)` numeric is returned for a bot, the character used as the value of the ISUPPORT `BOT` token is returned in the flags (alongside `H|G`).
 
 ## The `draft/bot` tag
-The `draft/bot` tag indicates that the given user is a bot. This tag SHOULD be added by the server to all commands sent by a bot (e.g. `PRIVMSG`, `JOIN`, `MODE`, `NOTICE`, and all others). The tag SHOULD also be added by the ircd to all numerics directly caused by the bot. This tag MUST only be sent to users who have requested the `message-tags` capability.
+The `draft/bot` tag indicates that the given user is a bot. This tag SHOULD be added by the server to all commands sent by a bot (e.g. `PRIVMSG`, `JOIN`, `MODE`, `NOTICE`, and all others). The tag SHOULD also be added by the ircd to all numerics directly caused by the bot. This tag MUST only be sent to users who have requested the `message-tags` capability. Servers MUST NOT send this tag with a value, and clients MUST ignore any value if it exists.
 
 ## Examples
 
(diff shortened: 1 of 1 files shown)


extensions/bot-mode.md
```diff
@@ -0,0 +1,64 @@
...
+When a client is marked as a bot, they are given a special numeric as part of their `WHOIS` response, it is indicated in their `WHO` flags, and servers may include the `bot` tag on that client's outgoing messages.
+
+## The `RPL_WHOISBOT` `(335)` numeric
+This numeric is returned as part of a bot's `WHOIS` reply.
+
+## The `B` `WHO` flag
+When a `RPL_WHOREPLY` `(352)` numeric is returned for a bot, the `"B"` character is returned in the flags (alongside `H|G`).
+
+## The `bot` tag
+The `bot` tag indicates that the given user is a bot. This tag SHOULD be added by the server to all commands sent by a bot (e.g. `PRIVMSG`, `JOIN`, `MODE`, `NOTICE`, and all others). The tag SHOULD also be added by the ircd to all numerics directly caused by the bot. This tag MUST only be sent to users who have requested the `message-tags` capability.
```
ProgVal, line 27: `This tag MUST only be sent to users who have requested` -> `This tag MUST be sent to users if and only if they have requested`

DanielOaks: double negatives feel a bit hard to process?

MaxLeiter: `This tag MUST NOT be sent to users who have not requested` also allows clients to not send the tag to clients that did request it.

SadieCat: @MaxLeiter It's a S2C tag not a C2S tag.

SadieCat:
LGTM. :shipit:

jwheare:
Don't see a strong reason for the draft/ to be skipped here.

jwheare:
Referring to the tag specifically. The other stuff is already out there in the wild.

ProgVal:
Also, mention that server mustn't/shouldn't send a value to the tag, but clients should ignore the value if any. (same reasons as usual)

SadieCat:
It might be a good idea to require the WHO flag to be the same as the value specified in the BOT isupport token? That way if a server implementation is already using `B` as a status rank they won't confuse any clients. This idea would be compatible with all existing server implementations afaik.

DanielOaks:
Alright, will change it to `draft/bot` and we could make the `WHO` flag be the same value I guess? requires a bit more work on clients' side but if it makes it more flexible in a way that makes sense and conforms with all current impls then sure why not.

extensions/bot-mode.md
DanielOaks: I'm gonna mark this as resolved for now, I prefer the original wording a bit more

DanielOaks:
I'd be interested in another server dev sanity-checking/commenting on the https://github.com/ircv3/ircv3-specifications/pull/439/commits/bdf5f5ac268da740b68d468b2bde5df04f80219d change, but yeah I'm pretty satisfied with this now after the changes.

SadieCat:
Is there anything blocking this?

jwheare:
LGTM

slingamn:
Let's take a census of implementations so we can ratify this? It sounds like this has at least two server implementations (Ergo, Inspircd) and at least two client implementations (Limnoria, Bobbit).

ProgVal:
Limnoria sets the bot mode, and IRCCloud uses the bot tag. According to https://ircv3.net/software/clients.html , BitBot doesn't set the mode.

PR #440, opened 2021-02-26
Specify the 'forward' mode.

progval:
It is already made available by many server implementations,
usually as `+f` (Charybdis, Oragono) or `+L` (UnrealIRCd, InspIRCd)

This specification adds a new ISUPPORT token so clients can detect
its name.

I wrote this based on Unreal's behavior, but it seems other IRCds behave similarly.

As for the bot spec, I think we can skip the draft step given the wide set of existing implementations

### DIFF
--- a/extensions/forward-mode.md
+++ b/extensions/forward-mode.md
@@ -22,7 +22,7 @@ channel when they try to join, by setting a channel mode.
 The value of the `FORWARD` token is the mode character which is used to enable
 this (e.g. `FORWARD=f` or `FORWARD=L`).
 
-IF the server does not support a forward mode, it MAY include the
+If the server does not support a forward mode, it MAY include the
 token with no value (`FORWARD=`).
 
 ## The FORWARD mode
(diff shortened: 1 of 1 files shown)


jwheare:
This is cool. What is our plan for importing stuff like this into ircv3 in general? There's a lot of de-factor standards out there that aren't really documented somewhere. Are we aiming to pull all of that onto IRCv3 or just some stuff? What's the rationale for what we choose to import.

Something like WHOX really should be documented, because other specs rely on it now e.g. account-tag account-notify etc.

It's less clear what the basis for forward is, mode based stuff in general is kind of fraught. Is it just because most implementations agree, or what?

jwheare:
On second thoughts, no need to reply to that comment, it's mostly me thinking aloud and not really relevant to this PR directly.

SadieCat:
>IF the server does not support a forward mode, it MAY include the token with no value (`FORWARD=`).

Is this necessary? If a server doesn't support the spec it won't be advertising the token anyway.

ProgVal:
This is to tell the difference between servers that don't support it at all, and existing servers that support it but don't advertise the token

progval:
The only novelty in this spec is the FORWARD ISUPPORT token; but after more thoughts; I don't think this is a great idea, as we'd just keep adding tokens like this for every mode. Sadie's named-modes looks like a better solution.

As for the rest of the spec; this should go in https://modern.ircdocs.horse/ instead of an IRCv3 spec.

Issue #329, opened 2017-11-09
Cleanup specs and structure


DanielOaks:
This is something I've been meaning to do for a while, just making an issue so I remember to submit the changes. Once #327 is done and SASL is merged I'll do this.

1. Remove version numbers from spec filenames and YAML title names.
2. Where they exist, remove markdown headers that duplicate the YAML title headers on specs (some specs for instance have an `## IRCv3 <blah> Extension` up the top in addition to the YAML header – rip that out.
3. Reorganise the specs under a few separate folders.

Since this repo's a submodule named `/specs` on the website, makes sense to have a few subfolders:

- `specs/client-tags`: Client tag specifications.
- `specs/batches`: Batch type specifications.
- `specs/sasl-mechs`: SASL-mechanism-specific specs.
- `specs/metadata-keys`: Metadata key specs.
- `specs/deprecated`: Deprecated specs we want to keep around (such as the current SASL mech specs under `specs/documentation`).
- `specs/extensions`: Everything not covered by the above.

Yep, under this arrangement everything currently in `specs/core` gets put into `specs/extensions` instead. It makes sure people know that things are optional (things like metadata aren't mandatory to implement to be IRCv3-compliant), and the main [Specs page](http://ircv3.net/irc/) on the website is better positioned to make the necessary separation between, for example, cap, message-tags, and specs like extended-join anyway.

DanielOaks:
Note to self when SASL gets merged, fit this example into the merged spec:

-----

Example showing new SASL mechanisms becoming available. This example requires the client to support `CAP LS 302`:		

        Client: CAP LS 302
        Server: :irc.example.com CAP * LS :sasl=PLAIN
        ...
        Server: :irc.example.com CAP client DEL :sasl
        Server: :irc.example.com CAP client NEW :sasl=PLAIN,EXTERNAL
		
The given `NEW`/`DEL` messages may be displayed together for simplicity's sake.

jwheare:
We could also note that 908 can be used to advertise a change of mechlist

DanielOaks:
For changing mechanisms, should we push CAP DEL+NEW or 908? Maybe if they've already negotiated the `sasl` cap and the mechanism list changes, just send them 908 yeah?

jwheare:
We could push 908 for changed mechlists but still have an example of sasl going away and then coming back on e.g. a services outage. Clients should handle both.

DanielOaks:
[...]

PR #441, opened 2021-02-27
Spec cleanups

DanielOaks:
This PR does steps 2+3+4 in #329 - cleans up the specs, puts them in a more defined folder structure, and unifies the titles (both how they're named, and also removing useless duplicated Markdown titles).

I've left Metadata and SASL alone for now, since there's some non-trivial PRs to those.


ProgVal (commented):
> Fix duplicate markdown headers.

Shouldn't you add a new header at the beginning, eg. `## Description` for consistency with https://ircv3.net/specs/extensions/echo-message-3.2.html ?

>  Reorganise specs under new folder structure. 

Shouldn't they all have a redirect?

DanielOaks:
re markdown headers I was thinking that yeah, `## Description` or `## Introduction` or whatever fits - will need to see what fits for each spec. Will throw another commit on top adding them, thanks for the yell.

Yeah I've added redirects from their old locations to their new ones, `redirect_from` key in the yaml header auto-creates redirect files.

DanielOaks:
Cool, I'm fairly happy with this now. Some of these changes are made for the sake of consistency across our specs, but I've tried to only touch the initial header terms / metadata / file structure and no actual spec text here. Good step forward, not the last, keep an eye on #329 for future work I plan to do on our existing specs, and throw additional suggestions in there. Reviews + merge pls.

jwheare:
Looking snazzy. Can you make a draft pr to the website repo that refs this branch for the submod so we can preview it?

DanielOaks:
Draft website view: https://deploy-preview-355--ircv3.netlify.app/irc/

DanielOaks:
So I'd like SNI, and future stuff like the [clients loading urls considerations document](https://github.com/ircv3/ircv3-ideas/issues/16) to live outside of `https://ircv3.net/specs/` since they... don't really fit there, I reckon. More appropriate under a different root folder on the site, say `https://ircv3.net/docs/*` instead. So I've taken that document out of this repo, and I'll throw it into the website repo when these changes get merged. Down the line, once we have more than like, one or two documents, we could have a generic repo for docs vs specifications, but that's work that can be really easily done in future given the very few documents along these lines that we have right now.

jwheare:
Is this good to go?

PR #442, opened 2021-02-27
chathistory: The command should not return a batch unless the batch cap was negotiated

ProgVal:


slingamn:
Strongly support this; my intention was that you should be able to issue `CHATHISTORY` queries without support for batch or tags, getting an appropriately degraded response.

jwheare:
Lgtm

PR #443, opened 2021-02-28
[sts] require clients to ignore missing policy parts.

jwheare:
Also clarify part requirements and connection types

This follows up from https://github.com/ircv3/ircv3-specifications/pull/390#issuecomment-785220509 and makes the requirements a bit clearer.

The goal here is to allow for potential policy extension in future without breaking existing implementations.

### DIFF
--- a/extensions/sts.md
+++ b/extensions/sts.md
@@ -36,9 +36,13 @@ Clients MUST ignore every token with a key that they don't understand.
 
 An STS policy has several parts:
 
-* An **upgrade policy**, expressed via the [`port` key](#the-port-key) over an insecure connection. *REQUIRED*
-* A **persistence policy**, expressed via the [`duration` key](#the-duration-key) over a secure connection. *REQUIRED*
-* A **preload policy**, expressed via the [`preload` key](#the-preload-key) over a secure connection. *OPTIONAL*
+* An **upgrade policy**, expressed via the [`port` key](#the-port-key). *REQUIRED* on an insecure connection.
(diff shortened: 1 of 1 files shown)


extensions/sts.md
```diff
@@ -36,9 +36,13 @@ Clients MUST ignore every token with a key that they don't understand.
 
 An STS policy has several parts:
 
-* An **upgrade policy**, expressed via the [`port` key](#the-port-key) over an insecure connection. *REQUIRED*
-* A **persistence policy**, expressed via the [`duration` key](#the-duration-key) over a secure connection. *REQUIRED*
-* A **preload policy**, expressed via the [`preload` key](#the-preload-key) over a secure connection. *OPTIONAL*
+* An **upgrade policy**, expressed via the [`port` key](#the-port-key). *REQUIRED* on an insecure connection.
+* A **persistence policy**, expressed via the [`duration` key](#the-duration-key) *REQUIRED* on a secure connection.
+* A **preload policy**, expressed via the [`preload` key](#the-preload-key) *OPTIONAL* on a secure connection.
```
ProgVal, line 41: nitpick: "*OPTIONAL* on a secure connection." seems to imply that it isn't optional on an insecure connection, ie. that is required. The existing wording was less ambiguous for that item.

jwheare: I think that interpret would have to be a deliberately dishonest misinterpretation.

jwheare:
Will merge this in a couple days if there are no outstanding comments.

PR #447, opened 2021-03-06
[multiline] Relax restriction on clients sending unrelated messages

ProgVal:
Upholding this requirement can be excessively hard to implement
for some clients (eg. if they have multiple queues and/or are threaded).

S2C batches do not have this constraint, so C2S batches
shouldn't either, unless this is a significant burden on server
implementations.

(I'm aware this conflicts with #436, but I'd like a discussion on the
idea behind the change)

### DIFF
--- a/extensions/multiline.md
+++ b/extensions/multiline.md
@@ -80,8 +80,6 @@ Servers MUST NOT reject blank lines other than in the following cases:
 * Clients MUST NOT send blank lines with the `draft/multiline-concat` tag.
 * Clients MUST NOT send messages consisting entirely of blank lines.
 
-Clients MUST NOT send messages other than PRIVMSG while a multiline batch is open.
-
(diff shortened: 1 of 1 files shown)


slingamn:
Oragono currently enforces this restriction:

https://github.com/oragono/oragono/blob/788b37b12ac10839bd177d11ef4d797e022b66e6/irc/commands.go#L42

the rationale being that fakelag is temporarily disabled while a multiline batch is open, so allowing other commands would let the client bypass fakelag.

In general multiline poses complex DoS/abuse issues; see #413 and related for some historical context (one of the concerns there was that clients should not be allowed to send arbitrarily long multiline batches, even in the case where `max-bytes` is enforced and `max-lines` is not).

I think a compromise could be possible here but it's definitely a tricky issue.

PR #448, opened 2021-03-07
chathistory: soften FAIL code guarantees for programming errors

slingamn:
@kylef alerted me to the fact that Oragono violates a MUST of the current spec: "If the server receives a `CHATHISTORY` command with missing parameters, the `NEED_MORE_PARAMS` error code MUST be returned." Oragono will send `461 ERR_NEEDMOREPARAMS` in some cases, based on generic command-handling logic.

Upon consideration, there doesn't seem to be much of a point in having a MUST here at all, or in requiring distinct error codes for `UNKNOWN_COMMAND` and `NEED_MORE_PARAMS`: these are both programming errors, rather than runtime errors, and they're unrecoverable at runtime, so the only job of the error message is to illuminate the problem for debugging. Hence this proposed change does the following:

1. All FAIL codes are now SHOULD
2. All unrecoverable programming errors are now `INVALID_PARAMS`
3. `INVALID_TARGET` (which is a potentially recoverable runtime error) now includes the target as its second parameter

### DIFF
--- a/extensions/chathistory.md
+++ b/extensions/chathistory.md
@@ -107,6 +107,8 @@ If the server receives a syntactically invalid `CHATHISTORY` command, e.g., an u
     FAIL CHATHISTORY INVALID_PARAMS the_given_command :Too many parameters
     FAIL CHATHISTORY INVALID_PARAMS the_given_command the_given_timestamp :Invalid timestamp
 
+If an existing error numeric such as `ERR_NEEDMOREPARAMS` is appropriate, it MAY be used instead.
+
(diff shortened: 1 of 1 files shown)


jwheare:
In channel renaming we have this wording after defining FAIL codes:
https://ircv3.net/specs/extensions/channel-rename#errors

> If existing error numerics (such as ERR_CHANOPRIVSNEEDED, ERR_NOTONCHANNEL, ERR_NEEDMOREPARAMS) are more appropriate, they SHOULD be used instead.

Might be more appropriate to follow that style.

Issue #444, opened 2021-03-01
CHATHISTORY: consider an API to discover DM correspondents


slingamn:
This is related to, but distinct from #438. It comes out of discussion on #434 (and out-of-band with @kylef).

The previous de facto standard for retrieving history was [znc.in/playback](https://wiki.znc.in/Playback). `draft/chathistory` is at full feature parity with this, except for one feature: the `LIST` command, which "lists available buffers" (including joined channels and "active" DM correspondents).

This functionality cannot be imported directly into `draft/chathistory` because of the design assumption that all client synchronization state must be stored on the client side: therefore, the server has no concept of "active buffers" or "open queries". The set of DM correspondents is only defined relative to a time window.

However, the lack of such an API creates a problem: DMs from one user get "buried" behind another, so the client has to retrieve all missed DMs before it can surface new DMs of interest from previously unknown correspondents.

We can design a new API for this (and indexes to support it), it just has to be explicitly time-relative (or potentially msgid-relative?), and we have to verify that it can be efficiently implemented on the server side.

hhirtz:
> However, the lack of such an API creates a problem: DMs from one user get "buried" behind another, so the client has to retrieve all missed DMs before it can surface new DMs of interest from previously unknown correspondents.

Also, the fact that "*" is an invalid chathistory batch name i guess? or was it resolved?

slingamn:
That's being considered on #437. I thought this merited a separate discussion thread.

slingamn:
First pass: `CHATHISTORY CORRESPONDENTS timestamp=<start> timestamp=<end> <limit>`:

1. Timestamps queried and referenced in selectors are the times of the *latest* message sent from or to the correspondent
1. Same selector semantics as `BETWEEN`: starting from and excluding the first selector, ending on and excluding the second selector, either forwards or backwards in time
1. Same pagination semantics as `BETWEEN`
1. Returns messages like `CHATHISTORY CORRESPONDENT <nick> <timestamp>`. This introduces a new `CHATHISTORY` command; continuation is handled by `BATCH` as usual (this may require a new batch type?)

slingamn:
If we can agree on an API here, discussion is trending in the direction of it replacing the `*` target (not totally sure about that though).

emersion:
Looks like a good start. Some thoughts:

[...]

PR #449, opened 2021-03-07
Initial bouncer implementation draft spec

prawnsalad:
A progression from https://github.com/ircv3/ircv3-ideas/issues/39 to have this in spec form and clearer WIP history.

Added since the gist:
- Clearer on what a "buffer" is. An implementation defined channel/query/active buffer/etc.
- A note on how this spec optionally keeps multiple connected clients in sync
- Clearer explanation on what the CAP can be used for.

Next thoughts:
- As per https://github.com/ircv3/ircv3-ideas/issues/39#issuecomment-783485317, some sort of idle/active/inactive client state to aid the server in its notification logic. This spec does mention an optional notification level setting for buffers so including this makes sense.
- Could the name "BOUNCER" for this spec be loading too much towards a traditional IRC bouncer? As with Oragano the lines get muddled between ircd/bouncer so perhaps a better term could be used.

### DIFF
--- a/extensions/bouncer.md
+++ b/extensions/bouncer.md
@@ -1,0 +1,236 @@
+---
+title: IRCv3 bouncer extension
+layout: spec
+work-in-progress: true
+copyrights:
+  -
+    name: "Darren Whitlen"
+    period: "2018-2021"
+    email: "darren@kiwiirc.com"
+---
+
+## BOUNCER
+
+This feature is intended to provide an API for clients to build a UI around bouncer IRC servers, to manage the users IRC networks and buffers in a user friendly way and to keep multiple clients and devices synced together.
+
+It is based around network and buffer entities which both have key/value tags, such as `joined=true` for a buffer, with commands to manage these. A list of common tags can be found under the tags section of this document.
+
+From a single bouncer connection, a complete implementation will allow users to:
+* Add, remove, modify and search networks under their bouncer account.
+* List and search channels and queries.
+
+
+## Notes
+References in this document:
+- A `network` is an IRC network that a bouncer connects to
+- A `buffer` is either a channel or query
+- A client is a users client or device that connects to the bouncer
+
+A network is identified by a per-user unique ID that MUST not change. A human readable label MUST also be associated with the network that the user can identify it with in a client UI - this is the `network` tag in most commands. The ID and label may be the same as long as the ID does not change. Neither the ID or label may contain a space or `:` characters.
+
+All commands and data for this bouncer feature MUST be associated to a logged-in and authorised user. It is the implementors choice on how the user authentication is handled and is outside the scope of this document.
+
+## Capabilities
+
+This adds a capability called `BOUNCER`. Once negotiated the server may change it's behaviour to better suit bouncer clients, such as not automatically sending `JOIN` commands for already joined channels.
+
+## RPL_ISUPPORT
+This adds an isupport token called `BOUNCER`. Its optional value is message-tag encoded data including:
+* `network=` The name or label of the network being connected to.
+* `netid=` The ID of the network being connected to.
+
+Eg: `BOUNCER=network=freenode;netid=1234`
+
+## `BOUNCER` command
+This introduces the `BOUNCER` command, followed by a case insensitive subcommand.
+
+All tags sent and received by this command are message-tags encoded.
+
+***Note***: Where a network ID is used to specify a network in a command, a `*` character may be used instead to signify the active network the client is connected to.
+
+
+### Connecting a network
+Syntax: `[c] BOUNCER connect <netid>`
+
+Replies:
+* Various errors:
+  - `[s] BOUNCER connect * :Invalid arguments given`
+  - `[s] BOUNCER connect netid ERR_NETNOTFOUND`
+
+The connection state of the network will be received as mentioned in the network state section.
+
+### Disconnecting a network
+Syntax: `[c] BOUNCER disconnect <netid> [:quit message]`
+
+Replies:
+* Various errors:
+  - `[s] BOUNCER disconnect * ERR_INVALIDARGS`
+  - `[s] BOUNCER disconnect netid ERR_NETNOTFOUND`
+
+The connection state of the network will be received as mentioned in the network state section.
+
+
+### Listing networks
+Syntax: `[c] BOUNCER listnetworks [*]`
+
+This command MUST return all networks under the users bouncer account. The optional second parameter of `*` is a wildcard search string that filters the returned networks. It is matched against the `network` tag of each network.
+
+This may be used for rudimentry network organisation, Eg. `work/*` or `social/*` where networks are labelled with `work/somenetwork` and `social/rizon` structures.
+
+The end of the network list MUST be indicated by an `RPL_OK` message, even if there are no networks to list. Eg:
+~~~
+[s] BOUNCER listnetworks netid network=freenode;host=irc.freenode.net;port=6667;state=disconnected;nick=bob;
+[s] BOUNCER listnetworks netid network=IRC.com;host=irc.irc.com;port=6697;state=connected;tls=1;nick=bob;
+[s] BOUNCER listnetworks RPL_OK
+~~~
+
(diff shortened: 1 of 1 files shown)


slingamn:
>Could the name "BOUNCER" for this spec be loading too much towards a traditional IRC bouncer? As with Oragano the lines get muddled between ircd/bouncer so perhaps a better term could be used.

As I understand this spec, it is in fact for conventional bouncers, and therefore Oragono won't implement it. Specifically, in Oragono there's no concept of active buffers that can be added or removed, or of multiple networks. (See #444 for a case where the "buffer" concept has to be adjusted to fit the Oragono setting.)

So, if my understanding is accurate, I have no objection to this spec using the term "BOUNCER". My concern is more that we carefully factor out whatever is shared between the two models. (It sounds like this spec defers all history retrieval to `draft/chathistory`, which seems good. There may be some overlap between `BOUNCER listbuffers` and #444?)

emersion:
pounce said they won't be implementing this spec, because pounce doesn't expose multiple networks this way and manages buffers differently.

This PR is missing a lot of improvements from https://git.sr.ht/~emersion/soju/tree/bouncer-ext/item/doc/ext/bouncer-networks.md (standard replies, batches, more idiomatic names/style, etc).

extensions/bouncer.md
```diff
@@ -0,0 +1,236 @@
...
+## `BOUNCER` command
+This introduces the `BOUNCER` command, followed by a case insensitive subcommand.
+
+All tags sent and received by this command are message-tags encoded.
+
+***Note***: Where a network ID is used to specify a network in a command, a `*` character may be used instead to signify the active network the client is connected to.
+
+
+### Connecting a network
+Syntax: `[c] BOUNCER connect <netid>`
```
emersion, line 53: This should just be a `enabled` network attribute.

extensions/bouncer.md
```diff
@@ -0,0 +1,236 @@
...
+    name: "Darren Whitlen"
+    period: "2018-2021"
+    email: "darren@kiwiirc.com"
+---
+
+## BOUNCER
+
+This feature is intended to provide an API for clients to build a UI around bouncer IRC servers, to manage the users IRC networks and buffers in a user friendly way and to keep multiple clients and devices synced together.
+
+It is based around network and buffer entities which both have key/value tags, such as `joined=true` for a buffer, with commands to manage these. A list of common tags can be found under the tags section of this document.
```
emersion, line 16: s/tags/attributes/

As discussed previously, "tags" is confusing.

extensions/bouncer.md
```diff
@@ -0,0 +1,236 @@
...
+- A `buffer` is either a channel or query
+- A client is a users client or device that connects to the bouncer
+
+A network is identified by a per-user unique ID that MUST not change. A human readable label MUST also be associated with the network that the user can identify it with in a client UI - this is the `network` tag in most commands. The ID and label may be the same as long as the ID does not change. Neither the ID or label may contain a space or `:` characters.
+
+All commands and data for this bouncer feature MUST be associated to a logged-in and authorised user. It is the implementors choice on how the user authentication is handled and is outside the scope of this document.
+
+## Capabilities
+
+This adds a capability called `BOUNCER`. Once negotiated the server may change it's behaviour to better suit bouncer clients, such as not automatically sending `JOIN` commands for already joined channels.
```
emersion, line 35: Caps are lowercase. Additionally, this is missing the "work in progress" section about using the `draft/`-prefixed cap.

extensions/bouncer.md
```diff
@@ -0,0 +1,236 @@
...
+  - `[s] BOUNCER state netid netname connecting`
+  - `[s] BOUNCER state netid netname connected`
+  - `[s] BOUNCER state netid netname disconnected`
+
+## Network and buffer tags
+This document has described several tags that SHOULD be supported. Implementations MAY introduce other tags but MUST contain a vendor prefix, eg. `kiwibnc/mytag=value`.
+
+#### Network tags
+Implementations MUST recognise the following:
+- `network` - The human readable name for the network. It may be changed.
```
emersion, line 208: This reads as "the network attribute of the network". Something like "name" would be better.

extensions/bouncer.md
```diff
@@ -0,0 +1,236 @@
...
+  - `[s] BOUNCER delnetwork * ERR_INVALIDARGS`
+  - `[s] BOUNCER delnetwork netid ERR_NETNOTFOUND`
+  - `[s] BOUNCER delnetwork netid ERR_UNKNOWN :Optional extra information`
+
+
+## Network state notifications
+As a network state changes on the server such as connecting or disconnection, it MUST send the changes as they happen to any users connected clients.
+
+* Connection state changes:
+  - `[s] BOUNCER state netid netname connecting`
```
emersion, line 199: This isn't great IMHO, as it's only useful for network state. Would be better to allow the bouncer to notify updates of arbitrary network/buffer attribute changes.

extensions/bouncer.md
```diff
@@ -0,0 +1,236 @@
...
+- `username` - The username for the active connection to the IRC network or the username to use when connecting.
+- `realname` - The realname for the active connection to the IRC network or the realname to use when connecting.
+
+
+#### Buffer tags
+
+Implementations MUST recognise the following:
+- `network` - The name of the network this buffer belongs to.
+- `buffer` - The name of this buffer.
+- `seen` - The ISO 8601 timestamp of when the user last saw this buffer. Used to determine the point at which the user as read up to.
```
emersion, line 227: This should be out-of-scope for the bouncer extension. Servers not implementing this spec (soju, pounce, oragono) may be interested in this functionality.

extensions/bouncer.md
```diff
@@ -0,0 +1,236 @@
...
+
+This command MUST return all networks under the users bouncer account. The optional second parameter of `*` is a wildcard search string that filters the returned networks. It is matched against the `network` tag of each network.
+
+This may be used for rudimentry network organisation, Eg. `work/*` or `social/*` where networks are labelled with `work/somenetwork` and `social/rizon` structures.
+
+The end of the network list MUST be indicated by an `RPL_OK` message, even if there are no networks to list. Eg:
+~~~
+[s] BOUNCER listnetworks netid network=freenode;host=irc.freenode.net;port=6667;state=disconnected;nick=bob;
+[s] BOUNCER listnetworks netid network=IRC.com;host=irc.irc.com;port=6697;state=connected;tls=1;nick=bob;
+[s] BOUNCER listnetworks RPL_OK
```
emersion, line 84: This seems alien for a modern IRCv3 spec. A batch would be better.

extensions/bouncer.md
```diff
@@ -0,0 +1,236 @@
...
+  - `[s] BOUNCER disconnect * ERR_INVALIDARGS`
+  - `[s] BOUNCER disconnect netid ERR_NETNOTFOUND`
+
+The connection state of the network will be received as mentioned in the network state section.
+
+
+### Listing networks
+Syntax: `[c] BOUNCER listnetworks [*]`
+
+This command MUST return all networks under the users bouncer account. The optional second parameter of `*` is a wildcard search string that filters the returned networks. It is matched against the `network` tag of each network.
```
emersion, line 76: I don't think the wildcard search string adds a lot of value here. What's the use-case? Do you expect the user to have a huge number of networks so that filtering can't be done client-side?

extensions/bouncer.md
```diff
@@ -0,0 +1,236 @@
...
+- `sasl_pass` - The SASL password to log into the IRC network with.
+- `nick` - The nick for the active connection to the IRC network or the nick to use when connecting.
+- `username` - The username for the active connection to the IRC network or the username to use when connecting.
+- `realname` - The realname for the active connection to the IRC network or the realname to use when connecting.
+
+
+#### Buffer tags
+
+Implementations MUST recognise the following:
+- `network` - The name of the network this buffer belongs to.
```
emersion, line 225: This should be a stable netid, not a network name.

jwheare:
I don't think this will be progressed, given the OP has quit IRC dev and other interested parties want to take this in a different direction. So I'll close it, @emersion if you want to reopen your version, please do.

PR #450, opened 2021-04-06
CHATHISTORY TARGETS

slingamn:
This comes from discussion on #444 (cc @kylef, @emersion, @hhirtz). In short:

1. CHATHISTORY needs a way to surface DMs with previously unknown users. The previous proposal for this was the `*` target (which "contained" all DMs sent to or from the user).
2. Client developers don't like the `*` target; it has the problem that high-priority DMs from friends can get blocked behind spam
3. Bouncer developers don't like the `*` target; in a typical bouncer implementation, there is no global time index over all DMs (nor any need for one)

So, this proposal adds an explicit way to list DM correspondents, with pagination support, and removes the `*` target. It's marked as a draft; the intent is that Oragono 2.6.0 will include some version of this API, but will continue to include the `*` target for the time being.

### DIFF
--- a/extensions/chathistory.md
+++ b/extensions/chathistory.md
@@ -93,7 +93,7 @@ Request up to `limit` number of messages between the given `timestamp` or `msgid
 
 #### `TARGETS`
 
-Unlike the other subcommands, `TARGETS` does not return message history. Instead, it lists channels and users with which the user has exchanged direct messages, ordered by the time of the latest message in the channel history or direct message conversation (i.e., sent from or to the other user). This allows the client to discover missed direct message conversations with users it is not currently aware of.
+Unlike the other subcommands, `TARGETS` does not return message history. Instead, it lists channels the user has visible history in, together with users with which the user has exchanged direct messages, ordered by the time of the latest message in the channel history or direct message conversation (i.e. sent from or to the other user). This allows the client to discover missed direct message conversations with users it is not currently aware of.
 
     CHATHISTORY TARGETS <timestamp=YYYY-MM-DDThh:mm:ss.sssZ> <timestamp=YYYY-MM-DDThh:mm:ss.sssZ> <limit>
 
@@ -101,7 +101,7 @@ The parameters have the same semantics as `BETWEEN`, except that they MUST be ti
 
     CHATHISTORY TARGETS <nickname | channel name> <YYYY-MM-DDThh:mm:ss.sssZ>
 
-where the timestamp is the time of the latest message sent from or to the other user.
+where the timestamp is the time of the latest message in the channel history or direct message conversation.
 
 #### Returned message notes
 The order of returned messages within the batch is implementation-defined, but SHOULD be ascending time order or some approximation thereof, regardless of the subcommand used. The `server-time` tag on each message SHOULD be the time at which the message was received by the IRC server. The `msgid` tag that identifies each individual message in a response MUST be the `msgid` tag as originally sent by the IRC server.
--- a/extensions/chathistory.md
+++ b/extensions/chathistory.md
@@ -42,21 +42,20 @@ The `draft/event-playback` capability MAY be negotiated. This allows the client 
 ### `CHATHISTORY` Command
 The client can request message history content by sending the `CHATHISTORY` command to the server. This command has the following general syntax:
 
-    CHATHISTORY <subcommand> <target> <timestamp | msgid> <limit>
+    CHATHISTORY <subcommand> <target> <timestamp | msgid> [<timestamp | msgid>] <limit>
+    CHATHISTORY LISTCORRESPONDENTS <timestamp> <timestamp> <limit>
 
 The `target` parameter specifies a single buffer (channel or nickname) from which history is to be retrieved. If a nickname is given as the `target` then the server SHOULD include history sent between the current user and the target nickname, including outgoing messages ("self messages"). The server SHOULD attempt to include history involving other nicknames if either the current user or the target nickname has changed during the requested timeframe.
 
-The special target `*` refers to all direct messages sent to or from the current user, regardless of the other party. This allows the client to retrieve conversations with users it is not yet aware of.
-
 A `timestamp` parameter MUST have the format `timestamp=YYYY-MM-DDThh:mm:ss.sssZ`, as in the [server-time][server-time] extension. A `msgid` parameter MUST have the format `msgid=foobar`, as in the [message-ids][message-ids] extension.
 
-If the `batch` capability was negotiated, the server MUST reply to a successful `CHATHISTORY` command using a [`batch`][batch]. The batch MUST have type `chathistory` and take a single additional parameter, the canonical name of the target being queried. If no content exists to return, the server SHOULD return an empty batch in order to avoid the client waiting for a reply.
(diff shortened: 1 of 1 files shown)


slingamn:
Summarizing some discussion, the tentative plan is:

1. The subcommand should be `LIST_TARGETS` (`LISTTARGETS`?) and the response subcommand should be `TARGET`
2. Channels should be included in the output

extensions/chathistory.md
```diff
@@ -92,6 +91,18 @@ This is useful for retrieving conversation context around a single message.
 
 Request up to `limit` number of messages between the given `timestamp` or `msgid` values. With respect to the limit, the returned messages MUST be counted starting from and excluding the first message selector, while finishing on and excluding the second. This may be forwards or backwards in time.
 
+#### `TARGETS`
+
+Unlike the other subcommands, `TARGETS` does not return message history. Instead, it lists channels the user has visible history in, together with users with which the user has exchanged direct messages, ordered by the time of the latest message in the channel history or direct message conversation (i.e. sent from or to the other user). This allows the client to discover missed direct message conversations with users it is not currently aware of.
+
+    CHATHISTORY TARGETS <timestamp=YYYY-MM-DDThh:mm:ss.sssZ> <timestamp=YYYY-MM-DDThh:mm:ss.sssZ> <limit>
+
+The parameters have the same semantics as `BETWEEN`, except that they MUST be timestamps and not msgids. Returned messages have the syntax:
```
kylef, line 100: > Returned messages have the syntax:

Have you considered using [standard replies](https://ircv3.net/specs/extensions/standard-replies)?

emersion: Standard replies are only for errors, warnings, and such. They can't really be used for returning data for success replies. See the first sentence of the spec:

> This document specifies the standard FAIL, WARN, and NOTE messages, intended to provide a clean, consistent interface for sending general errors, warnings, and informational messages to clients.

extensions/chathistory.md
```diff
@@ -92,6 +91,18 @@ This is useful for retrieving conversation context around a single message.
 
 Request up to `limit` number of messages between the given `timestamp` or `msgid` values. With respect to the limit, the returned messages MUST be counted starting from and excluding the first message selector, while finishing on and excluding the second. This may be forwards or backwards in time.
 
+#### `TARGETS`
+
+Unlike the other subcommands, `TARGETS` does not return message history. Instead, it lists channels the user has visible history in, together with users with which the user has exchanged direct messages, ordered by the time of the latest message in the channel history or direct message conversation (i.e. sent from or to the other user). This allows the client to discover missed direct message conversations with users it is not currently aware of.
+
+    CHATHISTORY TARGETS <timestamp=YYYY-MM-DDThh:mm:ss.sssZ> <timestamp=YYYY-MM-DDThh:mm:ss.sssZ> <limit>
```
kylef, line 98: Most of the time here the "until" timestamp is the current timestamp, would it make sense to allow the client to send a placeholder so it does not need to figure out the server's current time?

```
CHATHISTORY TARGETS <timestamp=YYYY-MM-DDThh:mm:ss.sssZ> * *
```

extensions/chathistory.md
kylef: > The NOTE message indicates some informational message about a given command/function, or about the current session

The description for `NOTE` appears to be applicable for this case.

emersion: NOTE is for sending an information message. There's no informational message to be sent here, only machine-readable data.

extensions/chathistory.md
emersion: Need to be careful wrt. race conditions.

extensions/chathistory.md
kylef: Other commands returning lists such as LIST (which uses RPL_LIST), WHO which uses RPL_WHOREPLY use numerics.

Standard replies appears to be intended as a replacement for new numerics as far as I understand given this statement in the spec:

> Implementers should not need to reserve new numerics for error, warning, or general informational messages, especially as numerics themselves and the mapping of numerics to names can be unclear or conflicting.

Standard replies are machine readable with specified code and context. I can't think of any other defined command which returns a list with the same command name (with exception of CAP itself, which uses `*`/sender prefix after command to indicate the reply).

This proposal here does not fit into any of the existing norms. I am wondering if we avoid making yet another pattern and instead follow some existing patterns?

extensions/chathistory.md
slingamn: My understanding is that the most common case of this is analogous to `CHATHISTORY LATEST`: list all targets starting from the current time and paginating backwards, possibly until a lower bound is reached (estimated from the time of the last message received on the connection).

Here's what I thought: unlike the other commands, `TARGETS` can only handle timestamps, not message IDs (because the spec is designed for compatibility with systems where there is no global ordering on message IDs). In this context, it makes sense to simplify the specification of the command as much as possible --- a single command, with parameters of a single form.

The intended way to express the current time is as in the pseudocode example: take the local current time and add some constant.

extensions/chathistory.md
slingamn: I think this is mostly addressed by discussion on #357, specifically [this comment](https://github.com/ircv3/ircv3-specifications/pull/357#issuecomment-485479876)

extensions/chathistory.md
emersion: My recommendation is to use the `time` tag from the last received message from the server as the upper bound. That way, if another user decides to send us a PRIVMSG while we're still in the process of fetching chat history, we won't get the message twice.

slingamn:
@jwheare motion to merge this? It now has a server and a client implementation, and we didn't find any spec bugs.

Right now, no one is implementing the `*` target (which is removed by this PR), so it would be nice to update ircv3.net to remove it before someone tries to implement it.

hhirtz:
I'll also implement it in senpai, I just need to finish the work on `soju.im/bouncer-networks` first since soju doesn't support this spec in multi-upstream mode (one connection to the bouncer to show all upstream channels/users)

jwheare:
Sounds good. I'm guessing no one thinks a version bump is necessary? Any other backwards incompatible changes other than removing the * target?

slingamn:
That should be the only backwards-incompatible change. (This change also improves compatibility with the published [chathistory batch type spec](https://ircv3.net/specs/batches/chathistory), since there is no longer any ambiguity about the value of the BATCH parameter.)

+1 for merging with no version bump.

extensions/chathistory.md
kylef: > My recommendation is to use the time tag from the last received message from the server as the upper bound (e.g. from RPL_WELCOME).

The welcome can be replayed with older timestamps than the current connection. For example with the ZNC bouncer, the time attached to 001/005/MOTDs numerics can be the time in which the ZNC server received the message from its upstream server. This can be days, months, years before the current server time. I don't see a way I can trust any time message from connection handshake as they are sometimes replayed in parts, and cannot be treated as "current" server time.

For the "until time" in this command, I don't see as a client we'd have anything to put here other than making up a date in the servers future time which isn't anymore useful than the server using its own time upon receiving the message (because at this point its the only part of this that knows the server's time).

emersion: > For example with the ZNC bouncer, the time attached to 001/005/MOTDs numerics can be the time in which the ZNC server received the message from its upstream server.

Sounds like a ZNC bug to me.

emersion:
I'm fine with merging without a version bump, even if it's a backwards-incompatible change.

PR #451, opened 2021-04-30
Drop 3.2/3.3 suffix in links, make sure .html suffix is consistently used

jwheare:
Avoids unnecessary redirects by using canonical URLs


DanielOaks:
Threw some small fixes on top (that apparently I forgot to fix when we shook up the folder structure). Looks good, thanks for this!

PR #454, opened 2021-05-28
Move client-initiated BATCH from 'multiline' to the new 'client-batch' spec

progval:


SadieCat:
Some things I'd like to see in this:

- A way for the server to tell clients the max nesting depth of client batches (if any).
- A way for the server to tell clients the max timeout they can have on batches (if any).

jwheare:
Batch nesting isn't really a thing at this point, but this could be amended if a future spec needs it.

A set of generic standard replies for things like timeout exceeded might make sense, but I'm not sure if there's anything meaningful a client can do with knowing a timeout value up front.

SadieCat:
Fair enough. I'd be happy with a generic "you timed out" error that servers can send.

extensions/client-batch.md
```diff
@@ -0,0 +1,49 @@
...
+Clients SHOULD NOT send `BATCH` commands to servers, unless they negotiated
+a capability that allows them to do so.
+
+## The client-to-server `BATCH` verb
+
+This specification introduces client initiated batches.
+These use the same syntax as server initiated batches.
+
+Once a client has opened a batch, it MUST NOT send any messages
+that are not part of the batch, i.e. it may only send PRIVMSG or NOTICE
```
slingamn, line 44: This "i.e." is specific to `draft/multiline` and not applicable to all client-to-server BATCH uses, right?

extensions/client-batch.md
```diff
@@ -0,0 +1,48 @@
...
+Clients SHOULD NOT send `BATCH` commands to servers, unless they negotiated
+a capability that allows them to do so.
+
+## The client-to-server `BATCH` verb
+
+This specification introduces client initiated batches.
+These use the same syntax as server initiated batches.
+
+Once a client has opened a batch, it MUST NOT send any messages
+that are not part of the batch.
```
jwheare, line 44: .. until it has been closed.

extensions/client-batch.md
```diff
@@ -28,14 +28,23 @@ batch type, but is designed as a framework for other specifications.
 Certain specifications call for clients to be able to send batched messages to
 servers, for instance [`draft/multiline`][]
 
-## Capabilities
+## Architecture
+
+### Dependencies
+
+This specification does not depends on the [`batch`][] capability,
+as it is only used for server-initiated batches.
```
jwheare, line 36: does not depend *[no s]* on the batch capability, which is used for server-initiated batches only.

extensions/client-batch.md
```diff
@@ -47,6 +56,15 @@ that are not part of the batch, until it is closed
 Clients MUST NOT send nested batches, i.e. clients may not send `BATCH`
 commands with a `batch` tag.
 
+### Errors
+
+Servers MUST use `FAIL` messages from the [standard replies][] framework
+to notify clients of errors with multiline batches.
```
jwheare, line 62: s/multiline/client-initiated

extensions/client-batch.md
```diff
@@ -0,0 +1,80 @@
...
+commands with a `batch` tag.
+
+### Errors
+
+Servers MUST use `FAIL` messages from the [standard replies][] framework
+to notify clients of errors with client-initiated batches.
+The command is `BATCH` and the following codes are defined:
+
+* `TIMEOUT`: the batch was left open for too long, all future messages
+  in this batch will be ignored.
```
DarthGandalf, line 67: Starting from which moment? As worded now, the client has no way of knowing which messages are ignored

ProgVal: True. Should all messages of the batch be ignored?

DarthGandalf: That's an option. However, it creates another corner case: server sends `FAIL TIMEOUT` just after client closes the batch. Now in the case of success client can't know if the that was a success; it will learn of a failure though.

ProgVal: Clients should get something echoed back to them (eg. with `echo-message`) if the batch was successful. And I can add the batch id as argument of `FAIL TIMEOUT` to disambiguate the failure case

DarthGandalf: Only for messages which do provide some feedback. But since it's a generic framework, the messages could be anything

ProgVal: Do we currently have any type of message that doesn't provide feedback? If yes, they should probably be "fixed" outside this spec, because feedback is also important outside a batch.

PR #455, opened 2021-06-01
typing: Clarify when typing status is cleared

ProgVal:
The old description was ambiguous, so this PR replaces it with the
most logical interpretation.

### DIFF
--- a/client-tags/typing.md
+++ b/client-tags/typing.md
@@ -39,7 +39,7 @@ A typing notification is represented by a [`TAGMSG`][tags] command sent with a `
 
 Input event handlers MUST be throttled so that any `typing` notification is not sent within 3 seconds of another one for a given target.
 
-After receiving a typing notification, clients SHOULD assume the sender is still typing until any of the following conditions is met:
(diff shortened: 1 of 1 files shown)


client-tags/typing.md
```diff
@@ -39,9 +39,9 @@ A typing notification is represented by a [`TAGMSG`][tags] command sent with a `
 
 Input event handlers MUST be throttled so that any `typing` notification is not sent within 3 seconds of another one for a given target.
 
-After receiving a typing notification, clients SHOULD assume the sender is still typing until any of the following conditions is met:
+After receiving a typing notification, clients SHOULD assume the sender is still typing until any of the following conditions is met on the same target (channel or private message):
 
-* A message is received from the sender
+* A `PRIVMSG` or `NOTICE` is received from the sender
```
jwheare, line 44: Don't think this needs to be changed. "message" is already defined further up as:
> For the purpose of this specification, a message is defined as a PRIVMSG, NOTICE, TAGMSG, or batch end.

ProgVal: Ok. Feels weird to include TAGMSG though...

PR #461, opened 2021-07-01
bot-mode: Specify RPL_WHOISBOT

progval:


ProgVal:
(please don't merge, I'm investigating whether "Remove <target> from the format, to match InspIRCd" should be reverted)

PR #462, opened 2021-07-03
chathistory: Add an ISUPPORT token to indicate what targets are not valid

ProgVal:
This allows clients to skip some queries that would FAIL anyway.

eg. this can be useful on UnrealIRCd, as it does not allow
CHATHISTORY with private targets.

### DIFF
--- a/extensions/chathistory.md
+++ b/extensions/chathistory.md
@@ -118,6 +118,21 @@ If no message history can be returned due to an error, the `MESSAGE_ERROR` error
 
     FAIL CHATHISTORY MESSAGE_ERROR the_given_command the_given_target [extra_context] :Messages could not be retrieved
 
+### RPL_ISUPPORT Token
+
+This specification defines the `CHATHISTORY_TARGETS` token for use in `RPL_ISUPPORT` (005) responses.
+
+Servers SHOULD use this token to communicate to clients restrictions on what targets are valid.
+
+The `CHATHISTORY_TARGETS` token value is a (possibly empty) comma `,` (0x2C) separated list of the following values:
+
+* `no-channels`, if `CHATHISTORY` cannot be used with channel targets
+* `no-nick`, if `CHATHISTORY` cannot be used with nickname/private targets
+
+Clients MUST ignore any unknown value. Servers SHOULD not send any value other than the ones defined above.
+
+The token may be missing if the server implements an earlier version of this draft.
+
 ### Examples
 
 Requesting the latest conversation upon joining a channel


SadieCat:
Isn't this covered by CHATHISTORY TARGETS? #450

slingamn:
I don't think all cases for this are covered by `TARGETS`. Consider a client that doesn't keep local logs, just the timestamp of the last received message. Such a client might request `CHATHISTORY TARGETS` on a new connection and see no DM targets, because there were no DMs since the last received message. Later on, the user might open a query (either manually or in response to a DM) and then request scrollback for that conversation with `CHATHISTORY LATEST` or `CHATHISTORY BEFORE`.

I'm undecided at this time on the need for this.

syzop:
My initial reaction was purely cosmetic... I don't think there are 005 tokens with underscores, it's just always concating(sp?) names without that, like ```USERLEN``` in case of user length (2 words), ```MAXNICKLEN``` in case of maximum nick length (3 words), etc. 

Anyway, wouldn't it be better to modify the existing ```CHATHISTORY``` token? I mean, right now what you propose is about some possibly "missing" functionality that "needs" to be communicated (and I understand why it could possible be useful, it saves some client traffic at least). But who knows in the future you would want some kind of "feature" in a later CHATHISTORY spec or update being communicated? So it makes sense to make that token extensible in some way. And since this is all about the CHATHISTORY command.... I think it belongs there. Since all this is relatively fresh and in progres I think it's perfectly fine to update it now still.

As for adding any 005's in general... it's pretty much always fine for me if it benefits client coders. And even if you ignore all what I said above, that is fine too, I'm quite relax with implementing any 005's :D.

SadieCat:
> I don't think all cases for this are covered by `TARGETS`. Consider a client that doesn't keep local logs, just the timestamp of the last received message. Such a client might request `CHATHISTORY TARGETS` on a new connection and see no DM targets, because there were no DMs since the last received message. Later on, the user might open a query (either manually or in response to a DM) and then request scrollback for that conversation with `CHATHISTORY LATEST` or `CHATHISTORY BEFORE`.
> 
> I'm undecided at this time on the need for this.

Wouldn't this PR not be enough for that use case? Not all channels or users are going to want to have server logging enabled so the client will need to know which individual targets don't have logs available.

emersion:
With my client developer hat on, I'll probably not implement this hint. I don't really see a benefit, I'm fine with having my CHATHISTORY requests failing.

ProgVal:
Alright, it looks like it isn't as useful as I expected :p

(But feel free to resurrect this later if you find a use-case)

PR #463, opened 2021-07-07
Add certfp and spkifp options to WEBIRC

jobe1986:
Add option parameters to WEBIRC specification to allow forwarding of tls client certificate fingerprints through WebIRC gateways.

### DIFF
--- a/extensions/webirc.md
+++ b/extensions/webirc.md
@@ -45,6 +45,8 @@ These options are defined and may be sent by clients while connecting:
 - `secure`: This flag indicates that the client has a TLS-secured connection to the gateway. Servers MUST ONLY treat the connection as secure if this flag is sent and the connection from the gateway to the server is also secure with TLS.
 - `remote-port=<port>`: This flag indicates the remote port the client has connected to the gateway from.
 - `local-port=<port>`: This flag indicates the port the gateway accepted the client connection on (e.g. `6697`, `6667`).
+- `certfp=<fingerprint>`: This flag indicates the tls client certificate fingerprint supplied to the WebIRC gateway by the user's actual client application.
+- `spkifp=<fingerprint>`: This flag indicates the public key fingerprint for the tls client certificate supplied to the WebIRC gateway by the user's actual client application.
 
 Servers MUST be able to handle options that don't currently have defined values gaining values in the future. For example, they MUST treat the options `secure` and `secure=examplevalue123` in exactly the same way.
 
--- a/extensions/webirc.md
+++ b/extensions/webirc.md
@@ -50,6 +50,8 @@ These options are defined and may be sent by clients while connecting:
 
 `<algo>` should be the hash algorithm used to produce the fingerprint supplied such as sha256.
 
+`<fingerprint>` should be the hash of the certificate (certfp) or public key (spkifp) encoded in hexadecimal without colon (:) sepertators. Servers should accept either upper case or lower case hexadecimal characters.
+
 Servers MUST be able to handle options that don't currently have defined values gaining values in the future. For example, they MUST treat the options `secure` and `secure=examplevalue123` in exactly the same way.
 
 ### Examples
--- a/extensions/webirc.md
+++ b/extensions/webirc.md
@@ -45,8 +45,10 @@ These options are defined and may be sent by clients while connecting:
 - `secure`: This flag indicates that the client has a TLS-secured connection to the gateway. Servers MUST ONLY treat the connection as secure if this flag is sent and the connection from the gateway to the server is also secure with TLS.
 - `remote-port=<port>`: This flag indicates the remote port the client has connected to the gateway from.
 - `local-port=<port>`: This flag indicates the port the gateway accepted the client connection on (e.g. `6697`, `6667`).
-- `certfp=<fingerprint>`: This flag indicates the tls client certificate fingerprint supplied to the WebIRC gateway by the user's actual client application.
-- `spkifp=<fingerprint>`: This flag indicates the public key fingerprint for the tls client certificate supplied to the WebIRC gateway by the user's actual client application.
+- `certfp-<algo>=<fingerprint>`: This flag indicates the tls client certificate fingerprint supplied to the WebIRC gateway by the user's actual client application.
+- `spkifp-<algo>=<fingerprint>`: This flag indicates the public key fingerprint for the tls client certificate supplied to the WebIRC gateway by the user's actual client application.
+
+`<algo>` should be the hash algorithm used to produce the fingerprint supplied such as sha256.
 
 Servers MUST be able to handle options that don't currently have defined values gaining values in the future. For example, they MUST treat the options `secure` and `secure=examplevalue123` in exactly the same way.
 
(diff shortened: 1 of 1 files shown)


SadieCat:
+1. Working on an implementation of this for InspIRCd.

grawity:
Should this describe the hash format somewhat, i.e. is it hexadecimal or base64, or does that not matter? I'm thinking about networks switching to a different webclient, or users using the same cert between web and direct connection.

(I'd prefer base64 personally, due to length, but I guess it has to be hex for compatibility with what ircds do themselves...)

SadieCat:
Basically every IRCd uses hexadecimal with no separating : but it wouldn't hurt to define it yeah.

jobe1986:
Any suggestions on how to specify the format? Perhaps a prefix to the fingerprint like "base64;" or "b64;" if not hex?

jobe1986:
Perhaps it would be better to not include the algorithm in the option name, and instead create a separate IRCv3 specification documenting a format for cert fingerprints that includes the ability to include the hash algorithm and encoding in the format of the fingerprint itself.

syzop:
Good idea.

I'm glad the hash algorithm is mentioned in the keyname (eg ```certfp-sha256```).  I mean, we use sha256 now, but that won't be forever. Similarly, there could be a transition point some day where we have both sha256 and Algorithm X in use for a few years, where both need to be communicated. Your spec allows all that. Great :)

I agree with @SadieCat to make ```certfp-sha256``` use hexadecimals without using the colon sign. I think we can just define it as that and not allow different different encodings (eg base64) for ```certfp-sha256```. No need to make the spec more complex in that respect.

As for whether to include the algorithms (and how they should be encoded) in the main specification itself or elsewhere. I don't mind either way. My gut feeling says a separate spec (or registry) sounds most logical, though. Avoids the clutter/talk about hash algorithms and encoding in the main WEBIRC spec.. and allows more room for examples and references on how to calculate them.

jwheare:
@jobe1986 have you got enough to go on in the comments here to make any of the modifications discussed?

jobe1986:
I'm still not sure whether it should be "certfp-<algo>=<fingerprint>" or "certfp-<algo:<fingerprint>".

Personally I prefer <algo> in the key name to prevent duplicate keys with different values for situations where you may want to provide the fingerprint in more then one algorithm.

jwheare:
@SadieCat @syzop @grawity pls confirm you're happy to see this merged.

grawity:
Looks good to me, although it might be better to be more explicit about algorithm names so that we don't end up with `certfp-sha256` and `certfp-SHA2-256`.

I think it's better to keep the algorithm name as part of the key name. (Although it would also be possible to avoid the duplicate key problem using `certfp=sha1:foo,sha256:bar,sha3:baz` but that seems unnecessarily complex and possibly risky, since you'd have to parse the value, while `certfp-sha1=foo certfp-sha256=bar` achieves the same thing using existing code.)

syzop:
```certfp-sha256=abc``` sounds good.

I do have a comment on the latest edit. Hexadecimal for certfp is normal and we already concluded is used by pretty much all IRCds, so that is good. However, for spki fingerprint using base64 is normal, see [RFC7469](https://datatracker.ietf.org/doc/html/rfc7469) and ctrl+f base64. I'm not aware of any implementation doing that hexadecimal. You can google and see the difference if you search on "spki fingerprint" "base64" versus "spki fingerprint "hexadecimal". The hits for both are low, but the base64 one has 10 times more hits than hexadecimal, and if you click on the hits for hexadecimal they are actually not about hexadecimal in the spkifp end result but about something else (unrelated).

Sorry for not mentioning that earlier, I only noticed it now that you put that text in.

I also agree with @grawity about this (which Jobe already raised as well):
> it might be better to be more explicit about algorithm names so that we don't end up with certfp-sha256 and certfp-SHA2-256.

If not going with that registry, which I can totally understand too, then maybe just illustrate with the most common ones. So: sha1, sha256, and perhaps sha3-256. Almost everyone will use one (or more) of those hashes for the next few years.

grawity:
> I do have a comment on the latest edit. Hexadecimal for certfp is normal and we already concluded is used by pretty much all IRCds, so that is good. However, for spki fingerprint using base64 is normal, see [RFC7469](https://datatracker.ietf.org/doc/html/rfc7469) and ctrl+f base64. I'm not aware of any implementation doing that hexadecimal. You can google and see the difference if you search on "spki fingerprint" "base64" versus "spki fingerprint "hexadecimal". The hits for both are low, but the base64 one has 10 times more hits than hexadecimal, and if you click on the hits for hexadecimal they are actually not about hexadecimal in the spkifp end result but about something else (unrelated).

Hmm to be honest I was kind of wishing that this would be just a transfer encoding, not something that gets directly exposed to users, and the ircd would internally keep and compare fingerprints as binary blobs instead of strings...

SadieCat:
InspIRCd's SPKI implementation will use hex but as long as it's in a format we can convert from I'm not bothered if it's hex or base64.

jobe1986:
> > I do have a comment on the latest edit. Hexadecimal for certfp is normal and we already concluded is used by pretty much all IRCds, so that is good. However, for spki fingerprint using base64 is normal, see [RFC7469](https://datatracker.ietf.org/doc/html/rfc7469) and ctrl+f base64. I'm not aware of any implementation doing that hexadecimal. You can google and see the difference if you search on "spki fingerprint" "base64" versus "spki fingerprint "hexadecimal". The hits for both are low, but the base64 one has 10 times more hits than hexadecimal, and if you click on the hits for hexadecimal they are actually not about hexadecimal in the spkifp end result but about something else (unrelated).
> 
> Hmm to be honest I was kind of wishing that this would be just a transfer encoding, not something that gets directly exposed to users, and the ircd would internally keep and compare fingerprints as binary blobs instead of strings...

Something like that was my thinking as the purpose of the WEBIRC command is to send technical information to the IRCd that the IRCd can then decide what it wants to do with. As for the Base64 SPKI thing, I have seen a solanum  network using spki fingerprints encoded in hex.

There is no reason why an IRCd cannot decode the hex into a binary blob and then re-encode into base64 if it so wishes. In fact writing a simple routine to decode hex is far simpler then doing the same for base64. So in my mind this has always been about being a transfer encoding.

syzop:
Ah okay, didn't know spki fp in hexadecimal notation existed. Then it doesn't matter, so yeah fine to use hexadecimal to keep both cert and spkifp consistent.

@jwheare Yup, I'm fine with merging

jwheare:
OK thanks, I can't work out if any more changes were requested re `certfp-<algo>=<fingerprint>` Sounds like the format has agreement but there was a call for more specific examples?

jobe1986:
Hopefully that last commit should cover any remaining concerns, unless anyone has anything else to say about it.

jwheare:
LGTM, will allow for final comments and merge tomorrow.

SadieCat:
This is now implemented in InspIRCd and is available on the testnet. We currently only have gateway configs for KiwiIRC.com enabled but if anyone else wants to add their gateway for testing then they can poke me in #inspircd.dev on ChatSpike and I can add it.

PR #464, opened 2021-07-08
Add draft account-extban spec

causal-agent:
I asked about this on IRC and people seemed in agreement that such an ISUPPORT token should exist.

### DIFF
--- a/extensions/account-extban.md
+++ b/extensions/account-extban.md
@@ -1,0 +1,48 @@
+---
+title: Account Extended Ban
+layout: spec
+work-in-progress: true
+copyrights:
+  -
+    name: "C. McEnroe"
+    email: "june@causal.agency"
+    period: "2021"
+---
+
+
+## Notes for implementing work-in-progress version
+
+This is a work-in-progress specification.
+
+
+## Introduction
+
+The [account-notify](account-notify.html), [account-tag](account-tag.html) and [extended-join](extended-join.html) specifications allow clients to much more easily know the account names of logged in users.
+The [EXTBAN][] ISUPPORT token indicates the types of extended ban masks that the server supports, but not their meanings.
+This specification allows clients to construct ban masks for accounts by indicating which extended ban type matches account names.
+
+[EXTBAN]: https://modern.ircdocs.horse/#extban-parameter
+
+
+## The `ACCOUNTEXTBAN` ISUPPORT token
+
+Servers publishing the `ACCOUNTEXTBAN` [ISUPPORT][] token allow clients to construct ban masks matching account names.
+The value of the `ACCOUNTEXTBAN` token is the extended ban type character which matches account names.
+Servers publishing the `ACCOUNTEXTBAN` token MUST also publish the `EXTBAN` token,
+and the value of the `ACCOUNTEXTBAN` token MUST appear in the value of the `EXTBAN` token.
+
+[ISUPPORT]: https://modern.ircdocs.horse/#feature-advertisement
+
+
+## Examples
+
+Example with the [account-tag](account-tag.html) capability enabled:
+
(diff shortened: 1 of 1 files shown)


jwheare:
I think this is great. I'd like to get some indication that this is something that would be implemented by server devs. Ideally Insp, Unreal, Solanum, ergo. Can't remember if there was interest when this was first being discussed.

progval:
I also think this is great, as extbans currently can't be implemented by any form of automation (bots/scripts or GUIs) without scrapping the server name from the version in RPL_MYINFO.

Two issues that may need to be addressed first, though:

* Unreal does not allow any mask, only complete names and two special values (`0` and `*`) https://www.unrealircd.org/docs/Extended_bans#Group_3:_selectors
* both Unreal and Insp are planning to introduce longer forms for their extban names (a small string instead of a single char), so they would not be able to fit both in this token. Let's make it comma-separated?

emersion:
Ref https://bugs.unrealircd.org/view.php?id=5991 for the new (inconsistent) extban names.

progval:
Yeah, that's reasonable

emersion:
@progval Would this be enough to address your concern?

"Servers MUST support matching exact account names. Servers MAY support matching account masks and other server-specific extensions."

I don't think we need to mention the longer forms of extbans.

progval:
> @progval Would this be enough to address your concern?
> 
> "Servers MUST support matching exact account names. Servers MAY support matching account masks and other server-specific extensions."

Sure.

> I don't think we need to mention the longer forms of extbans.

I'd rather we did. Unreal 6 will transform short extbans to long extbans, so clients need to know the long form in order to display it.

emersion:
> I'd rather we did. Unreal 6 will transform short extbans to long extbans, so clients need to know the long form in order to display it.

Hm. This is unfortunate. Would it be fine for these servers to only advertise the longer form, and not the short form? This would avoid having to advertise multiple account extban tokens.

progval:
But Insp 4 won't convert whatever the user wrote, so clients may see either

progval:
@causal-agent Could you make this a comma-separated list of names? eg. `ACCOUNTEXTBAN=a,account` for Unreal >=6 and `ACCOUNTEXTBAN=R,account` for Insp >= 4, and `ACCOUNTEXTBAN=a` for Solanum

causal-agent:
> @causal-agent Could you make this a comma-separated list of names? eg. `ACCOUNTEXTBAN=a,account` for Unreal >=6 and `ACCOUNTEXTBAN=R,account` for Insp >= 4, and `ACCOUNTEXTBAN=a` for Solanum

What should I do with this sentence:

> Servers publishing the `ACCOUNTEXTBAN` token MUST also publish the `EXTBAN` token,
> and the value of the `ACCOUNTEXTBAN` token MUST appear in the value of the `EXTBAN` token.

Is there an equivalent for long extban names? Should I just remove the second requirement?

progval:
There is no equivalent. Yes, I think you should remove it; `EXTBAN` is completely useless anyway, because it does not say anything about the semantics.

progval:
Implemented by Limnoria in a feature branch: https://github.com/progval/Limnoria/compare/testing...account-ban

progval:
Let's merge this?

vanosg:
So happy this is included!

Reviewing this draft, is there a limit to how many tokens are allowed in an ACCOUNTEXBAN line? I don't see a limit explicitly called out, and I know common sense suggests a server will only publish at most one short version (a) and one long version (account), but are multiples of each (ACCOUNTBAN=a,b,account,baccount) allowed as all valid prefixes?

Question 2, I think this is also implied but should I be able to do ACCOUNTBAN=account (no short version) and be in line with this spec? My interpretation of the spec as written now says yes, but just want to bring it up to ensure it was considered and intentional. 

Thanks so much for the time and effort put into this! It will make life so much easier for clientside folks

PR #465, opened 2021-07-09
Params should not be required

realJoshByrnes:
I know it's somewhat trivial, and `<params>` (erroneously) is required in RFC1459, but I believe it shouldn't be.

RFC 1459 defines params as:
```ebnf
<params>   ::= <SPACE> [ ':' <trailing> | <middle> <params> ]
```

Which would meant at minimum, we would need to format a message like this
```ebnf
<command> <SPACE> <crlf>
```

I know that IRC doesn't require a space between `<command>` and `<crlf>` where there are no middle or trailing parameters, but that's what seems to be indicated.

It's also very possible I could be misreading this - I'm certainly not a Backus–Naur form expert.

### DIFF
--- a/extensions/message-tags.md
+++ b/extensions/message-tags.md
@@ -39,7 +39,7 @@ Tagged messages can be sent by both servers and clients. The usage of individual
 
 The message pseudo-BNF, as defined in [RFC 1459, section 2.3.1][rfc1459] is extended as follows:
 
-    <message>       ::= ['@' <tags> <SPACE>] [':' <prefix> <SPACE> ] <command> <params> <crlf>
+    <message>       ::= ['@' <tags> <SPACE>] [':' <prefix> <SPACE> ] <command> [params] <crlf>
     <tags>          ::= <tag> [';' <tag>]*
     <tag>           ::= <key> ['=' <escaped_value>]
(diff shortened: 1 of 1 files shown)


ProgVal:
IMO that's a mistake in the RFC and shouldn't be "fixed" here. FWIW, the Modern spec has fixed this bug: https://modern.ircdocs.horse/#wire-format-in-abnf

realJoshByrnes:
> IMO that's a mistake in the RFC and shouldn't be "fixed" here. 

If it's redefining '<message>' it's certainly the place to put it.

> EDIT: alternative idea (possibly bad): use this opportunity to redefine `<params>` here to make this BNF more self-contained.

I did consider this, but without changing the `<message>` it'd still appear that '<Params>` is required, when it is in fact optional.

The `<SPACE>` should be required in `<params>` and would match the rest of the document, I'm all for being more inclusive in this document though.

jwheare:
I think this is probably fine. I agree that this doesn't "fix" the base RFCs, but probably best not to propagate the error. @progval any further thoughts? Do you still think this is a bad idea?

progval:
Nah, it's better than to keep it the way it is. I'll see if I can come up with a better idea later, but let's merge this PR asap

Issue #79, opened 2014-03-23
Away and account-notify integration with MONITOR


TheSaunders:
Adding away-notify and account-notify support to MONITOR. Users will get the 'updates' from away and account notify with MONITOR if they have those capabilities enabled.

The client should send `MONITOR + <nick>` when the user PMs another person, NOT when the user recieves a PM. After so much inactivitiy the client should auto-remove the MONITOR

Normal usage:

```
:nick!user@host PRIVMSG nick2 :initial pm
MONITOR + nick2
:nick2!user@host ACCOUNT nick2
:nick2!user@host AWAY :I'm not currently here
```

and then after so much inactivity, the client can send `MONITOR - nick2`

In the case of the current MONITOR list being full, clients would need to respect the `MONITOR=XXX` in the 005 numeric and remove a user before sending another, eg:

```
:nick!user@host PRIVMSG nick2 :initial pm
MONITOR - oldnick
MONITOR + nick2
:nick2!user@host ACCOUNT nick2
:nick2!user@host AWAY :I'm not currently here
```

Only clients that support `away-notify` or `account-notify` should be getting the AWAY/ACCOUNT messages, the rest would get the MONITOR-specific stuff.

Clients should maintain two lists for MONITOR, one that users specify in the clients that should be on the MONITOR list 100% of the time, and another that is used for the query-specific stuff.

If/When clients have to remove old nicknames they should never remove nicks from the list of nicks that users want to be on the list 100% of the time, only from the second list.

---

I made this topic to track the status of the IRC discussion, I think it all the non-metadata specific stuff in it, because quite frankly I don't understand the metadata stuff yet. I will update it with new information when/if new information is given.

PR #466, opened 2021-07-22
Add extended-monitor spec

emersion:
Closes: https://github.com/ircv3/ircv3-specifications/issues/79

### DIFF
--- a/extensions/extended-monitor.md
+++ b/extensions/extended-monitor.md
@@ -1,0 +1,53 @@
+---
+title: "`extended-monitor` Extension"
+layout: spec
+work-in-progress: true
+extends:
+  - monitor
+  - away-notify
+  - account-notify
+  - chghost
+  - setname
+copyrights:
+  -
+    name: "Simon Ser"
+    email: "contact@emersion.fr"
+    period: "2021"
+---
+
+
+## Notes for implementing work-in-progress version
+
+This is a work-in-progress specification.
+
+Software implementing this work-in-progress specification MUST NOT use the
+unprefixed `extended-monitor` capability name. Instead, implementations SHOULD
+use the `draft/extended-monitor` capability name to be interoperable with other
+software implementing a compatible work-in-progress version.
+
+The final version of the specification will use an unprefixed capability name.
+
+
+## Description
+
+With the help of extensions such as [`away-notify`](away-notify.html),
+[`account-notify`](account-notify.html), [`chghost`](chghost.html) and
+[`setname`](setname.html), clients are able to keep other users' metadata
+up-to-date with their local state when they share a channel. However, clients
+are not able to do so when they don't share a channel.
+
+The [`monitor`](monitor.html) extension allows clients to track when another
+user goes offline or comes online. This specification extends MONITOR to also
+include AWAY, ACCOUNT, CHGHOST and SETNAME notifications.
+
+The `extended-monitor` capability advertises that the server supports sending
+such extended notifcations for monitored nicks. When enabled by the client:
+
+- If `away-notify` is also enabled, the client will get AWAY notifications for
+  monitored nicks.
+- If `account-notify` is also enabled, the client will get ACCOUNT
+  notifications for monitored nicks.
+- If `chghost` is also enabled, the client will get CHGHOST notifications for
+  monitored nicks.
+- If `setname` is also enabled, the client will get SETNAME notifications for
+  monitored nicks.
--- a/extensions/extended-monitor.md
+++ b/extensions/extended-monitor.md
@@ -51,3 +51,23 @@ such extended notifcations for monitored nicks. When enabled by the client:
   monitored nicks.
 - If `setname` is also enabled, the client will get SETNAME notifications for
   monitored nicks.
+
+## Privacy considerations
+
+This extension allows users to monitor personal information of other users.
+Since the IRC protocol doesn't provide any way to atomically change this
+personal information (nick, host, realname all at the same time), it may be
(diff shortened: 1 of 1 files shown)


ProgVal:
Would it make sense to add `NICK` changes too?

emersion:
MONITOR explicitly avoids sending NICK updates:

> The MONITOR implementation also enhances user privacy by disallowing subscription to hostmasks, allowing users to avoid nick-change stalking.

ProgVal:
`account-notify` and `chghost` may have similar issues then; as they are likely to contain the new nickname.

SadieCat:
If someone changes nickname it is typical behaviour to mark them as offline. If a user is marked as offline then they shouldn't receive anything enabled by this cap.

delthas:
> account-notify and chghost may have similar issues then; as they are likely to contain the new nickname.
CHGHOST and SETNAME contain the current nick. If the nick is changed, then we won't receive any more CHGHOST or SETNAME messages that would contain the new nick (because MONITOR wouldn't track the new nick). Or am I missing something?

ProgVal:
Not if they change host and name before changing nick.

emersion:
>      Do we want to keep these semantics for the specs this adds?

I don't think so, since none of the specs (away-notify, account-notify, chghost, setname) specify that the server must send the initial state on JOIN. Instead, clients are expected to query the initial state with WHO/WHOX.

SadieCat:
>Do we want to keep these semantics for the specs this adds?

I can see how this would be useful for clients. It would allow you to monitor someone when opening a query window (something some clients already do) and immediately get their relevant info.

emersion:
Same applies to JOIN messages, without MONITOR. When another user joins a channel, clients want to immediately get relevant info for that user, to e.g. display the username differently in the list if it's away, among other things.

SadieCat:
JOIN is slightly different imo as:

1. It's not feasible to send a ton of messages to users on join because of flood queues. Requiring clients that want that information to use WHO scales better.
2. Immediately having the account name, away message, etc is of less interest for channels than private messages.

emersion:
I see. So basically sending the initial metadata is not technically reasonable for JOIN, but may fine be for MONITOR.

IMHO, I'd still prefer not to do it: rhe cost of the inconsistency between the specs outweights the minor annoyance of having to send WHO/WHOX.

k4bek4be:
Including account names is not strictly necessary in the case of private conversation start, because it's already communicated in a message tag. It won't hurt though, and maybe there are clients that do not want to implement tags.

delthas:
Thinking about this again, a modern client that wants to track user state would want the user away status, but also their realname. In which case we'd have to do a WHO anyways. So sending the AWAY state on MONITOR + wouldn't be so useful b/c we'd send out a WHO right after anyways. @SadieCat thoughts?

SadieCat:
@delthas WHO doesn't include the away message.

delthas:
From the implementation of extended-monitor in the only 3 popular, maintained server that support MONITOR (see linked MRs), it follows that implementing extended-monitor can be relatively easy and non intrusive, if we don't send the initial state (AWAY, ACCOUNT, CHGHOST, SETNAME) on MONITOR +. Otherwise we'd have to regenerate these messages in the server somehow from the MONITOR + handler, but this would need intrusive changes to the servers. Instead, adding a function for sending a message to all monitors of a nick and calling that function when a user away, account, chghost, setname is sent is much simpler and works well.

Modern clients that track the user state will want the user away status, message, realname, and such. Today these clients already send a WHOIS and have handlers for the various WHOIS replies. If we wanted to replace that with all state being pushed automatically, these clients would have to add support for chghost, and a handler for chghost, in order to receive the user real name. I expect that clients would find it easier to just keep sending WHOIS as they did before.

So TL;DR:
- much easier for servers to implement extended-monitor if we keep it simple, not sending the initial away/account/chghost/setname state
- clients will still send a WHOIS when first getting the state, but it's fine, they already did it. otherwise they'd have to support chghost to get the realname. even then, it's a small point compared to server implementations.

emersion:
Ping: any more feedback on this spec?

progval:
I don't think this was addressed: https://github.com/ircv3/ircv3-specifications/pull/466#issuecomment-885005724

emersion:
I don't understand, can you rephrase? Is your concern that there is an information leak when the user changes their host or realname before changing their nick?

If that's your concern, do you want to add a paragraph to recommend clients to change their nick before their host or realname?

progval:
If they change host and realname after changing nick, then it will show up for people monitoring the new nick (although that's still likely).

It should at least be mentioned that the only way to avoid leaks is to change the info while reconnecting

emersion:
But if someone monitors both the old and new nicks, reconnecting won't help either? Once the target user re-connects with the new nick, the monitoring user will receive `RPL_MONONLINE` and will able to query the new host/realname?

Note, extended-monitor doesn't make this scenario worse. MONITOR alone allows polling the host/realname.

emersion:
Added a section about privacy considerations.

emersion:
Ping: any more feedback on this spec?

jwheare:
This seems like it has decent consensus and can be merged. Will do that next week if no further comments. Ping me if I forget.

jwheare:
I just remembered this discussion https://github.com/ircv3/ircv3-ideas/issues/78

I'm a bit hesitant to merge new drafts until we have a decision on that. I'll try to resolve that first.

jwheare:
Dropping draft/ isn't clear cut so unblocking this.

Issue #467, opened 2021-08-15
sasl spec should clarify that AUTHENTICATE is a normal IRC message


slingamn:
The [SASL 3.1 spec](https://ircv3.net/specs/extensions/sasl-3.1) makes it sound as though AUTHENTICATE is its own subprotocol with its own grammar. This is a persistent source of confusion because implementations fail to handle AUTHENTICATE with prefixes or tags. Some examples:

1. [ZNC](https://github.com/znc/znc/issues/1212)
2. [Pidgin/libpurple](https://keep.imfreedom.org/pidgin/pidgin/rev/86e7c9ca0bcd)
3. [Weechat](https://github.com/weechat/weechat/issues/1679)

The spec should be amended to make it clear that `AUTHENTICATE` is a normal IRC message.

ailin-nemui:
4. https://github.com/pircbotx/pircbotx/issues/401

ailin-nemui:
points:
* `:+` and `:*` are all valid and need to be parsed
* optional `:server.` can be given
* message tags, when enabled

PR #468, opened 2021-08-15
AUTHENTICATE: Use a higher-level grammar consistent with other IRCv3 specs

progval:
This low-level grammar enforced a serialization stricter than usual
IRC commands; while most implementations use the latter (accepting,
and sometimes emitting extra colons), leading to misunderstandings.

Closes GH-467.


slingamn:
Should we add an explicit example with a prefix?

slingamn:
+1, ship it

grawity (approved):
Overall looks good to me.

Slightly unsure whether it makes sense for the example to use a different source than the local server (there's no corresponding target parameter in client commands), but clients will just ignore it anyway so it's fine.

PR #470, opened 2021-09-09
Clarify only sending values with CAP LS and NEW

### LINKED ISSUE #469: CAP DEL undefined behavior
Based on the docs, it does not seem to explicitly say in the main documentation if a CAP DEL can contain a value or not. Reading the Errata section, I see "Previous versions of this spec did not specify when the <name>=<value> format could be used. This was clarified to limit capability values to the LS and NEW subcommands." and "Previous versions of this spec did not mention whether NEW and DEL can have values or not." - BUT, despite that, I don't see those changes actually being made in the spec- nothing seems to mention where it "clarified to limit capability values to the LS and NEW subcommands" nor where it mentions if "DEL can have values or not". 

I would suggest a modification to the CAP DEL section adding the sentence "The server shall only send the CAP DEL subcommand for an entire capability, not individual values of a capability" or something similar.

vanosg:
Closes #469 

Add sentence clarifying values are not to be sent via CAP DEL

### DIFF
--- a/extensions/capability-negotiation.md
+++ b/extensions/capability-negotiation.md
@@ -248,7 +248,7 @@ As an overview, these are the new features introduced with each `CAP LS` version
 #### Capability Values
 
 If the client supports CAP version `302`, the server MAY specify additional data for each
-capability using the `<name>=<value>` format in `CAP LS` and `CAP NEW` replies.
+capability using the `<name>=<value>` format in `CAP LS` and `CAP NEW` replies, but MUST NOT specify additional data for any other CAP subcommands.
 
 Each capability, if it supports a value, defines what this value means in its specification.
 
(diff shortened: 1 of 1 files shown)


progval:
From https://ircv3.net/specs/extensions/capability-negotiation#capability-values : 

> If the client supports CAP version 302, the server MAY specify additional data for each capability using the <name>=<value> format in CAP LS and CAP NEW replies.

So this seems redundant.

vanosg:
It doesn't mention DEL; where is the redundancy?

progval:
It says it's allowed for "CAP LS", "CAP NEW", so implicitly disallowed for everything else. Otherwise, you would have to mention it's also disallowed for "CAP REQ", "CAP NAK", "CAP LIST".

Herringway:
an implementation sending values with DEL would be compatible with the spec as currently written. implied things can and will be ignored. this must be made explicit

progval:
Why only DEL, then?

grawity:
I think the spec _should_ be clarified, but I think it would be better to add it to the text quoted by progval, e.g. "… but MUST NOT do so for any other CAP subcommands" or something along those lines.

progval:
Can you make the edit suggested by gravity?

vanosg:
Certainly!

PR #471, opened 2021-09-17
Add Web Push Extension

emersion:
A primary goal of this spec is to not tie the extension to any proprietary push service. This is achieved by using the standard Web Push protocol. Web Push also provides encryption of the payload.

IRC clients are responsible for providing a push endpoint to the IRC server. This is usually provided by the platform (e.g. by the browser for the Web). IRC servers don't need to register or anything like that, all they need is send an HTTP request when they wish to deliver a push notification.

See the spec for more details.

Not mentioned in the spec is the integration with proprietary push services. For instance, Android apps can't (unfortunately) use Web Push: they must use Firebase Cloud Messaging (FCM). Since FCM is also used on Chrome for their Web Push implementation, I've tried to see if there was a way to figure out some Web Push parameters from the FCM ones, but it sounds like the Web Push functionality is limited to the Chrome API keys.

My approach is to introduce a relay which listens for Web Push messages and forwards them to FCM. This does require an additional moving part that IRC clients need to rely on, however this sounds like a reasonable trade-off to me. Since the relay doesn't see any clear-text privacy-sensitive information (payload is encrypted by the IRC server, no client IP address is visible), I could provide a free relay service for client developers who don't want to setup their own.

```
      ┌────────────┐              ┌────────────┐
      │            │  Subscribe   │            │
      │   Android  ├─────────────►│            │
      │ IRC client │              │ IRC server │
      │            │              │            │
      │            │              │            │
      └────────────┘              └─────┬──────┘
             ▲                          │
             │                          │
        Push │                          │Push
notification │                          │notification
             │                          ▼
       ┌─────┴─────┐              ┌──────────┐
       │           │              │          │
       │ Firebase  │◄─────────────┤ Web Push │
       │ Messaging │              │   Relay  │
       │           │              │          │
       └───────────┘              └──────────┘
```

So far, I've implemented proof-of-concepts for soju and two client platforms:

- gamja (Web): the implementation is pretty straightforward, it's mostly a matter of piping the JS API to IRC.
- Android: this requires a relay, and requires to implement Web Push decryption logic in the client. This isn't too bad, the Web Push encryption and decryption logic is mostly symmetrical. So as long as a library to send Web Push notifications exists, it shouldn't be too hard to turn it the other way around and make it decrypt notifications.

References:

- Ergo discussion: https://github.com/ergochat/ergo/issues/1527
- Web Push JS API spec: https://w3c.github.io/push-api/
- soju PoC: https://git.sr.ht/~emersion/soju/commit/0dfc1b66b9f5bfff7a10add9c4320517c1156a59
- gamja PoC: https://git.sr.ht/~emersion/gamja/commit/e324638ae895aa409c84063c20b36065953bc78e
- Web Push to Firebase Cloud Messaging relay: https://git.sr.ht/~emersion/pushgarden
- Android client PoC (remember, this is just a PoC, for now the code is terrible, but it works!): https://git.sr.ht/~emersion/PushGardenTestApp

* * *

**This is an incomplete specification**. I opened this PR to gather feedback on the approach. Let me know what you think!


hhirtz:
Sorry if this is obvious, but if the server needs to wait for the client to send the `endpoint`, what is the use of the `webpush` capability? Also it's not present in soju and gamja's commits.

emersion:
It's used to signal the client that the commands are supported. Could use an `ISUPPORT` I suppose… but need to make sure servers don't want to change behavior when push notifications are enabled.

extensions/webpush.md
```diff
@@ -0,0 +1,96 @@
+---
+title: "Web Push Extension"
+layout: spec
+copyrights:
+  - name: "Simon Ser"
+    period: "2020"
```
DarthGandalf, line 6: Check the calendar :)

DarthGandalf:
capability doesn't prevent server from changing the behavior after cap was enabled. There's even `CAP DEL` especially for that

extensions/webpush.md
emersion: Oops, thanks

emersion:
> capability doesn't prevent server from changing the behavior after cap was enabled. There's even CAP DEL especially for that

Nah, I mean that a cap would inform the server that the client supports the ext. For instance the chat history spec uses this to make servers *not* send playback on connect when the `chathistory` cap is enabled.

DarthGandalf:
Okay, and why this information is needed upon connect? I understand why it's needed for chathistory

emersion:
As said above, I'm not necessarily opposed to making this an `ISUPPORT` instead of a cap. We just need to make sure that no server is interested in knowing whether the client supports the ext or not.

extensions/webpush.md
```diff
@@ -0,0 +1,96 @@
...
+```
+
+Web Push is defined in [RFC 8030], [RFC 8291] and [RFC 8292].
+
+Although Web Push has been designed for the Web, it can be used on other platforms as well. Web Push provides a vendor-neutral standard to send push notifications.
+
+## Implementation
+
+The `webpush` capability allows clients to subscribe to Web Push and receive
+notifications for messages of interest.
```
kylef, line 52: How does a user/client indicate which messages would be of interest? There doesn't appear to be mechanisms in-place in the currently offered commands to indicate which messages would be of interest.

extensions/webpush.md
```diff
@@ -0,0 +1,96 @@
...
+The `REGISTER` subcommand creates a new Web Push subscription.
+
+    WEBPUSH REGISTER <endpoint> <keys>
+
+The `<endpoint>` is an URL pointing to a push server, which can be used to send push messages for this particular subscription.
+
+`<keys>` is a string encoded in the message-tag format. The values are [URL-safe base64-encoded][RFC 4648 section 5]. It MUST contain at least:
+
+- One public key with the name `p256dh` set to the client's P-256 ECDH Diffie-Hellman public key.
+- One shared key with the name `auth` set to a client-generated authentication secret.
```
kylef, line 87: I would recommend wording the spec a little differently so that algorithm(s) can be updated over time. Ensuring the design of the protocol is easy to update based on discovered weaknesses or when stronger algorithms come around. Prevent using terms like "these fields are required", but "for the algorithm X these fields are required".

extensions/webpush.md
```diff
@@ -0,0 +1,96 @@
...
+Although Web Push has been designed for the Web, it can be used on other platforms as well. Web Push provides a vendor-neutral standard to send push notifications.
+
+## Implementation
+
+The `webpush` capability allows clients to subscribe to Web Push and receive
+notifications for messages of interest.
+
+Once a client has subscribed, the server will send push notifications for a
+server-defined subset of IRC messages. Each push notification MUST contain
+exactly one IRC message as the payload, without the final CRLF.
```
kylef, line 56: Would the message follow the same capabilities and respect the same 005 isupport as when the client registered?

extensions/webpush.md
emersion: Yup, as discussed in #ircv3, this part isn't addressed yet. For now, it's spec'ed as a "server-defined subset of IRC messages".

I think there are multiple ways to go about this:

1. Keep as-is, let servers decide completely. Leave it to a future extension to let clients indicate messages of interest. This can be useful for chathistory as well, to avoid fetching all unread history when connecting.
2. Indicate a minimal set of messages to forward (e.g. PRIVMSG/NOTICE/INVITE with user as target, plus PRIVMSG/NOTICE highlights). Let servers send more if they want to. Clients can always ignore some push notifications.
3. Build a webpush-specific command to indicate the messages of interest. This is a slippery slope, because the rules can be pretty complicated. If we go down this road I'd rather keep it simple and have just simple word matching.

extensions/webpush.md
emersion: This part follows exactly the Web Push specs, which makes these two required.

extensions/webpush.md
emersion: Ah, good point. I wonder if there's an alternative here.

Your suggesting seems the simplest, and clients can always unregister/re-register if they notice new caps they're interested in.

extensions/webpush.md
kylef: The set of specifications surrounding web push are separated into individual RFCs. RFC8030 does not mandate encryption nor an single encryption standard. 

Whereas here we have one specification that defines both the web push capabilities, how to negotiate clients which requires one specific encryption mechanisms. This doesn't appear to be the case for web push.

RFC 8030 "Generic Event Delivery Using HTTP Push" specifies:

> Examples do not include specific methods for push message encryption or application server authentication because the protocol does not define a mandatory system.  The examples in Voluntary Application Server Identification [VAPID] and Message Encryption for WebPush [ENCRYPT] demonstrate the approach adopted by the W3C Push API [API] for its requirements.

[RFC8291](https://datatracker.ietf.org/doc/html/rfc8291) on its own only describes how to use this particular algorithm and does not describe how web push works. This RFC can be replaced on its own.

This is similar to how IRCv3 has specified SASL, the mechanisms are separate specs which live in their own lifecycle from core SASL.

slingamn:
Two issues that came up in discussion with @kylef:

1. Signal only sends a ping message (i.e. with no user data) over the push channel, and then the client reconnects to the server and downloads the actual message data. This was felt to be unworkable in the IRC context, since iOS is imposing increasingly onerous restrictions on what can be done in a push notification handler (it's too hard to connect to an IRC server, and soon it may be outright impossible).
2. However, [RFC8030](https://datatracker.ietf.org/doc/html/rfc8030) seems to impose a de facto size limit of 4096 bytes on push messages. This is shorter than the maximum message length under message-tags, so provisions for the server shortening the message are necessary. (The server might have a list of tags to prioritize?) Given this, we might want to specify an alternative format for the message (JSON would allow the server to send extensible metadata with the push message, although at this time I don't have a candidate use case).

emersion:
Yeah, the size limit is going to be an issue. FCM alone also has a 4096 bytes limit for the payload. 

> we might want to specify an alternative format for the message

I don't think JSON would help, the limit still applies regardless of the format of the payload. Also, servers can extend messages already with tags. Am I missing something?

slingamn:
Sorry, that was unclear. I meant something like: abandon RFC1459-and-derivatives entirely in the format of the push message, parse prefix/tags/command/params directly into JSON.

emersion:
How is this going to help with the size limit of the push notification payload?

extensions/webpush.md
emersion: I see, thanks for taking the time to explain. I'm not sure how to word this. Would this do it? "For the encryption scheme described in RFC 8291, these fields are required"

extensions/webpush.md
```diff
@@ -0,0 +1,101 @@
...
+
+### `REGISTER` Subcommand
+
+The `REGISTER` subcommand creates a new Web Push subscription.
+
+    WEBPUSH REGISTER <endpoint> <keys>
+
+The `<endpoint>` is an URL pointing to a push server, which can be used to send push messages for this particular subscription.
+
+`<keys>` is a string encoded in the message-tag format. The values are [URL-safe base64-encoded][RFC 4648 section 5]. It MUST contain at least:
```
emersion, line 87: This uses the URL-safe base64 encoding because that's what the [W3C API uses](https://w3c.github.io/push-api/#dom-pushsubscription-tojson). So this allows Web clients to just use the result of `subscription.toJSON().keys` as-is.

But maybe that isn't such a good idea. The W3C API also doesn't provide any function to encode an arbitrary byte buffer to URL-safe base64, so users need to roll out their own if we use it elsewhere.

Instead we could just use regular base64, just like SASL does. Users can call `subscription.getKey()`.

extensions/webpush.md
emersion: Pushed a commit to add a note about this.

kylef:
Do you intent to add a command to view the list of subscriptions? I could see how as a user this could be useful to debug problems and be able to manually clean up or unsubscribe old clients for which you do not know the subscription URL for. This could very well be something that a NickServ or similar service could facilitate for a user and not be included in the spec though.

emersion:
Yeah, if it's just a debugging tool, I think I'd rather let servers include it as a special service or custom commands. Also need to investigate if/how subscription expiration can be integrated.

extensions/webpush.md
emersion: I think (1) is the superior option here.

emersion:
My plan is to go ahead and ship this as a soju vendored extension, then gather field testing feedback and report back what I've learnt here.

emersion:
Goguma and pushgarden now support Apple Push Notification service, as more evidence that this approach works on all platforms.

emersion:
The vendored extension is now supported by Igloo and Ergo.

slingamn (commented):
Let me say, thanks very much for your work on this spec; it's well-designed and solves an important problem.

More discussion of implementation considerations and security considerations would be helpful. For example, the implications of sending a POST to a user-chosen address should be discussed in more detail. The Soju restrictions (which I copied in Ergo) are to only allow https endpoints, and to disallow contacting loopback or internal IP addresses.

emersion:
> More discussion of implementation considerations and security considerations would be helpful. For example, the implications of sending a POST to a user-chosen address should be discussed in more detail. The Soju restrictions (which I copied in Ergo) are to only allow https endpoints, and to disallow contacting loopback or internal IP addresses.

Yes, agreed! Also some recommendations around expiration on the server side, how a client should renew registrations, etc.

extensions/webpush.md
```diff
@@ -0,0 +1,123 @@
...
+
+The messages follow the same capabilities and the same `RPL_ISUPPORT` as when the client registered for Web Push notifications.
+
+Because of size limits on the payload of push notifications, servers MAY drop some or all message tags from the original message. Servers MUST NOT drop the `msgid` tag if present.
+
+## `VAPID` ISUPPORT token
+
+If the server supports [Voluntary Application Server Identification (VAPID)][RFC 8292] and the client has enabled the `draft/webpush` capability, the server MUST advertise its public key in the `VAPID` ISUPPORT token. This key can be used to verify notifications upon reception by the Web Push server.
+
+The value MUST be the [URL-safe base64-encoded][RFC 4648 section 5] public key usable with the Elliptic Curve Digital Signature Algorithm (ECDSA) over the P-256 curve. The value MUST NOT change over the lifetime of the connection to avoid race conditions.
```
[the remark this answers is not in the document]
emersion: It's the IRC TCP connection.

The spec has been designed to allow key rotation. The server can advertise a different key to new connections. However, a server needs to keep the old key working or risk breaking previous subscriptions. For instance, maybe a client hasn't connected via TCP for a month but still receives push notifications every week or so.

The race condition is the following:

- Client connects and receives VAPID 1. Client sends a `WEBPUSH REGISTER` command.
- Server sends a new VAPID 2.
- Server receives `WEBPUSH REGISTER`, but can't tell whether it must use VAPID 1 or VAPID 2 when sending push notifications.

It's important to use the correct VAPID, because the push endpoint should validate the VAPID.

We could add ways to fix the race (e.g. by adding some kind of ACK mechanism, simplest would be for the client to send the VAPID in `WEBPUSH REGISTER` perhaps), but I figured it wasn't worth it since a server needs to keep the old key working anyways and the typical TCP connection lifetime is very small compared to the typical push registration lifetime.

extensions/webpush.md
```diff
@@ -0,0 +1,123 @@
...
+
+Web Push is defined in [RFC 8030], [RFC 8291] and [RFC 8292].
+
+Although Web Push has been designed for the Web, it can be used on other platforms as well. Web Push provides a vendor-neutral standard to send push notifications.
+
+## Implementation
+
+The `draft/webpush` capability allows clients to subscribe to Web Push and receive notifications for messages of interest.
+
+Once a client has subscribed, the server will send push notifications for a server-defined subset of IRC messages. Each push notification MUST contain exactly one IRC message as the payload, without the final CRLF.
```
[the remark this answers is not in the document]
emersion: I haven't considered draft/multiline at all when writing the spec.

I think sending the first line is fair: UX-wise notifications are not designed to display big blobs of text, and the user can always tap to see the whole thing.

extensions/webpush.md
```diff
@@ -0,0 +1,123 @@
...
+
+    WEBPUSH UNREGISTER <endpoint>
+
+Servers MUST silently ignore `UNREGISTER` commands for non-existing subscriptions.
+
+If the unregistration is successful, the server MUST echo back the `WEBPUSH UNREGISTER` message. On error, the server MUST reply with a `FAIL` message.
+
+### Errors
+
+Errors are returned using the standard replies syntax.
```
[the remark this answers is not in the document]
slingamn: Would we advertise the CAP to unauthenticated users pre-registration, and then remove it after the 001? That seems awkward.

And yeah, `FORBIDDEN` seems justified for other cases like limits.

extensions/webpush.md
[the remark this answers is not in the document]
slingamn: OK, that makes sense. Here's the disconnect I was having:

>the typical TCP connection lifetime is very small compared to the typical push registration lifetime

This is not true of traditional or hybrid IRC setups where connections can live for weeks. The current spec language seems to imply that the VAPID key cannot be rotated as long as there is *any* open TCP connection (that hasn't already sent WEBPUSH REGISTER?). This would make rotation impossible in practice.

It seems to me that the race window from rotation is very small as long as the server notifies active clients about the rotation, by sending them a new 005 line with the new value.

What is the expected behavior if the push endpoint fails to validate the VAPID signature? 404?

p1gp1g:
I've submitter a draft RFC for an IMAP WebPush extension. This is, as expected, very close to this extension for IRC: https://datatracker.ietf.org/doc/draft-gougeon-imap-webpush/

I have notably added another command to silent a registration, this can be useful when the client sync for some time after a push message and if the client define a "do not disturb" period. This may interest you to add it as well.

Since you are already using your IRC extension, I wonder if there are things I forgot to take in considerations ?

> Not mentioned in the spec is the integration with proprietary push services. For instance, Android apps can't (unfortunately) use Web Push: they must use Firebase Cloud Messaging (FCM). Since FCM is also used on Chrome for their Web Push implementation, I've tried to see if there was a way to figure out some Web Push parameters from the FCM ones, but it sounds like the Web Push functionality is limited to the Chrome API keys.

This is actually possible to send Web Push message to FCM servers directly, without a gateway, I've described it here: https://unifiedpush.org/news/20250131_push_for_decentralized/

extensions/webpush.md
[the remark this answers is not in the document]
emersion: No, the server would hide the cap pre-conn-reg, and send `CAP NEW` after authentication.

A `FORBIDDEN` error will be bubbled up to the user, which is unlikely to be what you want here.

emersion:
> I have notably added another command to silent a registration, this can be useful when the client sync for some time after a push message and if the client define a "do not disturb" period. This may interest you to add it as well.

What is the difference with unregistering (and potentially re-registering later)?

> This is actually possible to send Web Push message to FCM servers directly, without a gateway, I've described it here

Great to see you've found a way to do this? A few years back, I've tried reading the Chromium source code and send Web Push payloads directly to FCM servers, but it seemed like that endpoint was restricted to Chrome API keys.

extensions/webpush.md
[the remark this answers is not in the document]
emersion: I've added a `MAX_REGISTRATIONS` code for this case.

emersion:
Added sections for security and implementation considerations.

p1gp1g:
> > I have notably added another command to silent a registration, this can be useful when the client sync for some time after a push message and if the client define a "do not disturb" period. This may interest you to add it as well.
> 
> What is the difference with unregistering (and potentially re-registering later)?

The most advantageous is `SILWEBPUSH session`, when it is used, the client don't have to implement anything when it closes its connection, and it can let itself be killed without worrying to restore the subscription

When `SILWEBPUSH duration` is used, it doesn't have to implement a scheduled event to restore the subscription; it avoid a connection to send a new REGISTER

emersion:
I'm not entirely sure I undertsand the use-case here. I don't recommend switching notifications on and off because this can easily result in race conditions and missed notifications. I recommend either always relying on push notifications and not showing any notification from the TCP connection (preferred), or ignoring push messages when the TCP connection is opened (but this is racy).

PR #474, opened 2021-11-15
account-registration: Add examples of the 'working' cases.

progval:
Copy-pasted from what Ergo sends

### DIFF
--- a/extensions/account-registration.md
+++ b/extensions/account-registration.md
@@ -233,8 +233,6 @@ The client can then proceed with the connection:
     C: REGISTER * * hunter2
     S: REGISTER SUCCESS tester :Account successfully registered
 
-Note that email addresses may be required by the server, but not verified.
-
 # Client considerations
 
 This section is non-normative.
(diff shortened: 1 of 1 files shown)


jwheare:
Nice, good to show one without verification too. And with optional email.

progval:
I used a single example for both, so it does not get too long

extensions/account-registration.md
```diff
@@ -218,6 +223,18 @@ The client can then proceed with the connection:
...
+
+    C: CAP LS 302
+    S: CAP * LS :draft/account-registration=before-connect
+    C: CAP REQ :draft/account-registration
+    C: CAP END
+    ...
+    C: REGISTER * * hunter2
+    S: REGISTER SUCCESS tester :Account successfully registered
+
+Note that email addresses may be required by the server, but not verified.
```
jwheare, line 236: What does this mean? As in the server sends email-required but no VERIFICATION_REQUIRED?

progval: Yes. Any suggestion to phrase it better?

jwheare: I'd just leave it out, less confusing.

jwheare: Better to just split the examples if you think people might be confused.

progval: ok, done

PR #480, opened 2021-12-02
Explicitly allow CAP NEW to update existing caps

### LINKED ISSUE #477: Unclear how servers should send cap updates
Is this valid if a server looses the ability to perform SASL EXTERNAL auth?

```
CAP NEW sasl=PLAIN,EXTERNAL
[…]
CAP NEW sasl=PLAIN
```

Or should the server send a `CAP DEL sasl` in-between? The latter would have some annoying consequences, e.g. the client will need to `CAP REQ sasl` again, even though the server never lost the ability to perform SASL PLAIN auth.

emersion:
Closes: https://github.com/ircv3/ircv3-specifications/issues/477


jwheare:
Is there a server that does this at the moment that can be used to test clients?

emersion:
soju does this, but there's no public test instance.

jwheare:
I imagine it's quite a rare real world situation tbh, and may not manifest in an immediately obvious bug. I'll have a play with a dummy netcat server.

jwheare:
Confirmed that IRCCloud handles this as well.

emersion:
Any objections against landing this?

jwheare:
Would prefer to see some tests with other clients, but I'm fine with it.

SadieCat:
Any updates on this? InspIRCd currently sends `CAP DEL` then `CAP NEW` but I can easily update it to just send `CAP NEW` if this is merged.

jwheare:
Someone just needs to test other clients.

4e554c4c:
confirming that halloy supports this behaviour

Issue #479, opened 2021-12-02
Standardize pre-welcome FAIL ACCOUNT_REQUIRED


emersion:
It seems like some ergo servers will send the following when performing connection registration without logging in:

    FAIL * ACCOUNT_REQUIRED :You must log in with SASL to join this server
    ERROR :You must log in with SASL to join this server

It would be nice to standardize this error code so that clients can reliably match it. It's not in the registry atm.

Ref https://github.com/ergochat/ergo/blob/c9b54ee2b801909cc3944bedbd04bc4079c54d8a/irc/server.go#L305

cc @slingamn

jwheare:
This should probably be added to the account-registration spec.

slingamn:
It's not inherently tied to that spec, though; it applies to situations where the problem can't be corrected by registering an account (i.e. private servers with closed registration). We've been sending it since well before that spec was proposed.

I would submit a PR to add this to https://ircv3.net/registry#fail-codes as long as no one objects to the use of `*`.

jwheare:
I'm not sure it matters tbh, it's a convenient and clearly related place to document it in an actual spec, rather than a dangling registry entry.

PR #481, opened 2021-12-03
add FAIL * ACCOUNT_REQUIRED to account-registration

slingamn:
From discussion on https://github.com/ircv3/ircv3-specifications/issues/479

### DIFF
--- a/extensions/account-registration.md
+++ b/extensions/account-registration.md
@@ -181,7 +181,7 @@ unavailable.
 
 Sent by the server in response to the end of connection registration (e.g.
 to `USER`, `NICK`, or `CAP END`) if connection registration cannot proceed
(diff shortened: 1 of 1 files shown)


emersion:
Maybe link to the SASL spec? Should we state that the connection will be closed with `ERROR`, or should we allow servers to leave it open to let clients issue a SASL command? (just like `ERR_PASSWDMISMATCH` doesn't necessarily close the connection)

slingamn:
I mentioned SASL explicitly, but didn't add a link for now, since SASL is mentioned in other places in the spec without a link.

I think it's fine to leave it unspecified whether the connection is closed in this case.

emersion:
LGTM!

Issue #81, opened 2014-03-23
Import WHOX spec


jwheare:
Note that some implementations in the wild have vendor specific variations. E.g. rules on case sensitivity, the `H` (real host names) option, join-delayed channel members.

It would be useful to document the base compatibility case for new implementations and highlight any areas where vendor-extensions are possible.

https://github.com/quakenet/snircd/blob/master/doc/readme.who
http://faerion.sourceforge.net/doc/irc/whox.var

slingamn:
@ilbelkyr expressed interest in this :1st_place_medal:

SadieCat:
It might be a neat WHOX extension to add a value to the ISUPPORT token like `WHOX=[flags],[fields]` so that if server vendor has custom fields/flags they can let clients know about them

We do something similar to this with ESILENCE in InspIRCd for our SILENCE flags.

(cc @ilbelkyr)

emersion:
I'm in the process of documenting WHO on modern.ircdocs.horse (https://github.com/ircdocs/modern-irc/pull/138), and I'm wondering: would WHOX make more sense as a modern.ircdocs.horse patch or as an IRCv3 ext? cc @progval

jwheare:
I think it'd help to have it in IRCv3 given its long standing non-standardisation. But no reason it can't be on modern too.

emersion:
Opened https://github.com/ircv3/ircv3-specifications/pull/482

PR #482, opened 2021-12-07
Add WHOX

emersion:
This PR documents a common subset of the WHOX extension. The goal here is not to draft a new standard, but document what already exists in the wild. We can extend this spec as a second step if desirable.

"querytype" has been renamed to "token", because it's a much more descriptive name IMHO (and happens to have the same first letter). Let me know if you want to change it back.

Intentionally left out:

- WHOX flags, unimplemented by some servers
- ~~'d' (distance/hopcount) and 'o' (oplevel), unreliable and not well-defined~~

References:

- https://github.com/quakenet/snircd/blob/master/doc/readme.who
- http://faerion.sourceforge.net/doc/irc/whox.var
- Ergo impl: https://github.com/ergochat/ergo/pull/1184
- Tentative WHO spec: https://github.com/ircdocs/modern-irc/pull/138

Closes: https://github.com/ircv3/ircv3-specifications/issues/81

### DIFF
--- a/extensions/whox.md
+++ b/extensions/whox.md
@@ -30,20 +30,25 @@ The second argument contains a percent character followed by a list of fields
 requested by the client. Each field is represented by a letter. The standard
 fields are:
 
-- 't': return the `<token>` specified by the client
-- 'c': return an arbitrary channel the client is joined to
-- 'u': return the username
-- 'i': return the IP address
-- 'h': return the hostname
-- 's': return the server name
-- 'n': return the nickname
-- 'f': return the `WHO` flags (away, server operator, etc)
-- 'l': return the number of seconds the user has been idle for
-- 'a': return the account name
-- 'r': return the realname
+- `t`: return the `<token>` specified by the client
+- `c`: return an arbitrary channel the client is joined to
+- `u`: return the username
+- `i`: return the IP address
+- `h`: return the hostname
+- `s`: return the server name
+- `n`: return the nickname
+- `f`: return the `WHO` flags (away, server operator, etc)
+- `d`: return the hop count (distance)
+- `l`: return the number of seconds the user has been idle for
+- `a`: return the account name
+- `o`: return the channel op level
+- `r`: return the realname
 
 Servers MAY support additional non-standard fields. Servers MUST NOT rely on
 the ordering of the fields.
+
+Clients SHOULD avoid requesting the hop count (`d`) and the channel op
+level (`o`), because they are ill-defined and unreliable.
 
 Clients can also specify a token which will be returned by the server in the
 replies. The token MUST contain only digit characters and MUST contain at most
@@ -57,7 +62,7 @@ by a final `RPL_ENDOFWHO` reply. Servers MUST NOT send `RPL_WHOREPLY` replies.
 
 ## `RPL_WHOSPCRPL` (354) numeric reply
 
-    :<server> 354 <client> [token] [channel] [user] [ip] [host] [server] [nick] [flags] [idle] [account] [:realname]
+    :<server> 354 <client> [token] [channel] [user] [ip] [host] [server] [nick] [flags] [hopcount] [idle] [account] [oplevel] [:realname]
 
 Servers MUST send the fields in the order specified above. Servers MUST include
 in their reply all of the standard fields requested by the client, and MUST NOT
@@ -72,6 +77,7 @@ are used:
 - `255.255.255.255` is used for `[ip]` when the server refuses to disclose the
   IP address
 - `0` is used for `[account]` when the user is logged out
+- `n/a` is used for `[oplevel]` when the server doesn't support op levels
 
 ## Examples
 
--- a/extensions/whox.md
+++ b/extensions/whox.md
@@ -1,0 +1,83 @@
+---
+title: WHOX
+layout: spec
+copyrights:
+  -
+    name: "Simon Ser"
+    period: "2021"
+    email: "contact@emersion.fr"
+---
+
+## Introduction
+
+The `WHO` command allows clients to obtain information about other clients.
+However, `WHO` doesn't allow clients to request extra data (e.g. account names)
+nor a subset of the data (e.g. omit user and host).
+
+## `ISUPPORT` token
+
+Servers supporting this specification MUST include the `WHOX` token in
+`RPL_ISUPPORT` without any value. Clients MUST accept `WHOX` tokens with a
+value for forwards compatibility.
+
+## Extended `WHO` command
+
+The `WHO` command is extended with an optional argument:
+
+    WHO <mask> [%<fields>[,<token>]]
+
+The second argument contains a percent character followed by a list of fields
+requested by the client. Each field is represented by a letter. The standard
+fields are:
+
+- 't': return the `<token>` specified by the client
+- 'c': return an arbitrary channel the client is joined to
+- 'u': return the username
+- 'i': return the IP address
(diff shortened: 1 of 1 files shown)


extensions/whox.md
```diff
@@ -0,0 +1,83 @@
...
+token.
+
+Servers MAY support additional non-standard flags before the percent character.
+
+The server will reply with zero, one or more `RPL_WHOSPCRPL` replies, followed
+by a final `RPL_ENDOFWHO` reply. Servers MUST NOT send `RPL_WHOREPLY` replies.
+
+## `RPL_WHOSPCRPL` (354) numeric reply
+
+    :<server> 354 <client> [token] [channel] [user] [ip] [host] [server] [nick] [flags] [idle] [account] [:realname]
```
slingamn, line 59: AIUI the ordering of the fields sent by the client MUST be ignored, and then the server MUST return the requested fields in the order specified here? If so, it might be worth stating this explicitly, and possibly also adding an example where the client requests in a nonstandard order and the server returns in the standard order.

jobe1986: > 

This is correct as per both my understanding and the implementations I have observed

emersion: Done!

extensions/whox.md
```diff
@@ -30,21 +30,26 @@ The second argument contains a percent character followed by a list of fields
...
+- `d`: return the hop count (distance)
+- `l`: return the number of seconds the user has been idle for
+- `a`: return the account name
+- `o`: return the channel op level
+- `r`: return the realname
 
 Servers MAY support additional non-standard fields. Servers MUST NOT rely on
 the ordering of the fields.
 
+Clients SHOULD avoid requesting the hop count (`d`) and the channel op
```
jwheare, line 50: I would say "avoid relying on the value of" instead. Requesting them can be useful to increase entropy when sending automated tokenised WHOX requests to distinguish them from manual client sent ones.

extensions/whox.md
```diff
@@ -72,6 +77,7 @@ are used:
 - `255.255.255.255` is used for `[ip]` when the server refuses to disclose the
   IP address
 - `0` is used for `[account]` when the user is logged out
+- `n/a` is used for `[oplevel]` when the server doesn't support op levels
```
jwheare, line 80: In the wild, for example on libera.chat: "n/a" actually means the user is not an op, while "999" means the user is an op. Probably best to match a common implementation in the spec.

extensions/whox.md
emersion: Hm, can you elaborate? I'm not sure I understand why it can be desirable for a client to request these.

extensions/whox.md
slingamn: FWIW, this is unsupported in Ergo, and we always send `0`:

https://github.com/ergochat/ergo/blob/v2.8.0/irc/handlers.go#L3311-L3314

extensions/whox.md
jwheare: It's not a particularly strong reason tbh, and I don't think it needs to be mentioned in the spec, but the more I can make an automated WHO request differ from a potential user requested one the better.

More importantly, clients could have any number of valid reasons for requesting this information that we can't predict so I don't know if it warrants a stern telling off in the spec for doing so.

extensions/whox.md
emersion: Hm. I'd prefer to avoid adding anything in the spec explaining what the values mean, since implementations behave differently. This bullet point just suggests servers that don't have a concept of "op level" to send "n/a". Would it be better to send "0" instead like Ergo does? Or "*" as a completely new value so that clients can tell when the value is completely meaningless or not?

jwheare: I'm happy to not go into the recommended value here. But if we are, it should probably reflect reality.

jwheare: I'd suggest testing on more networks (or browsing their source codes) to see whether they match the solanum/charybdis behaviour and if they do, go with that. But given we're saying it's not reliable, new implementations can effectively put whatever they want here, and clients shouldn't break.

slingamn: Maybe delete this as a normative recommendation and then add something like the following to the next paragraph?

>Servers that do not support reporting channel op levels MUST send a syntactically valid, implementation-defined placeholder value if it is requested, for example `*` or `0`.

jobe1986: The problem with a ```0``` value is ```0``` is a valid oplevel, it is in fact the highest oplevel, ```999``` being the lowest (ircu is backwards in that regard), so any integer between 0 and 999 inclusive is indicative of being opped in the channel.

So in my opinion the best choices would be ```*``` or ```n/a````

slingamn: Fair enough, a normative recommendation of `*` (whether or not it matches existing in-the-wild implementations) seems good also.

emersion: - Unrealircd uses `999` for op and `n/a` as a fallback: https://github.com/unrealircd/unrealircd/blob/2afc57aa38f7d78759b646ccb1f9fc2f9ea4cc85/src/modules/whox.c#L817 
- Inspircd uses `0` as a fallback: https://github.com/inspircd/inspircd/blob/6ff8f38a6d55970af1400cc78994e017ca351c39/src/coremods/core_who.cpp#L499
- Hybrid doesn't support WHOX

It doesn't seem like any server uses `*` yet. That makes me a bit uneasy to use it in this spec, because this spec is just supposed to document existing behavior. So I'd prefer to stick with `n/a`.

emersion: Updated for wording to allow servers to use any placeholder for oplevel.

emersion:
Any more comments on this spec?

extensions/whox.md
```diff
@@ -0,0 +1,96 @@
...
+
+The `WHO` command is extended with an optional argument:
+
+    WHO <mask> [%<fields>[,<token>]]
+
+The second argument contains a percent character followed by a list of fields
+requested by the client. Each field is represented by a letter. The standard
+fields are:
+
+- `t`: return the `<token>` specified by the client
```
progval, line 33: Incompatible with Insp: https://github.com/inspircd/inspircd/blob/ec37d6bb313f1c15df000fc635e0f2693209ddc7/src/coremods/core_who.cpp#L331-L332

progval (commented):
LGTM, just one issue.

extensions/whox.md
genius3000: @progval you've linked to a (match) flag, the 't' field logic is compatible: https://github.com/inspircd/inspircd/blob/ec37d6bb313f1c15df000fc635e0f2693209ddc7/src/coremods/core_who.cpp#L423...L425

extensions/whox.md
```diff
@@ -0,0 +1,96 @@
...
+  -
+    name: "Simon Ser"
+    period: "2021"
+    email: "contact@emersion.fr"
+---
+
+## Introduction
+
+The `WHO` command allows clients to obtain information about other clients.
+However, `WHO` doesn't allow clients to request extra data (e.g. account names)
```
slingamn, line 14: Bikeshedding the language a little here, I think this sentence might read better as:

>The `WHOX` extension allows clients to request additional fields in the `WHO` response (e.g. account name) or omit default fields (e.g. username and hostname).

emersion: That's better! Applied, thanks!

jwheare:
Looks ok to merge unless anyone has further comments.

PR #483, opened 2021-12-10
Add DNS SRV records

emersion:
I've seen a lot of users try to connect to "libera.chat" instead of "irc.libera.chat". This results in connection timeouts.

This is an attempt to improve the status quo.

Previous proposals:

- https://github.com/ircv3/ircv3-specifications/pull/59
- https://www.devever.net/~hl/md/irc-srv

### DIFF
--- a/extensions/dns-srv.md
+++ b/extensions/dns-srv.md
@@ -1,0 +1,43 @@
+---
+title: DNS SRV records
+layout: spec
+copyrights:
+  -
+    name: "Simon Ser"
+    period: "2021"
+    email: "contact@emersion.fr"
+---
+
+## Introduction
+
+IRC servers typically use a sub-domain so that IRC connections can be routed
+separately from the rest of the services. For instance, an IRC network using
+the domain name `example.org` may ask its users to connect to `irc.example.org`
+and use DNS load balancing to spread incoming connections to multiple servers.
+
+Unfortunately, this means that users need to make sure to connect to
+`irc.example.org` instead of `example.org` as the latter won't accept IRC
+connections.
+
+Additionally, other connection details such as the port and TLS usage might
+differ from one network to another.
+
+Other services (e.g. IMAP, SMTP) have standardized DNS SRV records which can be
+used to discover connection details. This specification defines DNS SRV records
+for IRC.
+
+SRV records are defined in [RFC 2782].
+
+## Description
+
+IRC networks conforming to this specification MUST publish an SRV record with
+the "ircs" service label. The record identifies an IRC server where TLS is
+initiated directly upon connection to the server.
+
+Example:
+
+    _ircs._tcp SRV 0 1 6697 irc.example.org.
+
+There is no defined service label for unencrypted connections.
+
+[RFC 2782]: https://datatracker.ietf.org/doc/html/rfc2782


progval:
As much as I like SRV, I wonder if this is the right direction. These days, the internet seems to be moving to "well-known" HTTP URLs, because web apps can't use the DNS.

Obviously web apps wouldn't connect to irc.libera.chat/6697 anyway; but we could define a single well-known for both normal sockets and websockets so that it's less technical overhead for network admins.

aaronmdjones:
You may want to clarify that the example record should be in the `IN` (Internet) namespace.

```
    _ircs._tcp IN SRV 0 1 6697 irc.example.org.
```

emersion:
> These days, the internet seems to be moving to "well-known" HTTP URLs, because web apps can't use the DNS.

I'd prefer SRV records, because I don't want to depend on an HTTP library in my IRC client.

emersion:
> You may want to clarify that the example record should be in the IN (Internet) namespace.

[RFC 2782](https://datatracker.ietf.org/doc/html/rfc2782) doesn't use that syntax, nor does [RFC 6186](https://datatracker.ietf.org/doc/html/rfc6186).

grawity:
I'm ok with including IN, but don't think it's necessary nowadays, as the other two classes (Hesiod and chaosnet) are so incredibly unlikely. Hell, even Hesiod switched to IN.

However, I'd like the spec to explicitly say which hostname – the original input or the SRV target – should be used for TLS host verification. That's something that took me a while to figure out when setting up Matrix, and it tends to vary between SRV-using protocols in general.

aaronmdjones:
Verifying the certificate against the SRV target would be dangerous; it allows an active MTIM to alter the SRV reply to get you to validate against a domain name they control.

grawity:
> Verifying the certificate against the SRV target would be dangerous; it allows an active MTIM to alter the SRV reply to get you to validate against a domain name they control.

Yes, but on the other hand, it's what most SRV users (except for Matrix) do. Also, *~DNSSEC~*

> Verifying the certificate against the SRV target would be dangerous; it allows an active MITM to alter the SRV reply to get you to validate against a domain name they control.

Indeed, which is probably why Matrix chose to handle it like CNAME and use the original domain name.

But many other SRV consumers do use the target domain instead (for example, LDAPS), either for historical reasons or because DNSSEC.

In the end, I don't care which way you decide to do it, I care about whether it'll be *documented* in the spec.

slingamn:
>Indeed, which is probably why Matrix chose to handle it like CNAME and use the original domain name.

For many users (in particular, people using ACME with the http-01 challenge), if the root domain and the IRC domain point to different hosts, it doesn't seem practical to get a certificate covering both domains onto the IRC host. So it seems in order to use this, you'd have to transfer one of the certificates between servers (and then expose both certificates and rely on SNI).

IMO the concerns about MITM are sufficiently serious that this should not be used as a means of automatically redirecting users to the correct server --- it should be at most a mechanism to suggest to the user that they reconfigure, e.g. a dialog box saying "there's no IRC server on libera.chat; did you mean irc.libera.chat?"

>I'd prefer SRV records, because I don't want to depend on an HTTP library in my IRC clients.

Just clarifying: AFAIK there's no way to look up a SRV record in JavaScript, so we're talking about desktop clients?

emersion:
For reference, the Matrix spec (https://spec.matrix.org/latest/server-server-api/#resolving-server-names):

> […] a server is found by resolving an SRV record for _matrix._tcp.<hostname>. This may result in a hostname (to be resolved using AAAA or A records) and port. Requests are made to the resolved IP address and port, using 8448 as a default port, with a Host header of <hostname>. The target server must present a valid certificate for <hostname>.

I agree we should require the certificate to be valid for the original hostname.

> For many users (in particular, people using ACME with the http-01 challenge), if the root domain and the IRC domain point to different hosts, it doesn't seem practical to get a certificate covering both domains onto the IRC host. So it seems in order to use this, you'd have to transfer one of the certificates between servers (and then expose both certificates and rely on SNI).

Let's go through the possible setups for network operators here:

1. A single server for `example.org` and `irc.example.org`: just share the cert on the filesystem
2. One server for `example.org`, multiple IRC servers behind `irc.example.org` sharing the load (e.g. Libera Chat): then http-01 can't be used anyways, since the server which will have to complete the challenge is random. I bet these networks are already using dns-01, can someone from Libera confirm?
3. A single server for `example.org` and a single server for `irc.example.org`: these are the problematic setups.

How often does (3) happen in practice? It sounds like sharing the certs would be a minor annoyance in this specific case.

> IMO the concerns about MITM are sufficiently serious that this should not be used as a means of automatically redirecting users to the correct server --- it should be at most a mechanism to suggest to the user that they reconfigure, e.g. a dialog box saying "there's no IRC server on libera.chat; did you mean irc.libera.chat?"

This severely degrades the client's UX. I think with the TLS cert requirement the MITM concerns are resolved.

Another way to resolve them would be to add a subdomain requirement: the SRV target MUST be a subdomain of the original hostname. This wouldn't cover rarer cases like `ubuntu.com` → `irc.libera.chat` and this doesn't seem as strong as the TLS cert requirement.

> Just clarifying: AFAIK there's no way to look up a SRV record in JavaScript, so we're talking about desktop clients?

Yes, only clients connecting via TCP are taken into account here. JavaScript clients don't connect via TCP as noted above, and are typically configured by the network operator themselves, so wouldn't benefit from a discovery mechanism regardless.

aaronmdjones:
> One server for example.org, multiple IRC servers behind irc.example.org sharing the load (e.g. Libera Chat): then http-01 can't be used anyways, since the server which will have to complete the challenge is random. I bet these networks are already using dns-01, can someone from Libera confirm?

We are. Adding an extra hostname to our certificates requires no further effort on our part.

aaronmdjones (approved):

slingamn:
>I think with the TLS cert requirement the MITM concerns are resolved.

ACK, this resolves my concerns.

grawity:
Latest change looks good to me.

emersion:
Client implementation: https://git.sr.ht/~emersion/soju/commit/cb695edecb0aaf8b4d3b2d3a500ada78a5a67621

grawity:
Hmm out of curiosity, are the networks in question planning to use SRV purely as a redirect (single record pointing to the existing `irc` round-robin), or would they use SRV directly for load-balancing (multiple records pointing to individual servers)?

That is, should clients expect to be able to just take `srv[0]` like in soju, or should they at least build a flat list from all records, or should they go all the way and handle priorities/weights?

emersion:
The Go library already takes care of the shuffling, `srv[0]` is random from the list: https://cs.opensource.google/go/go/+/refs/tags/go1.17.5:src/net/dnsclient.go;drc=refs%2Ftags%2Fgo1.17.5;l=194

extensions/dns-srv.md
```diff
@@ -0,0 +1,47 @@
...
+
+## Description
+
+IRC networks conforming to this specification MUST publish an SRV record with
+the "ircs" service label. The record identifies an IRC server where TLS is
+initiated directly upon connection to the server.
+
+Example:
+
+    _ircs._tcp SRV 0 1 6697 irc.example.org.
```
lanodan, line 39: I think it would be better to have the record explicitely be for `example.org` for the TLS explanation below:

```suggestion
    _ircs._tcp.example.org. SRV 0 1 6697 irc.example.org.
```

kylef:
> IRC networks conforming to this specification MUST publish an SRV record with the "ircs" service label

The expected behaviour for a conforming network is very clear and explicit, however we do not cover what circumstances a conforming IRC client should query for a SRV record, or if they must etc always check it etc. The linked client implementation in soju uses heuristics such as the end user did not specify a port to use.

Should we strengthen up the desired client behaviours in the spec?

slingamn:
It seems appropriate to me to leave this implementation-defined. I had actually imagined this as a fallback that can be used when the initial connection attempt fails.

What would be gained by mandating a specific client behavior here?

kylef:
> What would be gained by mandating a specific client behavior here?

Clients can offer a consistent experience in regards to DNS SRV records. There isn't a suggestion, recommendation, or should clause at the moment.

For network or server operators making a decision if they want to add DNS SRV records, one consideration would be how they are used by clients and the net user experience which cannot be determined by this specification. Each client doing something different with the record may hamper the usability of the feature.

extensions/dns-srv.md
```diff
@@ -38,6 +38,10 @@ Example:
 
     _ircs._tcp SRV 0 1 6697 irc.example.org.
 
+After discovering an IRC server via SRV records, the client MUST check that the
+TLS server certificate is valid for the original hostname (`example.org` in the
+example above).
```
ksu-0, line 43: This is not very practical when `example.org` hosts other services than the IRC network. If the private certificate of IRC server is leaked, TLS communications of other services are compromised as well.

slingamn: I don't understand the concern here. Is the claim that IRC servers have a larger attack surface for compromises than, say, the HTTP servers or load balancers that would normally serve a certificate covering `example.org`?

ksu-0: Yes, attack surface of IRC server is much larger than of reverse proxy HTTP server.

slingamn: I don't agree, but an operator who felt this way could put a TLS-terminating reverse proxy in front of their IRC server. (For example, nginx supports this with the [stream](https://docs.nginx.com/nginx/admin-guide/security-controls/terminating-ssl-tcp/) directive.)

ksu-0: It could indeed be a good solution.

slingamn:
@kylef I take your point, but, given that there is no clear candidate for a recommendation, a SHOULD is toothless and a MUST incentivizes implementations that disagree with the MUST to ignore the specification altogether.

Maybe we should start collecting potential client behaviors so that either (a) one of them could be selected as a SHOULD (b) they could all be listed in a non-normative section?

slingamn:
I guess I should state my own priorities: I care a lot about the handshake being fast, so I don't want to add a recommendation (even a SHOULD) that would mandate a SRV lookup (even with caching) in the case where the client is already "correctly" configured.

extensions/dns-srv.md
aaronmdjones: > This is not very practical when `example.org` hosts other services than the IRC network. If the private certificate of IRC server is leaked, TLS communications of other services are compromised as well.

As explained above; if you do not enforce that the certificate is valid for the original hostname (the one the client looked up), then a trivial DNS MITM is possible, to direct the client to a hostname of the attacker's choosing, defeating the entire point of using TLS.

grawity: > This is not very practical when `example.org` hosts other services than the IRC network.

It seems to be practical enough for XMPP server operators. XMPP also requires the certificate to validate against the domain from JID, not against the SRV target.

kylef: Just throwing an idea up there (haven't given it much thought). Could the TLS hostname validation use the SRV hostname instead (for exampole `_ircs._tcp.example.org`). That way we would be restricting the certificate for only use with the irc service and not any other service running in the `example.org` domain (where comprimise of cert could allow it to be used for other services in the root domain).

ksu-0: While you can't have underscores in dNSName, RFC 4985 exists as an alternative. Sadly, Baseline Requirements of CA/B Forum don't allow issuance of SRVName yet, but there is a [discussion](https://github.com/cabforum/servercert/issues/268) to allow them.

edk0:
FWIW we've added a basic SRV record for libera.chat, which is now also in our certificate SANs. I'd love to know if people are finding this useful. We might consider using the load balancing features of SRV instead of just pointing to the round robin if a significant number of clients intend to respect them.

```console
$ dig +short srv _ircs._tcp.libera.chat
0 1 6697 irc.libera.chat.
```

It's also duplicated on irc.libera.chat, which I'm less sure about, but it'd be nice to get away from well-known ports altogether.

As for the cert debate, MTAs have been relying on DNSSEC for this forever. Can we do the same? (But I also really don't mind just using the original hostname for validation if that's what everyone wants)

extensions/dns-srv.md
emersion: [RFC 6125 section 6](https://datatracker.ietf.org/doc/html/rfc6125#section-6] explains that the certificate must be valid for the domain provided by the user, regardless of any DNS SRV records.

emersion:
> Also, why not standardize a SRV record for unencrypted connections?

Unencrypted connections should never be used. This spec is optional, users can always configure their client if they really want unencrypted connections. (I recommend clients should never attempt unencrypted connections by default unless explicit opt-in by the user.)

Celelibi:
> > Also, why not standardize a SRV record for unencrypted connections?
> 
> Unencrypted connections should never be used. This spec is optional, users can always configure their client if they really want unencrypted connections. (I recommend clients should never attempt unencrypted connections by default unless explicit opt-in by the user.)

I totally agree with not trying unencrypted connections first. But I wouldn't be as strict about denying unencrypted fallback.
To me, using this argument for not defining `_irc.tcp` records sounds like *"I can't figure a use case, therefore I forbid it"*.
So here are some use cases in addition to temporary outages and misconfigurations:
- implementing a lax policy allowing unencrypted connections to join only `#help`;
- not caring about encryption because the server is only accessible on a trusted LAN anyway;
- not caring at all about encryption because you think it should be only the users' concern.

I am currently in the third case, and might move to the first one in the coming years.

IRC is usually pretty low stakes. Most channels are public and the server can log every message. I personally think it's up to the network admins and client developers to decide on an encryption policy.
But such policies seem *waaay* out of scope for this spec.

That's why I would prefer to see the `_irc._tcp` records specified, even if their omission is possible. So that at least there's a common ground for network admins and client implementations to work on when this feature is needed. My recommendation would actually be that:
- clients support this record and make use of it as an opt-in fallback (possibly interactive opt-in);
- networks implement this record only if a valid unencrypted connection can be made (not one that kills the user immediately)

emersion:
Sorry, but defining `_irc._tcp` is a hard no for me. This is not a  "I can't figure a use case, therefore I forbid it" case, this is a "I forbid it because I think allowing it is harmful" case.

SadieCat:
:+1: for not having plaintext records.

PR #484, opened 2021-12-13
Add named-modes specification

progval:
This is a spec that has been in the making for a while, and it's finally ready for review.

It is inspired by [InspIRCd's namedmode](https://docs.inspircd.org/3/modules/namedmodes/) module, though they are unfortunately not compatible.

In terms of implementation, I wrote a proof-of-concept as a [third-party InspIRCd 4 module](https://github.com/progval/inspircd-contrib/blob/namedmodes/4.0/m_ircv3_namedmodes.cpp) that I hope to productionize when this draft is accepted.

Finally, as this is a pretty complicated spec, please avoid sending copyediting comments right away, this is probably going to be a very long thread with many changes.

[rendered view](https://github.com/progval/ircv3-specifications/blob/named-modes/extensions/named-modes.md)

### DIFF
--- a/extensions/named-modes.md
+++ b/extensions/named-modes.md
@@ -50,11 +50,11 @@ These numerics MAY occur more than once. If the reply consists of multiple lines
 where `<type>` is one of the following that tells the client about the nature
 of the mode:
 
-* 1 - mode is a list mode, i.e. requires a parameter when setting and unsetting (group 1 in 005 CHANMODES).
-* 2 - mode is a parameter mode, i.e. requires a parameter when setting and unsetting (group 2 in 005 CHANMODES).
-* 3 - mode is a parameter mode, i.e. requires a parameter when setting and no parameter when unsetting (group 3 in 005 CHANMODES).
-* 4 - mode is a flag, i.e. requires no parameter (group 4 in 005 CHANMODES).
-* 5 - mode is a prefix mode, i.e. requires a parameter when setting and unsetting, target is always a user on the channel.
+* 1 - mode is a list mode that requires a parameter when setting and unsetting (group 1 in 005 CHANMODES).
+* 2 - mode is a parameter mode that requires a parameter when setting and unsetting (group 2 in 005 CHANMODES).
+* 3 - mode is a parameter mode that requires a parameter when setting and no parameter when unsetting (group 3 in 005 CHANMODES).
+* 4 - mode is a flag mode that requires no parameter (group 4 in 005 CHANMODES).
+* 5 - mode is a prefix mode that requires a parameter when setting and unsetting, target is always a user on the channel (any mode from 005 PREFIX).
 
 and `<letter>` is an equivalent mode name for the `MODE` command.
 
--- a/extensions/named-modes.md
+++ b/extensions/named-modes.md
@@ -1,0 +1,323 @@
+---
+title: IRCv3 Named Modes Extension
+layout: spec
+copyrights:
+  -
+    name: "Attila Molnar"
+    period: "2014"
+    email: "attilamolnar@hush.com"
+  -
+    name: "Sadie Powell"
+    period: "2017"
+    email: "sadie@witchery.services"
+  -
+    name: "Val Lorentz"
+    period: "2021"
+    email: "progval+ircv3@progval.net"
+
+---
+
+## Intro
+
+The name of this client capability MUST be named `draft/named-modes`.
+
+This capability enables receiving PROP messages instead of MODE messages as well as introducing the `RPL_CHMODELIST` and `RPL_UMODELIST` numerics to replace the user and channel mode parameters of the `RPL_MYINFO` numeric.
+
+## Motivation
+
+The IRC protocol originally had a `MODE` command to configure channels, using single character as configuration keys.
+Implementation extend the base protocol with their own modes, causing some of them to exhaust the namespace available. Additionally, independent implementations can accidentally implement the same feature with different characters, or use the same characters for different features.
+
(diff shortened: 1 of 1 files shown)


progval:
A possible change: currently, the `RPL_LISTPROPLIST` syntax is like this:

```
RPL_LISTPROPLIST <nick> <channel> <modename> <mask> [<setter> <settime>]
```

to mirror the existing behavior, and to make it easier to handle optional parameters.

However, if we change it to this:

```
RPL_LISTPROPLIST <nick> <channel> <modename> [<setter> <settime>] <mask>
```

It would allow spaces in the mask, as it is always the trailing parameters

I don't want to do allow spaces in this spec (to make it easy to convert to MODE), but we may want to change this order so that we can easily allow spaces later in a new spec.

What do you think?

extensions/named-modes.md
```diff
@@ -0,0 +1,356 @@
...
+### RPL_CHMODELIST
+
+    :<server name> XXX <nick> [*] {<type>:<modename>[=<letter>]}+
+
+where `<type>` is one of the following that tells the client about the nature
+of the mode:
+
+* 1 - mode is a list mode, i.e. requires a parameter when setting and unsetting (group 1 in 005 CHANMODES).
+* 2 - mode is a parameter mode, i.e. requires a parameter when setting and unsetting (group 2 in 005 CHANMODES).
+* 3 - mode is a parameter mode, i.e. requires a parameter when setting and no parameter when unsetting (group 3 in 005 CHANMODES).
```
jwheare, line 55: This reads strange. "a parameter mode" is being defined in two different ways but with the same nomenclature. Can we add something to differentiate them before the "i.e."?

Maybe just replace ", i.e." in all 5 of these with "that". e.g.

> * 2 - mode is a parameter mode that requires a parameter when setting and unsetting (group 2 in 005 CHANMODES).

extensions/named-modes.md
```diff
@@ -0,0 +1,356 @@
...
+* 4 - mode is a flag, i.e. requires no parameter (group 4 in 005 CHANMODES).
+* 5 - mode is a prefix mode, i.e. requires a parameter when setting and unsetting, target is always a user on the channel.
```
jwheare, line 57: Maybe add this at the end of the line:

> (any mode from 005 PREFIX)

slingamn:
Any thoughts on requiring that servers MUST translate MODE to PROP for clients that have enabled the capability? This opens up the possibility of future greenfield clients that don't know about MODE at all.

progval:
@slingamn maybe a separate cap to allow gradual upgrades?

progval:
On second thought, nevermind. We don't need one more moving part.

slingamn:
Here's an edgy take: I think we should preemptively make this specification syntactically tolerant of spaces in mode parameters. (They can still be semantically disallowed, especially for named modes that are the equivalent of an RFC1459 mode.)

"Non-final parameters" are an ever-flowing wellspring of correctness issues. The backwards compatibility issues don't seem too bad: since the `MODE` parameters to `+k` and `+b` can already be sent by clients as IRC final parameters, server implementations already need to validate them to ensure that they don't contain spaces.

@emersion might have been joking but at one point he suggested using the [tag value escaping syntax](https://ircv3.net/specs/extensions/message-tags.html#escaping-values) for this. What if we required all mode parameters --- in `RPL_PROPLIST`, `RPL_LISTPROPLIST`, and `PROP` --- to be tag-value-escaped?

grawity:
> Here's an edgy take: I think we should preemptively make this specification syntactically tolerant of spaces in mode parameters. (They can still be semantically disallowed, especially for named modes that are the equivalent of an RFC1459 mode.)
> 
> "Non-final parameters" are an ever-flowing wellspring of correctness issues. The backwards compatibility issues don't seem too bad: since the `MODE` parameters to `+k` and `+b` can already be sent by clients as IRC final parameters, server implementations already need to validate them to ensure that they don't contain spaces.
> 
> @emersion might have been joking but at one point he suggested using the [tag value escaping syntax](https://ircv3.net/specs/extensions/message-tags.html#escaping-values) for this. What if we required all mode parameters --- in `RPL_PROPLIST`, `RPL_LISTPROPLIST`, and `PROP` --- to be tag-value-escaped?

Sounds like an interesting idea. I would actually like the entire wire protocol to move towards something like that, leaving the `:trailing` wart in history.

progval:
What about a single PROP change per line, so we can always use the space character instead? It can simplify implementations on both ends and doesn't have much of an overhead

slingamn:
I don't have strong feelings about this but it seems suboptimal.

I feel like everyone who would implement this has ready access to a tag-value (un)escaping implementation, no?

progval:
It mostly just "feels wrong" to me. But why not. It may not be usable as-is everywhere though; as current implementations can assume serialized values don't contain `;`, but otherwise sure.

Though I also like "single PROP change per line" because it simplifies implementations on both ends.

No strong opinion either way.

ValwareIRC:
Maybe bit silly request, any chance to make this use hyphen-delimited names on things that might contain spaces? Only rationale here is being able to easily automate a human-readable name from it by uppercasing the first char of each hyphen-delimited token, replacing the hyphens with spaces and (in my case) moving any vendor prefix to a badge

```
// Would be nice
"no-color" => "No Color"
"invite-only" => "Invite Only"

// Current when trying
"nocolor" => "Nocolor"
"inviteonly" => "Inviteonly"
```

It's only a little extra UI work to automatically recognize and display human readable for standardized named-modes, but thought it'd be nice and less work in the future if more things ever got standardized

Has the potential to get a little out of hand with like `no-external-messages` but I don't mind long mode names like that. Thoughts?

image

progval:
That only works for English-only clients, though.

ValwareIRC:
> That only works for English-only clients, though.

This is very true... Still fine for clients with client-side automatic translation of things

PR #347, opened 2018-11-04
Add the read client-tag specification

MuffinMedic:

PR #489, opened 2022-03-04
Add read-marker extension

emersion:
This is adapted from the [soju extension](https://git.sr.ht/~emersion/soju/tree/master/item/doc/ext/read.md). Ergo has expressed interest in standardizing this extension.

- Are the security/privacy guarantees enough?
- There is no way to mark a target as unread (ref https://todo.sr.ht/~emersion/soju/186)

Related, but different: https://github.com/ircv3/ircv3-specifications/pull/347


jwheare:
Worth making it more explicit that this is for bouncers or bouncer-integrated servers. The command and cap names are too general at the moment and suggest that this is for read receipts between different users.

emersion:
This isn't just for bouncers, it's for any server supporting multiple connections with the same nick. Or is this what you're calling "bouncer-integrated servers"?

jwheare:
Yes

emersion:
Okay. I don't really like using the term "bouncer" here -- Ergo can't connect to other servers for instance. But yeah, they use the term "bouncer-like features" to describe this in their README. I wonder if we can find a better name to describe this category of servers? Or maybe we could go with a name like `self-read`?

progval (commented):
Nice spec, thanks again.

Name bikeshedding: I'd rather have a more explicit name for the cap, such as `draft/read-marker`, so it is not confused with a potential `draft/read-receipt` specification. (And you really need to rename the PR and the spec title, #347 already uses the name "read receipt")

The command name is fine though, as a read-receipt spec would most likely either build upon this command or use client tags.

slingamn:
>But yeah, they use the term "bouncer-like features" to describe this in their README.

This is tricky. I used "bouncer functionality" and "bouncer-like features" in the README to describe the combination of three different features:

1. Server-side chat history
2. Multiple connections per nickname: the [end user guide](https://github.com/ergochat/ergo/blob/master/docs/USERGUIDE.md#multiclient), operator manual, config files, online help, etc. refer to this specific functionality as "multiclient"
3. Remaining present on the server in the absence of an active TCP connection: the end user guide etc. refer to this as "always-on"

I think "multiclient" is probably a suboptimal name for this, but I don't really have a better candidate right now.

DanielOaks:
I like `draft/read-marker`, feels like it makes sense.

wrt positioning, some words like this should do the trick: "This is for bouncers or servers providing chat history. It lets chat history providers know what a user has read, and provide that to other clients that user has, so that they can adjust their behaviour accordingly (e.g. not send highlight notifications for messages that have already been read by the user). It's a local read marker, not sent between different users."

extensions/read.md
```diff
@@ -0,0 +1,145 @@
...
+~~~~
+[c] READ target
+[s] :irc.host READ target timestamp=2019-01-04T14:33:26.123Z
+~~~~
+
+## Use Cases
+
+Clients can know whether a user has already read newly received messages. For clients that display notifications about new messages or highlights, knowing when messages have been read can enable them to clear notifications for messages that were already read on another device.
+
+Clients never have to actively get the read timestamp because it is provided to them on join and as updated by the server, except for user targets where they have to request the initial read timestamp by sending a `READ` client get command.
```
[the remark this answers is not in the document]
emersion: Moved into the "implementation" section. The introduction is a high-level overview and doesn't describe the command.

extensions/read-marker.md
```diff
@@ -0,0 +1,145 @@
...
+
+The server MUST reply to a successful `MARKREAD` get command using a `MARKREAD` server command, or using an error message.
+
+#### `MARKREAD` server command
+
+When sent from a server, the `MARKREAD` command signals to the client that the last message read by the user, to the best knowledge of the server, has the specified timestamp. In that case, the command has the following syntax:
+
+    <prefix> MARKREAD <target> <timestamp>
+
+The `prefix` is the prefix of the client the message is sent to.
```
slingamn, line 76: Just noticed this:

1. `<prefix>` is syntactically the source/prefix of the IRC message, correct? e.g. `:n!u@h MARKREAD #test 2022-03-31T01:32:21.380Z`
2. I don't understand the purpose of specifying the prefix here. The prefix typically indicates the origin of the message, not the recipient, right? My sense is that this should not be specified, and server implementations should probably use the server name as the source/prefix of these messages.

slingamn:
I have a draft implementation of this at ergochat/ergo#1926; it's on my staging server at `ircs://files.stronghold.network:6697` . I also have a rough irctest covering the server behavior (I'll post that once the implementation is merged).

extensions/read-marker.md
emersion: Originally this spec was introducing a `READ` command, and we figured that the same command might get re-used by a future spec to send read receipts between users. In that case it would've made sense in the new spec to allow the source/prefix to be different from the current user's. Note, other commands like `JOIN` indicate success by having the server reply with a `JOIN` message with the current user's source/prefix.

Although I agree it's a bit premature at this point, we're not even sure it's a good idea to re-use the same message for read receipts. I'll drop this requirement.

slingamn: That makes sense, thanks. I support dropping the requirement for now.

My sense is that read receipts would be better implemented as TAGMSG.

emersion:
All done!

slingamn:
My bad actually, I missed this language in the earlier draft, which is still present and covers my concern about no-op updates:

>The last read timestamp of a target SHOULD only ever increase. If a client sends a MARKREAD command with a timestamp that is below or equal to the current known timestamp of the server, the server SHOULD reply with a MARKREAD command with the newer, previous value that was stored and ignore the client timestamp.

so now it's a MUST in one section and a SHOULD in another. Some consolidation might be in order? Either MUST or SHOULD is fine by me.

emersion:
Hm, it might be handy in the future to allow the timestamps to decrease when the user manually marks some messages as unread. However this should not be done by just accepting blindly `MARKREAD` commands with a lower timestamp, because of race conditions between clients. So let's switch this to a MUST in this spec. A future spec can always add another way to make the timestamps decrease.

emersion:
Gentle ping

slingamn:
An implementation of this was released in [Ergo 2.10.0](https://github.com/ergochat/ergo/releases/tag/v2.10.0). So I think that makes two server implementations (ergo and soju) and two client implementation (gamja and goguma)?

jwheare:
I don't think so. Seems to have a decent amount of approval. Will merge in a few days.

PR #490, opened 2022-03-15
fix ircv3/ircv3-ideas#73

slingamn:
Add `account-required` cap value to `draft/account-registration` as per
discussion here:

https://github.com/ircv3/ircv3-ideas/issues/73

### DIFF
--- a/extensions/account-registration.md
+++ b/extensions/account-registration.md
@@ -66,7 +66,7 @@ The only defined capability keys so far are:
    from the user's current nickname
  * `account-required` - if present, indicates that the connection to the
     server cannot be completed unless the client logs into an account with
-    SASL.
+    SASL
 
 ### Commands
 
--- a/extensions/account-registration.md
+++ b/extensions/account-registration.md
@@ -64,6 +64,9 @@ The only defined capability keys so far are:
    to process
  * `custom-account-name` - if present, the account name can be different
    from the user's current nickname
+ * `account-required` - if present, indicates that the connection to the
+    server cannot be completed unless the client logs into an account with
+    SASL.
 
 ### Commands
 
@@ -261,6 +264,13 @@ like a `PRIVMSG` to NickServ), which are limited in length, clients may want to
 prevent or discourage users from setting passwords so long they may not fit
 in these messages. 300 bytes should be a safe reasonable limit.
 
+The capability value `account-required` indicates that all connections require
+SASL to proceed. The absence of this value does not indicate that connection
+registration can necessarily proceed in the absence of SASL; it may be
+disallowed due to specific properties of the connection (e.g. an untrustworthy
+IP address), which will be indicated instead by `FAIL * ACCOUNT_REQUIRED`.
+If the capability value `before-connect` is advertised, clients should respond
+to both of these conditions by suggesting that the user register an account.
 
 # Server considerations
 


progval:
Wouldn't it make more sense as value on the `sasl` cap?

emersion:
That would work, but the `sasl` cap values are already used for mechanism names, so we'd be mixing up two different things here. There is a theoretical potential for a conflict between the additional cap values we define (here, `account-required`) and the SASL mechanism names. Not sure we should care, maybe this is fine.

progval:
mechanisms are in caps, this kind of value isn't

slingamn:
>This would allow servers to advertise that they require an account even if they don't support account registration.

This is actually covered by the existing semantics, I think: if you advertise `account-required` without advertising `before-connect`, then there is no way to register an account.

emersion:
> mechanisms are in caps uppercase, this kind of value isn't

That's a fair point.

>This is actually covered by the existing semantics, I think: if you advertise account-required without advertising before-connect, then there is no way to register an account.

Hm, right. It still feels a bit weird to advertise `account-registration` when neither REGISTER nor VERIFY are supported at all…

slingamn:
From IRC discussion, using the SASL cap value for this violates a MUST of the current ratified SASL 3.2 specification:

>The value, if present, MUST be a comma (,) separated list of SASL authentication mechanisms.

since `account-required` is not a mechanism.

>It still feels a bit weird to advertise account-registration when neither REGISTER nor VERIFY are supported at all…

It's a little weird. Here's how I'm thinking about it: the SASL spec covers, narrowly, how to tunnel SASL authentication inside the IRC C2S protocol. This spec covers policy issues related to accounts (whether they can be registered, what is required to register them, whether they're required to connect, etc.).

emersion:
@jwheare, do you have any preferences here?

jwheare:
I feel like this shouldn't be tied to either the SASL or draft/account-registration capabilities. It's not tied to the AUTHENTICATE, REGISTER or VERIFY flows. It *might* be followed up by a REGISTER flow but that's not mandatory.

I'm not sure I understand where the issue is with just using the FAIL ACCOUNT_REQUIRED error code. Catering to a use case where a client has decided to only pre connect for a list of CAPs to cache and then disconnect  seems odd to me. Surely such clients could cache the presence of the FAIL code too?

progval:
@jwheare They don't get the FAIL code until they have sent NICK+USER+CAP END.

emersion:
@jwheare My client displays a connection UI with the following fields: "server", "nickname", "password (optional)". What I'd like to do is peek at the server caps once the "server" field is filled, and:

- Hide the "password (optional)" field if the server doesn't support SASL
- Make the "password (optional)" field mandatory if the server advertises that authentication is required

because it's confusing for users to see "password (optional)" when connecting to a server where the password is mandatory.

emersion:
@slingamn, would you be fine with that?

slingamn:
I have no objection, but I think I'll leave it to someone else to draft :-)

kylef:
>  I'm still not sure if I'm fully on board for the necessity for this

I'm also in a similar boat, would an appropriate numeric or standard reply give you the same information pretty quickly during your registration attempt that having an additional capability is not needed?

For example:

```
C: CAP LS 302
C: NICK kylef
C: USER a b c d
S: CAP responses ...
S: ... registration is required ...

# user fixes registration and eventually triggers CAP END
```

emersion:
The goal here is to figure out whether authentication is required _before_ connection registration. User enters hostname to connect to, client performs a `CAP LS`, checks for `sasl`/`account-required`, and then shows/hides the password field accordingly. I feel like most people skeptical about this new cap don't really understand the use-case? I can post UI screenshots if that helps.

kylef:
@emersion In my example above, the client would receive the information that client is received alongside the CAP LS output. The client has that informtion about the same time, doesn't this suffice for your use-case?

emersion:
I don't know which nickname/username/realname the user wants to pick at that point. I can't send `NICK` nor `USER`. I only know about the hostname, nothing else.

emersion:
I've posted some screenshots at https://github.com/ircv3/ircv3-specifications/pull/492#issuecomment-1079659128.

PR #493, opened 2022-03-31
Add client consideration for labeled-response (batch vs non-batch)

awfulcooking:
it's difficult to know which will be generated for a given request.

It might depend on server configuration, user modes, or in some cases, may change as a result of future extensions.

Clients may therefore prefer to handle all labeled responses as if they were a BATCH, even in the single message case, for robustness and compatibility.


progval:
That's implied by "If a response consists of more than one message, a batch MUST be used to group them into a single logical response"; but I guess it doesn't hurt to say it explicitly here.

PR #494, opened 2022-04-02
Add `draft/oper-tag` spec

examknow:
This is based on the solanum [implementation](https://github.com/solanum-ircd/solanum/blob/main/extensions/cap_oper.c). The solanum spec can be found [here](https://solanum.chat/oper).

### DIFF
--- a/extensions/oper-tag.md
+++ b/extensions/oper-tag.md
@@ -27,7 +27,7 @@ The tag MUST be named `draft/oper`. The value of the tag, if specified, MUST ide
 The tag MUST be added by the ircd to all commands sent by a user (e.g. PRIVMSG,
 MODE, NOTICE, and all others) and numeric replies sent on behalf of the user.
 
-Servers supporting this capabilty MAY be configured to restrict visibility of this tag or its value.
+Servers supporting this capability MAY be configured to restrict visibility of this tag or its value.
 
 ## Example
 
--- a/extensions/oper-tag.md
+++ b/extensions/oper-tag.md
@@ -25,7 +25,7 @@ The `draft/oper` capability causes the server to add a [message tag][] to messag
 The tag MUST be named `draft/oper`. The value of the tag, if specified, MUST identify the operator.
 
 The tag MUST be added by the ircd to all commands sent by a user (e.g. PRIVMSG,
-MODE, NOTICE, and all others) and numeric replies sent on behalf of the user.
+MODE, NOTICE, and all others) and SHOULD be added to any numeric replies sent on behalf of the user.
 
 Servers supporting this capability MAY be configured to restrict visibility of this tag or its value.
 
--- a/extensions/oper-tag.md
+++ b/extensions/oper-tag.md
@@ -1,0 +1,40 @@
+---
+title: "`oper-tag` Extension"
+layout: spec
+work-in-progress: true
+copyrights:
+  -
+    name: "David Schultz"
+    period: "2022"
+    email: "me@zpld.me"
+---
+
+## Notes for implementing work-in-progress version
+
+This is a work-in-progress specification.
+
+Software implementing this work-in-progress specification MUST NOT use the
+unprefixed `oper` tag name. Instead, implementations SHOULD use the
+`draft/oper` tag name to be interoperable with other software
+implementing a compatible work-in-progress version.
+
+## Description
+
+The `draft/oper` capability causes the server to add a [message tag][] to messages sent by a user who is currently an IRC operator.
+
+The tag MUST be named `draft/oper`. The value of the tag, if specified, MUST identify the operator.
+
+The tag MUST be added by the ircd to all commands sent by a user (e.g. PRIVMSG,
+MODE, NOTICE, and all others) and numeric replies sent on behalf of the user.
+
+Servers supporting this capabilty MAY be configured to restrict visibility of this tag or its value.
+
+## Example
+
+Consider that I am identified to the `launchd` operator account while chatting with the `lunchd` nickname. I will send messages to three users with different configured levels of visibility.
+
+    @draft/oper=launchd :lunchd PRIVMSG full_access_friend :it's me!
+    @draft/oper :lunchd PRIVMSG partial_access_friend :you can see I'm an oper
+    :lunchd PRIVMSG unprivileged_friend :don't mind me
+
+[message tag]: ../extensions/message-tags
--- a/extensions/oper-tag.md
+++ b/extensions/oper-tag.md
@@ -20,7 +20,7 @@ implementing a compatible work-in-progress version.
 
 ## Description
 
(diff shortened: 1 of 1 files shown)


fo1981 (approved):

fo1981 (approved):

emersion:
It would be nice to elaborate on the use-case of this spec. `WHO` flags (in particular the `*` flag) already provides a way to check whether someone is an operator.

What is the value-add of this spec?

awfulcooking:
@emersion It's a blessing as a bot author. Instead of maintaining WHO state, and sending one-off WHO requests, we can check for this tag synchronously when processing commands.

extensions/oper-tag.md
```diff
@@ -0,0 +1,40 @@
...
+    name: "David Schultz"
+    period: "2022"
+    email: "me@zpld.me"
+---
+
+## Notes for implementing work-in-progress version
+
+This is a work-in-progress specification.
+
+Software implementing this work-in-progress specification MUST NOT use the
```
awfulcooking, line 16: Given the simplicity of this tag, ircv3/ircv3-ideas#78, and the time spent already in production as solanum.chat/oper, I propose we save everyone time and just call this the `oper` tag.

It can still be refined in future with CAP or tag values.

emersion:
> It's a blessing as a bot author. Instead of maintaining WHO state, and sending one-off WHO requests, we can check for this tag synchronously when processing commands.

Okay, so ease-of-use is the main motivation for this, just like `account-tag` and `bot`. Maybe it would be worth including the motivation in the spec description, to make it clear what problem the spec is trying to solve?

emersion:
One value-add from this spec is when a client needs to figure out on JOIN whether a user is an operator or not. Right now the only way is to WHO all users who JOIN, which doesn't work in practice because of rate limits. If an oper tag is attached to JOIN messages clients would be able to properly keep their state up-to-date on JOIN.

extensions/oper-tag.md
```diff
@@ -0,0 +1,40 @@
...
+implementing a compatible work-in-progress version.
+
+## Description
+
+The `draft/oper` capability causes the server to add a [message tag][] to messages sent by a user who is currently an IRC operator.
+
+The tag MUST be named `draft/oper`. The value of the tag, if specified, MUST identify the operator.
+
+The tag MUST be added by the ircd to all commands sent by a user (e.g. PRIVMSG,
+MODE, NOTICE, and all others) and numeric replies sent on behalf of the user.
```
emersion, line 28: Ref https://github.com/ircv3/ircv3-specifications/pull/505: maybe turn this into a SHOULD for other numerics.

ValwareIRC:
I really liked the sound of this so I PR'd a sort-of implementation to UnrealIRCd (https://github.com/unrealircd/unrealircd/pull/226), I hope this gets more reach.

ValwareIRC:
FWIW: I think this could include some additional oper information, like a potential operclass.
```
draft/oper=bobsmith,netadmin
```
or something similar. I'm not entirely sure how global operclasses are in the IRC world, but it's probably common enough to include catering for it :smiley:

SadieCat:
Is there any more interest in this? I'd like to add support for this to InspIRCd.

>FWIW: I think this could include some additional oper information, like a potential operclass.

I support this but I'd like to bikeshed `oper-role` as a more generic tag name for this as `operclass` may have a different meaning depending on your IRCd (e.g. in InspIRCd an oper class is a set of privileges and the equivalent of an Unreal oper class is an oper type).

SadieCat:
I've implemented this with the addition of my oper-role proposal: https://github.com/inspircd/inspircd/compare/master...SadieCat:inspircd:master+ircv3-oper-tag

jwheare:
Are we still keen to standardise this? Looks like decent amount of support from server devs. Sensible solutions to any open bikesheds welcome.

ValwareIRC:
how often are oper "classes" used in IRC servers? if it's a majority, I'd like to see an oper-class tag too

jesopo:
i think the concept of "oper classes" is so implementation-specific that it shouldn't be part of a protocol-wide standard

jesopo:
(i understand why it would be helpful, by the way, and maybe having another spec for it would make me feel better)

spb:
The only thing I don't like is the use of 'oper' as a capability name - it's too vague and doesn't describe what's actually being enabled by it.

jesopo:
would you be fine with `oper-tag`, congruent with `account-tag`?

spb:
Yes, that's what I'd call it.

jesopo:
@examknow are you still around enough to make changes to this PR? if not i can take it on

SadieCat:
> i think the concept of "oper classes" is so implementation-specific that it shouldn't be part of a protocol-wide standard

I would like to see the `oper-role` proposal I made included but if there's nobody else who is interested I'm fine with including it under a vendor tag.

jesopo:
> > i think the concept of "oper classes" is so implementation-specific that it shouldn't be part of a protocol-wide standard
> 
> I would like to see the `oper-role` proposal I made included but if there's nobody else who is interested I'm fine with including it under a vendor tag.

could we do it on another spec, so that implementing this spec doesn't require implementing `oper-role`? i'm not saying a hard no here, but i have a slight preference to doing it this way. more than happy to be outvoted

spb:
`oper-role` can probably be incorporated with no real negative impact - something like "the server MAY also attach a `draft/oper-role` tag containing information about the user's role in network operations, such as a privilege class" puts no obligation on anyone to do anything except not crash when they see that tag but standardises the tag name for anyone who does care.

emersion:
What upside is there to have a standard tag with implementation-defined contents? What would clients do with this information?

spb:
It doesn't have to have a fixed meaning across ircds to be useful - a bot framework might offer it as a variable that can be matched against when defining custom access controls, for example.

sadiepowell:
I'm happy to see this draft merged as-is currently, we can discuss the oper-role stuff later.

extensions/oper-tag.md
```diff
@@ -0,0 +1,40 @@
...
+This is a work-in-progress specification.
+
+Software implementing this work-in-progress specification MUST NOT use the
+unprefixed `oper` tag name. Instead, implementations SHOULD use the
+`draft/oper` tag name to be interoperable with other software
+implementing a compatible work-in-progress version.
+
+## Description
+
+The `draft/oper` capability causes the server to add a [message tag][] to messages sent by a user who is currently an IRC operator.
```
dgw:
```suggestion
The `draft/oper-tag` capability causes the server to add a [message tag][] to messages sent by a user who is currently an IRC operator.
```

dgw (commented):
Besides the oper-role idea, I would say don't merge this with `draft/oper` as the capability name. The PR title says `draft/oper-tag` and IMO that should be the WIP cap name.

The _tag name_ `draft/oper` works.

jwheare:
Will aim to merge this next week.

PR #495, opened 2022-04-12
Ratify bot-mode

progval:
## Known implementations

### Server (3/2)

* [x]  InspIRCd: (@SadieCat ) ircs://testnet.inspircd.org:6697
* [x]  Ergo (@slingamn) ircs://testnet.oragono.io:6697
* [x] UnrealIRCd (@syzop) ircs://irc.unrealircd.org:6697

### Client (2/1)

* [x]  IRCCloud client (@jwheare)
* [x]  Goguma (@emersion)

### Bouncer

* [x] Soju (@emersion)

### Bots

* [x] Limnoria (@progval)
* [x] Sopel (@dgw)

### Networks

* [x]  https://ergo.chat
* [x] irc.unrealircd.org

### DIFF
--- a/extensions/bot-mode.md
+++ b/extensions/bot-mode.md
@@ -10,7 +10,10 @@ copyrights:
 ---
 
 ## Introduction
-This is a standardised mode that lets clients mark themselves as bots, and lets other clients see them as bots.
+This specification defines a standardised mechanism to allow clients to mark themselves as bots.
+
+## Motivation
+This functionality was previously implemented by several IRC servers in independent ways using fixed mode letters and conflicting WHOIS numerics. This specification seeks to standardise the functionality, adding a message tag for increased visibility of bot status and an ISUPPORT token to allow for more flexible implementations.
 
 ## The `BOT` ISUPPORT token
 Servers publishing the `BOT` [ISUPPORT](https://modern.ircdocs.horse/#feature-advertisement) token let clients mark themselves as bots by setting a user mode. The value of the `BOT` token is the mode character which is used to enable bot mode and is also the flag used in `WHO` responses of bots (e.g. `BOT=B` or `BOT=b`).
--- a/extensions/bot-mode.md
+++ b/extensions/bot-mode.md
@@ -1,7 +1,6 @@
 ---
 title: Bot Mode
 layout: spec
-work-in-progress: true
 meta-description: A user mode that indicates a user is a bot.
 copyrights:
   -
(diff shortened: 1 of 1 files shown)


dgw:
Sopel implements both the mode (sopel-irc/sopel#2088) and tag (sopel-irc/sopel#2089).

Yes, it accepts the unprefixed tag. Figured this spec would be ratified long before we made it to release, and it seems I'm about to be proven correct.

SadieCat:
I can push my support for this the moment its ratified. 👍🏻

kylef (approved):

SadieCat:
Is there any objections to ratifying this? I'd like it if I could get our implementation into InspIRCd v3.13.0 which is due to be released next Friday.

jwheare:
There was an objection raised in IRC from solanum/libera staff. As I understand it, they felt that IRCv3 shouldn't be pursuing specs that feel like a stop gap for a true metadata implementation.

This spec doesn't define a mechanism for keeping up to date with a change in a user's bot status other than periodic WHOIS or WHO polling. True metadata notifications on status changes would allow clients to keep member lists in sync.

In reality, a real bot is unlikely to ever change status, so I feel like running an initial WHO check on channel join and checking for the tag on new user JOINs should be sufficient for the vast majority of use cases.

There are outstanding issues with PMs when no shared channel exists, but those issues extend to many other things, and a future METADATA spec won't necessarily cater to that situation either.

I acknowledge and agree that effort should be spent on reviving a true metadata spec, so that we can focus in that direction for future cosmetic user status things, . But I think for this specific spec, where there is prior art for non-standard mode/whois numerics indicating bot status, and we have a broad consensus of implementation, we should ratify this.

It might be worth mentioning the prior art in the spec, similar to the Motivation section in the SETNAME spec: https://ircv3.net/specs/extensions/setname.html#motivation

See initial discussion here: https://github.com/ircv3/ircv3-ideas/issues/43

extensions/bot-mode.md
```diff
@@ -10,17 +9,6 @@ copyrights:
...
-
-Software implementing this work-in-progress specification MUST NOT use the
-unprefixed `bot` tag name. Instead, implementations SHOULD use the
-`draft/bot` tag name to be interoperable with other software
-implementing a compatible work-in-progress version.
-
-The final version of the specification will use an unprefixed tag name.
-
 ## Introduction
 This is a standardised mode that lets clients mark themselves as bots, and lets other clients see them as bots.
```
jwheare, line 13:
```suggestion
This specification defines a standardised mechanism to allow clients to mark themselves as bots.

## Motivation
This functionality was previously implemented by several IRC servers in independent ways using fixed mode letters and conflicting WHOIS numerics. This specification seeks to standardise the functionality, adding a message tag for increased visibility of bot status and an ISUPPORT token to allow for more flexible implementations.
```

jwheare (changes requested):
How's this?

jwheare:
Thanks for feedback and implementations all.

Issue #497, opened 2022-04-19
Client-tag for specifying in which shared channel a private NOTICE should be displayed


delthas:
Many IRC bots respond to commands in a channel with a private NOTICE.

```
>>> PRIVMSG #foo :!help
<<< :bot NOTICE me :i can't help
```

This is a common practice that is done to prevent flooding a channel with messages that are only useful to a single user. But there is no way for the client to actually know that this message was meant to be sent as a response to a query in #foo.

Most clients have developed heuristics to match incoming NOTICEs with a corresponding channel context that makes the most sense. In some clients this is built-in, in some others these are third-party scripts.

The usual order in which these channels are matched with these heuristics is typically:
- the current channel shown in the client, if the sender is in that channel
- a channel in which the sender is present (this could be the last opened channel in the client, the first one by order, the last one in which the client sent a message, ...)
- optionally, the current channel even if the bot is not in it
- an actual private buffer with the sender

This is a bit of a hack and is a bit harder to work with on clients that use CHATHISTORY and do not necessarily have a "client context" (which buffer did we open last at the time of receiving) at hand. eg if I connect after 1 week to my bouncer and run CHATHISTORY, my heuristic may not return the same result as if I had just received the message "live".

Now that we have client-only tags, we could make the bot send the actual channel the message refers to, in a client-only tag. This is a client-only tag because it does not require any processing from the server. Client-only tags are blacklisted from some servers/networks, but this is fine because it is a best-effort piece of information.

```
>>> PRIVMSG #foo :!help
<<< @+channel=#foo :bot NOTICE me :i can't help
```

The heuristic now becomes:
- the channel specified in the channel tag, if the sender and the client are in that channel
- ... (the usual other heuristics)

There is no particular security issue (as long as the client checks that the sender is indeed in the channel it pretends to refer a message to).

The actual tag name is open to discussion. A short list:
- +channel
- +channel-context

Thoughts?

Relates to/closes: #163

progval:
You can use CPRIVMSG/CNOTICE (aka WHISPER) for this: https://github.com/evilnet/nefarious2/blob/master/doc/features.txt#L89

grawity:
> On networks which support it, you can use CPRIVMSG/CNOTICE (aka WHISPER) for this: https://github.com/evilnet/nefarious2/blob/master/doc/features.txt#L89

[...]

PR #498, opened 2022-04-22
Add @+channel-context

delthas:
See: https://github.com/ircv3/ircv3-specifications/issues/497

Known implementations:
- [ ] goguma (not merged)

---- 

cc @emersion


client-tags/channel-context.md
```diff
@@ -0,0 +1,60 @@
...
+
+This tag MUST be attached to NOTICE messages to users only. Clients MUST ignore this tag if it is not attached to a private NOTICE message or if its value is not a valid channel name.
+
+## Client implementation considerations
+
+This section is non-normative.
+
+When clients receive a private NOTICE message from a user with this tag, they can display that message in the buffer specified in the tag, as if it was a channel message.
+
+If possible, the client can check whether both the sender and itself were present in the specified channel at the time of the message, before adding the message to its buffer, in order to prevent the sender from showing messages in buffers he has not joined.
```
emersion, line 53: s/he/they/

emersion:
Looks good to me. I was a bit worried about security considerations, but I don't think we can really do better here. Ultimately clients are free to ignore/drop the tag.

How should this interact with CHATHISTORY? Should the messages be returned as part of e.g. `CHATHISTORY LATEST #chan`, or `CHATHISTORY LATEST bot`? Since this is a client-tag, then I guess the latter, but then these messages will get omitted when scrolling up channel buffers, then will magically appear when scrolling up the bot's buffer.

jesopo:
https://github.com/bitbot-irc/bitbot/pull/344

client-tags/channel-context.md
```diff
@@ -0,0 +1,60 @@
...
+
+Clients wishing to use this tag MUST negotiate the [`message-tags`](../extensions/message-tags.html) capability with the server.
+
+### Format
+
+The channel-context tag is sent by a client with the client-only prefix `+`. The value MUST be a valid channel name.
+
+    +draft/channel-context=<channel>
+
+This tag MUST be attached to NOTICE messages to users only. Clients MUST ignore this tag if it is not attached to a private NOTICE message or if its value is not a valid channel name.
```
progval, line 45: Could you allow PRIVMSG too? Some people prefer to configure their bots to use them instead of NOTICE.

client-tags/channel-context.md
```diff
@@ -0,0 +1,60 @@
...
+
+## Client implementation considerations
+
+This section is non-normative.
+
+When clients receive a private NOTICE message from a user with this tag, they can display that message in the buffer specified in the tag, as if it was a channel message.
+
+If possible, the client can check whether both the sender and itself were present in the specified channel at the time of the message, before adding the message to its buffer, in order to prevent the sender from showing messages in buffers they have not joined.
```
progval, line 53: No love for mode `-n`? :(

progval (commented):
Nice spec, thanks.

progval:
Implemented in Limnoria: https://github.com/progval/Limnoria/commit/109f938b0c9423885b8726c2d3f5dc238a4c2f30

This applies to all plugins using the standard reply facility and is gated behind the same feature switch as `+draft/reply`

client-tags/channel-context.md
ValwareIRC: FWIW: I think allowing PRIVMSG is a good idea for this reason, and I have written my unrealircd module to allow this tag on PRIVMSG, and would vote, if I could, for the tag to apply to PRIVMSG also.

client-tags/channel-context.md
ValwareIRC: in unrealircd, channels which record channelhistory propagate channel messages to other servers regardless of whether any of its users are on the channel.
I would like to take advantage of still being able to channel-context any `!help` replies to such a channel.
also, as val said, for mode `-n`

emersion: > channels which record channelhistory propagate channel messages to other servers regardless of whether any of its users are on the channel

Hm but messages with a `channel-context` tag wouldn't be sent to the whole channel, they'd be sent directly to a user?

emersion: Oh I see, that makes sense.

client-tags/channel-context.md
delthas: OK makes sense.

ValwareIRC:
> [Third-party UnrealIRCd implementation](https://github.com/ValwareIRC/valware-unrealircd-mods/tree/main/channel-context)

Actually this is being committed to source now, instead of third party (¹)[https://github.com/unrealircd/unrealircd-contrib/pull/64](²)[https://github.com/unrealircd/unrealircd/pull/205]

jwheare:
This is supported in IRCCloud now.

Try it in #limnoria-bots on ircs://testnet.oragono.io:6697

image

jwheare:
This is probably ok to merge. Does anyone have any strong bikesheddy thoughts around the name of the tag? This is clearly based on the precedent of the WHISPER IRCX feature, but I think I prefer channel-context as a more explicit name. Something like target-channel might be even more direct, but I don't feel strongly either way. A rename can wait for ratification stage anyway and doesn't need to block merge.

client-tags/channel-context.md
```diff
@@ -0,0 +1,58 @@
...
+
+    +draft/channel-context=<channel>
+
+This tag MUST be attached to PRIVMSG and NOTICE messages to users only. Clients MUST ignore this tag if it is not attached to a private PRIVMSG or NOTICE message or if its value is not a valid channel name.
+
+## Client implementation considerations
+
+This section is non-normative.
+
+When clients receive a private PRIVMSG or NOTICE message from a user with this tag, they can display that message in the buffer specified in the tag, as if it was a channel message.
```
kylef, line 51: Given the use case of the tag, would it make sense to add a constraint to only perform this when the recipient for the PRIVMSG target is the current nick. To prevent abuse from a sender using this tag in other channels?

jwheare: It already says "a private PRIVMSG"

jwheare:
Should there be a client implementation constraint that this should only be redirected to a channel if the sender is also a member of that channel?

progval:
it was present in the original version, but delthas removed it, following https://github.com/ircv3/ircv3-specifications/pull/498#discussion_r856352124

jwheare:
I guess that was already discussed and changed https://github.com/ircv3/ircv3-specifications/pull/498#discussion_r856352124

Kinda why I don't like force pushes within PRs so changes can be tracked more obviously. (Yes I know force pushes show a compare link)

jwheare:
The main concern with non-shared channel redirections is that it lets spammers clutter up your channel history, which is harder to clean up than just deleting a PM. /ignore is a thing but not always retroactive in all clients.

ValwareIRC:
Just to mention, the unrealircd implementation (in the next release) allows this behaviour only from U-Lines, so as to take into account cases like where ChanServ replies with channel-context about a channel that a botserv bot is in, but we still want to allow it to be displayed in the contexted channel =]

progval:
What about requiring `+draft/reply` with the id of a message in the channel? It's not perfect (message ids could be leaked), but it solves the spam issue

jwheare:
That would limit the usefulness of the feature I think. Sometimes you'd just want to whisper to someone without replying to another message.

jwheare:
Perhaps the risk and any possible recourse (e.g. a setting) would be worth mentioning in the non-normative section at least.

ValwareIRC:
could it be helpful to do something like this, but say for example you send a message to a bot/service outside of the channel about the channel, you include the channel context and msgid. and if the msgid in the reply and channel context match, then show it in the channel, if not, don't show it in the channel?
doesn't feel too complicated, maybe I am overlooking something?

ValwareIRC:
allows general filtering from people without 'access' to put things in that channel unless you have specifically asked something recently about it by specifying the context

jwheare:
@ValwareIRC that's kinda convoluted imo.

ValwareIRC:
@jwheare turns out I'm really bad at explaining things.
Check out this -slightly better- explanation I gave to @progval

jwheare:
> where we asked for the channel to be contexted

^ this is the part that's convoluted.

ValwareIRC:
Again, maybe I didn't explain correctly =]

maybe it is simpler to say, have the client remember (for a short period) a list of users it has messaged outside of the channel with the channel as channel-context. This message could go to ChanServ for example, or it could go to a user called Bob.

Upon the reply from the user outside of the channel with the channel-context, check to see if the channel and user match what we remember, and then show it or not show it.

I say "a short time", I would say only a few seconds, which limits showing these types of messages in the channel to entities which can reply quick enough to prove they have a pre-worded response (like ChanServ, not like Bob).

I don't think that's at all convoluted...

delthas:
> maybe it is simpler to say, have the client remember (for a short period) a list of users it has messaged outside of the channel with the channel as channel-context. This message could go to ChanServ for example, or it could go to a user called Bob.

It is up to each client to decide how to prevent abuse of this feature. I'd like the spec to enable clients to specify which channel the message should go to, in a best-effort manner, then let clients decide if they want to add specific logic for limiting when this can be used / for displaying a specific tooltip on this kind of messages. It doesn't have to be specced.


Any other thoughts on this spec?

emersion (approved):
LGTM

jwheare:
@delthas I agree we don't need to change any spec behaviour, and to let clients decide. But to help them decide, I still think we need to mention something more about the risk of abuse in client implementation considerations. See my earlier comment https://github.com/ircv3/ircv3-specifications/pull/498#issuecomment-1133774304

> Perhaps the risk and any possible recourse (e.g. a setting) would be worth mentioning in the non-normative section at least.

emersion:
Ping

progval (approved):

delthas:
@jwheare Updated according to your comment :grinning:

jwheare:
Sounds good. OK to merge I think.

PR #499, opened 2022-04-22
multiline: Specify behavior with respect to formatting.

progval:
This is the only way to format the fallback for multiline-unaware clients the same way as for multiline-aware clients


slingamn:
lgtm

jwheare:
I'm not sure. This puts a significant burden on clients when interpreting user input and composing a multiline message.

progval:
Update: weechat released v4.0 without changing its behavior with respect to formatting.

progval:
@FlashCode As implementer of the multiline spec, could you weigh in on this question?

JustAnotherArchivist:
I suppose so. Would a recommendation to insert formatting reset codes during such rearrangements (if needed) be reasonable?

Steering7253:
Amazing. 4 years old, no alternative has ever been proposed, languishing for no reason. How's that for consensus?

flashcode:
@Steering7253: thank you for offering your help!
We are listening to your suggestions.

Steering7253:
Sure, here's my suggestion: do something about this. Your spec is broken.

extensions/multiline.md
```diff
@@ -184,7 +186,9 @@ This section is non-normative.
 
 NOTE: In these examples, `<SPACE>` indicates a space character which would otherwise not be clearly visible.
 
-Client sending a multiline batch
+### Normal usage
+
+Client sending a mutliline batch
```
Herringway: tyop here (mutliline)

jwheare:
Have any client developers made an attempt at implementing this suggestion and have any thoughts?

skizzerz:
> What makes it easier for servers to insert `<0x0F>` than for receiving clients who implement `draft/multiline`?

imo it's not easier to do this server-side, doing it client-side would be easier because there'd be no concerns about impacting send queue and the like (plus a lot of extra overhead on the wire to add one character). However if this is stalled because clients don't want to go through the effort or we're afraid of client fragmentation with some following spec and others not doing the format reset, then there are a lot fewer servers in active development than clients, so having consistency would involve fewer people to do said implementation.

andymandias:
Implementing these suggestions in Halloy brings up a complication.  For spoiler text (and potentially monospaced text, depending on how the client handles it) two spans directly next to each other are not equivalent to one span covering the same text.  The former will reveal the spoiler for each span individually (on user interaction), while the latter will reveal the entire text at once. This means we can't just shove a reset between each received message if we want to preserve message intent, instead it's necessary to check the formatting at the end of the previous message and strip that from the beginning of the current message (if present).

slingamn:
As a server vendor, I'm very skeptical of any proposal that requires server modification of messages here: avoiding server modification was one of the core design goals of multiline in the first place.

progval:
@andymandias What you are saying is consistent the spirit of this PR. Can you suggest a better wording?

andymandias:
I agree it is consistent with the spirit of the PR.  I don't believe it to be necessarily prohibitive either, just an increase in complexity I felt was worth noting since it wasn't apparent to me on first read (and based on some of the earlier comments I suspect I wasn't the only one).  A note along the lines of "Clients will be expected to strip the formatting that carries over from the previous message in the batch in order to reproduce the original formatting spans (e.g. when creating a spoiler text span from multiple messages)." perhaps?

skizzerz:
> I agree it is consistent with the spirit of the PR. I don't believe it to be necessarily prohibitive either, just an increase in complexity I felt was worth noting since it wasn't apparent to me on first read (and based on some of the earlier comments I suspect I wasn't the only one). A note along the lines of "Clients will be expected to strip the formatting that carries over from the previous message in the batch in order to reproduce the original formatting spans (e.g. when creating a spoiler text span from multiple messages)." perhaps?

The goal is so that the text your client sends renders the same regardless of whether the receiving side supports multiline or not. In the fallback case, each line of the multiline batch (including multiline-concat lines) is rendered as its own independent line, and clients historically do not merge formatting between independent lines; it's reset for each line. The suggested wording in this PR serves to emulate that behavior for receiving clients which *do* support multiline.

As a sending client, you should therefore be repeating any formatting that needs to cross protocol line boundaries in each relevant line. As a receiving client, you are free to merge those sections together, and if you have suggested wording for that then feel free to propose it.

The option that doesn't work is "do nothing and interpret format codes across protocol line boundaries" because that makes it impossible to specify a rendering that works for all receiving clients.

PR #500, opened 2022-04-29
Reword 'Security considerations' to be less misleading about validation

progval:
Some servers (Unreal) do perform validation.


emersion:
+1. I think filtering/validation on the server side should be optional, not forbidden.

progval:
Note that it is already explicitly allowed earlier in the spec: "Servers MAY apply moderation to client-only tags using existing or newly specified modes or configuration."

PR #501, opened 2022-06-25
Add metadata-2 spec

progval:
As far as I am aware, it addresses all known issues with the current deprecated metadata spec (listed in #339).

It is based on @DanielOaks' latest version, which is itself based on #339. Notable changes since dan's version are:

* standard replies (thanks to @ValwareIRC)
* it's now a batch type instead of old-style RPL_ENDOFxxxx
* capability is renamed to `metadata-2`, since it is a breaking change compared to the deprecated spec
* metadata keys are now vendored to avoid clashes

[rendered](https://github.com/progval/ircv3-specifications/blob/metadata/extensions/metadata.md)


syzop:
Based on a quick chat on IRC i noticed that the position of UnrealIRCd on this isn't well known yet, so here we go:
I like the METADATA idea. @k4bek4be is also a big fan of it. Thing is, I felt the old specification had too many outstanding issues that i had no time to look into, issues which were already known by the time (hence the deprecation of the first spec). METADATA almost made it in UnrealIRCd 6.0.0. In fact, if you would look back in the git log for it you would see it the metadata module was actually in UnrealIRCd 6 but was taken out right before we went beta. So it's currently (still) a third party module from @k4bek4be.
So yes, we are a fan :D

Also to hijack this thread a bit. I have long been wondering why there is no language preference communicated by IRC clients to servers. Right now people are stuck with setting language at services, and then only that is localized. Better would be at the server level.. then we can communicate that to services.. and at the same time we can use it for localized things like serving HELPOP documentation in the local language at UnrealIRCd. Anyway, that was the old idea, and I see @progval had exactly that same idea as well at https://github.com/ircv3/ircv3-specifications/pull/424. I think there were even people before us, possibly more than 10 years ago with this too, credit doesn't really matter, just only saying it pops up once in a while in some head and then nothing happens.
Now that metadata is coming up I feel like the language preference thing would be ideal to put in METADATA. This way it is not only for servers (eg serving localized HELPOP and other strings) or services (like anope/atheme), but also bots can use the information to reply back to the user in their language preference. That's the benefit of METADATA.. it's public information, unless told otherwise.

What I also like about the current spec is that it's clear that servers will likely work with whitelists/blacklists, as that I think is how we would roll it out at unrealircd, with some popular values being allowed by default. I can also see us having some filters similar to like we currently have one for client tags. I know that is always controversial when I mention it in IRCv3 public, but I can see use cases like restricting avatar URLs to a certain domain and/or other filtering. Should probably add a `VALUE_INVALID` reply to `METADATA SET` for it.

I also like the new rate limiting options where the server basically controls that. Maybe I will have a comment about a small thingy or enhancement here or there but in general yeah I like it.

Not sure if @k4bek4be already mentioned it but he also permits METADATA SET on yourself (only yourself) in the unregistered phase. This makes total sense to me as well. Like with the language example, you could do the usual CAP REQ stuff for metadata and other features, then set the METADATA on yourself (like language), then do a SASL session and (provided server to server traffic handles this at the right moment) services could or would speak back to you with error messages in your language preference. And then finish with the CAP END.
Doing METADATA in the handshake is also useful for things that happen very quickly after the handshake, like the standard services notices you get post-MOTD, or the auto join stuff, etc, it is handy when all this information (like language) is already known at that time, otherwise the METADATA SETting is too late for various situations.
Of course, there come some obvious dangers with allowing commands in pre-auth stage, so have to be extra careful with that and possibly contain some things here and there but i think the pros outweight the cons.

progval:
> Anyway, that was the old idea, and I see @progval had exactly that same idea as well at https://github.com/ircv3/ircv3-specifications/pull/424.

Heh not really, that was about indicating the language of a PRIVMSG someone just wrote; not selecting what language you want.

> METADATA SET on yourself (only yourself) in the unregistered phase.

Interesting, I didn't consider this. I assume it uses `METADATA * SET <Key> <Value>` instead of `METADATA <Nick> SET <Key> <Value>`?

syzop:
Whoops. Sorry about that, indeed I have rectified that and now mention as related. Also did some more searching and found the LANGUAGE thingy from 2003/2004 from codemastr. Updated my post above. Anyway, just more for history :D.

>
>> METADATA SET on yourself (only yourself) in the unregistered phase.
>
> Interesting, I didn't consider this. I assume it uses METADATA * SET <Key> <Value> instead of METADATA <Nick> SET <Key> <Value>?

Yeah that sounds like a good syntax to me. Fits with the general pattern of using an asterisk in the handshake where the nick is not known yet. (As for if k4be currently does it like that I dont' know but i don't think that's terribly important ;D)

ValwareIRC:
FWIW: I could see both being implemented, the METADATA for being able to specify the language the server and other such service should speak to you in, and the message-tag for specifying that "this particular message is in this particular language" as a cute little addition.

if (when hopefully) this gets implemented I think it would make a great modern addition and would definitely be including support for it in my upcoming services package.

syzop:
The only thing that did raise some eyebrows is the Visibility field being implementation-defined. Visibility controls who gets to see something which is kinda related to security. I don't think it is good practice to be ambiguous about security, or in this case: to be ambiguous about whether to expose or not expose a value to other people.
I don't have any particular suggestions on how it then should be done though (simply because I have not thought about this at all) and I presume there was some thinking about it that you people did, that made you write it as implementation defined, maybe you can elaborate on those? Thanks!

k4bek4be:
> I assume it uses `METADATA * SET <Key> <Value>` instead of `METADATA <Nick> SET <Key> <Value>`?
I would use * instead of a nick always when setting own metadata. This way, pre-registration makes no difference.

Also, I suggested allowing pre-registration `METADATA SUB` too, to allow receiving metadata on channel history messages if channel autojoin is enabled (autojoin via services gives the client no chance to `SUB` before the join event). This way we can render channel history with all the data.

syzop:
Indeed you said that and you are right, doing SUB in pre-registration is useful too because of auto join, otherwise the client has to send SYNC again (even for a small channel). Not devastating but still. And, indeed, perhaps at some point people would want to include it in channel history or there is some other use case for it that we cannot think of right now. Better to get it right.

extensions/metadata.md
```diff
@@ -0,0 +1,740 @@
...
+    C: METADATA * UNSUB website website
+    S: :irc.example.com RPL_METADATAUNSUBOK modernclient :website website
+
+#### Subscribing to a key which requires privileges but without privileges
+
+    C: METADATA * SUB avatar secretkey website
+    S: FAIL METADATA KEY_NO_PERMISSION secretkey modernclient :You do not have permission to do that.
+    S: :irc.example.com RPL_METADATASUBOK modernclient :secretkey website
+    C: METADATA * SUBS
+    S: :irc.example.com RPL_METADATASUBS modernclient :secretkey website
```
Herringway, line 695: These should probably be avatar and website, since secretkey was already rejected

extensions/metadata.md
```diff
@@ -0,0 +1,740 @@
...
+
+    C: METADATA * SUB $invalid1 secretkey1 $invalid2 secretkey2 website
+    S: FAIL METADATA KEY_NO_PERMISSION secretkey1 modernclient :You do not have permission to do that.
+    S: FAIL METADATA KEY_INVALID $invalid1 modernclient :Invalid key
+    S: FAIL METADATA KEY_NO_PERMISSION secretkey2 modernclient :You do not have permission to do that.
+    S: FAIL METADATA KEY_INVALID $invalid2 modernclient :Invalid key
+    S: :irc.example.com RPL_METADATASUBOK modernclient :secretkey1 secretkey2 website
+    C: METADATA * SUBS
+    S: :irc.example.com RPL_METADATASUBS modernclient :secretkey1 secretkey2 website
+
```
Herringway, line 707: just website, since secretkey1 and secretkey2 were rejected

extensions/metadata.md
```diff
@@ -0,0 +1,740 @@
...
+    S: @batch=gWkCiV RPL_KEYNOTSET user1 blargh :No matching key
+    S: @batch=gWkCiV RPL_KEYNOTSET user1 splot :No matching key
+    S: @batch=gWkCiV :irc.example.com RPL_KEYVALUE user1 im.xmpp * :user1@xmpp.example.com
+    S: :irc.example.com BATCH -gWkCiV
+
+#### Client joins a channel and syncs metadata immediately
+
+    C: JOIN #smallchan
+    S: modernclient!modernclient@example.com JOIN #smallchan
+    S: :irc.example.com 353 modernclient @ #smallchan :user1 user2 user3 user4 user5 ...
```
Herringway, line 472: missing : on the prefix here

extensions/metadata.md
```diff
@@ -0,0 +1,740 @@
...
+    S: :irc.example.com BATCH -UwZ67M
+
+#### Client joins a channel and but needs to sync metadata later
+
+To-do: Think of a better sync-later flow
+
+Client joins channel:
+
+    C: JOIN #bigchan
+    S: modernclient!modernclient@example.com JOIN #bigchan
```
Herringway, line 492: Another missing : on the prefix

extensions/metadata.md
```diff
@@ -0,0 +1,740 @@
...
+| Label                              | GET | LIST | SET | CLEAR | SUB | UNSUB | SUBS | SYNC | Other   |
+| ---    | :-: | :-: | :-: | :-: | :-: | :-: | :-: | :-: | :-: |
+| `RPL_WHOISKEYVALUE`                |     |      |     |       |     |       |      |      | `WHOIS` |
+| `RPL_KEYVALUE`                     | *   | *    | *   | *     |     |       |      |      |         |
+| `RPL_KEYNOTSET`                    | *   |      |     |       |     |       |      |      |         |
+| `RPL_METADATASUBOK`                |     |      |     |       | *   |       |      |      |         |
+| `RPL_METADATAUNSUBOK`              |     |      |     |       |     | *     |      |      |         |
+| `RPL_METADATASUBS`                 |     |      |     |       |     |       | *    |      |         |
+| `RPL_METADATASYNCLATER`            |     |      |     |       | *   |       |      | *    | `JOIN`  |
+| `RPL_METADATASUBS`                 |     |      |     |       |     |       | *    |      |         |
```
Herringway, line 354: Duplicate RPL_METADATASUBS here

extensions/metadata.md
```diff
@@ -0,0 +1,740 @@
...
+* zero of more of this standard reply: `FAIL METADATA INVALID_KEY`
+
+### METADATA SUBS
+
+    METADATA * SUBS
+
+This subcommand returns which keys the client is currently subscribed to.
+
+The server responds with zero or more `RPL_METADATASUBS` numerics. The server MAY return the keys in any order. The server MUST NOT list the same key multiple times in a response to this subcommand.
+
```
Herringway, line 280: Without a batch or ending numeric, this is kinda tricky to deal with.

extensions/metadata.md
```diff
@@ -0,0 +1,740 @@
...
+    S: CAP * LS :draft/metadata-2=maxsub=25 multi-prefix invite-notify
+
+-----
+
+### METADATA command examples
+
+#### Setting metadata on self
+
+    C: METADATA * SET url :http://www.example.com
+    S: :irc.example.com RPL_KEYVALUE * url * :http://www.example.com
```
Herringway, line 391: The numeric names in these examples should be replaced with the actual numeric values

Herringway (commented):
This wasn't tough to implement in my library, though there's a few things I think need addressing

extensions/metadata.md
```diff
@@ -0,0 +1,739 @@
...
+unprefixed `metadata-2` or `metadata-notify-2` capability names.
+Instead, implementations SHOULD use the `draft/metadata-2` and
+`draft/metadata-notify-2` capability name to be interoperable with other
+software implementing a compatible work-in-progress version.
+
+The final version of the specification will use an unprefixed capability name.
+
+## Introduction
+
+It is useful to associate metadata with one's IRC presence, e.g. to
```
SadieCat, line 46: Most of the spec is using soft newlines but this section is using hard newlines.

extensions/metadata.md
```diff
@@ -0,0 +1,739 @@
...
+
+If the key is valid, but not set, and the client tries to remove the key, the server responds with `FAIL METADATA KEY_NOT_SET` and fails the request.
+
+If the user cannot set keys on the given target, the server responds with `FAIL METADATA KEY_NO_PERMISSION` and fails the request.
+
+Servers MAY respond to certain keys considered not settable by the requesting user, or otherwise disallowed by the server, with `FAIL METADATA KEY_NO_PERMISSION` and fail the request.
+
+Servers MAY respond with `FAIL METADATA RATE_LIMITED` and fail the request. When a client receives `FAIL METADATA RATE_LIMITED`, it SHOULD retry the `METADATA SET` request at a later time. If the `FAIL METADATA RATE_LIMITED` event contains the `<RetryAfter>` parameter, the parameter value MUST be a positive integer indicating the minimum number of seconds the client should wait before retrying the request.
+
+If the request is successful, the server carries out the requested change and responds with one `RPL_KEYVALUE` event, representing the new value (or lack of one).
```
SadieCat, line 216: It might be useful to note that the server tweak the value and that clients should check the response before storing it in any local cache.

Herringway:
> Is there any reason for metadata-notify-2 to use a cap rather than an 005 token? From what I can see you can't subscribe before connection registration anyway, there's no need for immediate behaviour that needs to be performed on seeing it, and you now need to manually sub to any caps you want to see so it doesn't make sense for it to be a cap.

It's an opt-in feature. There's no standard way of opting into features advertised by ISUPPORT tokens.

I don't see the value in having it separate from metadata-2, though

progval:
`METADATA` now explicitly allowed in connection-registration, so the cap does make sense now

extensions/metadata.md
progval: Batches are a dependency of this spec

delthas:
When I'm joining a new channel (or receiving a join from a bouncer), I might want to wait until I've gathered all the information about a channel before displaying it. Right now I wait until RPL_ENDOFNAMES before considering the join complete and showing it.

The current spec gives 2 examples:
```
C: JOIN #smallchan
S: :modernclient!modernclient@example.com JOIN #smallchan
S: :irc.example.com 353 modernclient @ #smallchan :user1 user2 user3 user4 user5 ...
S: :irc.example.com 353 modernclient @ #smallchan :user51 user52 user53 user54 ...
S: :irc.example.com 353 modernclient @ #smallchan :user101 user102 user103 user104 ...
S: :irc.example.com 353 modernclient @ #smallchan :user151 user152 user153 user154 ...
S: :irc.example.com 366 modernclient #smallchan :End of /NAMES list.
S: :irc.example.com BATCH +UwZ67M metadata
S: @batch=UwZ67M :irc.example.com METADATA user2 bar * :second example value 
S: @batch=UwZ67M :irc.example.com METADATA user1 foo * :third example value
S: @batch=UwZ67M :irc.example.com METADATA user1 bar * :this is another example value
S: @batch=UwZ67M :irc.example.com METADATA user3 website * :www.example.com
...and some more metadata messages
S: :irc.example.com BATCH -UwZ67M
```

```
C: JOIN #bigchan
S: :modernclient!modernclient@example.com JOIN #bigchan
S: :irc.example.com 353 modernclient @ #bigchan :user1 user2 user3 user4 user5 ...
S: :irc.example.com 353 modernclient @ #bigchan :user51 user52 user53 user54 ...
S: :irc.example.com 353 modernclient @ #bigchan :user101 user102 user103 user104 ...
S: :irc.example.com 353 modernclient @ #bigchan :user151 user152 user153 user154 ...
S: :irc.example.com 366 modernclient #bigchan :End of /NAMES list.
S: :irc.example.com 774 modernclient #bigchan 4
```

Is it guaranteed that I will receive either a metadata batch or an RPL_METADATASYNCLATER on join? In which case I can wait on either of those before considering the entire join complete? Maybe that should be made clearer in the spec?

(If it's a RPL_METADATASYNCLATER I'll probably give up waiting and show the channel immediately, then sync the contents later.)

jwheare:
Is this good to be merged as draft so we can encourage more implementation, and remove the confusing "deprecated" status metadata has on the website?

progval:
I think so, and it will make it easier to iterate with new PRs.

Issue #456, opened 2021-06-02
ISUPPORT UTF8ONLY is not backwards-compatible.


keaston:

SadieCat:
The UTF8ONLY token only exists to let clients detect that the server is UTF-8 only. It is backwards compatible with the existing situation where servers that require UTF-8 break with clients which are not configured to use UTF-8.

keaston:
> The UTF8ONLY token only exists to let clients detect that the server is UTF-8 only. It is backwards compatible with the existing situation where servers that require UTF-8 silently break with clients which are not configured to use UTF-8.

Such servers aren't really following the spirit of the backwards-compatibility principle, so it seems harmful to endorse that approach in IRCv3.  The way it appears now it looks like a desired and encouraged part of the specification - ideally it would at least say that servers SHOULD not drop the client for sending non-UTF8, though they may ignore individual protocol messages.

DanielOaks:
> since clients that do not support the specification will happily send non-UTF8 and be disconnected for violating the protocol.
Ideally such servers would always hand these cases without disconnecting the client. However, given the amount of discussion that'd likely result from trying to specify one specific way of handling these cases, I thought it'd be best to just let the servers handle it in whatever way they find appropriate.

> To be backwards-compatible, this should be opt-in with a CAP exchange. Once a client has ACK'd UTF8ONLY, it is reasonable to expect it not to send anything that violates the UTF8ONLY specification.
Unfortunately we can't make this opt-in with a CAP, since servers that only accept UTF-8 traffic already exist and they need to transcode, reject, or in some other way handle non-UTF-8 traffic from clients in line with the definition written in the spec.

> ideally it would at least say that servers SHOULD not drop the client for sending non-UTF8, though they may ignore individual protocol messages
Definitely makes sense to discourage disconnecting the client outright. I'll play with the language there and try to PR some alternative language that encourages that only as a last resort. Thanks for the note, much appreciated!

slingamn:
>Such servers aren't really following the spirit of the backwards-compatibility principle, so it seems harmful to endorse that approach in IRCv3.

It's a tricky issue, yeah. I think a compatibility break is inherent in the intent of the specification --- if a server implements the spec, it's never really going to interoperate acceptably with clients that use non-UTF8 encodings (even if you can robustly transcode input, the server will only emit UTF8, likely violating client expectations that the output encoding will agree with the input encoding).

[...]

PR #502, opened 2022-07-03
UTF8ONLY: remove suggestion to disconnect the client

slingamn:
cc @DanielOaks @vanosg 

This is a response to the discussion on #456. It deletes the suggestion that servers disconnect noncompliant clients; this was unnecessarily controversial. Other language implying that UTF8 is now the preferred encoding for IRC remains in place.

Since there are no normative implications to this change (the server behavior remains implementation-defined; servers MAY still disconnect the client, although no server does so and I don't think any ever will), this shouldn't require an erratum notice.


vanosg:
Thanks for taking the time to review the suggestions from #456, by removing this language I think a stronger case can be made for this being a backward-compatible implementation.

PR #503, opened 2022-07-19
draft/persistence

slingamn:
This is a draft of an extension for automating certain interactions around persistent connections (i.e. connections where the user remains present on the server even after the connection is lost, as with a typical bouncer or Ergo's "always-on" functionality). The core goals as per the introduction:

* Servers can communicate to clients whether their connection is "persistent" in the above sense, so clients can adjust their behavior accordingly
* Clients can request that the server make their connection persistent or non-persistent, and the server can grant or deny the request

### DIFF
--- a/extensions/persistence.md
+++ b/extensions/persistence.md
@@ -85,7 +85,7 @@ If persistence cannot be accessed or modified due to an error, the `INTERNAL_ERR
 
     FAIL PERSISTENCE INTERNAL_ERROR :An error occurred
 
-Finally, servers MAY accept the change in client setting while not changing the effective setting. For example, persistence may be mandatory for all users, or it may be reserved for privileged users. In this event, the server MUST NOT send a `FAIL` response, but MUST send a `PERSISTENCE STATUS` response describing the current settings.
+Finally, servers MAY accept the change in client setting while not changing the effective setting. For example, persistence may be mandatory for all users, or it may be reserved for privileged users. In this event, the server MUST NOT send a `FAIL` response, but MUST send a `PERSISTENCE STATUS` response describing the current settings. The server MAY send a `WARN` or `NOTE` response explaining why the change was refused.
 
 ### Examples
 
--- a/extensions/persistence.md
+++ b/extensions/persistence.md
@@ -73,7 +73,7 @@ where `<client-setting>` takes one of the three values specified above; the serv
 
 #### Errors and Warnings
 
-Servers MUST NOT allow unauthenticated clients to become persistent, since there is no way for the client to reconnect in that case. Unauthenticated clients attempting to issue `GET` or `SET` MUST receive:
+Servers SHOULD NOT allow unauthenticated clients to become persistent, since there is currently no way for the client to reconnect in that case. Unauthenticated clients attempting to issue `GET` or `SET` SHOULD receive:
 
     FAIL PERSISTENCE ACCOUNT_REQUIRED :An account is required
 
--- a/extensions/persistence.md
+++ b/extensions/persistence.md
@@ -1,0 +1,160 @@
+---
+title: Persistence
+layout: spec
+work-in-progress: true
+copyrights:
+  -
+    name: "Shivaram Lingamneni"
+    period: "2022"
+    email: "slingamn@cs.stanford.edu"
+---
+
+## Notes for implementing work-in-progress version
+
+This is a work-in-progress specification.
+
+Software implementing this work-in-progress specification MUST NOT use the unprefixed `persistence` CAP name. Instead, implementations SHOULD use the `draft/persistence` CAP name to be interoperable with other software implementing a compatible work-in-progress version. The final version of the specification will use unprefixed CAP names.
+
+## Introduction
+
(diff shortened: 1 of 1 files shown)


slingamn:
A question that has come up a few times about this specification: why isn't it simply a CAP?

CAPs have always been per-connection (and this distinction is especially important in the bouncer context). For example, ZNC separately tracks client-to-ZNC CAPs and ZNC-to-server CAPs. It is not the expectation that requesting a client-to-ZNC CAP will directly affect the persistent connection (the ZNC-to-server connection).

Since this affects the "global" or "persistent" presence state, it's more like a NickServ command than a CAP.

slingamn:
Another objection that's come up: this specification is more general than required by any implementation other than Ergo.

This is true, but IMO the spec doesn't place an excessive burden on less general implementations:

1. Clients that don't want to introduce a user-controllable toggle can treat this the same way they would treat an informational CAP. The only difference is, an informational CAP would be seen during connection registration, but this comes as part of the registration burst (but since it's before RPL_ENDOFMOTD/ERR_NOMOTD, it will be seen alongside other successful registration numerics like RPL_ISUPPORT, so there's still time to make decisions based on the result).
2. Servers that don't support toggling (because the user is always persistent) just have to store the value of the user's requested setting (it should be simple to store this alongside other persistent user state).

extensions/persistence.md
```diff
@@ -0,0 +1,160 @@
...
+
+    PERSISTENCE SET <client-setting>
+
+where `<client-setting>` takes one of the three values specified above; the server replies with a `PERSISTENCE STATUS` command as specified above.
+
+#### Errors and Warnings
+
+Servers MUST NOT allow unauthenticated clients to become persistent, since there is no way for the client to reconnect in that case. Unauthenticated clients attempting to issue `GET` or `SET` MUST receive:
+
+    FAIL PERSISTENCE ACCOUNT_REQUIRED :An account is required
```
progval, line 78: I don't see a reason to make it a MUST. If a server can figure a way to do it, it wouldn't be protocol-breaking to support it.

extensions/persistence.md
```diff
@@ -0,0 +1,160 @@
...
+
+On invalid parameters, the server MUST return:
+
+    FAIL PERSISTENCE INVALID_PARAMETERS :Invalid parameters
+
+If persistence cannot be accessed or modified due to an error, the `INTERNAL_ERROR` error code SHOULD be returned.
+
+    FAIL PERSISTENCE INTERNAL_ERROR :An error occurred
+
+Finally, servers MAY accept the change in client setting while not changing the effective setting. For example, persistence may be mandatory for all users, or it may be reserved for privileged users. In this event, the server MUST NOT send a `FAIL` response, but MUST send a `PERSISTENCE STATUS` response describing the current settings.
```
progval, line 88: What if servers want to send a human-readable message to explain why they don't apply `PERSISTENCE SET`?

extensions/persistence.md
```diff
@@ -0,0 +1,160 @@
...
+[c] CAP REQ :draft/persistence sasl
+[s] :irc.example.com CAP * ACK :draft/persistence sasl
+[c] NICK testuser
+[c] USER u s e r
+[c] AUTHENTICATE PLAIN
+[c] AUTHENTICATE amlsbGVzAGppbGxlcwBzZXNhbWU=
+[c] CAP END
+[s] AUTHENTICATE +
+[s] :irc.example.com 900 * * jilles :You are now logged in as jilles
+[s] :irc.example.com 903 * :Authentication successful
```
progval, line 128: Could you reorder SASL messages to be in the "canonical" order?

extensions/persistence.md
slingamn: `WARN` or `NOTE` seem suitable for this purpose.

extensions/persistence.md
slingamn: This was intentional, but yeah, it was unnecessarily edgy :-)

emersion:
-1 from me. I think the cap is good, but I don't see the need for a standard command to get/set persistence setting.

spb:
> Another objection that's come up: this specification is more general than required by any implementation other than Ergo.

My problem is that it's not general enough - it assumes that there can be at most one persistent session per account. I'd rather see a newly-connecting client able to authenticate, request a list of available sessions, and choose which one to attach to.

slingamn:
@spb I had this conversation with Jess a while back; I think this specification can actually be decoupled from that assumption. See proposed change here: https://github.com/slingamn/ircv3-specifications/commit/16e6cbc49e672e32a0f62650813648e6dfac489b

spb:
That change does go some way towards it, but then you have the problem that a client needs to implement this spec and also whatever other spec defines the session selection mechanism. Combining them would make the ecosystem as a whole simpler.

slingamn:
From discussion in #ircv3, I'm open to an extension of this spec that would:

1. Mandate SASL as a prerequisite for reattaching to a session (allowing a bearer token mechanism to cover hypothetical cases where there is no conventional account system)
2. After this prerequisite is met, allow optionally enumerating the sessions and choosing one. It will still be possible for the client to proceed to `CAP END` without choosing a session, in which case they will get the "default" behavior (e.g. for a conventional bouncer or Ergo, they will attach to the unique session associated with the account)

thatcher-gaming:
This is [mostly implemented](https://codeberg.org/pronounmail/tubes/commit/77ca7c873db9cd9c16cc55e8142ed5a92b58c6fe) in Tubes

PR #505, opened 2022-09-18
Change the requirement to add an account tag to numerics to SHOULD.

SadieCat:
Whilst it might be useful to have an account tag on server-originated numerics that relate to a specific user it seems that everyone who implemented account tag in a server missed this bit of the spec.

I'm planning on updating our implementation (which was written by someone else) to implement this but I suspect it might be difficult for some implementations to do and as 100% of the implementations in the wild don't implement it currently it can't be relied on.

Therefore, I propose that we weaken the MUST to a SHOULD.


jwheare:
I'm fine with this, would like a few more upvotes from IRCv3 contributors before merging though.

SadieCat:
@syzop @edk0 How do you feel about this change?

syzop:
Yeah, I agree.

ElementalAlchemist:
Makes sense to me. I don't think this has come up for me yet somehow, but I can keep this in mind in the future.

SadieCat:
This has been acked by every implementation on our servers page. Merge time?

PR #506, opened 2022-10-12
Add a capability for sending arbitrary standard replies.

SadieCat:
This capability allows clients to signal that they support receiving arbitrary standard replies that are not otherwise enabled by a specific specification.

InspIRCd has something similar to this as a vendor capability and I'm planning on switching all of our custom numerics and notices over to it in v4. It'd be nice if we could standardise it so clients don't have to rely on a proprietary feature.

### DIFF
--- a/extensions/standard-replies.md
+++ b/extensions/standard-replies.md
@@ -54,7 +54,9 @@ Reply codes (especially more widely-applicable ones such as `NO_SUCH_NICK`) may 
 
 This specification adds the `standard-replies` capability. This capability allows servers to send arbitrary standard replies other than ones enabled by a specific specification. It also allows clients to indicate support for receiving any well-formed standard reply, whether or not it is recognised and used.
 
-It is RECOMMENDED that servers use standard replies instead of vendor-specific error numerics or server notices for non-standardised features when responding to a client which supports this capability. It is NOT RECOMMENDED that servers which support this capability add new custom error numerics. Also, servers which support this capability SHOULD NOT replace standardised error numerics with a standard reply unless enabled to by some other specification.
+It is RECOMMENDED that servers use standard replies instead of vendor-specific error numerics or server notices for non-standardised features when responding to a client which supports this capability. It is NOT RECOMMENDED that servers which support this capability add new custom error numerics. Servers which support this capability SHOULD NOT replace standardised error numerics with standard replies, unless the replacement is explicitly described by some other specification.
+
+
 
 ## Examples
 
--- a/extensions/standard-replies.md
+++ b/extensions/standard-replies.md
@@ -50,6 +50,11 @@ Reply codes (especially more widely-applicable ones such as `NO_SUCH_NICK`) may 
 
 `<description>` is a required final parameter, containing a plain-text description of the reply message. For example, for a `FAIL` message, the description should clearly state what error occurred, and how the user should act to resolve it. Implementers should write their descriptions to be easily understood by end users, for example _"Passwords must contain lowercase and uppercase letters, and be 8 or more characters"_ is a more clear description than _"Password consistency and strength enumerability (sec/* internal) check failed"_.
 
+### Capabilities
+
+This specification adds the `standard-replies` capability. This capability allows servers to send arbitrary standard replies other than ones enabled by a specific specification. It also allows clients to indicate support for receiving any well-formed standard reply, whether or not it is recognised and used.
+
+It is RECOMMENDED that servers use standard replies instead of error numerics or notices when responding to a client which supports this capability. It is NOT RECOMMENDED that servers which support this capability add new error numerics.
 
 ## Examples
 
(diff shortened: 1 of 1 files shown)


extensions/standard-replies.md
```diff
@@ -50,6 +50,11 @@ Reply codes (especially more widely-applicable ones such as `NO_SUCH_NICK`) may
 
 `<description>` is a required final parameter, containing a plain-text description of the reply message. For example, for a `FAIL` message, the description should clearly state what error occurred, and how the user should act to resolve it. Implementers should write their descriptions to be easily understood by end users, for example _"Passwords must contain lowercase and uppercase letters, and be 8 or more characters"_ is a more clear description than _"Password consistency and strength enumerability (sec/* internal) check failed"_.
 
+### Capabilities
+
+This specification adds the `standard-replies` capability. This capability allows servers to send arbitrary standard replies other than ones enabled by a specific specification. It also allows clients to indicate support for receiving any well-formed standard reply, whether or not it is recognised and used.
+
+It is RECOMMENDED that servers use standard replies instead of error numerics or notices when responding to a client which supports this capability. It is NOT RECOMMENDED that servers which support this capability add new error numerics.
```
emersion, line 57: This potentially requires standardizing error codes for all existing numerics?

SadieCat: It's only a recommendation (i.e. should) and it says **new** errors numerics in the next sentence. If you have any ideas for alternative phrasing here I'd be happy to tweak it though.

emersion: I mean, as a client dev, I really need to know what kind of error code I'd get from commands like JOIN or PRIVMSG in order to match them properly. IOW, enabling this suggested cap would regress my client without any way to un-break it. IOW, this cap is more ill-defined than error numerics.

For reference, example client code: https://git.sr.ht/~emersion/goguma/tree/master/item/lib/client.dart#L981

emersion: BTW -- "It's only a recommendation" makes it worse IMHO because clients don't know what to expect and will need to handle all cases (both error numerics and standard replies), only increasing the number of lines of code without any other benefit.

SadieCat: I just updated it. Is the new language better?

emersion: Yes, thanks!

SadieCat:
> FWIW, Ergo sends standard replies to all clients; I don't think it was ever an issue.
> 
> (The worst I have seen is Srain showing loud errors in the logs)

KiwiIRC, Irssi, The Lounge will just print a bunch of garbage to the server window. I think other clients that I haven't tested as well as older versions of clients that now have generic support for standard replies (e.g. HexChat) will do the same.

emersion:
Is their behavior with standard replies worse than their behavior with unknown error numerics?

SadieCat:
>Is their behavior with standard replies worse than their behavior with unknown error numerics?

Yes due to the extra info (command name, machine readable code) in a standard reply vs a numeric it can be very hard to read. Its also way worse than how server notices are handled by those clients currently.

SadieCat:
Does anyone have any objections/amendments/comments/etc regarding this?

SadieCat:
The specification did not allow that as-is but I amended it to make it more clear. Is that acceptable?

jwheare:
I'd be alright with this change.

extensions/standard-replies.md
```diff
@@ -50,6 +50,11 @@ Reply codes (especially more widely-applicable ones such as `NO_SUCH_NICK`) may
...
 
+### Capabilities
+
+This specification adds the `standard-replies` capability. This capability allows servers to send arbitrary standard replies other than ones enabled by a specific specification. It also allows clients to indicate support for receiving any well-formed standard reply, whether or not it is recognised and used.
+
+It is RECOMMENDED that servers use standard replies instead of vendor-specific error numerics or server notices for non-standardised features when responding to a client which supports this capability. It is NOT RECOMMENDED that servers which support this capability add new custom error numerics. Also, servers which support this capability SHOULD NOT replace standardised error numerics with a standard reply unless enabled to by some other specification.
```
slingamn, line 57: Belated nitpick, "unless enabled to" reads a little strangely to me. Maybe:

>Servers which support this capability SHOULD NOT replace standardised error numerics with standard replies, unless the replacement is explicitly described by some other specification.

SadieCat:
Is this just missing implementations now? InspIRCd supports this under a vendor cap and I have an implementation for [HexChat](https://github.com/hexchat/hexchat/compare/master...SadieCat:hexchat:stdrpl).

progval:
Limnoria already de facto supports it, I'd just need to add it to CAP REQ.

slingamn:
I'd be happy to add it Ergo (but requesting it wouldn't change Ergo's behavior at all, since we send arbitrary standard replies to users unconditionally).

SadieCat:
## Known implementations

### Server (2/2)

* [X] InspIRCd (@SadieCat)
* [x] Ergo &mdash; https://github.com/ergochat/ergo/pull/2026

### Client (1/2)

* [x] HexChat &mdash; https://github.com/hexchat/hexchat/compare/master...SadieCat:hexchat:stdrpl

### Bot

* [ ] Limnoria (@progval)

delthas:
Support added to senpai: https://git.sr.ht/~taiite/senpai/commit/1318e784bd2bba3765aee97811a3f0053d3a6723

SadieCat:
I think we have enough implementations to merge this now?

jwheare:
I guess we skipped the draft/ prefix for this one. I'll add it to IRCCloud, test and merge.

PR #510, opened 2022-11-27
chathistory: Add MSGREFTYPES ISUPPORT token

progval:
It appears that both Soju and Matrix2051 have issues implementing one of the types (resp. msgid and timestamp).

While they could add full support eventually, these would cause higher resource usage, so it will be beneficial to indicate a preference even if both types are supported.

@emersion and I considered adding [optional `msgid`s to read-marker](https://github.com/ircv3/ircv3-specifications/compare/master...progval:ircv3-specifications:read-marker-msgreftype) but it would add a lot of complexity to clients, without much benefit.

### DIFF
--- a/extensions/chathistory.md
+++ b/extensions/chathistory.md
@@ -14,6 +14,10 @@ copyrights:
     name: "Shivaram Lingamneni"
     period: "2020"
     email: "slingamn@cs.stanford.edu"
+  -
+    name: "Val Lorentz"
+    period: "2022"
+    email: "progval+ircv3@progval.net"
 ---
 
 ## Notes for implementing work-in-progress version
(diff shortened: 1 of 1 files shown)


extensions/chathistory.md
```diff
@@ -36,10 +40,22 @@ Full support for this extension requires support for the [`batch`][batch], [`ser
...
+This specification defines two ISUPPORT tokens.
+
+`CHATHISTORY` MUST be sent to the client to state the maximum number of messages a client can request in a single command, e.g., `CHATHISTORY=50`. If `0`, the client SHOULD assume that there is no maximum number of messages.
+
+`MSGREFTYPES` SHOULD be sent, with a comma-separated list of types of message references they support as parameter to `CHATHISTORY` commands, in order of decreasing preference. Clients MUST ignore any type they do not support. The currently defined list of types is:
+
+* `timestamp`, to reference messages by their timestamp
+* `msgid`, to reference them by their `msgid`
+
+If `MSGREFTYPES` is missing, clients SHOULD assume the server supports both and has no preference.
```
emersion, line 56: This "legacy path" isn't quite correct IMHO. If the server doesn't send `msgid` tags, then the client has no way to use the `msgid` reference type. Same goes for the `time` tag. Maybe that should be clarified? ie, state that the server will support `timestamp` refs if `time` tags are sent, and will support `msgid` refs if `msgid` tags are sent?

progval: Is it worth mentioning it in the spec, though? If either tag is missing, then clients don't really have a choice anyway

emersion: Right, just feels a bit weird to specify that the server supports both. I guess it's not that of a big deal since it's a SHOULD.

syzop:
Yeah, sounds reasonable.

emersion:
soju patch: https://git.sr.ht/~emersion/soju/commit/c0bbcdcc0af94438cbc9c05578ce2bcb48a84a26

progval:
and in matrix2051: https://github.com/progval/matrix2051/commit/6c4f8d477654718d8ae534f8eef086c13d452901

emersion:
And a Goguma patch which disables chathistory when `MSGREFTYPES` is missing `timestamp`, because it only supports that at the moment: https://git.sr.ht/~emersion/goguma/commit/5466e6ee58154bdada45baf47b34a4c8d50b87a3

PR #511, opened 2022-12-21
reword introduction

vanosg:
This is an interesting idea, misdirected notices are a bane we've all had to suffer!

Reading the channel-context draft specification, I was interested in what it was communicating but had to initially ask in #ircv3 for more information regarding what this client tag was attempting to solve. Based on that conversation, I'd offer the following elaboration on what @delthas did a great job starting out on. From an "outsider" standpoint to this spec, reorganizing it something similar to this might give the casual reader a better idea of what this spec is setting out to solve.

Take it, leave it, edit it, improve it, I take no pride in ownership. I hope that clarifying some of these points can be useful to others


vanosg:
just a bump, happy for a merge or a close.

PR #513, opened 2023-01-24
add an `identd` key to WEBIRC options

jesopo:
the motivation behind this is the IRCd can't do this lookup itself; an ident lookup must come from the IP to which the end user is connecting, which means the gateway has to be the one to do it.


slingamn:
I disapprove of ident, but this change seems fine.

SadieCat:
I don't think this is necessary (in 2023 gateways should be giving users their own IPv6) and we will be ignoring it (most of our users don't even have ident support enabled and we discourage its use) but I don't really have any objection with the key existing as long as its not mandatory to do anything with it.

jesopo:
> I disapprove of ident, but this change seems fine.

i tend to agree that it's a poor solution for what it's used for, but i also recognise that there isn't a workable alternative right this second. i hope one-ipv6-per-user becomes more ubiquitous, but as it stands this small spec tweak will help plumb two existing things together

SadieCat:
Bikeshed: an alternative to this would be some kind of unique identifier field which may or may not correlate to an identd response depending on implementation.

PR #514, opened 2023-02-02
add pre-away extension

slingamn:
This comes out of https://github.com/ircv3/ircv3-ideas/issues/96 . It provides a mechanism to indicate that a new connection was not initiated by an end user and does not indicate that the end user is present. The core objective is to work with auto-away systems, but the mechanism is potentially more general.

### DIFF
--- a/extensions/pre-away.md
+++ b/extensions/pre-away.md
@@ -1,0 +1,46 @@
+---
+title: "`pre-away` Extension"
+layout: spec
+meta-description: A capability allowing client connections to be marked AWAY during connection registration
+work-in-progress: true
+copyrights:
+  -
+    name: "Shivaram Lingamneni"
+    period: "2023"
+    email: "slingamn@cs.stanford.edu"
+---
+
+## Notes for implementing work-in-progress version
+
+This is a work-in-progress specification.
+
+Software implementing this work-in-progress specification MUST NOT use the unprefixed `pre-away` CAP name. Instead, implementations SHOULD use the `draft/pre-away` CAP name to be interoperable with other software implementing a compatible work-in-progress version. The final version of the specification will use an unprefixed CAP name.
+
+## Introduction
+Some IRC server implementations offer a mode of operation where a single nickname can be associated with multiple concurrent client connections, or no client connections. Typically, such implementations are bouncers (i.e., intermediaries between the client and another server), but some are full server implementations.
+
+Such implementations may wish to update publicly visible state depending on the status of the user's actual client connections. For example, if the user has no active connections, it may be desirable to mark them as AWAY, then mark them un-AWAY if they reconnect. However, a client implementation may wish to connect without active involvement from the user, e.g. to retrieve [chathistory][chathistory], in which case it would be undesirable to suggest that the user is present. This extension provides a mechanism for such clients to flag their connections as automatically initiated, so servers can disregard them for this or other purposes related to user presence.
+
+## Implementation
+This specification introduces a new informational capability, `draft/pre-away`. Clients MAY request this capability; servers implementing this specification MUST acknowledge it if it is requested.
+
+Servers implementing this specification MUST accept the `AWAY` command before connection registration has completed, whether or not the `draft/pre-away` capability has been negotiated. The `AWAY` command has its normal semantics in this context, except servers SHOULD treat the form:
+
+    AWAY *
+
+i.e. an `AWAY` message consisting of the single character `*`, as indicating that the user is not present for an unspecified reason.
+
+In its conventional form:
+
+    AWAY :Gone to lunch.  Back in 5
+
+the `AWAY` command MAY be used pre-registration to set a human-readable away message associated with the connection as usual. Similarly, `AWAY` with no parameters, which normally indicates that the user is present, has no effect.
+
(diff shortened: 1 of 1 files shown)


slingamn:
cc @emersion @causal-agent

SadieCat:
This seems like it just overcomplicates the problem. Why not use an unrequestable cap that tells clients they can send AWAY with a normal message before connection registration is complete like with `sts`?

progval:
@SadieCat This spec does that *and* defines `AWAY *` which tells the bouncer/server not to change the away status because of this connection

emersion:
This looks good!

Implementations:

- soju: https://git.sr.ht/~emersion/soju/commit/9f3596824112b577f732137459df2612ef5978fe
- Goguma: https://git.sr.ht/~emersion/goguma/commit/127bc622e61f6ae1131fdf4fb8678898be2c9b08

slingamn:
cc @kylef : can `FOREGROUND` / `BACKGROUND` be translated into `AWAY` and `AWAY *` under this specification?

slingamn:
From implementing this in Ergo (https://github.com/ergochat/ergo/pull/2044) and discussion with @emersion, two changes:

1. Negotiating the capability is now required; this gives server implementations more information, at negligible cost
2. Explicitly mention `AWAY *` post-registration (this is the intended flow for "backgrounding" a mobile app)

kylef:
@slingamn its not quite the same thing. They have different semantics as BACKGROUND/FOREGROUND are (client -> bouncer) whereas AWAY/UNAWAY is (client -> bouncer -> server). BACKGROUND/FOREGROUND do not leak the away state of the client to other clients.

If that could be a privacy concern for a user I do not know. In Palaver "auto away" is not enabled by default. BACKGROUND/FOREGROUND will be sent for supporting servers regardless of the user's "auto away" preference as it is for managing notification state and not exposing an away state.

Slightly related, in ZNC (using the [clientaway module](https://github.com/kylef-archive/znc-contrib/blob/master/clientaway.cpp)), a per-client away state can be configured, where if all clients have decided to be away then the away can (optionally) be propagated to the upstream server (to be visible by other clients). I presume that is the desirable behaviour with the proposed specification and what we might see in Ergo utilising `pre-away`?

kylef:
> 2. Explicitly mention AWAY * post-registration (this is the intended flow for "backgrounding" a mobile app)

With a mobile application, you may also be going away/unaway very frequently and may not want this to by directly visible to others IRC users when jumping back and fourth between applications. In this context if AWAY/UNAWAY is a proposed alternative to BACKGROUND/FOREGROUND, it may be desirable for server implementations to offer some form of "delay" so that if a client aways and unaways within 10 seconds for example, it is not the user's away state.

slingamn:
Thanks, that makes sense. What information do `BACKGROUND` and `FOREGROUND` convey to to the bouncer? Or rather, how does the bouncer make use of that information?

extensions/pre-away.md
```diff
@@ -0,0 +1,46 @@
...
+
+In its conventional form:
+
+    AWAY :Gone to lunch.  Back in 5
+
+the `AWAY` command MAY be used pre-registration to set a human-readable away message associated with the connection as usual. Similarly, `AWAY` with no parameters indicates that the user is present.
+
+If the client's nickname was not already present on the server, then `AWAY` pre-registration sets the away message but does not inhibit reporting of the change in nickname status, e.g. via [monitor][monitor].
+
+Clients receiving `*` as an away message (for example, in `301 RPL_AWAY` or [away-notify][away-notify]) SHOULD treat it as indicating that the user is not present for an unspecified reason. Servers MAY substitute a human-readable message for the `*` if it would otherwise be relayed as an away message.
```
SadieCat, line 41: Can you clarify whether this means a client receiving it for themself or other users? If it's the latter then this gonna cause weird behaviours on servers that don't support the spec. Might be best to require servers to rewrite it to something sensible?

slingamn: If I changed it to say "Clients having negotiated the capability", would that alleviate concerns about behavior on servers not supporting the spec?

slingamn:
I think the language is more clear now:

* If the capability is negotiated, `*` has a special meaning in both directions (albeit only at the SHOULD level)
* If the capability is not negotiated, the behavior is implementation-defined; servers may give `*` its special meaning regardless, or may implement the traditional meaning

emersion (approved):

emersion:
Now shipped as a vendored soju ext: https://git.sr.ht/~emersion/soju/commit/c36bb342fb13604996214951bfec477abfa38843

delthas (approved):
LGTM

jesopo (approved):
this functionality isn't super important to me, but it seems to be to others and i have no objections to it. lgtm

emersion:
Thanks a lot!

awfulcooking:
That [USER bitmask in rfc2812](https://datatracker.ietf.org/doc/html/rfc2812#section-3.1.3), tho :)

SadieCat:
As a minor (dont let this block anything) bikeshed before this is ratified: given that `AWAY *` can be sent after registration might there be a more appropriate capability name for this than `pre-away`? e.g. generic-away, extended-away, etc.

PR #517, opened 2023-02-02
multiline: Clarify what counts toward max-bytes

### LINKED ISSUE #516: Multiline messages: Clarify what counts towards max-bytes and what doesn't
Does the entire PRIVMSG message, command and/or tags included, count towards the max-bytes limit? Is it only the PRIVMSG's message parameter?

progval:
This matches the behavior [in Ergo](https://github.com/ergochat/ergo/blob/4317016a09a2629603fb6198cbeddf6afa28c8ce/irc/handlers.go#L2163-L2170) and [in Limnoria](https://github.com/progval/limnoria/blob/96b7f51e7162f540475ed9dd8763dff38a256e64/src/callbacks.py#L927-L946).

[BitBot ignores `max-bytes`](https://github.com/bitbot-irc/bitbot/blob/4a6037c77405f3584efadc10ae75826b6b9ac422/modules/ircv3_multiline.py); and I did not check how the IRCCloud server implements it (though the client does not seem to enforce it at all before sending).

Resolves #516.

### DIFF
--- a/extensions/multiline.md
+++ b/extensions/multiline.md
@@ -51,8 +51,8 @@ Clients MUST ignore every token with a key that they don't understand.
 
 The only defined capability key so far is:
 
-* `max-bytes` - This defines the maximum allowed total byte length of multiline batched content *REQUIRED*
-* `max-lines` - This defines the maximum allowed number of lines in a multiline batch content *RECOMMENDED*
+* `max-bytes` - This defines the maximum allowed total byte length of multiline batch, as defined below *REQUIRED*
+* `max-lines` - This defines the maximum allowed number of lines in a multiline batch *RECOMMENDED*
 
 Only the last parameter of PRIVMSG and NOTICE are counted toward `max-bytes`; not origin, command, or target.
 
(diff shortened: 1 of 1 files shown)


extensions/multiline.md
```diff
@@ -54,6 +54,8 @@ The only defined capability key so far is:
 * `max-bytes` - This defines the maximum allowed total byte length of multiline batched content *REQUIRED*
```
slingamn, line 54: Suggestion: "This defines the maximum allowed total byte length of a multiline batch's message value, as defined below"

extensions/multiline.md
```diff
@@ -54,6 +54,8 @@ The only defined capability key so far is:
...
 * `max-lines` - This defines the maximum allowed number of lines in a multiline batch content *RECOMMENDED*
```
slingamn, line 55: Suggestion: "This defines the maximum allowed number of lines in a multiline batch"

slingamn (commented):
The definition of "message value" below was intended to resolve similar ambiguities, I suggest leveraging it here.

PR #518, opened 2023-02-22
fixed typo

orkim:
I believe this to be a typo in the escaping.  The 'raw value' did have a semi-colon shown, but looking at the example I believe this should be a colon.


jwheare:
The example is correct, the escaping rules state that a raw semi-colon should be escaped as a backslash and colon, so:

`;` is escaped as `\:`

orkim:
OK, understood.  I guess I missed this re-reading it a few times.  Thanks.

PR #520, opened 2023-03-21
Add sasl-ir extension

emersion:
Pretty similar to the [IMAP SASL-IR extension](https://www.rfc-editor.org/rfc/rfc4959).

### DIFF
--- a/extensions/sasl-ir.md
+++ b/extensions/sasl-ir.md
@@ -1,0 +1,43 @@
+---
+title: SASL initial response
+layout: spec
+copyrights:
+  -
+    name: "Simon Ser"
+    period: "2023"
+    email: "contact@emersion.fr"
+---
+
+## Notes for implementing work-in-progress version
+
+This is a work-in-progress specification.
+
+Software implementing this work-in-progress specification MUST NOT use the unprefixed `sasl-ir` CAP name. Instead, implementations SHOULD use the `draft/sasl-ir` CAP name to be interoperable with other software implementing a compatible work-in-progress version. The final version of the specification will use unprefixed CAP names.
+
+## Introduction
+
+Clients can use the [SASL extension](sasl-3.1.html) to authenticate against IRC servers. Some SASL mechanisms are client-first, ie. the client sends an initial response to start the authentication exchange.
+
+This extension updates the first `AUTHENTICATE` command to accept an initial response, to avoid a roundtrip in the SASL exchange.
+
+## Implementation
+
+Servers supporting this extension MUST advertise the `sasl-ir` capability. Servers MUST accept the new form of the `AUTHENTICATE` command even if the capability hasn't been explciitly enabled by the client.
+
+The initial `AUTHENTICATE` command is extended to accept an extra argument for the initial response:
+
+    AUTHENTICATE <mechanism> [initial-response]
+
+The initial response is optional. When provided, the initial response MUST be encoded as defined by SASL 3.1 (base64 or `+` with a maximum size of 400 bytes). If the initial response is larger than 400 bytes, the client MUST send separate `AUTHENTICATE` commands as usual with the rest of the chunks.
+
+An `AUTHENTICATE` command with an initial response is equivalent to the following exchange:
+
+    C: AUTHENTICATE <mechanism>
+    S: AUTHENTICATE +
+    C: AUTHENTICATE <initial-response>
+
+## Example protocol exchange
+
+    C: AUTHENTICATE PLAIN amlsbGVzAGppbGxlcwBzZXNhbWU=
+    S: :irc.example.org 900 emersion emersion!emersion@emersion.fr emersion :You are now logged in as emersion
+    S: :irc.example.org 903 emersion :SASL authentication successful


extensions/sasl-ir.md
```diff
@@ -0,0 +1,43 @@
...
+
+    C: AUTHENTICATE <mechanism>
+    S: AUTHENTICATE +
+    C: AUTHENTICATE <initial-response>
+
+## Example protocol exchange
+
+    C: AUTHENTICATE PLAIN amlsbGVzAGppbGxlcwBzZXNhbWU=
+    S: :irc.example.org 900 emersion emersion!emersion@emersion.fr emersion :You are now logged in as emersion
+    S: :irc.example.org 903 emersion :SASL authentication successful
```
progval, line 43: Could you add an example with > 400 bytes?

extensions/sasl-ir.md
```diff
@@ -0,0 +1,43 @@
...
+
+## Implementation
+
+Servers supporting this extension MUST advertise the `sasl-ir` capability. Servers MUST accept the new form of the `AUTHENTICATE` command even if the capability hasn't been explciitly enabled by the client.
+
+The initial `AUTHENTICATE` command is extended to accept an extra argument for the initial response:
+
+    AUTHENTICATE <mechanism> [initial-response]
+
+The initial response is optional. When provided, the initial response MUST be encoded as defined by SASL 3.1 (base64 or `+` with a maximum size of 400 bytes). If the initial response is larger than 400 bytes, the client MUST send separate `AUTHENTICATE` commands as usual with the rest of the chunks.
```
progval, line 31: > If the initial response is larger than 400 bytes, the client MUST send separate `AUTHENTICATE` commands as usual with the rest of the chunks.

This means it shouldn't repeat the mechanism, right?

progval:
Implemented in Matrix2051: https://github.com/progval/matrix2051/commit/e03f2be5f1dba14951158a0738dbee54f651f21b

extensions/sasl-ir.md
```diff
@@ -0,0 +1,43 @@
...
+
+## Notes for implementing work-in-progress version
+
+This is a work-in-progress specification.
+
+Software implementing this work-in-progress specification MUST NOT use the unprefixed `sasl-ir` CAP name. Instead, implementations SHOULD use the `draft/sasl-ir` CAP name to be interoperable with other software implementing a compatible work-in-progress version. The final version of the specification will use unprefixed CAP names.
+
+## Introduction
+
+Clients can use the [SASL extension](sasl-3.1.html) to authenticate against IRC servers. Some SASL mechanisms are client-first, ie. the client sends an initial response to start the authentication exchange.
```
dgw, line 19:
```suggestion
Clients can use the [SASL extension](sasl-3.1.html) to authenticate against IRC servers. Some SASL mechanisms are client-first, i.e. the client sends an initial response to start the authentication exchange.
```

extensions/sasl-ir.md
emersion: Correct, added a note about that

slingamn:
Independently of this spec, thoughts on adding a non-normative caution to the SASL spec, recommending that unrecognized mechanisms should not be logged?

emersion:
Done here: https://github.com/ircv3/ircv3-specifications/pull/522

grawity:
could there be some guidance regarding mechanisms that are server-first? e.g.. explicitly abort if the server's initial challenge is not empty (i.e. if it's not "AUTHENTICATE +")

I don't remember such mechanisms offhand (and the existing ones wouldn't be relevant for IRC to be fair) but afaik they do exist

extensions/sasl-ir.md
```diff
@@ -0,0 +1,52 @@
...
+
+## Introduction
+
+Clients can use the [SASL extension](sasl-3.1.html) to authenticate against IRC servers. Some SASL mechanisms are client-first, i.e. the client sends an initial response to start the authentication exchange.
+
+This extension updates the first `AUTHENTICATE` command to accept an initial response, to avoid a roundtrip in the SASL exchange.
+
+## Implementation
+
+Servers supporting this extension MUST advertise the `draft/sasl-ir` capability. Servers MUST accept the new form of the `AUTHENTICATE` command even if the capability hasn't been explciitly enabled by the client.
```
SadieCat, line 25:
```suggestion
Servers supporting this extension MUST advertise the `draft/sasl-ir` capability. Servers MUST accept the new form of the `AUTHENTICATE` command even if the capability hasn't been explicitly enabled by the client.
```

dgw: Use of `sasl-ir` would still require requesting `sasl`, as I read it. I just can't think of a reason not to send `CAP REQ :sasl sasl-ir` instead of just `CAP REQ :sasl` when the server advertises this and a client wants to use it. It's only a few extra bytes, not a whole extra roundtrip.

_I thought I saw a previous review/suggestion flag `explciitly`, whoops.)_

slingamn:
Summarizing some discussion from #ircv3, it seems like in order for this extension to provide a speedup, the client has to cheat slightly (by caching a previous `CAP LS 302` response from the server, so that it knows that `sasl-ir` is available even without a CAP LS roundtrip). But if the client is willing to cheat a little more than that (by caching the availability of `sasl=PLAIN`), then it can get the speedup even without this extension, by preemptively sending its SASL messages without waiting for acknowledgement. (To be clear: I think this kind of cheating is good and should be encouraged in mobile clients.)

emersion:
The goal of this extension is not to provide a speedup (clients can already do it), it's to make the protocol more reliable. If the client caches `sasl=WHATEVER` and the server happens to change their supported capabilities, then the credentials are interpreted as a SASL mechanism by the server.

slingamn:
>If the client caches sasl=WHATEVER and the server happens to change their supported capabilities, then the credentials are interpreted as a SASL mechanism by the server.

Are we talking about a TOCTOU race between the server sending a CAP LS 302 response including `sasl=PLAIN` and the client attempting to use it, within the same session?

Or is the idea that a cached cross-session observation of `sasl-ir` is more stable than a cached observation of `sasl=PLAIN`? (In theory, a cached observation of `sasl-ir` can also become invalid.)

emersion:
The idea of this extension is to make it safer to pipeline `AUTHENTICATE` commands without waiting for the `CAP LS` reply, by making it explicit which parameters are credentials and sensitive info in an `AUTHENTICATE` command. Right now a server can't make the difference between the start of a SASL exchange and the first initial response sent by the client.

DarthGandalf:
Maybe stop using the same keyword `AUTHENTICATE` to pass both mechanism and response, and start using different keywords for them?

emersion:
If we want to make bigger changes, I'd suggest adding a `*` parameter indicating a continuation like `CAP` does:

```
AUTHENTICATE PLAIN
AUTHENTICATE * <base64>
```

slingamn:
>The idea of this extension is to make it safer to pipeline AUTHENTICATE commands without waiting for the CAP LS reply, by making it explicit which parameters are credentials and sensitive info in an AUTHENTICATE command. Right now a server can't make the difference between the start of a SASL exchange and the first initial response sent by the client.

This issue would still exist for line continuations, right?

It seems to me that the concerns here don't really justify adding complexity to one of the most widely implemented and deployed IRCv3 specs.

(For what it's worth, the draft as written still has a performance rationale in the introduction section. I think I understand most of the robustness rationale now, but I'm still not really convinced.)

PR #521, opened 2023-03-24
sasl-3.1: add example with split client response

emersion:


progval:
The payload of the first message is 536 bytes long, that's doubly illegal.

You have to truncate at 400 bytes of the base64 encoding, not the encoded bytes

emersion:
Bleh I read the spec backwards.

emersion:
Fixed!

emersion:
Gentle ping

PR #522, opened 2023-03-24
sasl-3.1: recommend against logging AUTHENTICATE arguments

emersion:


progval:
Wouldn't an "Implementation considerations" section be a better fit?

slingamn:
+1 for a non-normative "implementation considerations", with lowercase should.

PR #523, opened 2023-04-05
Add : to standard-reply

vanosg:
Similar to the message format examples in other specs (away-notify, setname, etc), this adds the required : to the message format example prior to the description argument


jwheare:
This isn't required. The : is part of general IRC message argument grammar and the rules are clearly defined elsewhere. This format is just listing the parameters in order, irrespective of how the grammar might be applied in specific cases.

PR #524, opened 2023-05-26
Add message redaction

progval:
[Rendered](https://github.com/progval/ircv3-specifications/blob/redaction/extensions/message-redaction.md)

This is https://github.com/ircv3/ircv3-specifications/pull/425 with the following changes:

* no edits, because it's way more complicated to handle them safely, so IMO it should be deferred to a future spec
* targetmsgid is moved to a param (instead of a tag, because it didn't really make sense with a dedicated verb)
* added an optional reason (just because I can)
* renamed "delete" to "redact" so the cap name doesn't conflict with the other one (and there is some argument esp. in Matrix circles that "redact" better reflects that it's best effort and other clients might not respect it, but I don't really buy it)
* explicitly gives a lot of freedom to servers wrt chathistory
* new `UNKNOWN_MSGID` fail tag (because I noticed that's needed while implementing the spec)

Implementations:

* Ergo (server): https://github.com/ergochat/ergo/pull/2065
* Matrix2051 (gateway): https://github.com/progval/matrix2051/pull/53
* Limnoria (client): https://github.com/progval/Limnoria/pull/1538 (sending only)
* Unreal (server): https://github.com/unrealircd/unrealircd/pull/253 (new core API) + https://github.com/unrealircd/unrealircd-contrib/pull/91 (third-party module)
* Goguma (client): https://git.sr.ht/~emersion/goguma/log/redact
* IRCCloud

### DIFF
--- a/extensions/message-redaction.md
+++ b/extensions/message-redaction.md
@@ -88,6 +88,7 @@ The following codes are defined, with sample plain text descriptions.
 
 * `FAIL REDACT REDACT_FORBIDDEN <target> <target-msgid> :You are not authorised to delete this message`
 * `FAIL REDACT REDACT_WINDOW_EXPIRED <target> <target-msgid> <window> :You can no longer edit this message`
+* `FAIL REDACT UNKNOWN_MSGID <target> <target-msgid> <window> :This message does not exist or is too old`
 
 ## Client implementation considerations
 
--- a/extensions/message-redaction.md
+++ b/extensions/message-redaction.md
@@ -108,6 +108,11 @@ allow users to attempt to delete any message via some mechanism.
 
 Clients SHOULD NOT provide a default reason if users do not provide one.
 
+When a `REDACT` command's `msgid` parameter references a known message not in
+the `target`'s history, clients MUST ignore it.
+This allows servers to safely relay `REDACT` commands targeting messages which they
+did not keep in their history.
+
 ## Server implementation considerations
 
 *This section is non-normative.*
--- a/extensions/message-redaction.md
+++ b/extensions/message-redaction.md
@@ -37,12 +37,12 @@ guarantees.
 
 ### Dependencies
 
-Clients wishing to use these capabilities MUST negotiate the [`message-tags`][]
+Clients wishing to use this capability MUST negotiate the [`message-tags`][]
 capability with the server.
 Additionally, this capability relies on messages being sent with the
 [`msgid`][] tag, and message ids MUST NOT contain spaces.
-Clients SHOULD negotiate the [`echo-message`][] capabilities in order to receive
-message IDs for their own messages, to allow them to be redacted.
+Clients SHOULD negotiate the [`echo-message`][] capability in order to receive
+message IDs for their own messages, so they can be redacted.
 
 
 ### Capability
@@ -50,8 +50,8 @@ message IDs for their own messages, to allow them to be redacted.
 This specification adds the `draft/message-redaction` capability.
 Clients MUST ignore this capability's value, if any.
 
-Implementations that negotiate these capabilities indicate that they are
-capable of handling the respective commands and message tags described below.
+Implementations that negotiate this capability indicate that they are
+capable of handling the command described below.
 
 
 ### Command
@@ -68,9 +68,9 @@ Where `<msgid>` is the id of the message to be redacted, which MUST be a
 If the client is authorised to delete the message, the server:
 
 * SHOULD forward this `REDACT`, with an appropriate prefix, to the target
-  recipients which negotiated the `draft/message-redaction` capability, in the
+  recipients that have negotiated the `draft/message-redaction` capability, in the
   same way as PRIVMSG messages.
-* MUST not forward this `REDACT` to target recipients which did not negotiated
+* MUST not forward this `REDACT` to target recipients that have not negotiated
   this capability (see "Fallback" below)
 
 ### Chat history
@@ -88,7 +88,7 @@ This specification defines `FAIL` messages using the [standard replies][]
 framework for notifying clients of errors with message editing and deletion.
 The following codes are defined, with sample plain text descriptions.
 
-* `FAIL REDACT INVALID_TARGET the_given_target :You cannot delete messages from the_given_target`
+* `FAIL REDACT INVALID_TARGET <target> :You cannot delete messages from <target>`
 * `FAIL REDACT REDACT_FORBIDDEN <target> <target-msgid> :You are not authorised to delete this message`
 * `FAIL REDACT REDACT_WINDOW_EXPIRED <target> <target-msgid> <window> :You can no longer edit this message`
 * `FAIL REDACT UNKNOWN_MSGID <target> <target-msgid> :This message does not exist or is too old`
@@ -117,7 +117,7 @@ did not keep in their history.
 
 ## Server implementation considerations
 
-*This section is non-normative.*
+This section is non-normative.
 
 A key motivation for specifying this capability as a server tag, rather than
 a client-only message tag, is to enable more granular redaction permissions.
--- a/extensions/message-redaction.md
+++ b/extensions/message-redaction.md
@@ -79,7 +79,9 @@ After a message is redacted, [`chathistory`][] responses SHOULD either:
 * exclude it entirely
 * replace its content and/or tags with a placeholder and
   add the `REDACT` message to the response (not counting toward message limits)
+  after the redacted message
 * add the `REDACT` message to the response (not counting toward message limits)
+  after the redacted message
 
 ### Errors
 
--- a/extensions/message-redaction.md
+++ b/extensions/message-redaction.md
@@ -77,7 +77,8 @@ If the client is authorised to delete the message, the server:
 After a message is redacted, [`chathistory`][] responses SHOULD either:
 
 * exclude it entirely
-* replace it with the `REDACT` message with the same value as `@msgid` tag
+* replace its second parameter and/or tags with a placeholder and
+  add the `REDACT` message to the response (not counting toward message limits)
 * add the `REDACT` message to the response (not counting toward message limits)
 
 ### Errors
--- a/extensions/message-redaction.md
+++ b/extensions/message-redaction.md
@@ -131,7 +131,7 @@ If a message is redacted while a client is disconnected (or parted from a channe
 a message it saw is redacted, servers may send the `REDACT` command in a `chathistory`
 batch when it re-joins the channel.
 
-If servers use predictable or guessable `msgid`s, they should considered whether errors
+If servers use predictable or guessable `msgid`s, they should consider whether errors
 returned on `REDACT` may leak a message's existence to users who did not receive it
 (in a channel they are/were not in or in private messages).
 
--- a/extensions/message-redaction.md
+++ b/extensions/message-redaction.md
@@ -177,10 +177,10 @@ Deleting a TAGMSG:
 
 Deleting someone else's PRIVMSG:
 
-    C: PRIVMSG #channel :join my network for cold hard chats
-    S: @msgid=123 :nick!u@h PRIVMSG #channel :join my network for cold hard chats
-    C: REDACT #channel 123 spam
-    S: @msgid=124 :nick!u@h REDACT #channel 123 spam
+    C1: PRIVMSG #channel :join my network for cold hard chats
+    S:  @msgid=123 :nick!u@h PRIVMSG #channel :join my network for cold hard chats
+    C2: REDACT #channel 123 spam
+    S:  @msgid=124 :chanop!u@h REDACT #channel 123 spam
 
 
 [`echo-message`]: ../extensions/echo-message.html
--- a/extensions/message-redaction.md
+++ b/extensions/message-redaction.md
@@ -49,6 +49,7 @@ them to be redacted.
 ### Capability
 
 This specification adds the `draft/message-redaction` capability.
+Clients MUST ignore this capability's value, if any.
 
 Implementations that negotiate these capabilities indicate that they are
 capable of handling the respective commands and message tags described below.
--- a/extensions/message-redaction.md
+++ b/extensions/message-redaction.md
@@ -90,7 +90,7 @@ The following codes are defined, with sample plain text descriptions.
 * `FAIL REDACT INVALID_TARGET the_given_target :You cannot delete messages from the_given_target`
 * `FAIL REDACT REDACT_FORBIDDEN <target> <target-msgid> :You are not authorised to delete this message`
 * `FAIL REDACT REDACT_WINDOW_EXPIRED <target> <target-msgid> <window> :You can no longer edit this message`
-* `FAIL REDACT UNKNOWN_MSGID <target> <target-msgid> <window> :This message does not exist or is too old`
+* `FAIL REDACT UNKNOWN_MSGID <target> <target-msgid> :This message does not exist or is too old`
 
 ## Client implementation considerations
 
--- a/extensions/message-redaction.md
+++ b/extensions/message-redaction.md
@@ -77,7 +77,7 @@ If the client is authorised to delete the message, the server:
 After a message is redacted, [`chathistory`][] responses SHOULD either:
 
 * exclude it entirely
-* replace its second parameter (if it is a `PRIVMSG` or `NOTICE`) and/or tags with a placeholder and
+* replace its content and/or tags with a placeholder and
   add the `REDACT` message to the response (not counting toward message limits)
 * add the `REDACT` message to the response (not counting toward message limits)
 
--- a/extensions/message-redaction.md
+++ b/extensions/message-redaction.md
@@ -62,7 +62,8 @@ The command is defined as follows:
 
     REDACT <target> <msgid> [:reason]
 
-Where `<msgid>` is the id of the message to be redacted.
+Where `<msgid>` is the id of the message to be redacted, which MUST be a
+`PRIVMSG`, `NOTICE`, or `TAGMSG`.
 
 If the client is authorised to delete the message, the server:
 
--- a/extensions/message-redaction.md
+++ b/extensions/message-redaction.md
@@ -100,7 +100,7 @@ Redacted messages MAY be hidden entirely from the primary message log,
 but a deletion log SHOULD be made available.
 
 For the purposes of user interface, clients MAY assume that their own messages
-are editable.
+are redactable.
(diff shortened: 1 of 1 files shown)


DarthGandalf:
1. client A sends a message
2. client B receives the message
3. client B disconnects
4. client A sends REDACT
5. client B reconnects

Client B has no chance of deleting the message even if it supports it. Subject to chathistory, if server sent such REDACT upon reconnect, if the buffer was big enough.

progval:
@DarthGandalf It's unreasonable to make it a requirement; but I mentioned in the "Server implementation considerations" that servers may send it in the `chathistory` batch when clients re-join (Ergo, Insp, and Unreal already send such a batch for clients without the `draft/chathistory` cap who aren't bots)

extensions/message-redaction.md
```diff
@@ -0,0 +1,183 @@
...
+any changes to the original message.
+
+## Examples
+
+Deleting a PRIVMSG:
+
+    C: PRIVMSG #channel :an example
+    S: @msgid=123 :nick!u@h PRIVMSG #channel :an example
+    C: DELETE #channel 123 :bad example
+    S: @msgid=124 :nick!u@h DELETE #channel 123 :bad example
```
delthas, line 160:
```suggestion
    C: REDACT #channel 123 :bad example
    S: @msgid=124 :nick!u@h REDACT #channel 123 :bad example
```

extensions/message-redaction.md
```diff
@@ -0,0 +1,183 @@
...
+
+Deleting a TAGMSG:
+
+    C: @draft/react=🤞TAGMSG #channel
+    S: @msgid=123;draft/react=🤞TAGMSG #channel
+    C: DELETE #channel 123
+    S: @msgid=124 :nick@u@h DELETE #channel 123
```
delthas, line 167:
```suggestion
    C: REDACT #channel 123
    S: @msgid=124 :nick@u@h REDACT #channel 123
```

extensions/message-redaction.md
```diff
@@ -0,0 +1,183 @@
...
+
+Deleting someone else's PRIVMSG:
+
+    C: PRIVMSG #channel :join my network for cold hard chats
+    S: @msgid=123 :nick!u@h PRIVMSG #channel :join my network for cold hard chats
+    C: DELETE #channel 123 spam
+    S: @msgid=124 :nick!u@h DELETE #channel 123 spam
```
delthas, line 174:
```suggestion
    C: REDACT #channel 123 spam
    S: @msgid=124 :nick!u@h REDACT #channel 123 spam
```

delthas (commented):
😋

extensions/message-redaction.md
```diff
@@ -0,0 +1,192 @@
...
+Deleting a TAGMSG:
+
+    C: @draft/react=🤞TAGMSG #channel
+    S: @msgid=123;draft/react=🤞TAGMSG #channel
+    C: REDACT #channel 123
+    S: @msgid=124 :nick@u@h REDACT #channel 123
+
+Deleting someone else's PRIVMSG:
+
+    C: PRIVMSG #channel :join my network for cold hard chats
```
DarthGandalf, line 180: C1:
S:
C2:
S:

Is nick in second S: line intentionally the same as in first?

extensions/message-redaction.md
```diff
@@ -0,0 +1,192 @@
...
+name to be interoperable with other software implementing a compatible
+work-in-progress version.
+
+The final version of the specification will use an unprefixed capability name.
+
+## Introduction
+
+This specification enables messages to be deleted.
+Use cases include retracting accidentally sent messages, moderation,
+and removing a [`+draft/react` client tag][], amongst others.
```
DarthGandalf, line 32: What should happen to existing reactions to the redacted message? And what should happen to attempts to redact such reacts later?

extensions/message-redaction.md
```diff
@@ -0,0 +1,192 @@
...
+
+## Architecture
+
+### Dependencies
+
+Clients wishing to use these capabilities MUST negotiate the [`message-tags`][]
+capability with the server.
+Additionally, this capability relies on messages being sent with the
+[`msgid`][] tag, and message ids MUST NOT contain spaces.
+Clients SHOULD negotiate the [`echo-message`][] and [`labeled-response`][]
```
emersion, line 44: Nit: `labeled-response` is not strictly required here, a client could rely on `echo-message` only.

extensions/message-redaction.md
```diff
@@ -0,0 +1,192 @@
...
+capable of handling the respective commands and message tags described below.
+
+
+### Command
+
+To redact a message, a client MUST negotiate the `draft/message-redaction`
+capability and send a `REDACT` command to a target nickname or channel.
+The command is defined as follows:
+
+    REDACT <target> <msgid> [:reason]
```
emersion, line 64: In #ircv3 we discussed this `<target>` parameter and found that it could be a footgun for clients.

extensions/message-redaction.md
```diff
@@ -0,0 +1,192 @@
...
+batch when it re-joins the channel.
+
+If servers use predictable or guessable `msgid`s, they should considered whether errors
+returned on `REDACT` may leak a message's existence to users who did not receive it
+(in a channel they are/were not in or in private messages).
+
+### Fallback
+
+Server implementations might choose to inform clients that haven't negotiated
+the capability that an edit or deletion has taken place.
```
emersion, line 141: s/edit//

progval:
```suggestion
the capability that an edit has taken place.
```

extensions/message-redaction.md
```diff
@@ -0,0 +1,190 @@
...
+  same way as PRIVMSG messages.
+* MUST not forward this `REDACT` to target recipients which did not negotiated
+  this capability (see "Fallback" below)
+
+### Chat history
+
+After a message is redacted, [`chathistory`][] responses SHOULD either:
+
+* exclude it entirely
+* replace it with the `REDACT` message with the same value as `@msgid` tag
```
emersion, line 80: What does this mean exactly? This seems confusing to handle.

extensions/message-redaction.md
```diff
@@ -0,0 +1,190 @@
...
+* add the `REDACT` message to the response (not counting toward message limits)
```
emersion, line 81: How does this interact with the `event-playback` cap?

extensions/message-redaction.md
emersion: In particular: the original author of the message is unknown in this case. If the intent is to communicate that a message was redacted without disclosing the exact original contents, maybe replacing the message body with `[redacted]` would be a better option? (As in, less edge cases to handle for clients.)

```
@msgid=123 :nick!u@h PRIVMSG #channel :[redacted]
@msgid=124 :nick!u@h REDACT #channel 123 :bad example
```

emersion:
Goguma impl (currently untested): https://git.sr.ht/~emersion/goguma/log/redact

extensions/message-redaction.md
progval:
```suggestion
* replace its second parameter and/or tags with a placeholder and
  add the `REDACT` message to the response (not counting toward message limits)
```

extensions/message-redaction.md
progval: As in, redacting JOIN/PART/... messages?

emersion: My question is basically: are `REDACT` messages always returned by chathistory regardless of `event-playback`?

progval: I'd say yes. The motivation for `event-playback` is that it's hard to deal with channel state (membership, op list, ...) and not necessarily very useful, right? Here this shouldn't be an issue.

extensions/message-redaction.md
```diff
@@ -0,0 +1,191 @@
...
+  this capability (see "Fallback" below)
+
+### Chat history
+
+After a message is redacted, [`chathistory`][] responses SHOULD either:
+
+* exclude it entirely
+* replace its second parameter and/or tags with a placeholder and
+  add the `REDACT` message to the response (not counting toward message limits)
+* add the `REDACT` message to the response (not counting toward message limits)
```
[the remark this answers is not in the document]
progval: Why do client need to differentiate between the two?

extensions/message-redaction.md
emersion: We had more discussions about this. There are two reasons why server developers want this parameter:

- Remove the need for a global msgid lookup table. The table can be per-target. Servers can check that the msgid belongs to the right target in this case.
- Make it possible for servers to support this ext even if they don't keep track of msgids. Servers are not able to check that the msgid belongs to the right target in this case, the responsibility is shifted to the client.

extensions/message-redaction.md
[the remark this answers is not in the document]
emersion: Per the linked discussion above, my clients check whether there is more history to fetch by counting the number of messages. If it's less than the limit, no more messages.

Hm, but now that I've written this, this would only cause the number of messages to exceed the limit, causing needless but harmless chathistory requests in the edge case where there are no more messages. So it should be all fine?

Does a standalone `REDACT` count towards the limit? I'd say yes. This draws a line between "standalone" and "context" messages.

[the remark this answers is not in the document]
emersion: The discussion hasn't really settled AFAIU, and my clients are still doing the limit check.

Please note, there is an interleaved discussion about what should happen when a channel has chathistory disabled, and this one has settled on empty batches instead of a `FAIL`. But it's unrelated.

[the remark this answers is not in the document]
emersion: Oh, and another reason why clients must be able to tell the difference between a "context" `REDACT` and a "standalone" one: timestamp-based chathistory queries. Let's say a batch ends with a deleted message:

```
@time=2022-01-01T20:00:00Z;msgid=foo PRIVMSG #chan :Hey there
@time=2023-05-01T20:00:00Z REDACT #chan foo
```

Then the next chathistory request will fetch all messages after 2023-05-01…

extensions/message-redaction.md
emersion: Right. I agree.

extensions/message-redaction.md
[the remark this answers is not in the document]
emersion: There are two ways to fix this issue that I see:

- Add a `@chathistory-context` tag to all "related"/"context" messages returned by chathistory. Clients can ignore these when figuring out the selector for their next query.
- Add cursor-based paging to chathistory: #524.

emersion:
Goguma impl has been tested with Ergo.

progval:
@jwheare Anything blocking merge?

progval:
@jwheare ping?

extensions/message-redaction.md
```diff
@@ -0,0 +1,192 @@
...
+
+### Errors
+
+This specification defines `FAIL` messages using the [standard replies][]
+framework for notifying clients of errors with message editing and deletion.
+The following codes are defined, with sample plain text descriptions.
+
+* `FAIL REDACT INVALID_TARGET the_given_target :You cannot delete messages from the_given_target`
+* `FAIL REDACT REDACT_FORBIDDEN <target> <target-msgid> :You are not authorised to delete this message`
+* `FAIL REDACT REDACT_WINDOW_EXPIRED <target> <target-msgid> <window> :You can no longer edit this message`
```
SadieCat, line 92: How is the window communicated to clients?

progval: For now, it's not. It could be an informational mode though.

emersion: It's an easy thing to add later if needed. I don't believe it's worth blocking this PR because of this.

extensions/message-redaction.md
```diff
@@ -0,0 +1,193 @@
...
+* MUST not forward this `REDACT` to target recipients which did not negotiated
+  this capability (see "Fallback" below)
+
+### Chat history
+
+After a message is redacted, [`chathistory`][] responses SHOULD either:
+
+* exclude it entirely
+* replace its second parameter and/or tags with a placeholder and
+  add the `REDACT` message to the response (not counting toward message limits)
```
jwheare, line 82:
```suggestion
```

copy pasta

jwheare: Hmm never mind ignore, just got thrown by the (still annoying) hard wrap.

jwheare (changes requested):

extensions/message-redaction.md
```diff
@@ -0,0 +1,193 @@
...
+use cases would not be feasible without server validation.
+
+Such use cases might include:
+
+* Allowing channel moderators or server admins to delete unwelcome messages from others
+* Specifying a cut-off time after which message edits are no longer allowed
+
+If a message is redacted while a client is disconnected (or parted from a channel) while
+a message it saw is redacted, servers may send the `REDACT` command in a `chathistory`
+batch when it re-joins the channel.
```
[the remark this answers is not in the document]
progval: > Is this just saying that REDACT is allowed in chathistory batches?

Yes.

> Are we expecting servers to keep track of whether or not clients have seen redacts?

No, why?

extensions/message-redaction.md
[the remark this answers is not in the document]
progval: I'd say anywhere after the message itself, unless you see a reason to do otherwise.

extensions/message-redaction.md
```diff
@@ -0,0 +1,193 @@
...
+  same way as PRIVMSG messages.
+* MUST not forward this `REDACT` to target recipients which did not negotiated
+  this capability (see "Fallback" below)
+
+### Chat history
+
+After a message is redacted, [`chathistory`][] responses SHOULD either:
+
+* exclude it entirely
+* replace its second parameter and/or tags with a placeholder and
```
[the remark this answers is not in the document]
jwheare: I'd say "final parameter" or even better "message content" to make it more clear.

> It's up to implementations.

OK, but the "and/or tags" bit does beg the question why?

extensions/message-redaction.md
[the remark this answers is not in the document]
jwheare: Yeah that's fine, should be clarified then.

extensions/message-redaction.md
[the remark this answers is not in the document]
progval: > I'd say "final parameter" or even better "message content" to make it more clear.

Ok, done

> OK, but the "and/or tags" bit does beg the question why?

Because I don't see the benefit of listing which tags should be redacted and which shouldn't in the spec

emersion:
Gentle ping

extensions/message-redaction.md
```diff
@@ -0,0 +1,193 @@
...
+A key motivation for specifying this capability as a server tag, rather than
+a client-only message tag, is to enable more granular redaction permissions.
+Clients might be able to determine which messages are their own, but other
+use cases would not be feasible without server validation.
+
+Such use cases might include:
+
+* Allowing channel moderators or server admins to delete unwelcome messages from others
+* Specifying a cut-off time after which message edits are no longer allowed
+
```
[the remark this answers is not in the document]
progval: can't think of any. re-added.

extensions/message-redaction.md
```diff
@@ -0,0 +1,206 @@
...
+If a message is redacted while a client is not present in a channel, servers may send the `REDACT` command in a `chathistory` batch when it re-joins the channel.
```
jwheare, line 133:
```suggestion
Servers may send the `REDACT` command in a `chathistory` batch.
```

Removes the implication that servers are tracking whether clients are present in a channel or not.

Should this section be in the normative part of the spec?

extensions/message-redaction.md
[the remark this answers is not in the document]
jwheare: Clarification suggestion here: https://github.com/ircv3/ircv3-specifications/pull/524/files#r1500038542

extensions/message-redaction.md
progval: The implication was intentional. And the normative already says `REDACT` can be sent in `chathistory` batches.

extensions/message-redaction.md
jwheare: Does this need to be specified in this spec?

extensions/message-redaction.md
```diff
@@ -0,0 +1,206 @@
...
+    C2: REDACT #channel 123 spam
+    S:  :chanop!u@h REDACT #channel 123 spam
+
+
+[`echo-message`]: ../extensions/echo-message.html
+[`+draft/react` client tag]: ../client-tags/react.html
+[standard replies]: ../extensions/standard-replies.html
+[`message-tags`]: ../extensions/message-tags.html
+[`msgid`]: ../extensions/message-tags.html
+[`chathistory`]: ../extensions/message-tags.html
```
jwheare, line 206:
```suggestion
[`msgid`]: ../extensions/message-ids.html
[`chathistory`]: ../extensions/chathistory.html
```

extensions/message-redaction.md
jwheare: The spec talks about the chathistory spec, not chathistory batches.

progval: ugh. I guess I meant all chathistory batches rather than just chathistory responses

jwheare: I'm still not sure why this implication is necessary. Which servers keep track of past membership when determining whether to send REDACTs in chathistory?

progval: none of them, but they could if they wanted to

jwheare: That could apply to a lot of things, what's the motivation for the special case here. Are you thinking of this in the context of e.g. the server can send a REDACT and omit the original PRIVMSG?

progval: yes, so that clients end up showing the same things as if they were in the channel the whole time

extensions/message-redaction.md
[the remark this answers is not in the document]
jwheare: Does a resolution to this issue require changes to the redaction spec, or just to the chathistory spec? i.e. is this a blocker for merge/ratification?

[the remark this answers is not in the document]
progval: only the chathistory spec IMO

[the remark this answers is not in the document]
jwheare: Just spotted this other comment from @emersion that appears unaddressed https://github.com/ircv3/ircv3-specifications/pull/524/files#r1213547049

This PR has become a bit of a nightmare to review lol.

PR #526, opened 2023-06-14
chathistory: add chathistory-context tag

emersion:
See discussion at https://github.com/ircv3/ircv3-specifications/pull/524#discussion_r1214155605


slingamn:
I support this! I would have relaxed "MUST immediately follow" to "MUST appear after", but the language here works for me as long as it's helpful to clients.

emersion:
The intent here is to allow clients to figure out which message is the parent of a context message. For instance, a client might want to populate a "parent_context" column in its DB, to retrieve all the context messages attached to a parent message when locally fetching history.

That said, with msgid tags, it should be possible to reconstruct the relationships client-side, at least for context messages that the client understands.

Is the mandated behavior an issue for server implementors?

PR #528, opened 2023-08-27
echo-message: Make 'label' tags required in echoed messages

progval:
This is a middle-ground as #485 is stuck because some servers can't guarantee labels on all messages; but guaranteeing labels on echoed messages should be easier, and is extremely useful for clients.


delthas (approved):

spb:
Special-casing one particular instance is the wrong answer here - the right answer is #485. If we keep adding specific cases like this without merging the proper strict labeled-response then the spec is going to end up just codifying what InspIRCd has implemented, which isn't how things are supposed to work.

PR #531, opened 2023-09-03
away-notify: Clarify away notifications should not be echoed

progval:
This matches every known server implementation (with the possible exceptions of IRCCloud Teams and txird, which I did not test).

### DIFF
--- a/extensions/away-notify.md
+++ b/extensions/away-notify.md
@@ -30,7 +30,7 @@ state, as well as when a user joins and has an away message set.
 (Note that AWAY will not be sent for joining users with no away
 message set.)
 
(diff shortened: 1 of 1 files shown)


extensions/away-notify.md
```diff
@@ -30,6 +30,9 @@ state, as well as when a user joins and has an away message set.
 (Note that AWAY will not be sent for joining users with no away
 message set.)
 
+Clients should not be sent AWAY messages to notify them of their own away
```
jwheare, line 33:
```suggestion
Clients SHOULD not be sent AWAY messages to notify them of their own away
```

PR #535, opened 2024-02-22
Ratify the reply client tag specification.

SadieCat:
I don't think there are any objections relating to this and we have enough implementations so lets ratify and ship it!

---

## Known implementations

### Client (3/1)

* [x] catgirl
* [x] Goguma
* [x] IRCCloud 

# Server (2/2)

* [x] Anope (pseudoclient)
* [x] Matrix2051
* [x] UnrealIRCd

## Bots

* [x] BitBot
* [x] Limnoria
* [x] Moon Moon

### DIFF
--- a/client-tags/react.md
+++ b/client-tags/react.md
@@ -32,7 +32,7 @@ This tag provides a means of communicating with context-sensitive, potentially n
 
 ### Dependencies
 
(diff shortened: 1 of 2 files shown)


jwheare:
The only concern I have is do we want to prohibit multi level replies. I would favour doing so, but if not then we might need another to indicate the top level of a reply chain. If we do prohibit them, then react would have to override that rule.

PR #543, opened 2024-05-24
Add extended-isupport

emersion:
Proposal for https://github.com/ircv3/ircv3-ideas/issues/124

### DIFF
--- a/extensions/extended-isupport.md
+++ b/extensions/extended-isupport.md
@@ -25,17 +25,24 @@ entries. However, `RPL_ISUPPORT` is only sent by servers after connection
 registration. This undermines the usefulness of `RPL_ISUPPORT`: some of the
 metadata would be useful to clients prior to connection registration. This
 extension adds a mechanism to send `RPL_ISUPPORT` messages before connection
-registration, as well as an end of `RPL_ISUPPORT` list indication.
+registration, as well as a dedicated command to request the `ISUPPORT` list and
+a batch for `RPL_ISUPPORT` messages.
 
 ## Implementation
+
+### `ISUPPORT` command
+
+A new `ISUPPORT` command is introduced to request the full `RPL_ISUPPORT` list.
+When receiving this command, the server MUST reply with one or more
+`RPL_ISUPPORT` messages (grouped inside a `draft/isupport` batch if the `batch`
+and `draft/extended-isupport` capabilities are enabled, see below).
 
 ### `draft/extended-isupport` capability
 
 When the `draft/extended-isupport` capability is enabled by the client, the
-server MUST send one or more `RPL_ISUPPORT` messages (grouped inside a
-`draft/isupport` batch if the `batch` capability is enabled). The capability
-MAY be enabled by the client before connection registration completes (ie,
-before the client sends `CAP END`, and before the server sends `RPL_WELCOME`).
+server MUST accept `ISUPPORT` commands before connection registration
+completes (ie, before the client sends `CAP END`, and before the server sends
+`RPL_WELCOME`).
 
 Before connection registration completes, while `draft/extended-isupport` is
 enabled, the server MAY send updates to the key-value entries via subsequent
@@ -48,10 +55,6 @@ registration completes, if it already sent all information.
 Before connection registration, the server MAY send only a subset of the full
 `RPL_ISUPPORT` list. In that case, the server MUST send a `RPL_ISUPPORT` list
 when connection registration completes with entries previously omitted.
-
-The server MUST send one or more `RPL_ISUPPORT` messages when the capability is
-enabled, including after connection registration or if the disables then
-re-enables the capability.
 
 ### `draft/isupport` batch
 
@@ -75,6 +78,7 @@ Enabling the capability:
     S: :irc.example.org CAP * LS :multi-prefix sasl batch draft/extended-isupport
     C: CAP REQ batch draft/extended-isupport
     S: :irc.example.org CAP * ACK :batch draft/extended-isupport
+    C: ISUPPORT
     S: :irc.example.org BATCH +asdf draft/isupport
     S: @batch=asdf :irc.example.org 005 * NETWORK=Example NICKLEN=30 FOO=bar
     S: :irc.example.org BATCH -asdf
--- a/extensions/extended-isupport.md
+++ b/extensions/extended-isupport.md
@@ -50,10 +50,8 @@ Before connection registration, the server MAY send only a subset of the full
 when connection registration completes with entries previously omitted.
 
 The server MUST send one or more `RPL_ISUPPORT` messages when the capability is
-enabled, including after connection registration or if the client enables,
-disables and then enables again the capability. If the client requests the
-`draft/extended-isupport` capability while it's already enabled, the server
-MUST NOT send any `RPL_ISUPPORT` messages in response.
+enabled, including after connection registration or if the disables then
+re-enables the capability.
 
 ### `draft/isupport` batch
 
--- a/extensions/extended-isupport.md
+++ b/extensions/extended-isupport.md
@@ -49,12 +49,11 @@ Before connection registration, the server MAY send only a subset of the full
 `RPL_ISUPPORT` list. In that case, the server MUST send a `RPL_ISUPPORT` list
 when connection registration completes with entries previously omitted.
 
-If the client enables `draft/extended-isupport` after connection registration,
-the server MUST send one or more `RPL_ISUPPORT` messages. If the client
-enables, disables and then enables again `draft/extended-isupport` at any time,
-the server MUST send one or more `RPL_ISUPPORT` messages. If the client
-requests the `draft/extended-isupport` capability while it's already enabled,
-the server MUST NOT send any `RPL_ISUPPORT` messages in response.
+The server MUST send one or more `RPL_ISUPPORT` messages when the capability is
+enabled, including after connection registration or if the client enables,
+disables and then enables again the capability. If the client requests the
+`draft/extended-isupport` capability while it's already enabled, the server
+MUST NOT send any `RPL_ISUPPORT` messages in response.
 
 ### `draft/isupport` batch
 
--- a/extensions/extended-isupport.md
+++ b/extensions/extended-isupport.md
@@ -1,0 +1,93 @@
+---
+title: "`extended-isupport` Extension"
+layout: spec
+copyrights:
+  -
+    name: "Simon Ser"
+    period: "2024"
+    email: "contact@emersion.fr"
+---
+
+## Notes for implementing work-in-progress version
+
+This is a work-in-progress specification.
+
+Software implementing this work-in-progress specification MUST NOT use the
+unprefixed `extended-isupport` capability name. Instead, implementations SHOULD
+use the `draft/extended-isupport` capability name to be interoperable with
+other software implementing a compatible work-in-progress version. The final
+version of the specification will use unprefixed capability names.
+
+## Introduction
+
+`RPL_ISUPPORT` is used to advertise metadata about the server as key-value
+entries. However, `RPL_ISUPPORT` is only sent by servers after connection
+registration. This undermines the usefulness of `RPL_ISUPPORT`: some of the
+metadata would be useful to clients prior to connection registration. This
+extension adds a mechanism to send `RPL_ISUPPORT` messages before connection
+registration, as well as an end of `RPL_ISUPPORT` list indication.
+
+## Implementation
+
+### `draft/extended-isupport` capability
+
+When the `draft/extended-isupport` capability is enabled by the client, the
+server MUST send one or more `RPL_ISUPPORT` messages (grouped inside a
+`draft/isupport` batch if the `batch` capability is enabled). The capability
+MAY be enabled by the client before connection registration completes (ie,
+before the client sends `CAP END`, and before the server sends `RPL_WELCOME`).
+
+Before connection registration completes, while `draft/extended-isupport` is
+enabled, the server MAY send updates to the key-value entries via subsequent
+`RPL_ISUPPORT` messages (the same way it would after connection registration
+completes without this extension).
+
+The server MAY skip the `RPL_ISUPPORT` replies usually sent when connection
+registration completes, if it already sent all information.
+
+Before connection registration, the server MAY send only a subset of the full
+`RPL_ISUPPORT` list. In that case, the server MUST send a `RPL_ISUPPORT` list
+when connection registration completes with entries previously omitted.
+
+If the client enables `draft/extended-isupport` after connection registration,
+the server MUST send one or more `RPL_ISUPPORT` messages. If the client
+enables, disables and then enables again `draft/extended-isupport` at any time,
+the server MUST send one or more `RPL_ISUPPORT` messages. If the client
+requests the `draft/extended-isupport` capability while it's already enabled,
+the server MUST NOT send any `RPL_ISUPPORT` messages in response.
+
+### `draft/isupport` batch
+
+The server MUST group all `RPL_ISUPPORT` messages inside a `draft/isupport`
+batch when the [`batch`][] and `draft/extended-isupport` capabilities are
+enabled. The server MUST NOT send any unbatched `RPL_ISUPPORT` message while
+both of these capabilities are enabled. The order in which the capabilities are
+enabled is not significant.
+
+The batch MUST only contain one or more `RPL_ISUPPORT` messages, it MUST NOT
+contain any other message.
+
+As usual, servers can update or delete existing values by sending additional
+`RPL_ISUPPORT` messages in a `draft/isupport` batch after the initial batch.
+
+## Examples
+
+Enabling the capability:
+
+    C: CAP LS 302
+    S: :irc.example.org CAP * LS :multi-prefix sasl batch draft/extended-isupport
+    C: CAP REQ batch draft/extended-isupport
+    S: :irc.example.org CAP * ACK :batch draft/extended-isupport
+    S: :irc.example.org BATCH +asdf draft/isupport
+    S: @batch=asdf :irc.example.org 005 * NETWORK=Example NICKLEN=30 FOO=bar
+    S: :irc.example.org BATCH -asdf
+    C: NICK emersion
+    C: USER emersion 0 * :Simon Ser
+    C: CAP END
+    S: 001 emersion :어서 오세요
+
+Sending a change:
+
+    S: :irc.example.org BATCH +asdf draft/isupport
+    S: @batch=asdf :irc.example.org 005 * CHANNELLEN=64 NICKLEN=42 -FOO
+    S: :irc.example.org BATCH -asdf


extensions/extended-isupport.md
```diff
@@ -0,0 +1,69 @@
...
+and `RPL_ENDOFISUPPORT` messages.
+
+The server MAY skip the `RPL_ISUPPORT` replies usually sent when connection
+registration completes, if it already sent all information.
+
+Before connection registration, the server MAY send only a subset of the full
+`RPL_ISUPPORT` list. In that case, the server MUST send a `RPL_ISUPPORT` list
+when connection registration completes with entries previously omitted.
+
+### `RPL_ENDOFISUPPORT` numeric (370)
```
emersion, line 47: I picked a numeric here because it felt a bit awkward to have a 005 list ended by a non-numeric. But I'd be happy to change that if a non-numeric is deemed better.

SadieCat:
Rather than adding more numerics can we just require that it gets batched?

emersion:
Yeah, that would be a cleaner solution. I wonder if this would be hard for older servers to implement?

extensions/extended-isupport.md
```diff
@@ -0,0 +1,69 @@
...
+extension adds a mechanism to send `RPL_ISUPPORT` messages before connection
+registration, as well as an end of `RPL_ISUPPORT` list indication.
+
+## Implementation
+
+### `draft/extended-isupport` capability
+
+When the `draft/extended-isupport` capability is enabled by the client, the
+server MUST send one or more `RPL_ISUPPORT` messages followed by a
+`RPL_ENDOFISUPPORT` message. While `draft/extended-isupport` is enabled, the
```
emersion, line 36: What happens if the client enables, disables, then re-enables the capability? Should we require servers to re-send the list then?

What happens if the client enables the capability after connection registration? Same question: should we require servers to re-send the list then?

The current approach tries to keep the spec as simple as possible (always require to re-send the list) by avoiding special cases.

extensions/extended-isupport.md
emersion: Honestly, I have no idea. I just picked a random free one.

emersion:
Switched to a batch. Does this look good?

extensions/extended-isupport.md
```diff
@@ -0,0 +1,71 @@
...
+
+## Notes for implementing work-in-progress version
+
+This is a work-in-progress specification.
+
+Software implementing this work-in-progress specification MUST NOT use the
+unprefixed `extended-isupport` capability name. Instead, implementations SHOULD
+use the `draft/extended-isupportt` capability name to be interoperable with
+other software implementing a compatible work-in-progress version. The final
+version of the specification will use unprefixed capability names.
```
progval, line 19:
```suggestion
Software implementing this work-in-progress specification MUST NOT use the
unprefixed `extended-isupport` capability name. Instead, implementations SHOULD
use the `draft/extended-isupport` capability name to be interoperable with
other software implementing a compatible work-in-progress version. The final
version of the specification will use unprefixed capability names.
```

progval:
Could you mention that:

1. as before, ISUPPORT messages are additive (ie. if a new ISUPPORT overwrites or adds values, but the absence of a value does not mean it's not present anymore)
2. specify the `-` prefix to remove a key? I don't think this existed prior to this spec.

emersion:
> as before, ISUPPORT messages are additive (ie. if a new ISUPPORT overwrites or adds values, but the absence of a value does not mean it's not present anymore)

Not sure how to phrase this. Maybe:

> As usual, servers can update or delete existing values by sending additional `RPL_ISUPPORT` messages in a `draft/isupport` batch after the initial batch.

But maybe you want something more precise?

> specify the - prefix to remove a key? I don't think this existed prior to this spec.

It's in modern: https://modern.ircdocs.horse/#rplisupport-005

SadieCat:
> > specify the - prefix to remove a key? I don't think this existed prior to this spec.
> 
> It's in modern: https://modern.ircdocs.horse/#rplisupport-005

Also worth noting that Modern gets it from from the draft isupport spec.

extensions/extended-isupport.md
```diff
@@ -0,0 +1,71 @@
...
+extension adds a mechanism to send `RPL_ISUPPORT` messages before connection
+registration, as well as an end of `RPL_ISUPPORT` list indication.
+
+## Implementation
+
+### `draft/extended-isupport` capability
+
+When the `draft/extended-isupport` capability is enabled by the client, the
+server MUST send one or more `RPL_ISUPPORT` messages (grouped inside a
+`draft/isupport` batch if the `batch` capability is enabled). While
```
SadieCat, line 36: It might be an idea to spec this up in its own section so it can be used for other 005 responses.

emersion: The intent is that the section about the batch covers this, specifically:

> The server MUST group all `RPL_ISUPPORT` messages inside a `draft/isupport`
batch when the `batch` and `draft/extended-isupport` capabilities are
enabled.

SadieCat: Fair enough I must have missed that.

jwheare: It's mentioned in the introduction that this is sent before registration, but I think this section is where it needs to be explicit. Also maybe worth using wording to clarify that this is sent before capability negotiation is completed with CAP END.

extensions/extended-isupport.md
```diff
@@ -0,0 +1,71 @@
...
+    C: CAP REQ draft/extended-isupport
+    S: :irc.example.org CAP * ACK draft/extended-isupport
+    S: :irc.example.org BATCH +asdf draft/isupport
+    S: @batch=asdf :irc.example.org 005 * NETWORK=Example NICKLEN=30
+    S: :irc.example.org BATCH -asdf
+
+Sending a change:
+
+    S: :irc.example.org BATCH +asdf draft/isupport
+    S: @batch=asdf :irc.example.org 005 * CHANNELLEN=64 -NETWORK
```
jwheare, line 70: -NETWORK isn't a very realistic example. Also, maybe show a key being changed via removal and readdition. Also these examples could do with the full CAP negotiation being shown, including CAP END and maybe the 001 being sent.

extensions/extended-isupport.md
```diff
@@ -0,0 +1,71 @@
...
+
+## Implementation
+
+### `draft/extended-isupport` capability
+
+When the `draft/extended-isupport` capability is enabled by the client, the
+server MUST send one or more `RPL_ISUPPORT` messages (grouped inside a
+`draft/isupport` batch if the `batch` capability is enabled). While
+`draft/extended-isupport` is enabled, the server MAY send updates to the
+key-value entries via subsequent `RPL_ISUPPORT` messages.
```
jwheare, line 38: This behaviour is standard ISUPPORT and isn't exclusive to this cap being enabled.

extensions/extended-isupport.md
SadieCat: I think the intended meaning here is that users will get updates before connection registration?

jwheare: Yeah the last part refers to the previous example sorry.

extensions/extended-isupport.md
```diff
@@ -0,0 +1,71 @@
...
+
+## Examples
+
+Enabling the capability:
+
+    C: CAP LS 302
+    S: :irc.example.org CAP * LS :multi-prefix sasl draft/extended-isupport
+    C: CAP REQ draft/extended-isupport
+    S: :irc.example.org CAP * ACK draft/extended-isupport
+    S: :irc.example.org BATCH +asdf draft/isupport
```
slingamn, line 63: This appears to show `BATCH` being sent without the client having enabled the `batch` CAP.

My reading of the current draft is that `BATCH` should only be sent if the client enables the `batch` CAP. Is that the correct understanding? (Another possibility is that requesting `draft/extended-isupport` implies the ability to receive batches, and therefore the server should always send `BATCH`.)

slingamn: Following up on discussion in #ircv3: this spec is similar to [draft/multiline](https://ircv3.net/specs/extensions/multiline) in that it depends on the `batch` CAP. draft/multiline says:

>This specification depends on the [batch](https://ircv3.net/specs/extensions/batch) capability which MUST be negotiated to use multiline messages. The order of capability negotiation is not significant and MUST not be enforced.

I read "MUST not be enforced" in this case to mean that if the client requests `draft/multiline` before `batch`, the server MUST send an `ACK` for `draft/multiline` and record that it was enabled. If the client continues with only `draft/multiline` enabled, server behavior is undefined.

However, the request order is necessarily significant for `draft/extended-isupport` because of the "edge-triggered" design. If the client successfully requests `draft/extended-isupport`, it should receive an 005 burst in response. If the client doesn't have `batch` enabled at that moment, what should the server do? (My vote would be for the server to send the 005 burst without batching.)

SadieCat: I don't see why the batch cap couldn't be optional for clients here tbh. Not every client needs the request in a batch.

slingamn: That works for me too --- just want the spec to be explicit about the behavior, and the example to be consistent with that behavior.

extensions/extended-isupport.md
emersion: Right. I wasn't sure how far the examples should go, I wouldn't want to list the full registration burst. Will add.

`-NETWORK` could happen if the network configuration file is reloaded with the network name removed. Do you have another example in mind which might be more realistic?

extensions/extended-isupport.md
emersion: Yeah, the intent is to document that servers might send updates pre-conn-reg as well. Will make more explicit.

extensions/extended-isupport.md
emersion: The examples seems wrong here indeed.

My preference would be to allow `draft/extended-isupport` without `batch` (in which case messages are sent unbatched) and make the order insignificant.

extensions/extended-isupport.md
emersion: Tried to make all of this stuff more explicit:

> If the client enables `draft/extended-isupport` after connection registration,
the server MUST send one or more `RPL_ISUPPORT` messages. If the client
enables, disables and then enables again `draft/extended-isupport` at any time,
the server MUST send one or more `RPL_ISUPPORT` messages. If the client
requests the `draft/extended-isupport` capability while it's already enabled,
the server MUST NOT send any `RPL_ISUPPORT` messages in response.

extensions/extended-isupport.md
emersion: > Also, maybe show a key being changed via removal and readdition.

The ISUPPORT spec says that changing a key is done by advertising it again, rather than removing it and then re-adding it.

emersion:
I've tried to address the comments. Let me know what you think!

extensions/extended-isupport.md
```diff
@@ -0,0 +1,93 @@
...
+Before connection registration, the server MAY send only a subset of the full
+`RPL_ISUPPORT` list. In that case, the server MUST send a `RPL_ISUPPORT` list
+when connection registration completes with entries previously omitted.
+
+If the client enables `draft/extended-isupport` after connection registration,
+the server MUST send one or more `RPL_ISUPPORT` messages. If the client
+enables, disables and then enables again `draft/extended-isupport` at any time,
+the server MUST send one or more `RPL_ISUPPORT` messages. If the client
+requests the `draft/extended-isupport` capability while it's already enabled,
+the server MUST NOT send any `RPL_ISUPPORT` messages in response.
```
jwheare, line 57: What's the purpose of this? Servers can already send RPL_ISUPPORT messages to notify changes without this cap. I don't see why enabling the cap should trigger isupport being sent again. Bear in mind clients can also trigger isupport being sent with the /version command.

emersion: The goal is to have (1) all of ISUPPORT tokens get sent pre-connection-registration and (2) keep things as dumb/simple as possible. See also https://github.com/ircv3/ircv3-specifications/pull/543#discussion_r1613437758

jwheare:
Changes look good. No need to force push updates, makes it a bit harder to review. I squash commits on merge anyway.

extensions/extended-isupport.md
jwheare: I get (1), but this para seems to be about post registration behaviour.

By (2) do you mean you want to avoid servers having to special case whether it's pre or post reg before deciding whether to sent isupport? If so, this para is worded as though it *is* a special case which might defeat the objective a bit.

emersion: Yeah, the intent is to have the exact same behavior regardless of registration state...

emersion: Tried to make this more explicit.

extensions/extended-isupport.md
slingamn: If I understand correctly, this is the expected behavior:

```
>>> CAP REQ draft/extended-isupport
:oragono.test CAP * ACK draft/extended-isupport
:oragono.test 005 * AWAYLEN=500 BOT=B CASEMAPPING=ascii CHANLIMIT=#:100 CHANMODES=Ibe,k,fl,CEMRUimnstu CHANNELLEN=64 CHANTYPES=# CHATHISTORY=100 ELIST=U EXCEPTS EXTBAN=,m EXTJWT=1 FORWARD=f :are supported by this server
:oragono.test 005 * INVEX KICKLEN=1000 MAXLIST=beI:60 MAXTARGETS=4 MODES MONITOR=100 MSGREFTYPES=msgid,timestamp NETWORK=OragonoTest NICKLEN=32 PREFIX=(qaohv)~&@%+ RPCHAN=E RPUSER=E STATUSMSG=~&@%+ :are supported by this server
:oragono.test 005 * TARGMAX=NAMES:1,LIST:1,KICK:,WHOIS:1,USERHOST:10,PRIVMSG:4,TAGMSG:4,NOTICE:4,MONITOR:100 TOPICLEN=1000 UTF8ONLY WHOX draft/CHATHISTORY=100 :are supported by this server
>>> CAP REQ batch
:oragono.test CAP * ACK batch
```

This sounds correct to me (at the moment `draft/extended-isupport` is enabled, `batch` is not yet enabled, so the burst is sent unbatched). But I think "the order in which the capabilities are enabled is not significant" is a misleading description of the behavior: the order is significant in that it affects whether the burst is batched.

SadieCat: This isn't typically what cap ordering means, afaik it means that

    CAP REQ :batch draft/extended-isupport

is equivalent to

    CAP REQ :draft/extended-isupport batch

slingamn: Hmm, I always understood `REQ` as an atomic and simultaneous request for all the capabilities on the line. Is order ever significant on the same line?

extensions/extended-isupport.md
```diff
@@ -0,0 +1,92 @@
...
+
+Before connection registration, the server MAY send only a subset of the full
+`RPL_ISUPPORT` list. In that case, the server MUST send a `RPL_ISUPPORT` list
+when connection registration completes with entries previously omitted.
+
+The server MUST send one or more `RPL_ISUPPORT` messages when the capability is
+enabled, including after connection registration or if the client enables,
+disables and then enables again the capability. If the client requests the
+`draft/extended-isupport` capability while it's already enabled, the server
+MUST NOT send any `RPL_ISUPPORT` messages in response.
```
jwheare, line 56:
```suggestion
The server MUST send one or more `RPL_ISUPPORT` messages when the capability is
enabled, including after connection registration or if the client disables then re-enables
the capability.
```

Tidied wording a bit. I don't think the thing about enabling when already enabled is necessary, it feels like
a) behaviour that should be defined (if at all) in the cap spec
b) unlikely to be something server cap implementations will trip over
c) not a huge problem if you get sent those messages again, if you implement the client correctly.

emersion:
Alternatively, instead of spec'ing that servers must send the ISUPPORT list when the cap is enabled, we could guarantee that VERSION works pre-reg and recommend that clients send it after enabling the cap.

extensions/extended-isupport.md
emersion: Yeah. I just didn't want clients to rely on this. But I don't mind dropping it.

extensions/extended-isupport.md
emersion: Ref https://github.com/ircv3/ircv3-specifications/pull/549#issuecomment-2267452934

extensions/extended-isupport.md
```diff
@@ -0,0 +1,90 @@
...
+
+The server MAY skip the `RPL_ISUPPORT` replies usually sent when connection
+registration completes, if it already sent all information.
+
+Before connection registration, the server MAY send only a subset of the full
+`RPL_ISUPPORT` list. In that case, the server MUST send a `RPL_ISUPPORT` list
+when connection registration completes with entries previously omitted.
+
+The server MUST send one or more `RPL_ISUPPORT` messages when the capability is
+enabled, including after connection registration or if the disables then
```
jwheare, line 53:
```suggestion
enabled, including after connection registration or if the client disables then
```

slingamn:
I would be slightly more comfortable with allowing `VERSION` pre-registration --- the "edge-triggered" API has always bothered me a little (on the other hand, I haven't been able to identify any concrete problems with it).

`VERSION` has a wrinkle where you also get `RPL_VERSION`, so presumably you would get `RPL_VERSION` outside the batch and then a batch of `RPL_ISUPPORT`? We could add a command like `ISUPPORT` that just sends the `RPL_ISUPPORT`.

jwheare:
Just specifying an ISUPPORT command that can work pre-reg works for me. Regarding batch, I feel like this was already discussed, but what use case does batch enable? My implementation just incrementally updates the ISUPPORT list whenever an 005 is received, I don't have a need for an end of isupport marker.

slingamn:
@jwheare there is a note on this here: https://github.com/ircv3/ircv3-ideas/issues/124#issuecomment-2130183063

tl;dr it allows the UI to be updated atomically, reducing "flicker".

extensions/extended-isupport.md
jwheare: In multiline the intention is just that a server shouldn't NAK a client that REQs multiline and then batch with two separate REQs.

jwheare:
Ok so maybe define the ISUPPORT command to require use of a batch if the cap is enabled? But leave the existing automatic isupport sent post reg unbatched. Would that work?

emersion:
That wouldn't work if the server wants to send an atomic update of more than 13 tokens.

extensions/extended-isupport.md
emersion: The edge-triggered behavior has been removed.

emersion:
A new `ISUPPORT` command has been introduced, and the `RPL_ISUPPORT` list is no longer sent when the capability is enabled.

extensions/extended-isupport.md
emersion: The example has been fixed, and it's been spelled out that the order doesn't matter.

slingamn:
Looks good, I merged support for this in ergo and deployed it on my staging server ircs://files.stronghold.network

emersion:
I've finally got some time to implement this in Goguma! Here's the patch: https://codeberg.org/emersion/goguma/pulls/2

Ergo's implementation seems to have a small bug: there is an extraneous `chathistory` param in the batch parameters.

```
flutter: [3] <- :files.stronghold.network BATCH +1 chathistory draft/extended-isupport
flutter: [3] <- @batch=1 :files.stronghold.network 005 * AWAYLEN=500 BOT=B CASEMAPPING=ascii CHANLIMIT=#:100 CHANMODES=Ibe,k,fl,CEMRUimnstu CHANNELLEN=64 CHANTYPES=# CHATHISTORY=100 ELIST=U EXCEPTS EXTBAN=,m FORWARD=f INVEX :are supported by this server
flutter: [3] <- @batch=1 :files.stronghold.network 005 * KICKLEN=1000 MAXLIST=beI:60 MAXTARGETS=4 MODES MONITOR=100 MSGREFTYPES=msgid,timestamp NETWORK=OraStage NICKLEN=32 PREFIX=(qaohv)~&@%+ STATUSMSG=~&@%+ TARGMAX=NAMES:1,LIST:1,KICK:,WHOIS:1,USERHOST:10,PRIVMSG:4,TAGMSG:4,NOTICE:4,MONITOR:100 TOPICLEN=1000 UTF8MAPPING=rfc8265 :are supported by this server
flutter: [3] <- @batch=1 :files.stronghold.network 005 * UTF8ONLY WHOX draft/CHATHISTORY=100 :are supported by this server
flutter: [3] <- :files.stronghold.network BATCH -1
```

With a small workaround for the broken batch type, it seems to work fine:

out

slingamn:
Sorry about that. I fixed it in https://github.com/ergochat/ergo/pull/2197 and redeployed.

emersion:
The batch is named "draft/isupport", not "draft/extended-isupport".

emersion:
Gentle ping :)

jwheare:
Fine to merge this as is, but would the non draft cap name be more self explanatory as prereg-isupport? The format is not being extended by this afaict.

emersion:
The name "extended-isupport" was picked because the cap also introduces a new batch and a new command. I'm fine using any other name if you prefer.

PR #546, opened 2024-06-10
Clarify that the WHOX reply RPL_WHOSPCRPL may contain multiple prefixes even without the multi-prefix capability

JustAnotherArchivist:
For historical context: ircu [first introduced WHOX in version 2.10.01](https://ircu.sourceforge.net/release.2.10.01-who.html) and appears to have originally only returned the highest prefix like regular WHO (verified in version 2.10.08.01). https://github.com/UndernetIRC/ircu2/commit/7ec4144c6b646292025b62ebeb1f974e02c03698 introduced returning all prefixes and was first released in version 2.10.11. Charybdis adopted this behaviour from the first version that implemented WHOX support (3.1, commits https://github.com/charybdis-ircd/charybdis/commit/48957a495191764aa3222fd8b15371c979621b61 and https://github.com/charybdis-ircd/charybdis/commit/02eca3f19a01a0fb55fdb65b83f3ec8fd9d7ad5f). I haven't inspected other ircds. Since both behaviours exist in released ircds, the wording is intentionally vague.

### DIFF
--- a/extensions/whox.md
+++ b/extensions/whox.md
@@ -77,6 +77,9 @@ MUST be used:
 - Any syntactically correct placeholder can be used for `[oplevel]` when the
   server doesn't support op levels, for instance `n/a`.
 
+Servers MAY return multiple or all applicable prefixes in the `[flags]` (`f`)
+field, regardless of whether [`multi-prefix`](multi-prefix.html) is enabled.
+
 Clients SHOULD ignore the values of the hop count (`d`) and the channel op
 level (`o`) fields, because they are ill-defined and unreliable.
 


SadieCat:
Perhaps it should say that clients SHOULD handle this but servers SHOULD NOT emit it?

extensions/whox.md
```diff
@@ -77,6 +77,9 @@ MUST be used:
 - Any syntactically correct placeholder can be used for `[oplevel]` when the
   server doesn't support op levels, for instance `n/a`.
 
+Servers MAY return multiple or all applicable prefixes in the `[flags]` (`f`)
+field, regardless of whether [`multi-prefix`](multi-prefix.html) is enabled.
```
nektro, line 81: this should link to `multi-prefix.md` and the site generator will rewrite it

JustAnotherArchivist:
@SadieCat I'm not sure about that. Both the ircd that invented WHOX (ircu) and at least one modern ircd (Solanum) do emit them – and have been doing so for over 23 years in the case of ircu. I'm sure a significant number of clients (including older or obscure ones without `multi-prefix` support) expect servers to send it, too, given that it was originally introduced specifically so clients could know the state of a channel's users accurately; this was the only way to get all prefixes at the time. Further, the WHOX document is one of the few that, to my understanding, is more of a documentation of existing technology than an IRCv3 specification. See also the PR that added it: #482

extensions/whox.md
JustAnotherArchivist: Hmm, this would be the only `.md` link in the repository. I based this link on what I found in other documents, which all use `.html`, e.g. in `extensions/echo-message.md` or `extensions/extended-monitor.md`.

vanosg:
I agree that this is a gap in documentation that would be beneficial to document- I don't think we at Eggdrop are fully handling multiple flags being sent because of our perhaps outdated assumptions and a lack of explicit definition telling us otherwise.  I defer to others to agree how  WHOX should perform, but my personal opinion is that "WHOX flags are sent the same way WHO flags are sent" makes sense, which allows the negotiation of multi-prefix to dictate whether those WHO flags are sent in full or in part.

PR #547, opened 2024-07-01
EXTJWT command for integrating external web services

ItsOnlyBinary:
> @prawnsalad 
> This spec provides a way for web services hosted externally to an IRC server to authenticate users that are connected to the IRC server by making use of the standard JWT tokens (https://jwt.io/).
> 
> This allows a web service to do things such as:
> 
>     * Granting admin access to a networks wiki page if a user has +o on the network
> 
>     * Granting write access to a channels wiki page if the user has +o in the channel
> 
>     * Automatically creating a user account if the user is logged into the IRC server and has an account name
> 
> 
> For a more indepth example we could use the free audio/video conference service - Jitsi Meet. This service has built in JWT verification in that an application can send a user to a URL that contains a JWT token, and if the Jitsi Meet server verifies this token successfully, the user is granted access to that conference room. When an IRC client wants to join a conference room, it would first call `EXTJWT #testchannel` to receive a JWT token from the IRCd. The client would then open a browser window navigating to the Jitsi Meet URL while passing that token. It is up to the client to decide how and where to use this token, eg. via a "Jitsi Call" button for example.

As prawnsalad has stepped back from IRC development and I have taken over [Kiwi IRC](https://github.com/kiwiirc/kiwiirc) development. I cannot allow this ircv3 specification to be abandoned as [Kiwi IRC](https://github.com/kiwiirc/kiwiirc) uses it for quite a few plugins. (jitsi conferencing, file uploads, account avatar uploads)

So I have also forked #341 and updated based on comments from within the previous pull-request.

### DIFF
--- a/extensions/extjwt.md
+++ b/extensions/extjwt.md
@@ -21,7 +21,7 @@ Once in use, this:
 
 ### How the proof works
 
-The IRC network creates the proof by generating JWT tokens and sending them to the client. The client may use this token to open an external service with the token in its URL.
+The IRC network creates the proof by generating signed JWT tokens and sending them to the client. The client may use this token to open an external service with the token in its URL.
 
 The JWT token (https://jwt.io/) is an encoded JSON payload that is signed with a shared secret string between the IRC server and the external service. The JSON payload consists of known properties (claims) that include:
 * `exp` `1529917513` Expiry time for this token. Usually less than 1 minute from the token generation.
@@ -42,9 +42,9 @@ Only one new command is introduced in this extension, `EXTJWT`.
 
 Syntax: `EXTJWT [channel]`
 
-Response syntax: `EXTJWT <requested_target> [*] <JWT_token>`
+Response syntax: `EXTJWT <requested_target> [*] <jwt_token>`
 
-The client may send `EXTJWT` or `EXTJWT *` to the server to request a new JWT token. The server must then reply with `EXTJWT *` and a JWT token as its second parameter, containing the following claims that are relevant to the client at that time:
+The client MAY send `EXTJWT` or `EXTJWT *` to the server to request a new JWT token. The server MUST then reply with `EXTJWT *` and a JWT token as its jwt_token parameter containing the following claims that are relevant to the client at that time:
 
 * `exp` Number; Unix timestamp for when this token expires. Usually less than 1 minute from the token generation.
 * `iss` String; The server name that generated this token.
@@ -52,18 +52,18 @@ The client may send `EXTJWT` or `EXTJWT *` to the server to request a new JWT to
 * `account` String; The account name of the user that requested this token. Empty if not available.
 * `net_modes` []String; An array of user modes the IRCd may want to disclose. Eg, if the user is an operator.
 
-The command must also support a single parameter of a channel name. Eg. `EXTJWT #channel`. The server must then reply with the channel name as its first parameter and the JWT token containing the above claims and also the following claims relevant to the channel at that time:
+The command MUST also support a single parameter of a channel name. Eg. `EXTJWT #channel`. The server MUST then reply with the channel name as its requested_target parameter, the JWT token containing the above claims and also the following claims relevant to the channel at that time:
 
 * `channel` String; The channel name this token is related to.
 * `joined` Boolean; True if the client that requested this token is joined to the channel.
 * `time_joined` Number; The time in which the user joined the channel.
 * `modes` []String; An array of the channel modes the client has in this channel.
 
-The IRC server must include the above claims but may include any extra claims.
+The IRC server MUST include the above claims but MAY include any extra claims.
 
 #### Handling long responses
 
-In some cases the encoded token may be longer than the maximum line length allowed between the client and server. In this case, the first parameter of the response must be `*` to indicate that further data will follow. The final chunk of the response sent to the client must not include `*` as the first parameter.
+In some cases the encoded token may be longer than the maximum line length allowed between the client and server. In this case, the first parameter of the response MUST be `*` to indicate that further data will follow. The final chunk of the response sent to the client MUST NOT include `*` as the first parameter.
 
 Eg:
 ~~~
--- a/extensions/extjwt.md
+++ b/extensions/extjwt.md
@@ -33,7 +33,7 @@ The JWT token (https://jwt.io/) is a base64 encoded JSON payload that is signed 
 
 When an external service is opened with this token in its URL, the external service verifies that the token has not been tampered with by one of two methods:
 1. If the external service has a shared secret for this IRC network, it can now verify the token directly.
-2. The external service can make a HTTP GET request to the URL given in the `vfy` claim, replacing `%s` with the token string. The request MUST be responded to with a HTTP 200 status if the token has been verified, or a HTTP 403 status if the token is invalid.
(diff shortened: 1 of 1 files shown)


ItsOnlyBinary:
Reasoning behind removing the `vfy` claim:

Having the verify url within the token would encourage implementations to use the url without verifying its trusted, allowing the spoofing of tokens. 

The verify url should be a pre-shared component for trust, just like the key would be pre-shared.

SadieCat:
Our implementation is updated to match the new pull request.

https://github.com/inspircd/inspircd-contrib/blob/master/4/m_ircv3_extjwt.cpp

extensions/extjwt.md
```diff
@@ -0,0 +1,139 @@
...
+30 seconds is usually enough time for the client to receive the token from the `EXTJWT` command and then open the external service webpage, however, non-webpage based services may require a longer expiration depending on its implementation.
+
+## Examples
+
+All examples may be verified using the secret of "your-256-bit-secret".
+
+#### A client logged into the IRC server with operator privileges
+~~~
+[C -> S] EXTJWT *
+[S -> C] EXTJWT * * eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJleHAiOjE1Mjk5MTc1MTMsImlzcyI6ImlyYy5leGFtcGxlLm9yZyIsInN1YiI6InNvbWVuaWNrIiwiYWNjb3VudCI6InNvbWVuaWNrIiwidW1vZGVzIjpbIm8iXX0.wZbYLX4rgDDB4svLQIsx5jq5_Dc0csdqgamVsgocOas
```
slingamn, line 111: Only one `*` parameter here, right?

extensions/extjwt.md
```diff
@@ -0,0 +1,139 @@
...
+
+Where the replied token is decoded into:
+~~~json
+{"exp":1529917513,"iss":"irc.example.org","sub":"somenick","account":"somenick","umodes":["o"]}
+~~~
+
+#### A client connected to the IRC server without a registered account or operator privileges
+~~~
+[C -> S] EXTJWT *
+[S -> C] EXTJWT * * eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJleHAiOjE1Mjk5MTc1MTMsImlzcyI6ImlyYy5leGFtcGxlLm9yZyIsInN1YiI6InNvbWVuaWNrIiwiYWNjb3VudCI6IiIsInVtb2RlcyI6W119.i_ak4qvb1BPdH0a0HyRNTz036rHE2lGrZ17SQV3LAFE
```
slingamn, line 122: Same here

jwheare:
Can we get an audit on implementations for this. Those I'm aware of:

* kiwi
* unreal (https://github.com/ircv3/ircv3-specifications/pull/341#issuecomment-1002240181)
* inspircd (https://github.com/ircv3/ircv3-specifications/pull/547#issuecomment-2200779024)

Anyone else?

extensions/extjwt.md
```diff
@@ -0,0 +1,139 @@
...
+* `umodes` []String; An array of user modes the IRCd may want to disclose. Eg, if the user is an operator.
+
+If the command sent by the client includes a service name, Eg. `EXTJWT * jitsi`, the server MUST then reply with the service name as its `service_name` parameter, along with the JWT token containing the above claims and also the following claims relevant to the service:
+
+* `service` String; The configured service name that can be used to verify the token, with either a pre-shared key or url.
+
+If the command sent by the client includes a channel name, Eg. `EXTJWT #channel`, the server MUST then reply with the channel name as its `requested_target` parameter, along with the JWT token containing the above claims and also the following claims relevant to the channel at that time:
+
+* `channel` String; The channel name this token is related to.
+* `joined` Number; Unix timestamp of the time at which the client joined the channel. 0 if not joined. 1 if no timestamp is available.
```
skizzerz:
```suggestion
* `joined` ?Number; Unix timestamp of the time at which the client joined the channel. `null` if not joined. 0 if no timestamp is available.
```

Similar to `account`, we can make values nullable so trying to do that to indicate lack of value makes more sense than identifying special sentinel values.

extensions/extjwt.md
```diff
@@ -0,0 +1,139 @@
...
+Response syntax: `EXTJWT <requested_target> <service_name> [*] <jwt_token>`
+
+The `service_name` is an optional name for what the token is being requested for. Eg, `EXTJWT * forum.example.org`. The server MAY take this service name to sign its token using different secrets. It defaults to `*` if not provided.
+
+The client will send at minimum `EXTJWT *` to the server to request a new JWT token. The server MUST then reply with the above response syntax with the `requested_target` and `service_name` defaulting to `*` if they were not provided in the request. The JWT token MUST contain the following claims that are relevant to the client at that time:
+
+* `exp` Number; Unix timestamp for when this token expires. See below for notes on the expiry claim.
+* `iss` String; The server name that generated this token.
+* `sub` String; The nick of the client that requested this token.
+* `account` String; The account name of the user that requested this token. Empty if not available.
```
skizzerz:
```suggestion
* `account` ?String; The account name of the user that requested this token. `null` if not available.
```

JSON allows `null` values, and JWT allows any JSON values. Making these values nullable to designate a lack of value rather than designating certain strings as "special" makes a lot more sense to me.

extensions/extjwt.md
```diff
@@ -0,0 +1,139 @@
...
+* `umodes` []String; An array of user modes the IRCd may want to disclose. Eg, if the user is an operator.
+
+If the command sent by the client includes a service name, Eg. `EXTJWT * jitsi`, the server MUST then reply with the service name as its `service_name` parameter, along with the JWT token containing the above claims and also the following claims relevant to the service:
+
+* `service` String; The configured service name that can be used to verify the token, with either a pre-shared key or url.
+
+If the command sent by the client includes a channel name, Eg. `EXTJWT #channel`, the server MUST then reply with the channel name as its `requested_target` parameter, along with the JWT token containing the above claims and also the following claims relevant to the channel at that time:
+
+* `channel` String; The channel name this token is related to.
```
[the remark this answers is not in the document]
skizzerz: I wasn't quite sure about what a channel-scoped JWT would actually be used for. It very well could be not correct in which case it'd be better to not implement this suggestion (and leave `aud` omitted potentially for such JWTs)

ItsOnlyBinary:
Kiwi IRC currently has 3 main plugins that use EXTJWT they are:

`kiwiirc/plugin-fileuploader` allows file uploads that can be restricted to logged in accounts only or allow logged in account files to be retained for longer.

`kiwiirc/plugin-conference` integrates jitsi-meet in to channels and queries allowing for video conferencing with IRC. when used with a jitsi-meet side auth plugin this can be used to control who is moderator of the conference based on IRC channel mode and also restrict channels etc.

`itsonlybinary/kiwiirc-plugin-avatar-upload` allows logged in users to upload avatars which can then be shown to all users.

Other network admins do appear to have also created quite complex account/profile systems that also use EXTJWT tokens to link different pieces.

I hope these demonstrate the usefulness of this spec for linking IRC with 3rd party services

sadiepowell:
Given its still a draft spec I'm fine with breaking compatibility to make the IRCv3 implementation behave more like the upstream spec, if necessary server implementations can emit both keys for a period.

PR #549, opened 2024-08-04
clarify that the order of CAP REQ parameters is not significant

slingamn:
As per discussion here: https://github.com/ircv3/ircv3-specifications/pull/543#discussion_r1702186139

my understanding is that this is not currently specified, but that it's not a behavior change for any actually existing servers, and there's no use case for having order matter within the same line (if the client cares about ordering or dependencies, it should use multiple lines, or rely on the `CAP LS` output).

### DIFF
--- a/extensions/capability-negotiation.md
+++ b/extensions/capability-negotiation.md
@@ -311,6 +311,7 @@ which is not enabled, the server MUST continue processing the REQ subcommand as 
 handling this capability was successful.
 
 The capability identifier set must be accepted as a whole, or rejected entirely.
+The order of capabilities within the list is not significant.
 
 Clients SHOULD ensure that their list of requested capabilities is not too long to be
 replied to with a single ACK or NAK message. If a REQ's final parameter gets sufficiently
--- a/extensions/capability-negotiation.md
+++ b/extensions/capability-negotiation.md
@@ -550,3 +550,6 @@ Previous versions of this spec did not state that capability names MUST NOT star
 
 Previous versions of this spec did not state that space-separated capability lists may end with
 a trailing space.
+
+Previous versions of this spec did not state that the order of capabilities within a single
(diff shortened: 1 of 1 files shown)


progval:
might be worth adding an errata ("Previous versions of this specification did not say order of capabilities in CAP REQ is not meaningful")

emersion:
Seems to contradict with this bit of the spec:

> The list of capabilities MUST be parsed and processed from left to right

slingamn:
Thanks, good point. This was added in 1e7df885e1fb6b21044021956004d7656a309f68. Does anyone know what the intent was?

slingamn:
55df8fdc81fb5e08875145bae22439ff4df34fb5 changed "parsed from left to right" to "parsed and processed from left to right". This came out of review activity on #327. Chronologically it might relate to this comment: https://github.com/ircv3/ircv3-specifications/pull/327#discussion_r179913823

My sense right now is that specifying left-to-right parsing is reasonable because it provides an unambiguous semantics for duplicated CAP values, but the semantic result of the parsing should be an unordered mapping.

jwheare:
Yeah I think this change is incorrect. We should keep the existing left to right significance.

slingamn:
I think the underlying question here is actually "is there a use case for atomic, simultaneous updates of capabilities via CAP REQ"? If the answer is yes, then that would be a reason to adopt the syntax/semantics split I described earlier, where parsing is left-to-right but the result of the parsing is an unordered mapping that gets applied atomically. If the answer is no, then the current behavior is fine and  `CAP REQ :baz bat` (at least when successful) is synonymous with `CAP REQ baz` followed by `CAP REQ bat`.

But also this is basically moot if #543 switches away from an edge-triggered design where `CAP REQ` itself triggers a side effect, and instead relies on an explicit command to trigger the burst. So there's no need to get into that question now.

PR #550, opened 2024-08-04
add message formats to extensions/capability-negotiation.md

winrg:


extensions/capability-negotiation.md
```diff
@@ -131,6 +131,14 @@ with several subcommands.  The command added is named `CAP`.  `CAP` takes a sing
 required subcommand, optionally followed by a single parameter. Each subcommand defines any
 further parameters.
 
+The format of a client-sent `CAP` message is:
+
+    CAP <subcommand> [<param>{,<param>}]
```
emersion, line 136: Hm, I find the `[<param>{,<param>}]` a bit confusing, is it used anywhere else? (`{` and `,` in particular)

winrg: Taken from https://modern.ircdocs.horse/#join-message

emersion: Ah, that one is different, because the JOIN messages accepts two comma-separated arguments:

    JOIN #chan1,#chan2 key1,key2

winrg: My mistake, will edit.

extensions/capability-negotiation.md
```diff
@@ -131,6 +131,14 @@ with several subcommands.  The command added is named `CAP`.  `CAP` takes a sing
 required subcommand, optionally followed by a single parameter. Each subcommand defines any
 further parameters.
 
+The format of a client-sent `CAP` message is:
+
+    CAP <subcommand> [<parameters>]
```
emersion, line 136: Maybe `[<parameters>...]` would make it clearer that these are multiple fields, not a single one?

extensions/capability-negotiation.md
```diff
@@ -176,6 +184,18 @@ The server MUST accept the `CAP` command at any time, including after registrati
 The LS subcommand is used to list the capabilities supported by the server.  The client
 should send an LS subcommand with no other arguments to solicit a list of all capabilities.
 
+The format of a server-sent `CAP LS` message is:
+
+    CAP LS <version>
```
emersion, line 189: Should probably be `[<version>]` because it's an optional argument.

extensions/capability-negotiation.md
```diff
@@ -176,6 +184,18 @@ The server MUST accept the `CAP` command at any time, including after registrati
...
+
+The format of a server-sent `CAP LS` message is:
+
+    CAP <nick> LS :[<extension 1> ... [<extension n>]]
```
emersion, line 193: Do we want to allow empty strings here? I suppose nothing is forbidding this in the spec…

extensions/capability-negotiation.md
```diff
@@ -343,6 +383,10 @@ change the capabilities of the client until the last ACK of the set has been sen
 
 ### The CAP NAK subcommand
 
+The format of a `CAP NAK` message is:
+
+    CAP <nick> NAK :[[-]<extension 1> ... [[-]<extension n>]]
```
emersion, line 388: I don't think the spec allows dashes here: compare "The last parameter is a space-separated list of capabilities." with the wording from CAP ACK where dashes are explicitly allowed.

emersion (commented):
Apart from this, looks good!

extensions/capability-negotiation.md
winrg: Good idea

extensions/capability-negotiation.md
winrg: Oops

extensions/capability-negotiation.md
winrg: Spec explicitly allows empty strings for `LS` and `LIST`

emersion: ACK!

winrg: Actually, this makes me wonder about `ACK` and `NAK`, should they support it?

winrg: `NAK` uses the phrase "space-seperated list of capabilities", which is the same as in `LS` and `LIST`.

extensions/capability-negotiation.md
winrg: I guess it would be pretty silly to allow a server to explicitly refuse disabling a capability.

extensions/capability-negotiation.md
winrg: Should I add language to explicitly permit empty `REQ`, `ACK`, and `NAK`?

extensions/capability-negotiation.md
jwheare: Unless it wasn't already requested, or is not a requestable cap (e.g. sts)

winrg: Possibly? The protocol does explicitly allow for erroneous REQ entries, but I think the server would always be required to ACK them.

> If a client requests a capability which is already enabled, or tries to disable a capability which is not enabled, the server MUST continue processing the REQ subcommand as though handling this capability was successful.

extensions/capability-negotiation.md
emersion: I don't see any use-case for allowing `REQ` with an empty string.

Do servers in the wild support it?

These three commands need to all accept it or all reject it, since a server will reply to a `REQ` with an `ACK`/`NAK` with the same parameter.

emersion (approved):
LGTM!

extensions/capability-negotiation.md
```diff
@@ -334,18 +370,27 @@ Example removing a capability:
 The ACK subcommand is sent by the server to acknowledge a client-sent REQ, and let the client
 know that their requested capabilities have been enabled.
 
+The format of a `CAP NAK` message is:
```
jwheare, line 373:
```suggestion
The format of a `CAP ACK` message is:
```

PR #552, opened 2024-08-19
Notify client tag

SadieCat:
[Rendered version available here.](https://github.com/SadieCat/ircv3-specifications/blob/notify/client-tags/notify.md)

## Introduction

This specification defines a client-only message tag to indicate to the receiving client how they should notify the user.

## Motivation

Sometimes when you send a message to a user you don't want to bother them immediately. Unfortunately, it is not currently possible to control whether a user will be notified by a message. This specification defines a method for controlling notifications in compatible clients.

---

This is still very WIP. Any comments are massively appreciated.

### DIFF
--- a/client-tags/notify.md
+++ b/client-tags/notify.md
@@ -1,0 +1,68 @@
+---
+title: "`notify` client tag"
+layout: spec
+work-in-progress: true
+copyrights:
+  -
+    name: "Sadie Powell"
+    period: "2024"
+    email: "sadie@witchery.services"
+---
+
+## Notes for implementing work-in-progress version
+
+This is a work-in-progress specification.
+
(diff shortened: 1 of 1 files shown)


dgw:
I don't understand the dependency on `reply`, maybe because the examples don't include the reply tag.

Is this to control notifications from messages that are replying to a specific message rather than controlling notifications triggered by a highlight?

What about messages that highlight more than one user, and you want to bother one/more but not the rest?

SadieCat:
> I don't understand the dependency on `reply`, maybe because the examples don't include the reply tag.

Oops I copy/pasted that section from another client tag spec. Removed.

SadieCat:
As an alternative to this I also considered a `+quiet` tag that just turns off notifications entirely but this felt marginally more useful.

emersion:
In this situation:

    C: PRIVMSG #ircdocs :emersion: I'm not sure, try asking val when they're around?

How would the client indicate that emersion needs to be notified but val should not?

PR #556, opened 2024-10-26
Add SAFERATE spec

delthas:
This is an "implementation" of the [rate limiting ircv3-idea](https://github.com/ircv3/ircv3-ideas/issues/116), for the first possible solution described in it: "Add a new cap to indicate that a server won't disconnect a client because of rate limits. The server would still be free to slowly process message bursts, this would result in writes blocking on the client-side.".

## Implementations

- Server: [soju](https://github.com/emersion/soju/commit/6e4fb181d87be7f0790562844c2a3d9d3dcb346e) (as client and server)

### DIFF
--- a/extensions/saferate.md
+++ b/extensions/saferate.md
@@ -1,0 +1,34 @@
+---
+title: SAFERATE ISUPPORT token
+layout: spec
+work-in-progress: true
+copyrights:
+  -
+    name: "delthas"
+    period: "2024"
+---
+
+## Introduction
+
+This is a work-in-progress specification.
+
+This specification offers a way for servers to advertise that clients will not
+be disconnected due to server rate-limiting / anti-flood, and so that the client
+can send messages without doing its own conservative rate-limiting.
+
+## The `SAFERATE` ISUPPORT token
+
+This specification introduces the `SAFERATE` ISUPPORT token.
+
+When advertised, the server ensures that a client will not be disconnected due
+to server rate-limiting.
+
+The token MUST NOT be advertised with a value.
+
+## Implementation considerations
+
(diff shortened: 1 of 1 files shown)


SadieCat:
I'm not sure this is usable in practice. Even with fakelag users will eventually hit some kind of hard limit on most implementations.

slingamn:
Just as an example: current versions of Ergo in the default/recommended configuration could publish this token. Writes will start to block on the client side (the limiting factor in practice is the TCP buffer size) but there is no "hard limit".

slingamn:
re. usefulness to implementations, I believe it is demonstrably useful to soju and senpai?

SadieCat:
I'm not sure another specification that is only usable by Ergo is a good idea. Maybe this should be a vendor extension?

slingamn:
This specification is implemented by Soju-as-a-server as well, but I didn't comment on the implementation details there since I'm not familiar with them.

delthas:
Edit: Added InspIRCd support.

emersion:
Now that we have 3 implementations, can we merge this @jwheare?

slingamn:
Someone in #ircv3 was arguing that in principle, the protocol requires every client to maintain two queues, a low-priority queue for ordinary messages and a high-priority queue for sending PONG in a timely fashion. I don't think most implementations worry about this, but it might be worth a note in the "implementation considerations" section that a server publishing this token should relax its expectations around PONG, i.e. tolerate PONG being delayed because it's stuck in the TCP buffer behind fakelag.

DarthGandalf:
> the protocol requires every client to maintain two queues

ZNC currently has a single queue, but PONG gets pushed to the front of the queue instead of back

PR #558, opened 2024-11-02
Clarify that users don't get metadata updates they didn't SUB to

progval:
This text conflicted with these two other parts of the spec:

> If `draft/metadata-notify-2` is negotiated, clients SHOULD receive the current values of keys they are subscribed to when they [`MONITOR`]

and

> The `metadata-notify` key subscribed you to all keys. Since we have now added the [`SUB`](#metadata-sub) and [`UNSUB`](#metadata-unsub) subcommands, `metadata-notify-2` does not act in this way.


jwheare:
Closed in favour of #559

PR #562, opened 2025-04-06
Add filehost

### LINKED ISSUE #100: File upload
On many messaging platforms, the server provider is also responsible for providing a way for clients to upload attachments. This removes the need for clients to maintain/pick their own upload service.

@progval has written a proposal: https://github.com/progval/ircv3-specifications/blob/filehost/extensions/filehost.md

Prior art:

- https://xmpp.org/extensions/xep-0363.html
- https://spec.matrix.org/v1.3/client-server-api/#content-repository

emersion:
This is based on [soju.im/FILEHOST](https://soju.im/filehost), which was itself based on an earlier draft by @progval.

Implementations: none so far.

Closes https://github.com/ircv3/ircv3-ideas/issues/100

### DIFF
--- a/extensions/filehost.md
+++ b/extensions/filehost.md
@@ -1,0 +1,108 @@
+---
+title: The Filehost ISUPPORT token
+layout: spec
+work-in-progress: true
+copyrights:
+  -
+    name: "Val Lorentz"
+    email: "progval+ircv3@progval.net"
+    period: "2022"
+  -
+    name: "Simon Ser"
+    period: "2024"
+---
+
+# filehost
+
+This is a work-in-progress specification.
+
+Software implementing this work-in-progress specification MUST NOT use the
+unprefixed `FILEHOST` ISUPPORT name. Instead, implementations SHOULD use the
+`draft/FILEHOST` ISUPPORT name to be interoperable with other software
+implementing a compatible work-in-progress version. The final version of the
+specification will use unprefixed ISUPPORT names.
+
+## Motivation
+
+This specification offers a way for servers to advertise a hosting service for
+users to upload files (such as text or images), so they can post them on IRC.
+
+## Architecture
+
+### ISUPPORT token
+
+This specification introduces the `draft/FILEHOST` ISUPPORT token.
+
+Its value MUST be a URI and SHOULD use the `https` scheme. Clients MUST ignore
+tokens with an URI scheme they don't support. Clients MUST refuse to use
+unencrypted URI transports (such as plain `http`) if the IRC connection is
+encrypted (e.g. via TLS).
+
+Servers MUST accept `OPTIONS` requests on the upload URI. Servers MAY return an
+`Accept-Post` header field to indicate the MIME types they accept. Servers MAY
+advertise support for [resumable uploads].
+
+When clients wish to upload a file using the server's recommended service, they
+can send a request to the upload URI. The request method MUST be `POST`.
+Clients MUST include an `Authorization` header field with the `Bearer` scheme
+and a fresh token obtained via the `FILEHOST TOKEN` command.
+
+Clients SHOULD use the `Content-Type`, `Content-Disposition` and
+`Content-Length` header fields to indicate the MIME type, name and size of the
+file to be uploaded.
+
+On success, servers MUST reply with a `201 Created` status code and with a
+`Location` header field indicating the URI of the uploaded file. Servers MUST
+support `HEAD` and `GET` requests on the uploaded file URI.
+
+Clients SHOULD gracefully handle other common HTTP status codes that could
+occur.
+
+### `FILEHOST TOKEN` command
+
+When the `draft/FILEHOST` ISUPPORT token is advertised, servers MUST support
+the `FILEHOST TOKEN` command. When sent from the client, the command takes no
+parameter:
+
+    FILEHOST TOKEN
+
+The server MUST reply with a successful `FILEHOST TOKEN` or a `FAIL` reply.
+A successful `FILEHOST TOKEN` server reply has the following syntax:
+
+    FILEHOST TOKEN <token>
+
+Servers MUST ensure that tokens are short-lived and MAY invalidate tokens once
+they are used.
+
+## Examples
+
+Example isupport token:
+
+    :irc.example.org 005 seunghye draft/FILEHOST=https://irc.example.org/upload
+
+Example `OPTIONS` response:
+
+    HTTP/1.1 204 No Content
+    Allow: OPTIONS, POST
+    Accept-Post: image/*, video/*
+
+Example token query:
+
+    C: FILEHOST TOKEN
+    S: FILEHOST TOKEN oophay9HohKeAhb1nohy1iec
+
+Example `POST` request:
+
+    POST /upload HTTP/1.1
+    Host: irc.example.org
+    Content-Type: image/jpeg
+    Content-Disposition: inline; filename="picture.jpeg"
+    Content-Length: 4242
+    Authorization: Bearer oophay9HohKeAhb1nohy1iec
+
+Example `POST` response:
+
+    HTTP/1.1 201 Created
+    Location: /upload/hoh5eFThae4e.jpeg
+
+[resumable uploads]: https://datatracker.ietf.org/doc/draft-ietf-httpbis-resumable-upload/
--- a/extensions/filehost.md
+++ b/extensions/filehost.md
@@ -1,0 +1,90 @@
+---
+title: The Filehost ISUPPORT token
+layout: spec
+work-in-progress: true
+copyrights:
+  -
+    name: "Val Lorentz"
+    email: "progval+ircv3@progval.net"
+    period: "2022"
+  -
+    name: "Simon Ser"
+    period: "2024"
+---
+
+# filehost
+
+This is a work-in-progress specification.
+
+Software implementing this work-in-progress specification MUST NOT use the
+unprefixed `FILEHOST` ISUPPORT name. Instead, implementations SHOULD use the
+`draft/FILEHOST` ISUPPORT name to be interoperable with other software
+implementing a compatible work-in-progress version. The final version of the
+specification will use unprefixed ISUPPORT names.
+
+## Motivation
+
+This specification offers a way for servers to advertise a hosting service for
+users to upload files (such as text or images), so they can post them on IRC.
+
+## Architecture
+
+This specification introduces the `draft/FILEHOST` ISUPPORT token.
+
+Its value MUST be a URI and SHOULD use the `https` scheme. Clients MUST ignore
+tokens with an URI scheme they don't support. Clients MUST refuse to use
+unencrypted URI transports (such as plain `http`) if the IRC connection is
+encrypted (e.g. via TLS).
+
+Servers MUST accept `OPTIONS` requests on the upload URI. Servers MAY return an
+`Accept-Post` header field to indicate the MIME types they accept. Servers MAY
+advertise support for [resumable uploads].
+
+When clients wish to upload a file using the server's recommended service, they
+can send a request to the upload URI. The request method MUST be `POST`.
+
+If the client has authenticated via [`sasl`], the client MUST include an
+`Authorization` header field. If the `PLAIN` SASL mechanism was used, the
+client MUST use the HTTP `Basic` authorization scheme. If the `OAUTHBEARER`
+SASL mechanism was used, the client MUST use the HTTP `Bearer` authorization
+scheme.
+
+Clients SHOULD use the `Content-Type`, `Content-Disposition` and
+`Content-Length` header fields to indicate the MIME type, name and size of the
+file to be uploaded.
+
+On success, servers MUST reply with a `201 Created` status code and with a
+`Location` header field indicating the URI of the uploaded file. Servers MUST
+support `HEAD` and `GET` requests on the uploaded file URI.
+
+Clients SHOULD gracefully handle other common HTTP status codes that could
+occur.
+
+## Examples
+
+Example isupport token:
+
+    :irc.example.org 005 seunghye draft/FILEHOST=https://irc.example.org/upload
+
+Example `OPTIONS` response:
+
+    HTTP/1.1 204 No Content
+    Allow: OPTIONS, POST
+    Accept-Post: image/*, video/*
+
+Example `POST` request:
+
+    POST /upload HTTP/1.1
+    Host: irc.example.org
+    Content-Type: image/jpeg
+    Content-Disposition: inline; filename="picture.jpeg"
+    Content-Length: 4242
+    Authorization: Basic c2V1bmdoeWU6bm8=
+
+Example `POST` response:
+
+    HTTP/1.1 201 Created
+    Location: /upload/hoh5eFThae4e.jpeg
+
+[`sasl`]: ../extensions/sasl-3.1.html
+[resumable uploads]: https://datatracker.ietf.org/doc/draft-ietf-httpbis-resumable-upload/


extensions/filehost.md
```diff
@@ -0,0 +1,75 @@
...
+unencrypted URI transports (such as plain `http`) if the IRC connection is
+encrypted (e.g. via TLS).
+
+Servers MUST accept OPTIONS requests on the upload URI. Servers MAY return an
+`Accept-Post` header field to indicate the MIME types they accept.
+
+When clients wish to upload a file using the server's recommended service, they
+can send a request to the upload URI. The request method MUST be POST. Clients
+SHOULD authenticate their HTTP request with the same credentials used on the
+IRC connection (e.g. HTTP Basic for SASL PLAIN, HTTP Bearer for SASL
```
progval, line 39: The spirit of my draft was that servers would embed credentials in the URL, to allow uploading to third-party services. With your design, third-party services get connection credentials that are valid for the IRCd in the most common SASL scheme (PLAIN).

emersion: Yes. This is an intentional design decision. The filehost server can forward the request to another server if need be.

progval: but it still knows the password

emersion: I'm not sure I understand?

What I meant is that the filehost server can be hosted by the same folks as the IRC server, can check the credentials in the request, and can forward requests without IRC credentials to another server if need be.

Here's why I deviated from your proposal:

1. Putting secrets in URLs is a bit scary. ISUPPORT is in general not security-sensitive.
2. The upload service cannot be any random third-party: it needs to implement the spec's endpoints. In general, third-party upload services have a custom API.
3. With your proposal, the external service still needs to be somehow tied to the IRC server to check the credentials embedded in the request URL. Credentials need to be per IRC connection. With signed tokens you hit issues regarding revocation etc.
4. It resulted in a lot of additional complexity in my server implementation. On the other hand, proxying is very simple (and implemented in soju).

progval: > What I meant is that the filehost server can be hosted by the same folks as the IRC server

My point was that it doesn't have to be.

> The upload service cannot be any random third-party: it needs to implement the spec's endpoints.

There's no endpoint, it's just a URL that accepts generic HTTP POST

> With your proposal, the external service still needs to be somehow tied to the IRC server to check the credentials embedded in the request URL.

My thinking was that the IRC server would create token on the external service, not the other way around.

> Putting secrets in URLs is a bit scary. ISUPPORT is in general not security-sensitive.

> It resulted in a lot of additional complexity in my server implementation. On the other hand, proxying is very simple (and implemented in soju).

Those however, are good points. I'm sold.

bb010g:
This is missing the normal `draft/` prefix warning section:

> ## Notes for implementing work-in-progress version
>
> This is a work-in-progress specification.
>
> Software implementing this work-in-progress specification MUST NOT use the unprefixed `FILEHOST` isupport token. Instead, implementations SHOULD use the `draft/FILEHOST` isupport token to be interoperable with other software implementing a compatible work-in-progress version.
>
> The final version of the specification will use an unprefixed isupport token.

This absence has already led to ObsidianIRC jumping the gun: <https://github.com/ObsidianIRC/ObsidianIRC/blob/9632acfb35cbd208bfc94e66d5e2882f47ef58ec/src/protocol/isupport.ts#L64>

```TypeScript
    if (key === "FILEHOST" || key === "draft/FILEHOST") {
```

bb010g:
Ah, I had no idea the spec change there was so recent and that both soju and this proposal were previously out of compliance. Apologies.

ValwareIRC:
I think another thing was that there was prior art on the filehost spec from progval which didn't have the prefix in it, so relying on it being unprefixed to determine if it's a final version won't work so great.

extensions/filehost.md
```diff
@@ -0,0 +1,75 @@
...
+users to upload files (such as text or images), so they can post them on IRC.
+
+## Architecture
+
+This specification introduces the `draft/FILEHOST` isupport token.
+
+Its value MUST be a URI and SHOULD use the `https` scheme. Clients MUST ignore
+tokens with an URI scheme they don't support. Clients MUST refuse to use
+unencrypted URI transports (such as plain `http`) if the IRC connection is
+encrypted (e.g. via TLS).
```
furudean, line 37: Is there any actual relation between an IRC server's use of TLS vs insecure connection, and the accompanying upload server's connection protocol?

Many IRC networks provide insecure connections for wider client support. If a client is going to put in the effort today to implement `draft/filehost`, wouldn't it be more sensible to _require HTTPS_, given that we're sending credentials over the network? HTTPS is not hard to implement today, and I think we can make that requirement of a web server. Otherwise we encourage insecure practices for a brand new feature. Either that, or not be opinionated about what protocol is used to connect to the upload server.

classabbyamp:
not sure if it would count towards ratification, but I wrote a thing to use FILEHOST to proxy to various pastebins: https://github.com/classabbyamp/convoyeur

extensions/filehost.md
furudean: These motivations make sense to me. Would it be a good idea to include them in the RFC? I could see others asking why it's that way otherwise.

emersion:
AFAIK soju only so far. Ergo has added a way to manually insert ISUPPORT tokens from the config file but that requires server operators to bring up a separate HTTP file upload server:

https://github.com/ergochat/ergo/blob/c79b65cf8ad25e77367c863f20ddf38cfeb52bf3/default.yaml#L401

bb010g:
UnrealIRCd has a 3rd party module: <https://github.com/unrealircd/unrealircd-contrib/blob/929da5a2d628370b86129543d6936c08d7c4efa3/files/filehost.c>

furudean:
When implementing support for this and `soju.im/filehost` in Halloy I [raised a security concern](https://github.com/squidowl/halloy/pull/1680#issuecomment-4103733359):

> I'm having second thoughts with sending the user's credentials and I'm strongly now leaning that this is a **bad** idea. A server being able to specify an ISUPPORT token for uploads (that may be mistakenly sent) is an abuse vector. 
> 
> There are no semantics in `draft/FILEHOST` that defines how authentication should be made with the server. The non-normative example contains an Authorization header, but does not define _how_ authentication should be done. `soju.im/filehost` on the other hand, does specify:
> 
> > Clients SHOULD authenticate their HTTP request with the same credentials used on the IRC connection (e.g. HTTP Basic for SASL PLAIN, HTTP Bearer for SASL OAUTHBEARER).
> 
> It follows by implementation that soju's upload server does require credentials, or the request will fail. 
> 
> The client should follow the semantics of the upload server, but this is not specified well for `draft/FILEHOST`. Guessing is a bad idea. We could dynamically activate `send_credentials` when `soju.im/filehost` and  send credentials in that case, though I don't like this very much either.
> 
> I would prefer if it instead explicit in the config and then documented how to use this feature with soju instead. It's more to configure, but we avoid potentially leaking people's IRC credentials just for connecting.

As a result of this, I opted to not send the user's credentials by default in Halloy. I do not know concretely what sorts of changes should be made to the spec if any, but I figured I would at least raise my thoughts here to prompt discussion.

sadiepowell:
I have previously raised the idea of the server providing credentials to the user for the upload server. I still think this (as well as my proposal to formally require that upload servers use TUS) are a good idea.

furudean:
@sadiepowell does tus provide additional security over regular uploads?

sadiepowell:
No, but its a standard system for handling resumable uploads which is better than the vibes-based approach in the current spec.

reesericci:
Just going to forward this contribution I made from the Halloy repo:

---

It might make sense here to have a standardized OAuth/OIDC-style handshake between the file hosting providers, the bouncer/server and the client. I'd be happy to write up an RFP to IRCv3 to add this as the standard if people would be interested. Here's a little mermaid diagram of how this could work:

```mermaid
sequenceDiagram
actor Client
participant Server
participant Filehost@{ "type": "database"}
	Filehost-->>Server: Requests OIDC configuration
	Server->>Filehost: Returns OIDC configuration with JWKS url
	Filehost-->>Server: Requests JSON Web Keyset (JWKS) to verify tokens issued by the server
	Server->>Filehost: Returns JWKS
	Client-->>Server: Establish connection
	Server->>Client: Advertises FILEHOST
	Client -->>Server: Requests access token
	Server -->>Filehost: Requests nonce and provides client IP address to expect upload from
	Filehost->>Server: Returns nonce (bound to client IP)
	Server->>Client: Returns signed JWT with expiration, nonce and hashed username as data
	Client->>Filehost: Sends file using multipart/form-data with key `file` and the JWT in the Authorization header
	Filehost->>Client: Verifies signature with cached JWKS, nonce, expiration, audience and if valid returns file URL
```

_Originally posted by @reesericci in https://github.com/squidowl/halloy/issues/1680#issuecomment-4104700277_

ValwareIRC:
ObsidianIRC uses JWT >.>

emersion:
@furudean I don't understand your concern. The user already sends the credentials to the IRC server each time they connect. This spec makes it so clients also send credentials to the server-controlled upload endpoint. What is the threat model here? Why would a server send an untrusted ISUPPORT token?

> There are no semantics in draft/FILEHOST that defines how authentication should be made with the server.

`draft/FILEHOST` in this PR has the same authentication requirements as `soju.im/FILEHOST`. (Note that unmerged specs are subject to breaking changes.)

Celelibi:
I can't talk for @furudean but it makes sens to me to want to reduce the amount of credentials on the wire. And the filehost service doesn't necessarily lives on the same physical server as the IRCd. Not sending credentials to the filehost server could limit the impact of a compromised server.

emersion:
Credentials are transmitted every time a client connects anyways, so the benefit of skipping it when uploading is pretty low.

I agree that securing credentials in general is a good idea, but I think it would be worthwhile to improve that for the IRC connection too instead of limiting it to HTTP file upload. This is possible via multiple ways, e.g. https://github.com/ircv3/ircv3-specifications/pull/584 or OAuth2 with dynamic client registration.

sadiepowell:
Networks may want to offload file storage to a server which they do not control so I think it would be good to provide temporary tokens for use with the upload server.

slingamn:
My two cents: this specification works for the use cases it was intended for, and lacks features that would allow it to scale to a public network:

* Token-based authentication (as per Sadie's suggestion that this facilitates using a file hosting server that is less trusted than the ircd)
* File deletion and enumeration
* Communicating limits and quotas to the user (individual file size limit, total upload limit)
* Optionally, access controls for viewing the files (or some other mechanism to prevent use of the server for hosting malicious or illegal content)

I think it would be appropriate for this spec to coexist with a greenfield design that has hardening and additional features (I like Sadie's suggestion of [tus.io](https://tus.io/)).

reesericci:
I agree that we should take steps towards supporting OpenID connect - it would provide us a solid foundation to build on. I think this filehost specification is a good place to start because IRCv3 file hosts don't currently exist (except Soju), so we can implement this however we'd like.

IreneKnapp:
I share the concern about credential management, I really don't like the notion that the user uses the same credentials to connect to the file service's http endpoint as they did to log into IRC. It's much too tight a coupling, it really limits the implementation options. I see how the OIDC handshake would mitigate that concern but it still feels more complex than needed, to me.

I have an alternative implementation strategy I'd like to suggest.

Instead of having the CAP entry give a static upload URL, what about adding an additional command that the client sends when it wants to upload an image, and the server responds with a one-time-use upload URL for that specific image?

The details of how the upload URL works would be left up to the server, but for example, any file storage service that implements the S3 API is able to produce signed, time-limited upload URLs, and there are multiple independent implementations of it. That would be usable in its present form, with no modifications needed to the existing functionality on the S3 side of things.

furudean:
@IreneKnapp I really like this idea because if a server wanted it could provide a static URL when responding to the upload request. It aligns with the simplicity of the original proposal, in cases where network security is not an issue

slingamn:
IMO this idea presents similar security concerns to [EXTJWT](https://github.com/ircv3/ircv3-specifications/pull/547) (or slightly more, since capability URLs are slightly easier to mishandle than JWTs), but is less flexible:

1. JWTs have a predefined mechanism for carrying additional ircd-validated data (account name or channel scope)
2. JWTs could be used to authenticate to delete a file, or enumerate stored files associated with the user
3. In deployments that want to require authentication to read as well as write, JWTs would be usable for this purpose as well. (I'm imagining that the client could request an EXTJWT for previewing purposes, or autodownload in some potential cases. If the user views the link in a browser, the filehost would implement some other mechanism for authenticating the request (not HTTP Basic), likely via cookies.)

reesericci:
Hey y'all - I'm going to start working on a "competing" proposal based on OAuth/OIDC and tus.io. Would anyone be interested in working on this with me? I'm @reesericci on the IRC as well.

emersion:
### OAuth/OIDC

I am very strongly against making file upload depend on an OAuth/OIDC server. It is a very important requirement for me to make it easy to deploy a file upload server with an existing server setup or to write a patch to a server to support this. OAuth/OIDC is way too high of a barrier to entry here.

A proper OAuth/OIDC server is a whole project to build or get running. OAuth/OIDC adds a whole bunch of interactions clients need to perform. The OAuth/OIDC framework is very complex to understand in the first place (and has a lot of gotcha's). And that's not even getting into the OAuth client registration discussion.

This doesn't mean I hate OAuth. In fact, I think it can be useful in some scenarii, and [I've even written a blog post about that](https://emersion.fr/blog/2022/irc-and-oauth2/). I'd really like to make OAuth easier to use with IRC, but I don't believe tying it to file uploads is a good way forward.

By the way, `soju.im/FILEHOST` is perfectly usable with OAuth. A server can accept `OAUTHBEARER` SASL authentication, and the file upload endpoint can accept `Bearer` HTTP authentication. Then no user password is ever seen nor transmitted by the client.

### JWT

Some people really like JWTs, some people really dislike these. I personally would very much prefer to avoid them in my software. I understand some people want to use them and I'd like to make that possible too.

From a technical point-of-view, I don't think the client really needs to care what kind of token it receives? Can't the client treat the token as an opaque string, and leave it up to the server to decide if it needs to be JWT or something else?

### Credentials

I still don't really understand why sending username/password to the file upload endpoint is so controversial, nonetheless replacing that with something else is a compromise I'm perfectly willing to make.

Some have suggested to add a command to grab a short-lived (perhaps one-time-use) upload token. That sounds reasonable to me.

Some have suggested to add a command to get a URL with an implicit authorization embedded (e.g. token in a query parameter). The URL would become confidential because it contains security-sensitive material. That sounds fine to me as well, since the security-sensitive material would only be carried in a new command, much like AUTHENTICATE (as opposed to ISUPPORT, which is often times logged/printed/cached/etc).

classabbyamp:
i think an opaque token (whether generated per-connection or by a command) is the easiest and most flexible way

ValwareIRC:
### Regarding sending plain credentials to services with a potential to be unrelated to the IRC network
Why would we be sending our plain credentials to/via some different software? Sending plain credentials implies that the filehost has to exclusively either:

1. Be inside a services package
2. Share/piggyback databases with the services package
3. Validate passwords using RPC or some admin method via the services package

Most people will be pointing things towards their already-existing favorite (probably self-hosted) pastebin, or even some other public pastebin. Sharing a un-decryptable JWT token with them is nowhere near as bad as assuming we can send our IRC credentials to it.

ObsidianIRC's implementation of a filehost is actually designed for ObsidianIRC and uses EXTJWT, supports file deletion, but even still it just feels like a really bad idea to rely on things like that for sending IRC credentials to. We can already use EXTJWT to prove to a pre-approved service that we are who we say we are without any extra password validation going on.

With EXTJWT, the token is either understood or not understood, and if understood then either accepted or rejected. A bad actor from a public filehost could not steal your passwords from an unknown EXTJWT token, and I expect IRC server admins to make mistakes, having the potential to turn a simple mistake into an abuse vector.

whitequark:
> Sending plain credentials implies that the filehost has to exclusively either:
> 
> 1. Be inside a services package
> 2. Share/piggyback databases with the services package
> 3. Validate passwords using RPC or some admin method via the services package

There's a fourth option that I've implemented in [ircv3-filehost-server](https://codeberg.org/whitequark/ircv3-filehost-server): it SASL authenticates using the provided credentials using the IRCd's public endpoint. If the connection succeeds then the upload is authorized. (This is not without its own issues.)

---

Personally, from a "principle of least authority", I would have preferred if I could ask Ergo whether an (ideally, session-bound bearer) token is valid: it would have heavily reduced the fallout of any compromise & also reduced load on the IRCd. Also, if this is added to the proposal, I'd like to request that there is some standard way to find out which (registered) corresponds to the token, if any.

I do know though that I have neither the desire, nor the expertise, nor the time to implement OIDC/JWT (doubly so if I have to add it to Ergo first) so if the two options are draft/FILEHOST and OIDC then I'll keep using the current system.

skizzerz:
> i think an opaque token (whether generated per-connection or by a command) is the easiest and most flexible way

Agreed. I'd like to see a command that a client can use to fetch such a token, to pass along to the filehost as a Bearer token using the HTTP Authorization header. The exact nature of this token does not need to be specified; an implementation may choose to give it a short expiration, ensure it is not reused, etc.

The filehost backend would need some mechanism of validating this token against the IRCd/services package. I'm not sure this needs to be specified either, since there's a whole bunch of different ways this could work and the "best" way likely differs by how the filehost is implemented (is it part of the ircd/services package? is it a third party site?). Some level of OOB configuration on the filehost level would likely be required here because of this lack of specification, but that seems like an ok tradeoff to me.

whitequark:
I think not specifying this creates pressure to integrate filehost functionality into each IRCd, since maintaining a compatibility matrix of "each IRCd without integrated filehost" × "each filehost solution" (I counted four last time I searched) is going to be costly. Is this intentional?

I don't think it needs to be difficult, either. It would not be too challenging to specify an HTTP endpoint offered by IRCd that returns a 200 or 401 status when provided an `Authorization:` header.

skizzerz:
> I don't think it needs to be difficult, either. It would not be too challenging to specify an HTTP endpoint offered by IRCd that returns a 200 or 401 status when provided an `Authorization:` header.

I could get behind a standardized HTTP endpoint for token validation, so the URI at least is well-known. Or, perhaps a .well-known and/or SRV-based discovery mechanism for said URI since there's no guarantee the ircd itself or the server the ircd is living on is also running an httpd.

However, I'm thinking that if we were to specify this, we'd want the flexibility for the ircd to pass optional claims to the endpoint doing the validation. For filehost, this could be things like the username to associate the upload to, quota, etc.

whitequark:
> I could get behind a standardized HTTP endpoint for token validation, so the URI at least is well-known. Or, perhaps a .well-known and/or SRV-based discovery mechanism for said URI since there's no guarantee the ircd itself or the server the ircd is living on is also running an httpd.

I would be happy with any of these options (or anything else reasonably lightweight).

I was thinking that the ircd could have an HTTP endpoint for this sole purpose, on a service (non-standard) port, with the discovery of this being left to configuration. But I like your suggestion better.

> However, I'm thinking that if we were to specify this, we'd want the flexibility for the ircd to pass optional claims to the endpoint doing the validation. For filehost, this could be things like the username to associate the upload to, quota, etc.

Yes, I'm [fully behind](https://github.com/ircv3/ircv3-specifications/pull/562#issuecomment-4189281826) this:

> Also, if this is added to the proposal, I'd like to request that there is some standard way to find out which (registered) corresponds to the token, if any.

reesericci:
My proposal, using OIDC is about creating short-lived tokens instead of sending credentials over the wire, but with a number of advantages:

- Spec compliance. This is the gold standard way for authenticating users to third-party software, and anything that supports OIDC would immediately be able to use an IRCd as a directory for authentication. For instance, this would allow for off the shelf filehosts that support OIDC and tus.io to work with IRC daemons out of the box.
- Informed consent. Tight scoping around what identity data is being accessed and surfacing that to the user before that data sharing is crucial for ensuring privacy for users. (Does the service want just a confirmation that there is an IRC user logged in, do they want their nick, real name, etc)
- Ease-of-use. All the user should see is a popup after a service requests to authenticate via IRC to approve the login, and the client should handle the exchange of tokens which simply identify it to the service. If anything, it may be possible to entirely remove the initial token exchange and replace it with a nick or cloaked IP instead to identify the client to ask for authorization from.

This flow isn't complicated, it's simple & secure in a way that brings IRC into the modern authentication/authorization realm instead of remaining bespoke on the sidelines. We're getting scared of the _idea_ of OIDC/OAuth instead of critiquing what the flow actually looks like. Here's the diagram I made to explain EXTOIDC:

```mermaid
sequenceDiagram
	actor Alice
	participant Server
	participant RP@{ "type" : "control" } as Relying Party
    Alice->>Server: /extoidc token
	Server-->>Alice: signed JWT with UUID
   	Alice->>RP: JWT from server
	RP->>Server:Authentication Request with JWT
	Server->>Server: Validate JWT signature
	Server->>Alice:Do you authorize this authentication request?
	Alice->>Server: Who is it? (/extoidc info)
	Server-->>Alice: Info about RP & claim requests
	Alice-->>Server: Authorize authentication (/extoidc authorize)
	Server->>RP:Authentication Response & claim data
	Server->>Server:Invalidate JWT
```

In this case, the relying party is the file hosting.

reesericci:
Additionally, to respond directly to some of @emersion 's points, I think there's a few mischaracterizations there:

> From a technical point-of-view, I don't think the client really needs to care what kind of token it receives? Can't the client treat the token as an opaque string, and leave it up to the server to decide if it needs to be JWT or something else?

JWTs are useful because they can be _signed_ (and encrypted even) which allows for verification that the token was not intercepted or modified in transit, thus they are safe to pass between many different actors (IRCd, client, file hosting, etc)

> A proper OAuth/OIDC server is a whole project to build or get running. OAuth/OIDC adds a whole bunch of interactions clients need to perform. The OAuth/OIDC framework is very complex to understand in the first place (and has a lot of gotcha's). And that's not even getting into the OAuth client registration discussion.

I don't think we need to implement the entire spec, but rather a simple directory and auth flow for IRC, with OAuth client registration/management happening out-of-band is completely manageable. What makes OAuth a pain is dealing with web redirects & client registration. We can knock both of those out here and create a very usable and spec-compliant experience for servers & file hosts.

> I'd really like to make OAuth easier to use with IRC, but I don't believe tying it to file uploads is a good way forward.

It's not "tying it to file uploads" - it's about solving the question of how do we integrate external services with IRCd safely and securely with respect for the user.

> By the way, soju.im/FILEHOST is perfectly usable with OAuth. A server can accept OAUTHBEARER SASL authentication, and the file upload endpoint can accept Bearer HTTP authentication. Then no user password is ever seen nor transmitted by the client.

This is quite misleading as almost no servers have adopted OAUTHBEARER, thus the problem remains for the rest of us.

Celelibi:
I agree that on the IRC side, we don't need much more than an opaque token.
However, the way the token is verified should be standardized IMO. Not standardizing it would most likely lead to heterogeneous incompatible implementations. Even though it feels kind of out of scope for IRCv3 to specify that part.
Also, the filehost server will most likely want to retrieve some data about the user like a username and maybe some authorizations, so this part needs standardization. And if the filehost server is shared between several IRC networks, then it needs to know which IRC server to check the token against. (Although this information could be carried by the domain name TBH.) And what if we want to implement user files listing and deletion? Does the client regenerate a token for each query and confirm their use? Then we need to introduce a reusable token with an expiration date.
And now it feels like we're reinventing OAuth/OIDC... :/

whitequark:
@reesericci

> We're getting scared of the _idea_ of OIDC/OAuth instead of critiquing what the flow actually looks like.

Please don't be presumptious. I started my career in webdev and I've implemented OAuth before (though that was mostly OAuth 1). My criticism of it is not uninformed.

While I could respond to several of your points, I find it sufficient that JWTs are widely regarded by cryptographers as a technology to avoid ([1](https://lobste.rs/s/dkk1ny/things_use_instead_jwt), [2](https://paragonie.com/blog/2017/03/jwt-json-web-tokens-is-bad-standard-that-everyone-should-avoid), [3](https://kev.inburke.com/kevin/things-to-use-instead-of-jwt/)--last one not by a professional cryptographer but still salient). I understand the potential problems with a randomly generated per-call or per-session token; I do not have the expertise to evaluate whether an implementation that includes JWTs is secure, nor do I want to be on-call for bugs for it (and that's the position I would be in if I ship it somewhere). Other people can have different assessments of *how much* of a problem it is, but I don't think the *existence* of this problem can be dismissed.

In other words: I will never wake up to a CVE resulting from insufficient cryptographic validation of a random single-use token, and that's how I want to keep it. If this means foregoing some interoperability, or some abstract standards compliance, I'm fine with it.

(Also, virtually nobody pays attention to the surfaced OIDC scopes. People just want to upload their file and they'll click through almost anything to make that happen.)

> This is quite misleading as almost no servers have adopted OAUTHBEARER, thus the problem remains for the rest of us.

Why has that been the case? Since this is the case, why should we expect servers to adopt an OIDC-based flow more actively?

---

> However, the way the token is verified should be standardized IMO. Not standardizing it would most likely lead to heterogeneous incompatible implementations. Even though it feels kind of out of scope for IRCv3 to specify that part.

I agree!

> Also, the filehost server will most likely want to retrieve some data about the user like a username and maybe some authorizations, so this part needs standardization.

It is necessary to standardize at least passing the (services) user ID along, to allow for attribution of files to registered users (for quota enforcement and deletion), and to make it possible to reject file uploads from unregistered users (which seems so prone to abuse that almost nobody would want it).

I think the other user data is a "nice to have" but I don't see it as a blocker, personally.

> And if the filehost server is shared between several IRC networks, then it needs to know which IRC server to check the token against.

Wouldn't a query parameter in the upload URL suffice for this? Domain name also works as you suggest.

> And what if we want to implement user files listing and deletion?

I think these two are separate concerns.

- For deletion, and if we specify a way to retrieve the services user ID for the opaque token, all you need is to send a DELETE request with the appropriate file URL and `Authorization:` header. This could be integrated into the client: for example, Halloy records some per-message metadata, which could include "this particular link has been filehost-uploaded and can be deleted".
- For file listing, I think this *doesn't* need to be specified. There is a number of reasonable implementations: doing it via a new services user (FileServ?), doing it via an external web interface, leaving it to the end user to physically locate the file, etc. If we specify how the token works, that makes it possible to authenticate with the external web interface, and then it's OK to leave it to the individual filehost implementation.

> Does the client regenerate a token for each query and confirm their use? Then we need to introduce a reusable token with an expiration date.

We can specify that the client must ask the server for a token every time it performs an action (uploading or deleting). The server will then be at liberty of generating and using single-user tokens or per-session tokens. I can't think of any reason for the client to ever *store* a token, since generating them is so cheap and the actions can only be performed while online anyway.

In summary, I think we should avoid overthinking this.

slingamn:
I like @skizzerz 's proposal to add a new command that generates an opaque bearer auth token. (I was fixated on EXTJWT as the mandated mechanism to avoid defining a new command, but a new command that allows tokens other than JWTs is a better tradeoff.)

I don't really like asking the IRCd to expose a token validation endpoint. I don't think we need standardized interoperability between arbitrary ircds and arbitrary filehost implementations; mandating the use of a validation endpoint is also a performance bottleneck for implementations that would prefer to use signed JWTs.

whitequark:
> I like @skizzerz 's proposal to add a new command that generates an opaque bearer auth token. (I was fixated on EXTJWT as the mandated mechanism to avoid defining a new command, but a new command that allows tokens other than JWTs is a better tradeoff.)

I agree!

> I don't really like asking the IRCd to expose a token validation endpoint. I don't think we need standardized interoperability between arbitrary ircds and arbitrary filehost implementations; mandating the use of a validation endpoint is also a performance bottleneck for implementations that would prefer to use signed JWTs.

Are there reasons to avoid standardizing it besides that you prefer JWTs? It does not seem desirable to me to create a pressure to integrate filehost and services implementations.

ArsenArsen:
> > I like @skizzerz 's proposal to add a new command that generates an opaque bearer auth token. (I was fixated on EXTJWT as the mandated mechanism to avoid defining a new command, but a new command that allows tokens other than JWTs is a better tradeoff.)
> 
> I agree!
> 
> > I don't really like asking the IRCd to expose a token validation endpoint. I don't think we need standardized interoperability between arbitrary ircds and arbitrary filehost implementations; mandating the use of a validation endpoint is also a performance bottleneck for implementations that would prefer to use signed JWTs.
> 
> Are there reasons to avoid standardizing it besides that you prefer JWTs? It does not seem desirable to me to create a pressure to integrate filehost and services implementations.

the user will never use that endpoint and it requires the ircd to run an additional http server.  for filehost implementations verifying JWTs it's also useless as they'll never need to perform a request to validate a token.  specifying that endpoint only makes sense if the tokens are made to require verification over that endpoint by the standard for some reason.

whitequark:
It also has the desirable property of avoiding divergence between IRCd implementations (it can be specified as "if implemented at all, the endpoint should look like this"), but I take your point.

emersion:
I've pushed a few changes ([diff](https://github.com/ircv3/ircv3-specifications/compare/6b75ce571273ab3085759ceb1fdcbf8e53fb7224..59b9601e15364986c1085a8faa361d07ac0fbdfa)):

- A new command to retrieve upload tokens.
- A note about resumable uploads.
- Switched `Content-Disposition` to `inline` from `attachment` to follow recommendation from `draft-ietf-httpbis-resumable-upload-11`.

whitequark:
@emersion Do you think `DELETE` (with the same authorization scheme) is a worthwhile addition?

emersion:
When designing the soju extension, we've been thinking about `DELETE` but we've decided to not include it for now because it opens a new can of worms:

- Deleting a message containing a URL to an uploaded file doesn't necessarily imply that the uploaded file can be deleted. For instance, the same URL could've been forwarded or quoted in other messages. It's difficult to keep track of all references to an uploaded file, especially for clients which don't store the full chat history.
- Multi-client support is tricky. If client A uploads a file and includes its URL in a message, another client B has no way to know that this is a file uploaded by the same user and cannot figure out that the upload can be deleted.
- This hasn't been a requested feature from users since we added file upload support. It seems like the value-add is fairly low.

whitequark:
Sorry, that was unintentionally ambiguous. I meant HTTP DELETE support removing the uploaded file / making file URL inaccessible. The messages would not be affected.

The use case is that if you've uploaded something you didn't want you can quickly tell your client to undo it (either before or after sending it).

emersion:
Right, I think a `DELETE` could be useful for this use-case (deleting a file right after uploading it) and nothing much more. I wouldn't mind adding it if we think it's worth it (should it be mandatory? or discoverable via `OPTIONS` or something?).

(On the other hand, if an uploaded file is not posted in a message other users can't access it.)

ArsenArsen:
IMO deleting a file at any point is fine, even if it leaves dead links, it is probably useful anyway just in case the wrong thing gets shared

whitequark:
I think the endpoint should be mandatory (it doesn't seem desirable to make a service which is by design incapable of deleting any uploaded data without administrative action) but the endpoint may return a 401 status (e.g. if the upload was done with a different services account, or without authorization at all). It amounts to the same case coverage with less sprawling decision tree.

0x5c:
> it doesn't seem desirable to make a service which is by design incapable of deleting any uploaded data without administrative action

This seems non-applicable to cases where all uploads are ephemeral, like when using [convoyeur](https://codeberg.org/classabbyamp/convoyeur) combined with a temporary file hoster like 0x0.st

classabbyamp:
though convoyeur can be combined with a personal pastebin like rustypsste that does have deletion capability, so I might build optional support for it.

but in the spec DELETE shouldn't be mandated supported

whitequark:
Hm, I guess the benefit of a discovery process for `DELETE` is that for filehosts without deletion support the action would not be surfaced in the UI at all? Come to think of it, being warned upfront that uploaded files cannot be deleted *is* valuable.

(If this is never surfaced in the UI upfront then there's no point in having a discovery mechanism over simply trying to `DELETE` and getting an answer. Since deletion might fail even if the method is supported—if the token is wrong for example—this failure has to be handled anyways.)

What should be the mechanism for it? `Allow: DELETE` in `OPTIONS` doesn't seem right since the `DELETE` would not be issued to the upload endpoint URL but rather the file URL. `Allow-Delete: true`?

Celelibi:
What URL should the `DELETE` method be called to?
It cannot be the shared URL because it might point to S3 or a CDN or something.

I think it might not be too complicated to define an HTTP URL like `$filehost/filelist` that returns a file containing some information about the files uploaded by the user (URL, size, data of upload, uploader, ...) and the delete URL.
(I'd prefer a line-based text file with space-separated fields, but I know JSON is appreciated. ^^)

slingamn:
I agree that the deletion mechanism should be a POST taking an arbitrary URL, for the reasons @Celelibi mentioned: the URL returned from the upload endpoint is the canonical dentifier of the uploaded file, but it may be on a completely different host.

Here's a take that may be more controversial: I think the URL advertised in 005 should be an extensible JSON manifest, not an endpoint in itself. Some reasons:

1. The manifest could advertise the state of various optional features (is auth required to upload? to view? is deletion available? is enumeration?)
2. The manifest could publish various pieces of useful metadata. File size limits and user quotas are obvious candidates. Another thing I've been thinking about is advertising a URL prefix that's endorsed by the ircd for autopreviewing (i.e. the ircd tells the client "anything at https://example.com/files/ is safe to autopreview and won't leak your IP address to people who can't already see it from the IRC connection")
3. The manifest could list multiple endpoints instead of us relying on a common naming scheme

furudean:
It does not seem clear to me how an external service may validate that the token is valid after reading the current version of the spec.

emersion:
This is an implementation detail of the server and external service. This spec focuses on the client-to-server interactions only.

thatcher-gaming:
I do think a better idea would be to have the client first send a GET to the upload server, receive a URL back (the URL being another location on the upload server, to be clear), and then POST the file to that URL. Then, when the file is uploaded, that URL becomes a permanent redirect to wherever the hell the file actually is. This way you can have proper DELETE support, and you can also solve another issue I have with the spec, which is that you have to wait until the upload is finished before you can actually send the message. If you have comically slow and unreliable internet, it's not uncommon for the conversation to have moved on in the 30 or so seconds it takes to send an image. If you already know the URL where the image will be located once the upload has finished, you can just include that in your message and everyone else can just see a loading spinner and know that a file is on the way. Which I reckon is probably a better experience all told. I don't know of any other service that does it like this, so there's every chance in the world that this is actually a bad idea for reasons I haven't thought about.

whitequark:
That sounds reasonable to me, it's more or less a builtin link shortener.

Celelibi:
Oh yeah! Now a good reason to restart the PUT debate. xD
Here are a few drawbacks of sending a message referencing a image not completely uploaded yet:
- the file upload may never complete (client crashes or abandon the transfer after too many network errors)
- the server may reject the file based on its content (so, rejection is notified very late)
- receiving clients would have to handle gracefully the period of upload
- not necessarily a good experience for receiving users that have a good connection : expecting all the images to load instantly, but their load time actually depend on the connection of the sending user. They'll see a spinner for a few minutes, not knowing if it's their own connection that's the issue or not.
Nothing that's a real dealbreaker, IMO. At least it gives clients the choice of either sending the message immediately or to schedule it for when the upload succeeded.
But I'm wondering still if the benefits are worth it.

thatcher-gaming:
> * the file upload may never complete (client crashes or abandon the transfer after too many network errors)

Assuming you would be able to try and POST to the URL again if the upload was interrupted previously, I don't think this would be a problem. The client can just keep its uploads in a persistent queue that can be restored in the event of a crash.

> * the server may reject the file based on its content (so, rejection is notified very late)

If it's being rejected based on the media type etc, you just need to make sure your client sets Content-Type correctly. If you're scanning images with PhotoDNA or whatever, I can see that being a problem, but I honestly don't think files being rejected on that basis would be all that common of an occurrence in real life usage. Most current chat apps will send you a notification if an upload times out or otherwise fails (iMessage example below), so that could be something developers should be encouraged to implement.
IMG_2501

> * receiving clients would have to handle gracefully the period of upload

I guess this would require some testing, but I'm not aware of any HTTP clients in use today that would struggle with something like this (although you might need to configure the timeout interval).

> * not necessarily a good experience for receiving users that have a good connection : expecting all the images to load instantly, but their load time actually depend on the connection of the sending user. They'll see a spinner for a few minutes, not knowing if it's their own connection that's the issue or not.

_Theoretically_ you could have some kind of mechanism for relaying the upload progress to recipients. I have no idea how that would work. Suggestions welcome!!

>   Nothing that's a real dealbreaker, IMO. At least it gives clients the choice of either sending the message immediately or to schedule it for when the upload succeeded.
>   But I'm wondering still if the benefits are worth it.

I think so! In my view, it's crucial that the text content of the message is sent as quickly as possible, and isn't delayed by auxiliary content that may or may not be needed to understand the message.

slingamn:
@emersion are you open to reverting this PR back to 6b75ce571273ab3085759ceb1fdcbf8e53fb7224? I think that version has multiple implementations. I think 59b9601e15364986c1085a8faa361d07ac0fbdfa will probably end up being superseded by something that uses the generalized `draft/authtoken` mechanism (#602).

0x5c:
> I think so! In my view, it's crucial that the text content of the message is sent as quickly as possible, and isn't delayed by auxiliary content that may or may not be needed to understand the message.

On all messaging platforms I've used, uploads either happen immediately once selected (preventing sending before completion or failure) or as part of sending the message.

In both cases, a file upload failure will not cause a message to be sent without its attachments.

bb010g:
> In both cases, a file upload failure will not cause a message to be sent without its attachments.

A client could choose to not send a message until the completion of the upload POST.

emersion:
Following @slingamn's [recommendation](https://github.com/ircv3/ircv3-specifications/pull/562#issuecomment-4898779321), I've [reverted](https://github.com/ircv3/ircv3-specifications/compare/138e398c6e89940c3dc94be47e7aa2ba59041518..9c583eaf143d38467df469da93d4297958bef804) this PR to the earlier proposal. I've added a stricter and more explicit paragraph about authentication.

thatcher-gaming:
Also: it would be good to have some way of knowing (perhaps another ISUPPORT token) that the fileupload server requires authentication to upload files (so e.g., we can grey out the upload button when logged out)

skizzerz:
> Also: it would be good to have some way of knowing (perhaps another ISUPPORT token) that the fileupload server requires authentication to upload files (so e.g., we can grey out the upload button when logged out)

Send a HEAD request to the identified filehost URL and check if you get a 401 Unauthorized?

thatcher-gaming:
As far as I know HEAD is supposed to send the headers of a GET request, not a POST, so not terribly useful here.

PR #563, opened 2025-05-15
Add network icon ISUPPORT token

emersion:
Closes: https://github.com/ircv3/ircv3-ideas/issues/123

### DIFF
--- a/extensions/network-icon.md
+++ b/extensions/network-icon.md
@@ -1,0 +1,28 @@
+---
+title: "Network Icon"
+layout: spec
+work-in-progress: true
+copyrights:
+  -
+    name: "Simon Ser"
+    period: "2025"
+    email: "contact@emersion.fr"
+---
+
+## Notes for implementing work-in-progress version
+
+This is a work-in-progress specification.
+
+## Description
+
+The `NETWORK` ISUPPORT token allows servers to advertise their network name. This specification adds a new ISUPPORT token to advertise a network icon.
+
+## The `draft/ICON` ISUPPORT token
+
+If a server publishes a `draft/ICON` ISUPPORT token, the value MUST be a URL to an image. The URL SHOULD use the "https" scheme.
+
+Servers MAY implement [extended-isupport](extended-isupport.html) to allow clients to fetch the network icon before connection registration.
+
+## Example
+
+    S: :irc.example.org 005 * NETWORK=Example draft/ICON=https://example.org/icon.svg :are supported by this server


delthas (approved):

extensions/network-icon.md
```diff
@@ -0,0 +1,28 @@
...
+
+This is a work-in-progress specification.
+
+## Description
+
+The `NETWORK` ISUPPORT token allows servers to advertise their network name. This specification adds a new ISUPPORT token to advertise a network icon.
+
+## The `draft/NETWORKICON` ISUPPORT token
+
+If a server publishes a `draft/NETWORKICON` ISUPPORT token, the value MUST be a URL to an image. The URL SHOULD use the "https" scheme.
```
nektro, line 22: should there be a way for the client to say what mime types it accepts for the image to be in?

emersion: MIME types are one thing, image size in pixels is another one… For the latter negotiation wouldn't work, the client might want to show the icon in different places, with a different size for each one.

The Web solves this by listing all possible icons, with all formats and sizes. I've decided against this approach for this spec because:

- It adds a whole bunch of extra complexity.
- It's a lot of characters. URLs might be quite long.

If we really wanted to solve this, maybe advertising the URL of a JSON manifest would be better (like [Web app manifests](https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps/Manifest)). But I don't think we actually need this complexity.

grimmy: Something like `.ico` that can pack multiple sizes could work in this regard.

emersion: `.ico` is a very old format, I don't think we should build on it. It's missing a lot of features one would expect from modern image formats.

lollipopman: one option, to move the complexity elsewhere, would be to only support svg icons.

emersion: My expectation is that server authors will use one of the widely supported image formats (e.g. PNG or SVG).

If format negotiation is really deemed necessary by a network, HTTP supports this feature natively already via [Accept](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Accept).

nektro (approved):

ValwareIRC:
I think we should go with `ICON`, and here's why.

I absolutely agree that the icon would and in 80-90% of cases should use the same icon all over their network, however I feel like this would disallow theming of certain servers icons.

For example, if I have `fire.*`, `water.*` and `ice.*`, I might want to theme each icon with a background of either fire, water or ice, but with a clear print of the network name on top of the logo.

I do think it would be weird if people used completely different icons for each of their servers and I don't imagine anyone would do that anyway, but I still think this should be allowed, and, although it's technically not forbidden by this here spec proposal to use different icons on each server, the name suggests it would be and could be confusing to client developers.

I think outright forbidding different icons for different servers, either by suggestive language or by spec, limits things unnecessarily and I vote to change the token name to `ICON` (without a `draft/` prefix since the IRC specification forbids any characters that are not specifically A-Z)

jwheare:
Regardless of any future decision to allow draft/ in isupport, that decision has not happened yet nor even been formally proposed. So let’s continue to submit specs according to the status quo rather than desired but undecided futures.

jwheare:
I also think we go with ICON and not preempt a need to specify whether it has to be the same network wide or per server. That can be up to network operators (this includes server operators, they are necessarily collaborators)

extensions/network-icon.md
Celelibi: I don't think this spec should leave the format entirely open-ended. If the spec allows any format, then are clients also expected to support each and every single format that could be rasterized to an image? SVG, PDF, PS, Xfig?
If we go for only HTTP icon location (which I'm not a big fan of, but anyway), then the `Accept` header should probably be mentioned explicitly so that clients and servers are advised to use it. And the server be required to provide **at least** one of (for instance) ICO, BMP, PNG. So that lightweight clients could support only these 3 formats and be assured to get the network icon, if any.

Also, FWIW, ICO format does support PNG as individual images and support alpha channel for both BMP and PNG individual images.

SadieCat: +1 for using the Accept header. The server can respond with 406 if its not available in that form.

SadieCat:
In one of the other proposals I seem to remember there was a {size} template variable which the client could use to scale the icon. Is this useful to client developers?

delthas:
> In one of the other proposals I seem to remember there was a {size} template variable which the client could use to scale the icon. Is this useful to client developers?

No, images are small anyway and it is trivial to resize them client side.

thatcher-gaming:
unsolicited +1 for the {size} template variable, if not for reduced data usage then for optical sizing adjustments in particularly complex logos. i need to adjust my network's coat of arms for use at 32x32

emersion:
Renamed to `ICON`.

emersion:
Opened https://github.com/ircdocs/modern-irc/issues/250 to discuss prefixes for ISUPPORT tokens.

slingamn:
I still see `draft/ICON` in d2daee5 (also I would personally recommend against force-pushing to spec PRs).

emersion:
Yes, I have intentionally left the `draft/` prefix for now.

emersion:
Prefixes are now explicitly allowed in the modern spec, so `draft/` can stay.

SadieCat:
InspIRCd's server implementation is available on the testnet and at https://github.com/SadieCat/inspircd/tree/insp4%2Bircv3-icon

@ValwareIRC and @delthas have client implementations (afaik).

jwheare:
IRCCloud supports this now too:

image

PR #565, opened 2025-06-15
clarify CHATHISTORY TARGETS semantics

slingamn:
Clarify CHATHISTORY TARGETS semantics as follows:

>The parameters have the same semantics as `BETWEEN`, except that they MUST be timestamps and not msgids, and they match against the latest message stored for the target (i.e. a target is not returned if its latest message is outside of the selection parameters, even if it has other messages inside them).

This appears to be a source of confusion, which is understandable because the previous text was ambiguous. However, the emendation here was always the intent (otherwise the selectors could not be used effectively for pagination).


slingamn:
Bump

PR #566, opened 2025-06-20
updates to metadata spec

slingamn:
These came out of my review of the initial metadata implementation in Ergo, contributed by @thatcher-gaming. They're more starting points for discussion than polished amendments. A brief summary:

* `before-connect` allows clients to modify both their metadata and their subscriptions during connection registration. For conventional IRC stacks without persistent clients, this makes perfect sense. For bouncers and implementations like Ergo with persistent clients, it makes sense to be able to modify subscriptions, but clients probably shouldn't be unilaterally pushing metadata updates on every reconnection, since they may overwrite more up-to-date metadata that was set from another client. This adds a new cap value key, `sub-before-connect`, that indicates that subscriptions are accepted during connection registration, but `SET` etc. are not.
* Adds a batch parameter: the name of the primary target of the metadata request. Without this, the empty batch is uninterpretable (at least without labeled-response).
* Removes the edge-triggered requirement to send the client's own metadata when the cap is requested
* As a replacement for this, specifies that the server sends metadata-negotiating clients a batch of their own metadata during the reg burst

Something not addressed here is that `772 RPL_METADATASUBS` is effectively unusable without `labeled-response`, since there is no indication of where it ends. An initial proposal: a `772` line with no parameters after the client's nick could be taken as the end of the subscriptions? We could also add a new `RPL_ENDOFMETADATASUBS`.

### DIFF
--- a/extensions/metadata.md
+++ b/extensions/metadata.md
@@ -724,7 +724,7 @@ Twice:
     S: :metadata.test 001 abc :Welcome to the MetadataTesting IRC Network abc
     S: [redacted registration burst messages]
     S: :metadata.test BATCH +1 metadata
-    S: @batch=1 :metadata.test 761 abc abc display-name * :a b c
+    S: @batch=1 :metadata.test METADATA abc display-name * :a b c
     S: :metadata.test BATCH -1
     S: [redacted registration burst messages]
     S: :metadata.test 376 abc :End of MOTD command
--- a/extensions/metadata.md
+++ b/extensions/metadata.md
@@ -88,6 +88,7 @@ The ABNF format of the `metadata` capability is:
 These are the defined tokens:
 
 * `before-connect`: if present, indicates the server supports `METADATA` commands during connection registration
+* `sub-before-connect`: if present, indicates the server supports `METADATA` commands related to subscriptions (`SUB`, `UNSUB`, and `SUBS`) during connection registration
 * `max-subs`: the maximum number of keys a client is allowed in its subscription list. See the [`SUB`](#metadata-sub) subcommand for more details.
 * `max-keys`: the maximum number of keys a client is allowed to set on its own nickname.
 * `max-value-bytes`: the maximum size of values a client is allowed to set. Servers MAY send longer values.
--- a/extensions/metadata.md
+++ b/extensions/metadata.md
@@ -733,7 +733,9 @@ Twice:
     S: @batch=2 :metadata.test 761 abc abc display-name * :a b c
     S: :metadata.test BATCH -2
     C: METADATA * SUBS
-    S: :metadata.test 772 abc display-name
+    S: :metadata.test BATCH +3 metadata-subs
+    S: @batch=3 :metadata.test 772 abc display-name
+    S: :metadata.test BATCH -3
 
 -----
 
--- a/extensions/metadata.md
+++ b/extensions/metadata.md
@@ -82,7 +82,7 @@ The ABNF format of the `metadata` capability is:
     capability ::= 'metadata' ['=' tokens]
     tokens     ::= token [',' token]*
     token      ::= key ['=' value]
-    key        ::= <sequence of a-zA-Z0-9_./:->
+    key        ::= <sequence of one or more a-z0-9_./->
     value      ::= <utf8>
 
 These are the defined tokens:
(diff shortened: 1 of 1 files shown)


slingamn:
>Re: sub-before-connect can't this just be a client setting?

If I understand the question correctly: the point of having a separate server cap value here is to inform clients that the server may be storing persistent metadata that they shouldn't preemptively overwrite. Making this a client setting would probably require the end user to be aware of this, which seems undesirable. (I'm also open to making this a value of `before-connect`, e.g. `draft/metadata=before-connect=subs,max-subs=50`, where no value or `*` is the original behavior.)

re. the CAP REQ behavior, the original language is:

>Upon requesting the `metadata` capability, clients receive their non-transient metadata (for example, metadata stored by the server or by services) in a `metadata` batch with their own nick as target. If none exists, the server MUST send an empty batch instead.

As per discussion on #543 I prefer that requesting a CAP should not have side effects. This is amended to:

>If the `metadata` capability was negotiated during connection registration, clients receive their current metadata (any metadata stored by the server or by services, plus any metadata set by the client during connection registration via `before-connect`) in a `metadata` batch with their own nick as target as part of the registration burst, i.e. before `RPL_ENDOFMOTD` or `ERR_NOMOTD`. If none exists, the server MUST send an empty batch instead.

The concrete changes in behavior being:

* There is no direct side effect of requesting the CAP
* Better integration with persistent clients and bouncers; by the time connection registration is complete, the server should know what persistent metadata is set. (In contrast, if the CAP is requested during connection registration but before SASL, the server will be unaware of the state of persistent metadata. Even after SASL, this will probably be implementation-dependent in some cases, e.g. if the same account can be associated with multiple nicknames / presences.)
* The batch is sent during the registration burst

jwheare:
What would happen if a client did try to overwrite the metadata? They may want to e.g. set a key that indicates where they're logging in from.

I don't follow how the CAP REQ change prevents side effects, it still says "clients receive their current metadata"

slingamn:
>I don't follow how the CAP REQ change prevents side effects, it still says "clients receive their current metadata"

They don't receive it in response to the `CAP REQ` command itself, they receive it in the registration burst. The philosophy here is (among other considerations) to avoid ordering dependencies (like the one with SASL mentioned above) during registration --- in contrast, the decision to complete registration (typically by sending CAP END) is deliberate, explicit, and occurs after the client has finished sending all relevant data.

>What would happen if a client did try to overwrite the metadata? They may want to e.g. set a key that indicates where they're logging in from.

I think for implementations with persistent presence and metadata support, and which are designed to accept short-lived connections, rejecting pre-connect updates is probably a good tradeoff since they are likely to be spammy as well as out-of-date. But the tradeoff is real, and not just for keys like the one you're describing, because non-persistent clients will have to connect without being able to pre-set any metadata.

jwheare:
Ah, in response to CAP END vs CAP REQ makes sense, that wasn't really clear from the wording.

> I think for implementations with persistent presence and metadata support, and which are designed to accept short-lived connections, rejecting pre-connect updates is probably a good tradeoff since they are likely to be spammy as well as out-of-date. But the tradeoff is real, and not just for keys like the one you're describing, because non-persistent clients will have to connect without being able to pre-set any metadata.

I feel like that tradeoff is exactly why it should be a client setting, should be in the user's control.

slingamn:
Withdrawing the proposal of `sub-before-connect` for now; interested in discussion on the other points (including adding an indication of termination for `772 RPL_METADATASUBS` --- another possibility I didn't bring up is to always batch it, similar to `761 RPL_KEYVALUE`).

slingamn:
Currently, key names are "restricted to the ranges A-Z, a-z, 0-9, and _./- and are case-insensitive." Case-insensitivity creates a number of potential correctness issues (for both clients and servers), seemingly to no real benefit. Could we just require that key names are lowercase instead, i.e. remove A-Z from the allowed characters?

extensions/metadata.md
```diff
@@ -96,23 +96,25 @@ Clients MUST silently ignore any unknown tokens.
...
-This specification adds the `metadata` batch type.
+The `metadata` batch type MUST be sent to clients under the following circumstances:
 
-This batch MUST be sent to clients on connection and as reply to successful `METADATA GET`, `METADATA LIST`, and `METADATA SYNC` subcommands.
+* To enclose 0 or more `761 RPL_KEYVALUE` responses to `METADATA GET` or `METADATA LIST`
+* To enclose 0 or more `METADATA` commands in response to `METADATA SYNC`
+* To enclose 0 or more `METADATA` commands in response to a metadata-capable client completing connection registration and receiving its own metadata
 
-This batch type does not take any parameter, and clients MUST ignore them if any.
+It takes one parameter and clients MUST ignore extra parameters. This parameter contains the target for which the metadata was requested, allowing the client to correctly process empty batches. Since a batch sent in response to `METADATA SYNC` on a channel may contain metadata on users as well as the channel itself, clients MUST NOT assume that all metadata in the batch applies to the entity referenced in the batch parameter.
```
progval, line 115: > Since a batch sent in response to `METADATA SYNC` on a channel may contain metadata on users as well as the channel itself, clients MUST NOT assume that all metadata in the batch applies to the entity referenced in the batch parameter.

Why? What if clients don't want them (eg. lazy-loading only the channel currently visible in the UI)?

progval (approved):

PR #567, opened 2025-06-21
Disallow : as the first character of metadata keys

progval:
It can't be serialized as a non-trailing parameter, eg. in 'METADATA <Target> <Key> <Visibility> <Value>'

Found by @slingamn


emersion (approved):

jwheare:
Do we maybe want to disallow `:` from keys entirely for simplicity's sake?

progval:
sure, why not

slingamn (approved):
Credit for finding this goes to @thatcher-gaming

PR #568, opened 2025-06-23
Add `+draft/unreact` tag to reaction specification

4e554c4c:
This edit is in response to my client implementations—especially [halloy](https://halloy.chat) which does not yet support message redaction—and much reaction testing. In the draft phase, most implementations have settled on reactions being nick-based, but message redaction is account based. This creates an asymmetry between creating and removing reactions, which most users consider to be equivalent features.

I think it's OK for authorized users to remove reactions through a `REDACT`, but this should be performed for moderation reasons. Meanwhile average users may freely use `+draft/react` and `+draft/unreact` to change their reactions.

This specification is implemented in squidowl/halloy#1039.

### DIFF
--- a/client-tags/react.md
+++ b/client-tags/react.md
@@ -14,19 +14,25 @@ copyrights:
 This is a work-in-progress specification.
 
 Software implementing this work-in-progress specification MUST NOT use the
-unprefixed `+react` tag name. Instead, implementations SHOULD use the
-`+draft/react` tag name to be interoperable with other software
+unprefixed `+react` or `+unreact` tag names. Instead, implementations SHOULD use the
+`+draft/react` and `+draft/unreact` tag names to be interoperable with other software
 implementing a compatible work-in-progress version.
 
-The final version of the specification will use an unprefixed tag name.
+The final version of the specification will use unprefixed tag names.
 
 ## Introduction
 
-This specification defines a client-only message tag to indicate reactions to other messages
+This specification defines a client-only message tag to indicate reactions and unreactions to other messages
 
 ## Motivation
 
 This tag provides a means of communicating with context-sensitive, potentially non-textual reactions. It allows chat participants to respond to each other with text, symbols, emoticons, or emoji that don't necessarily appear as full messages, but instead as lightweight annotations, displayed adjacent to a parent message.
+
+While most IRC users do not expect messages to be removable, reactions are meant to be sent quickly, changed based on opinion and removed if sent on accident. For this reason, most implementations of emoji reactions allow an unreaction feature.
+
+One may wonder why an unreaction tag is needed when the [`draft/message-redaction`](../extensions/message-redaction.md.html) also allows for message removal. However, these two methods of message removal serve different purposes. Redaction is a server-moderated process to hide previous client actions, e.g., such that they do not show up in history. This capability is usually only given to users authorized to hide the action, such as a logged-in author or operators. 
+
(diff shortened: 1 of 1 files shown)


jwheare:
This has also been supported on the receiving end by IRCCloud for about 8 years.

emersion:
Thanks for expanding on the motivations, @skizzerz. There is overwhelming support for this approach, so feel free to move forward.

hhirtz (commented):
Two comments mostly for server implementation considerations that might not be obvious at first glance:

If the server filters reactions or unreactions (e.g. moderation, or unneeded/spurious unreact messages where the unreact value doesn't match any reaction from the user), how does this work with `echo-message`? Should the server drop the whole message (then you need `labeled-response` to get the ACK), or only the tags?

Second, with regards to chathistory, when the client sends an unreact message, the server must keep it in history, otherwise some clients might not remove the reaction when they reconnect later.

Otherwise the spec looks good to me.

4e554c4c:
> Currently the chathistory spec doesn't force servers to store anything more than PRIVMSG, anything else is a SHOULD not a MUST (although it would be nice to be able to rely on it for such things)

If you carefully read the TAGMSG spec, it says that TAGMSGs should be treated as PRIVMSGs (supposedly this includes in chathistory), so I consider this to be a chathistory omission.

The chathistory issue is real, but again I think this burden should fall on the clients. Of course, it depends on how clients store their data, but I know many clients keep a history of of the last fetched date to now, in which case they would never fetch a reaction whose unreaction has not been processed. In this case a client simply needs to keep an integer `reaction_count` for each reaction, and not display it if it is less than or equal to zero.

client-tags/react.md
```diff
@@ -14,20 +14,26 @@ copyrights:
...
 ## Introduction
 
-This specification defines a client-only message tag to indicate reactions to other messages
+This specification defines a client-only message tag to indicate reactions and unreactions to other messages
 
 ## Motivation
 
 This tag provides a means of communicating with context-sensitive, potentially non-textual reactions. It allows chat participants to respond to each other with text, symbols, emoticons, or emoji that don't necessarily appear as full messages, but instead as lightweight annotations, displayed adjacent to a parent message.
 
+While most IRC users do not expect messages to be removable, reactions are meant to be sent quickly, changed based on opinion and removed if sent on accident. For this reason, most implementations of emoji reactions allow an unreaction feature.
```
jwheare:
```suggestion
While most IRC users do not expect messages to be removable, reactions are meant to be sent quickly, changed based on opinion and removed if sent accidentally. For this reason, it's common for implementations of emoji reactions in other protocols to allow an unreaction feature.
```

client-tags/react.md
```diff
@@ -14,20 +14,26 @@ copyrights:
...
+
+One may wonder why an unreaction tag is needed when the [`draft/message-redaction`](../extensions/message-redaction.md.html) also allows for message removal. However, these two methods of message removal serve different purposes. Redaction is a server-moderated process to hide previous client actions, e.g., such that they do not show up in history. This capability is usually only given to users authorized to hide the action, such as a logged-in author or operators. 
+
+On the contrary, unreaction is its own action designed to change a previous choice (e.g., in response to a “voting” message), or remedy a misclick. If sent to clients as a text message, previous reactions need not be removed on an unreact message.
+
```
jwheare: I would just drop this, it's a bit confusing.

4e554c4c:
rebased on  `master`

jwheare:
LGTM. Will aim to merge next week.

PR #569, opened 2025-06-30
Add user query extension

emersion:
cc @slingamn

### DIFF
--- a/extensions/user-query.md
+++ b/extensions/user-query.md
@@ -1,0 +1,116 @@
+---
+title: User query
+layout: spec
+work-in-progress: true
+copyrights:
+  -
+    name: "Simon Ser"
+    period: "2025"
+    email: "contact@emersion.fr"
+---
+
+## Notes for implementing work-in-progress version
+
+This is a work-in-progress specification.
+
+Software implementing this work-in-progress specification MUST NOT use the unprefixed `user-query` CAP name. Instead, implementations SHOULD use the `draft/user-query` CAP name to be interoperable with other software implementing a compatible work-in-progress version. The final version of the specification will use unprefixed CAP names.
+
+## Description
+
+This document describes the format of the `user-query` extension. This extension enables several clients of the same user connected to a server or bouncer to tell each other about which user queries (conversations between two users) are currently opened.
+
+Some servers or bouncers remember the channels joined by a user and restore them after connection registration. This can be used by clients to synchronize channels, but doesn't work for user queries. If another client opens or closes a user query, there is no way for other clients to become aware of this action. This extension fills this gap.
+
+The user queries opened by a user MUST NOT be disclosed to other users by the server without an explicit opt-in.
+
+## Implementation
+
+The `user-query` extension uses the `draft/user-query` capability, introduces a new `USERQUERY` command and a `draft/user-query-list` batch type.
+
+The `draft/user-query` capability MUST be negotiated before using the `USERQUERY` command, and affects which messages are sent by the server as specified below.
+
+### `USERQUERY LIST` sub-command
+
+The `USERQUERY LIST` sub-command can be sent by clients to request a list of all opened user queries.
+
+This sub-command has the following client syntax:
+
+    USERQUERY LIST
+
+The server MUST reply to a successful `USERQUERY LIST` sub-command with a `draft/user-query-list` batch if the `batch` capability has been negotiated. The batch MUST contain any number of `USERQUERY LIST` replies with the following server syntax:
+
+    USERQUERY LIST <nick>
+
+The `nick` parameter specifies a single nickname.
+
+### `USERQUERY OPEN` sub-command
+
+The `USERQUERY OPEN` sub-command can be sent by both clients and servers.
+
+This sub-command has the following general syntax:
+
+    USERQUERY OPEN <nick>
+
+The `nick` parameter specifies a single nickname.
+
+When sent from a client, this sub-command signals to the server that the user has opened a query with the specified user. The server MUST reply to a successful `USERQUERY OPEN` client set command using a `USERQUERY OPEN` server sub-command, or using an error message.
+
+When sent from a server, the `USERQUERY OPEN` command signals to the client that the user has a query opened with the specified user. If a `PRIVMSG` is sent to the user while no user query is opened, the server automatically opens a user query before dispatching the message. Servers SHOULD broadcast the `USERQUERY OPEN` notification to all clients connected under the same account.
+
+TODO: interactions with MONITOR?
+
+### `USERQUERY CLOSE` sub-command
+
+The `USERQUERY CLOSE` sub-command can be sent by both clients and servers.
+
+This sub-command has the following general syntax:
+
+    USERQUERY CLOSE <nick>
+
+This command has the same syntax and behavior as `USERQUERY OPEN`, but indicates that the user has closed a user query.
+
+### Errors and Warnings
+
+Errors are returned using the standard replies syntax.
+
+If the server receives a syntactically invalid `USERQUERY` command, e.g., an unknown subcommand, missing parameters, excess parameters, parameters that cannot be parsed, or invalid nickname, the `INVALID_PARAMS` error code SHOULD be returned:
+
+    FAIL USERQUERY <subcommand> INVALID_PARAMS [invalid_parameters] :Invalid parameters
+
+If `USERQUERY OPEN` was sent while the user query was already opened, the `ALREADY_OPENED` error code MUST be returned. If `USERQUERY CLOSE` was sent while the user query was already closed, the `ALREADY_CLOSED` error code MUST be returned.
+
+    FAIL USERQUERY OPEN ALREADY_OPENED <nick> :This user query was already opened
+    FAIL USERQUERY CLOSE ALREADY_CLOSED <nick> :This user query was already closed
+
+If the user query cannot be opened or closed due to an error, the `INTERNAL_ERROR` error code SHOULD be returned.
+
+    FAIL USERQUERY <subcommand> INTERNAL_ERROR <nick> :The user query could not be opened or closed
+
+### Implementation considerations
+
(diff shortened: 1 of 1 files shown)


extensions/user-query.md
```diff
@@ -0,0 +1,99 @@
...
+
+    FAIL QUERYUSER INTERNAL_ERROR <nick> :The user query could not be opened
+
+### Examples
+
+Sending the list of opened user queries after connection registration:
+
+    S: 422 yannick :No message of the day
+    S: yannick JOIN #ircv3
+    S: yannick QUERYUSER gustave
```
slingamn, line 84: What's the `yannick` in these examples? It seems syntactically incorrect

slingamn (commented):
Thanks, this looks good, I'll look in more detail later.

One thing that concerns me is that we need some provision for the server autoclosing / autoexpiring these, otherwise the list of opened buffers can grow without bound in the case where the user is not using a client that supports this cap. Maybe a non-normative implementation considerations section?

extensions/user-query.md
emersion: Ah, it's the nickname without the host+user. AFAIK this is syntactically valid (these are optional in the message source) but uncommon, so I'll add them.

Oh, I realized I missed the colon as well 🤦

extensions/user-query.md
```diff
@@ -0,0 +1,99 @@
...
+
+This command has the following general syntax:
+
+    QUERYUSER <nick>
+
+The `nick` parameter specifies a single nickname.
+
+When sent from a client, this command signals to the server that the user has opened a query with the specified user. The server MUST reply to a successful `QUERYUSER` client set command using a `QUERYUSER` server command, or using an error message.
+
+When sent from a server, the `QUERYUSER` command signals to the client that the user has a query opened with the specified user. This command MAY be sent by the server after connection registration, alongside the `JOIN` burst. If a `PRIVMSG` is sent to the user while no user query is opened, the server automatically opens a user query before dispatching the message.
```
progval, line 44: > If a `PRIVMSG` is sent to the user while no user query is opened, the server automatically opens a user query before dispatching the message.

I don't think this is desirable in all cases. Users running `/msg ChanServ OP #foobar` don't expect a query to open automatically, and especially not on all their clients.

progval: as a reply, but a `PRIVMSG` is sent (by the sending user) to the NickServ user.

This paragraph says "user" for both the recipient and the sender, so it's not clear which one is meant here.

slingamn: I understood it to refer to the recipient but yeah, some clarification would be useful.

For anything other than *Serv it would probably make sense to open the synced query window on send. Maybe it should be a SHOULD and implementations can decide when to do it (e.g. most implementations would want to exempt *Serv)?

slingamn (commented):
Some bikesheddy stuff you can take or leave:

* How about one `QUERYUSER` command with subcommands `+` (add), `-` (remove), and `L` (list), similar to `MONITOR`?
* Any thoughts on batching `QUERYUSER` from the server (particularly if there were an `L` command for listing?)

I still think this needs an implementation considerations section on limiting data growth.

extensions/user-query.md
emersion: Yeah I definitely meant that a query is opened when the connected user receives a PRIVMSG. See the last example. I'm not super sure how to phrase it though, the server performs the action to open a query, and from the server PoV the recipient doesn't _receive_ the message, the server merely forwards it if a client is connected.

I don't think opening a query when the current user sends a PRIVMSG is desirable, let alone *Serv special-casing. Clients can just QUERYUSER/UNQUERYUSER when the user opens a query with the client UI or `/query` + `/close`. Clients will not QUERYUSER on `/msg`, as such no user query will be opened on the server side as expected.

emersion:
> How about one QUERYUSER command with subcommands + (add), - (remove), and L (list), similar to MONITOR?

I'm not in love with the single-character subcommands, I would prefer explicit words as done in e.g. metadata-2, but that's purely cosmetic.

I think the list subcommand could be a better idea instead of implicitly sending a (potentially) big list upon conn registration: the list can be quite large and implicitly sending replies has been problematic in the past (e.g. no-implicit-names, or the messy JOIN burst management).

I wonder if we need to introduce some kind of paging…? I would like the spec to stay as simple as possible, and in practice the list should be limited in size or else it would be annoying to display in a sidebar UI.

> Any thoughts on batching QUERYUSER from the server (particularly if there were an L command for listing?)

Sounds like a good idea either way!

> I still think this needs an implementation considerations section on limiting data growth.

Yup, was definitely something I wanted to add. Done!

extensions/user-query.md
```diff
@@ -0,0 +1,103 @@
...
+
+TODO: interactions with MONITOR?
```
emersion, line 46: I've been wondering whether we should automatically/implicitly add opened user queries to the MONITOR list, to ease client implementations. Clients are very likely interested to get updates from opened user queries just like they get updates from opened channels. But adding some more implicit semantics also makes the whole system more difficult to understand, and the list of opened user queries may grow larger than the MONITOR limit. (FWIW, Goguma only monitors opened user queries interacted with in the last 5 days, up to the MONITOR limit.)

So I think I'd lean on just keeping these two completely separate.

slingamn:
I think under a normal workload, data size shouldn't really be a concern and it should be fine to send all open queries as part of (technically after) the registration burst, without pagination.

emersion:
Switched to a sub-command structure, added a `LIST` sub-command and a new batch type.

emersion:
`OPEN` is typically broadcast to all other connected clients. I wouldn't want a no-op `OPEN` to be broadcast to all other clients. An `OPEN` notification is an indication that the query was closed before.

Clients can trivially ignore `ALREADY_OPENED` errors.

I don't feel strongly about this.

emersion:
> do you have feelings about making the CAP a MUST in order to use any of the USERQUERY commands?

Oh yeah, for sure. Your reasoning makes sense, and more generally I'm in favor of requiring caps to be enabled before commands are used.

Updated to require cap negotiation and mention broadcasting.

skizzerz:
batch should also be a requirement for USERQUERY LIST, rather than an option. If omitted, there's no way for the client to know that the list is complete. Additionally, for traffic mitigation, it'd be best if USERQUERY LIST responses could list multiple nicks on a single line. I can otherwise see someone with 100 queries open getting killed by the server's send queue for running the command.

slingamn:
OK, sounds reasonable to me.

extensions/user-query.md
```diff
@@ -0,0 +1,116 @@
...
+    S: :yannick!yan@example.org USERQUERY OPEN sophie
+
+Closing an existing user query:
+
+    C: USERQUERY CLOSE sophie
+    S: :yannick!yan@example.org USERQUERY CLOSE sophie
+
+The server automatically opens a user query when receiving a message:
+
+    S: :yannick!yan@example.org USERQUERY OPEN sophie
```
slingamn: Is the source of `USERQUERY OPEN` or `CLOSE` always the user themself? If so, what information does it convey?

extensions/user-query.md
```diff
@@ -0,0 +1,116 @@
...
+
+This sub-command has the following general syntax:
+
+    USERQUERY OPEN <nick>
+
+The `nick` parameter specifies a single nickname.
+
+When sent from a client, this sub-command signals to the server that the user has opened a query with the specified user. The server MUST reply to a successful `USERQUERY OPEN` client set command using a `USERQUERY OPEN` server sub-command, or using an error message.
+
+When sent from a server, the `USERQUERY OPEN` command signals to the client that the user has a query opened with the specified user. If a `PRIVMSG` is sent to the user while no user query is opened, the server automatically opens a user query before dispatching the message. Servers SHOULD broadcast the `USERQUERY OPEN` notification to all clients connected under the same account.
```
slingamn, line 58: My understanding of the state of discussion here is that we concluded that when combined with server-side pruning, this (auto-opening queries) becomes a DoS vector. But I don't remember the state of the discussion for alternative strategies.

emersion: We've discussed this a long time ago (logs below). moonmoon has suggested only sending `USERQUERY OPEN` in response to an explicit user action, but AFAIU that alone doesn't solve the race I've outlined. Another suggestion is to add a timestamp to `USERQUERY OPEN`, and make it no-op if the user query has been closed with a later timestamp.

```
19:47:46 <moonmoon> imo query windows should only be opened/recorded at request of the user, so someone sending you a message wouldn't cause the server to record that you have a query opened with that user
19:50:00 <emersion> moonmoon: we've tried this before and i don't think it works
19:50:15 <emersion> let's say i have two devices, laptop and mobile
19:50:30 <emersion> i'm connected on my laptop, someone sends me a message
19:51:16 <emersion> my laptop client sees the PRIVMSG, sends USERQUERY OPEN as a result, since it displays the new message in a new buffer/tab
19:51:39 <emersion> i don't care about the message, so i just close the buffer without even opening it
19:51:53 <emersion> laptop sends USERQUERY CLOSE
19:52:25 <emersion> my mobile client comes online, sees a new unseen message in chathistory
19:52:59 <emersion> it opens a new buffer since it's a new message according to its local state, sends USERQUERY OPEN
19:53:13 <emersion> that last step is undesirable from a UX PoV
19:54:52 <emersion> tl;dr the decision to open a new buffer in response to a new message cannot be taken by individual clients, because they step on each other's toes
19:55:12 <moonmoon> sure, but if the server actively prunes old queries spammers could use auto-open to cause queries you care about to auto-close
19:55:58 <moonmoon> imo the appropriate solution is for clients to *not* automatically send USERQUERY OPEN, and instead only do that if some user action is performed such as accepting a PM window (if your client builds in some sort of UI for that), focusing on the PM query, etc.
19:56:04 <emersion> right, user state being controlled from the outside isn't great
19:56:11 <moonmoon> either that, or add a timestamp param after OPEN
19:56:39 <moonmoon> and an OPEN on a nick with a ts lower than the ts specified on CLOSE just makes the server reply back automatically with a CLOSE message. Or put OPEN/CLOSE in chathistory
19:57:05 <moonmoon> since we're talking PMs, the chathistory for those is already necessarily tailored towards those specific users
19:57:57 <emersion> these are good suggestions
```

PR #573, opened 2025-10-26
Add more FAIL codes for client-initiated batches

skizzerz:
While building out my implementation of this, I noticed that many instances of clients sending garbage to the server did not have a defined error code. My implementation provides a baseline support for the BATCH command while allowing other modules to define recognized batch types. As such, the handler for the BATCH command itself is generic and agnostic to the actual type being used. It still needs to do some basic tracking on the supplied reference tag and type before handing it off to the handler module, so cases where the reference tag and/or type are not valid need error messages.

Rather than making something implementation-defined, it'd probably be best to specify some messages that can be used in this case. An implementation that doesn't split things up like mine does isn't required to use these additional messages, but having them specified lets clients know to expect them.

### DIFF
--- a/extensions/client-batch.md
+++ b/extensions/client-batch.md
@@ -63,8 +63,14 @@ Servers MUST use `FAIL` messages from the [standard replies][] framework
 to notify clients of errors with client-initiated batches.
 The command is `BATCH` and the following codes are defined:
 
+* `INVALID_REFTAG <reference-tag>`: the provided reference tag contains
+  disallowed characters, all past and future messages in this batch
+  will be ignored.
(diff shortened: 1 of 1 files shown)


skizzerz:
As an implementation note, solanum now makes use of these additional codes for client-initiated batches as of https://github.com/solanum-ircd/solanum/commit/4cc9be1ddf1e14fc76bf433e23a3d528424b60ca (plus a couple more related to nesting which aren't part of this PR as the current draft spec forbids nesting but the solanum implementation conditionally allows it for potential future specs which add nested client-initiated batches)

PR #584, opened 2026-02-20
Add client certificate management extension

### LINKED ISSUE #90: Add extension to manage CertFP
The new `draft/account-registration` allows clients to programmatically register a new account and save the password.

It would be nice to allow clients to programmatically add and remove authorized certificates for `SASL EXTERNAL`. This would allow clients to not save the password locally.

emersion:
This has been implemented as a vendor soju extension: https://codeberg.org/emersion/soju/src/branch/master/doc/ext/client-cert.md

Closes https://github.com/ircv3/ircv3-ideas/issues/90

### DIFF
--- a/extensions/client-cert.md
+++ b/extensions/client-cert.md
@@ -1,0 +1,281 @@
+---
+title: Client certificate management extension
+layout: spec
+work-in-progress: true
+copyrights:
+  -
+    name: "Hubert Hirtz"
+    period: "2025"
+    email: "hubert@hirtz.pm"
+  -
+    name: "Simon Ser"
+    period: "2026"
+    email: "contact@emersion.fr"
+---
+
+# Client certificate management
+
+## Notes for implementing work-in-progress version
+
+This is a work-in-progress specification.
+
+Software implementing this work-in-progress specification MUST NOT use the
+unprefixed `client-cert` CAP name and batch type. Instead, implementations
+SHOULD use the `draft/client-cert` CAP name and batch type to be
+interoperable with other software implementing a compatible work-in-progress
+version. The final version of the specification will use unprefixed CAP name
+and batch type.
+
+## Introduction
+
+This document describes an IRC extension that enables clients to manage TLS
+client certificates that are accepted for authentication with SASL EXTERNAL.
+Such certificates are called "pinned certificates".
+
+Each pinned certificate is identified by what this document calls its
+fingerprint. The fingerprint of a TLS client certificate is the SHA-512 digest
+of its DER encoding as defined by [RFC 6234].
+
+Each pinned certificate is also associated with a client-provided name that is
+meant to be human-readable and shown to users.
+
+Two kinds of authentications are to be distinguished when reading this document:
+
+- SASL authentication using the SASL IRC extension; and
+- TLS client authentication (also called TLS authentication in this document),
+  using mechanisms defined in [RFC 8446].
+
+## Architecture
+
+This extension depends on the [SASL 3.2] extension and the [standard replies]
+framework.
+
+This extension optionally depends on the [batch] extension.
+
+This extension defines a `draft/client-cert` capability as well as a
+`CLIENTCERT` command.
+
+## `draft/client-cert` capability
+
+Before sending any `CLIENTCERT` command, both following conditions MUST be
+fulfilled:
+
+- the `draft/client-cert` capability is successfully negotiated,
+- the client is authenticated using SASL.
+
+## `CLIENTCERT` command
+
+The `CLIENTCERT` command sent by the client has one of the following formats:
+
+```
+CLIENTCERT CREATE <name>
+CLIENTCERT LIST
+CLIENTCERT DELETE [fingerprint]
+```
+
+The `CLIENTCERT` command sent by the server has one of the following formats:
+
+```
+CLIENTCERT CREATE
+CLIENTCERT LIST <fingerprint> <attributes>
+CLIENTCERT DELETE [fingerprint]
+```
+
+In the above listings, the parameters have the following formats:
+
+- `attributes` follows the format of `tag` defined by the [message-tags]
+  extension.
+- `fingerprint` is a hex-encoded SHA-512 digest as defined by [RFC 4648] and
+  [RFC 6234].
+
+If the server receives a subcommand other than those listed above, it MUST reply
+with:
+
+```
+FAIL CLIENTCERT UNKNOWN_COMMAND <subcommand> :Unknown subcommand
+```
+
+### `CLIENTCERT CREATE` subcommand
+
+The `CREATE` subcommand has the following format when sent by the client:
+
+```
+CLIENTCERT CREATE <name>
+```
+
+The `CREATE` subcommand has the following format when sent by the server:
+
+```
+CLIENTCERT CREATE
+```
+
+Clients that haven't authenticated using SASL EXTERNAL, and that are wishing to
+authenticate through SASL EXTERNAL using the TLS authentication of the current
+connection may send the `CREATE` subcommand.
+
+If the server cannot process the create request, it MUST send a standard failure
+message with one of the following formats:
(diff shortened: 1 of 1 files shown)


extensions/client-cert.md
```diff
@@ -0,0 +1,281 @@
...
+C: AUTHENTICATE bWFjcm9uAGJnAGJyaWdpdGVqdG0=
+S: :server 900 macron macron!macron@macron.fr macron :You are now logged in as macron
+S: :server 903 macron :SASL authentication successful
+
+C: CLIENTCERT CREATE :senpai on linux!! 😊
+S: CLIENTCERT CREATE
+
+C: CLIENTCERT LIST
+S: :server BATCH +asdf draft/client-cert
+S: @batch=ok CLIENTCERT LIST F70EC120E1603A58F5979A36221630D92AC75DBAEF8C205B5BB94463E1FBB73BB8E77172BB4CCC4B17FA9F7162AFA80F4E865798DF6CDE7F30E50B0362180E82 name=senpai\son\slinux
```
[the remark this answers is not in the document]
jwheare: Do you foresee a lot more of these key/values being added in future? I'd think it would be far more convenient for implementations to be able to lean on their existing tag parsing functionality than to add special case parsing in the rest of the message.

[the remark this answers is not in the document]
emersion: Then the command is not extensible anymore.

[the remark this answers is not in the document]
skizzerz: Any extensions would need to define new functionality anyway, at which point the new client cap would enable an additional parameter, or tag, or whatever. Tags wouldn't even need a new client cap. Those same issues would still exist with the current version (not able to add more data to the command without a new cap), but this way gets rid of needing special parsers and makes use of native IRC parsing without extra effort.

extensions/client-cert.md
```diff
@@ -0,0 +1,281 @@
...
+
+This extension defines a `draft/client-cert` capability as well as a
+`CLIENTCERT` command.
+
+## `draft/client-cert` capability
+
+Before sending any `CLIENTCERT` command, both following conditions MUST be
+fulfilled:
+
+- the `draft/client-cert` capability is successfully negotiated,
```
skizzerz: What's the advantage of defining a new client capability here? There is nothing in the spec that would involve the server sending unsolicited messages using these new commands, it's all just replies to the client sending the CLIENTCERT command. A server could advertise support for this command via a new ISUPPORT token instead.

progval (commented):
good idea, thanks! I'll try to implement it in Limnoria as soon as a server supports it because that will be great for usability

extensions/client-cert.md
skizzerz: Ok, missed that quote. If it can be sent unsolicited then yes a cap is needed. There is a real cost to adding new caps to a server, so they should not be added "by default" -- only when actually necessary. For example, I've recently worked on expanding cap code in solanum to be a 64-bit bitfield instead of 32-bit (meaning the maximum number of caps it supports is 63 instead of 31; one bit is reserved for detecting if we've hit the limit), but if every spec adds a new cap I can see even 63 becoming insufficient eventually sooner rather than later.

extensions/client-cert.md
```diff
@@ -0,0 +1,281 @@
...
+that has this fingerprint may be deleted. Otherwise, only the pinned certificate
+that has the fingerprint of the TLS client certificate of the current IRC
+connection is deleted.
+
+If the server fails to delete a pinned certificate for any other reason, it MUST
+send a standard failure message of the following format:
+
+```
+FAIL CLIENTCERT INTERNAL_ERROR DELETE :...
+```
```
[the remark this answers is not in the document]
progval: `INTERNAL_ERROR` implies it can be retried later (eg. because the server's DB connection is flaky). Either automatically, or with the client suggesting the users does it.

extensions/client-cert.md
```diff
@@ -0,0 +1,281 @@
...
+- `time`: the time of the last successful authentication attempt using the
+  pinned certificate. The value MUST follow the format of time values defined by
+  the [server-time] extension.
+- `ip`: the IP of the last client that successfully authenticated using the
+  pinned certificate. The value MUST be either an `IPv6address` or an
+  `IPv4address` as defined by [RFC3986].
+
+If the `batch` capability was negotiated, the server MUST reply to a `LIST`
+subcommand using a single batch of type `draft/client-cert`. If the list of
+pinned certificates is empty, the server MUST reply with an empty batch.
```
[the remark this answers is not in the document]
skizzerz: a NOTE message may be appropriate in this case if we want to keep the batch dependency optional. e.g. `NOTE CLIENTCERT NO_FINGERPRINTS`. That way the client has something they can listen to instead of just timing out or something and figuring no response = no fingerprints to list.

If we go the NOTE route, would we want that to be the response even if batch is negotiated? Or should the no fingerprints case differ based on whether batch is enabled?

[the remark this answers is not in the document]
slingamn: Make `batch` mandatory, IMO

extensions/client-cert.md
```diff
@@ -0,0 +1,281 @@
...
+and batch type.
+
+## Introduction
+
+This document describes an IRC extension that enables clients to manage TLS
+client certificates that are accepted for authentication with SASL EXTERNAL.
+Such certificates are called "pinned certificates".
+
+Each pinned certificate is identified by what this document calls its
+fingerprint. The fingerprint of a TLS client certificate is the SHA-512 digest
```
slingamn, line 36: Any particular reason to choose SHA-512 here? I think most existing implementations (Unreal, Insp, Ergo) hardcode or default to SHA-256. (It's not an insuperable obstacle, just a source of friction, and I don't see a security concern given that this only needs second preimage resistance, which is unbroken even for MD5, let alone SHA-256.)

sadiepowell: Personally I don't see the need to define the hash or encoding method tbh. It can just be an opaque string allowing the IRCd and services to choose the encoding.

extensions/client-cert.md
```diff
@@ -0,0 +1,281 @@
...
+fingerprint in the form of additional `key=value` parameters after the
+`fingerprint` parameter. Clients MUST ignore unknown keys. The following keys
+are defined:
+
+- `name`: The name of the pinned certificate. The value MUST be at least 1 and
+  at most 64 bytes.
+- `time`: the time of the last successful authentication attempt using the
+  pinned certificate. The value MUST follow the format of time values defined by
+  the [server-time] extension.
+- `ip`: the IP of the last client that successfully authenticated using the
```
sadiepowell, line 187: How would servers which support TLS over UNIX sockets handle this?

grawity: Seems like it is specified as MAY – but I assume this field is for user display purposes only? so if I were writing such an ircd, I'd ignore the fact that it's named `ip` and would have the server send a string like `(local user $UID on $HOSTNAME)` or `unix/$UID@$HOSTNAME` or something similar (i.e. socket peercred, maybe with a getpwuid lookup even). Or `0.0.0.0` as is tradition.

slingamn: Ergo falls back to `127.0.0.1` in this case if it doesn't have a proxied IP. Are there real use cases for TLS over Unix domain (without proxying)?

sadiepowell: Its a niche use case but its also something we support (InspIRCd's TLS support is entirely origin-agnostic) and we'd need to special case it here as users' client address is their UNIX socket path. Ideally the spec should be vague about the protocol used so servers can put whatever origin want in the field imo.

extensions/client-cert.md
slingamn: It saves the client from potentially having to store an association between the certificate and the server-pushed identifier...it's also more resilient to server software changes. (Not that these are decisive arguments, but I can see the point of specifying the hash algorithm.)

PR #586, opened 2026-02-23
Add custom emoji spec

skizzerz:
There's a lot here and is largely guided from some discussions in `#ircv3`. Given the already-chunky size of the spec, I opted to omit any methods of programmatically managing the linked-to emoji pack JSON documents for clients to build emoji-management UIs (and for servers to maintain stricter control over URLs contained within the emoji packs). I'll be sending that up as a separate spec at some point in the future unless someone else beats me to it first.

### DIFF
--- a/extensions/custom-emoji.md
+++ b/extensions/custom-emoji.md
@@ -1,0 +1,235 @@
+---
+title: "Custom Emoji"
+layout: spec
+work-in-progress: true
+copyrights:
+  -
+    name: "Ryan Schmidt"
+    period: "2026"
+    email: "moonmoon@libera.chat"
+---
+
+## Notes for implementing work-in-progress version
+
+This is a work-in-progress specification.
+
+Software implementing this work-in-progress specification MUST NOT use the unprefixed `EMOJI` ISUPPORT name, the `emoji` channel METADATA name, or the `+emoji` tag name. Instead, implementations SHOULD use the `draft/EMOJI` ISUPPORT name, the `draft/emoji` channel METADATA name, and the `+draft/emoji` tag name to be interoperable with other software implementing a compatible work-in-progress version.
+
+The final version of the specification will use unprefixed ISUPPORT, channel METADATA, and tag names.
+
+## Introduction
+
+This specification introduces the ability for servers and channels to define packs of custom emoji associated with shortcodes. These packs are accessible via URLs returning a JSON document following a well-defined schema.
+
+## Dependencies
+
+Clients wishing to display or send custom emoji MUST negotiate the [`message-tags`](message-tags.html) capability with the server. Clients SHOULD additionally negotiate the [`metadata-2`](metadata.html) capability with the server so they can interact with custom emoji packs defined on a per-channel basis.
+
+## Format
+
+The emoji tag is sent by a client with the client-only prefix `+`. The value is a comma-separated (`,`) list of shortcode=packid pairs, with an optional suffix for each shortcode, as defined by the following ABNF grammar.
+
+```
+value       = emoji-def *("," emoji-def)
+emoji-def   = shortcode ["~" suffix] "=" pack-id
+shortcode   = 1*40(ALPHA / DIGIT / %x21-2B / %x2E-2F / %x3B-3C / %x3E-3F / %x5B-60 / %x7B-7D)
+              ; All ASCII numbers, letters, and symbols EXCEPT FOR ,=@:~
+suffix      = 1*3DIGIT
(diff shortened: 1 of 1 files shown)


extensions/custom-emoji.md
```diff
@@ -0,0 +1,235 @@
...
+
+When receiving a PRIVMSG, NOTICE, or [reaction](../client-tags/react.html) with an emoji tag, clients search for all occurrences of the exact substrings of each shortcode and optional suffix surrounded by a colon (`:`) on each side outside of monospaced contexts, and MAY replace those strings with the associated custom emoji image for the provided shortcode from the emoji pack whose id matches pack-id. Whether any particular shortcode is replaced with an associated custom emoji is an implementation decision for the client (see Client implementation considerations below).
+
+A monospaced context is a portion of the message body beginning with the monospace formatting code 0x11 and ending with either the monospace formatting code 0x11 again or the format reset code 0x0F. Even if a client does not support 0x11 for formatting text as monospaced, it MUST NOT replace custom emoji shortcodes within such contexts with the associated images. This allows for an easy way of "escaping" emoji shortcodes that would otherwise be replaced by images.
```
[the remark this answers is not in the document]
classabbyamp: I mentioned using `\:lol:` to escape in #ircv3, what about that?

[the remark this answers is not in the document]
jwheare: I can't actually see anywhere in this spec that explains what the suffix is for, as defined.

extensions/custom-emoji.md
```diff
@@ -0,0 +1,254 @@
...
+            ; This corresponds to the regex class [^\p{Cc}\p{Cf}\p{Cs}\p{Zl}\p{Zp}\p{Zs}/:]
+```
+
+When receiving a PRIVMSG, NOTICE, or [reaction](../client-tags/react.html), implementations search for all occurrences of emoji-value, and MAY replace those strings with the associated custom emoji image for the provided shortcode from the emoji pack whose id matches pack-id. Implementations SHOULD perform Unicode normalization using Form C (canonical decomposition followed by canonical composition) on the message before attempting emoji replacements. If a given emoji-value did not specify a particular pack-id, implementations MUST use the following algorithm to determine the corresponding pack-id:
+
+1. If the message is in a channel, look up the shortcode within the emoji pack document specified by that channel's `draft/emoji` METADATA.
+2. If step 1 fails to produce a shortcode mapping, look up the shortcode within the emoji pack document specified by ISUPPORT `draft/EMOJI`.
+3. If steps 1 and 2 fail to produce a shortcode mapping, an implementation MAY attempt to resolve the pack-id via other means (such as a local cache of known packs).
```
sugar700, line 55: What's the point of this fallback?

To me it sounds like an interoperability issue, where different IRC clients could resolve emojis in different ways.

The specification provides an example of replacing `:wave:` with 👋, but I don't understand why couldn't the client simply replace `:wave:` with 👋 when sending the message.

extensions/custom-emoji.md
```diff
@@ -0,0 +1,254 @@
...
+        }
+    }
+]
+```
+
+*--End non-normative example]]*
+
+## Emoji images
+
+Emoji images SHOULD be a minimum of 128x128 pixels in size for raster images. This allows clients to size emoji appropriately in their UI without needing to worry about blurry images on high DPI displays. This specification does not dictate any particular file formats for emoji images, however clients SHOULD support static PNG and JPG images and MAY support any number of additional formats. See the client implementation considerations section for discussion on the security implications of supporting certain formats.
```
sugar700, line 154: It may make sense to recommend IRC clients to send HTTP `Accept` header with the list of formats they support when they send a request to retrieve emoji. This would allow IRC servers to, say, determine whether to send animated WebP file or a PNG emoji, depending on file format support.

extensions/custom-emoji.md
skizzerz: The client could certainly do that. I'm not going to mandate that clients do so in outgoing messages because it's impossible to specify in any clean fashion and there's no feature negotiation needed to make use of custom emoji.

The reason for the fallback is simple: I cannot predict where custom emoji may go in the future or what inventive things users and clients may eventually do with it, and the algorithm to resolve an emoji shortcode is a MUST with a strictly-defined ordering. This fallback offers an escape hatch to augment the resolution algorithm in the future, such as via a future specification which brings some additional optional features into custom emoji. These extensions would not be possible without this fallback without errata to this specification as well (breaking the ability for those extensions to be self-contained).

PR #589, opened 2026-03-11
Add specification for channel flood limits

ValwareIRC:
This document outlines the specification proposed for channel flood limits, detailing how servers can advertise flood protection limits to clients and how privileged clients can set additional limits

### DIFF
--- a/extensions/channel-flood-limits.md
+++ b/extensions/channel-flood-limits.md
@@ -1,0 +1,853 @@
+# Channel Flood Limits
+
+Copyright © 2026 Valware <v.a.pond@outlook.com>
+
+Unlimited redistribution and modification of this document is allowed provided
+that the above copyright notice and this permission notice remains intact.
+
+This specification is a work-in-progress and may have major incompatible changes
+without warning.
+
+## Notes for implementing work-in-progress version
+
+This is a work-in-progress specification.
+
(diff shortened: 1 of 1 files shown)


sadiepowell:
I don't think this is a workable idea. Half of it is just reinventing channel modes to fix a deficiency with UnrealIRCd not telling users about default channel flood limits and the method for communicating flood methods isn't workable in practice because how flood methods work is vendor-specific meaning clients have to fall back to per-IRCd guesswork like they do currently.

skizzerz:
On Libera we explicitly do not publish our server configurations regarding flood limits/fakelag/etc., because it would otherwise provide an easy guideline for spambots to stay behind so they are more likely to evade network-wide detection and prevention measures for longer. These configurations are also subject to change without notice to respond to threats. We would not implement any form of this specification as written.

I can potentially see a value for a specified channel metadata key for channel bots to publish their own limits in a consumable form, for channels who do wish to expose such limits publicly. This really wouldn't need any sort of client capability or additional commands though.

Celelibi:
Looks like a fairly complex spec for just communicating the flood limits and not enforcing them despite the server having to know and manipulate them all. I'm wondering if the extbans wouldn't be a better framework for both communicating and enforcing flood limits. All it would need is a new matching criterion on the number of message per unit of time.

PR #591, opened 2026-03-12
Add the {size} template variable to the network-icon spec.

sadiepowell:
As discussed on IRC.

### DIFF
--- a/extensions/network-icon.md
+++ b/extensions/network-icon.md
@@ -7,6 +7,10 @@ copyrights:
     name: "Simon Ser"
     period: "2025"
     email: "contact@emersion.fr"
+  -
+    name: "Sadie Powell"
+    period: "2026"
+    email: "sadie@witchery.services"
 ---
 
 ## Notes for implementing work-in-progress version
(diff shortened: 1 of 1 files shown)


jwheare:
Might be worth adding something like

> Servers advertising a URL with the {size} variable SHOULD serve a default size when the literal string `{size}` is requested, in order to support clients that do not perform URL template interpolation.

And change the client "MUST replace" to a MAY.

jwheare:
Will aim to merge this next week.

emersion:
Are there implementations of this?

furudean:
@emersion halloy has an open PR: https://github.com/squidowl/halloy/pull/1781

emersion:
It doesn't seem like it uses the `{size}` parameter added in this PR?

https://github.com/squidowl/halloy/pull/1781/changes#diff-efaa5e2d4f638382602a1303ccf0f80bc2cd54cd68d4c77f8381b2d3d5a2beeeR1352

furudean:
i thought this PR was for draft/icon and not an addition to draft/icon so i misread the context, sorry. disregard my message.

jwheare:
IRCCloud implements it.

image

Note in this example, the value is ignored and a different fixed size image is being returned, which is perfectly valid.

sadiepowell:
Are there any objections to merging this?

PR #592, opened 2026-03-16
Clarify message-tags namespaces

sadiepowell:
The spec said there are two tag namespaces but then lists three (vendor, draft, standardized).

### DIFF
--- a/extensions/message-tags.md
+++ b/extensions/message-tags.md
@@ -179,7 +179,7 @@ typical method is to add `-0.x` to the name, where `x` is a version number. For 
 
 ### Standardized
 
-Reserved names for which a corresponding document exists in the [IRCv3 Extension Registry][registry].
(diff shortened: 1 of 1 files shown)


emersion (approved):

nektro:
isn't it saying that standardized tags are essentially unnamespaced?

sadiepowell:
> isn't it saying that standardized tags are essentially unnamespaced?

I can't read that from the spec tbh. If anything, I think the person who wrote this envisioned `draft/` tags as a type of IRCv3 vendor-prefixed tag but then they got moved to their own block.

Regardless, it should probably be clarified.

delthas (approved):

extensions/message-tags.md
jwheare:
```suggestion
A non-prefixed global namespace of reserved names for which a corresponding document exists in the [IRCv3 Extension Registry][registry].
```

PR #593, opened 2026-03-18
Clarify that the INVALID_UTF8 standard reply can have no command.

sadiepowell:
Requested by @progval as InspIRCd handles messaging parsing in a separate step to command lookup.

Ref: https://github.com/progval/irctest/pull/359

### DIFF
--- a/extensions/utf8-only.md
+++ b/extensions/utf8-only.md
@@ -22,7 +22,7 @@ This specification introduces a new token `UTF8ONLY` that servers can include in
 If a client implementing this specification sees this token, they MUST set their outgoing encoding to UTF-8 without requiring any user intervention. This allows clients to work transparently on networks that only allow UTF-8 traffic.
 
 ## The `INVALID_UTF8` standard replies code
-This is a code that can be used with the [standard replies](https://ircv3.net/specs/extensions/standard-replies) specification. When sent with the `FAIL` command, it indicates that the client's message was rejected because it contained invalid UTF-8 data. When sent with the `WARN` command, it indicates that the message was modified but still accepted.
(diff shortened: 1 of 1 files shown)


progval (approved):

slingamn:
Trying to clarify what's at stake here: non-ASCII commands are not valid IRC under any circumstances, right?

Herringway:
Seems like this would be *required* if the command happened to be invalid UTF-8. There would be no way to echo that back to the client.

sadiepowell:
> Trying to clarify what's at stake here: non-ASCII commands are not valid IRC under any circumstances, right?

For compliant clients that is correct but that doesn't mean misconfigured clients won't try to send one and if the command name is not encodable as UTF-8 then it can't be sent in the standard reply.

This behaviour is already implicitly allowed by the spec (and UnrealIRCd will send it in some cases) but its probably best to have it explicitly stated.

slingamn:
One alternative would be to send `ERR_UNKNOWNCOMMAND`, for example.

slingamn:
The problem is not necessarily exclusive to UTF8ONLY, for example, how should servers respond to `:a :\r\n` ? Here we also have an invalid command (`:`) that cannot be returned in `ERR_UNKNOWNCOMMAND` or `FAIL`. (I think the correct response is to send `ERR_UNKNOWNCOMMAND` with `*`.)

sadiepowell:
The command being invalid UTF-8 is just one example of where this is useful. In the case of InspIRCd we validate the message entirely as one string so we don't actually know if the command is valid or not because we handle messaging parsing in a separate step to command lookup. This behaviour is already used by UnrealIRCd in some cases so it feels appropriate.

slingamn:
Seems reasonable to me, but it also seems that it's not the current spec we need to amend --- it's standard-replies, which says:

>&lt;command&gt;: Indicates the user command which this reply is related to, or is * for messages initiated outside client commands (for example, an on-connect message).

If that language were changed to include "or cases where the user command cannot be parsed", then we wouldn't need a spec change here.

sadiepowell:
Thats fine to me, I filed #594 with that change instead.

PR #594, opened 2026-03-19
Clarify that standard replies can have no command if one can't be parsed.

sadiepowell:
Replaces #593.

### DIFF
--- a/extensions/standard-replies.md
+++ b/extensions/standard-replies.md
@@ -33,7 +33,7 @@ Some non-normative examples:
 Here's what each parameter means:
 
 - `<type>`: Either `FAIL`, `WARN`, or `NOTE`, this indicates the message type.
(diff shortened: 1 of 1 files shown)


extensions/standard-replies.md
```diff
@@ -33,7 +33,7 @@ Some non-normative examples:
 Here's what each parameter means:
 
 - `<type>`: Either `FAIL`, `WARN`, or `NOTE`, this indicates the message type.
-- `<command>`: Indicates the user command which this reply is related to, or is `*` for messages initiated outside client commands (for example, an on-connect message).
+- `<command>`: Indicates the user command which this reply is related to, or is `*` for messages initiated outside client commands (for example, an on-connect message) or where the user command cannot be parsed.
```
slingamn: +1. Nit on language, I would say something more like:

`<command>`: Indicates the client command which this reply is related to. If the reply is initiated outside the context of a client command (for example, an on-connect message), or if the client command is not available or cannot be relayed (for example, if the client's message could not be parsed), `*` is substituted.

PR #595, opened 2026-03-21
Several improvements to the account-registration spec.

### LINKED ISSUE #555: Implementation feedback for the draft/account-registration specification
I've implemented both the server (InspIRCd) and services (Anope) side of this specification and I have some implementation feedback:

- Unlike the SASL specification the `draft/account-registration` specification doesn't mention whether the capability needs to be requested before the commands can be used. This should probably be required so the server knows to hold the connection until services responds like with SASL.

- `REGISTER` allows the user to specify `*` for their current nickname but `VERIFY` does not. This should probably be also allowed for consistency.

- There's currently no way for the server to communicate password restrictions to clients. We should probably at the very least expose the minimum and maximum password length. This would make it a lot easier for clients that perform automatic registration.

- The registration specification currently hardcodes the use of email as the the contact method. It would be good if we could make this more generic so implementations could implement alternate contact methods.

sadiepowell:
See #555 for the rationale.

Closes #555.

### DIFF
--- a/extensions/account-registration.md
+++ b/extensions/account-registration.md
@@ -69,10 +69,18 @@ The only defined capability keys so far are:
  * `before-connect` - if present, indicates the server supports early
    registration, so it MUST NOT use
    `FAIL REGISTER COMPLETE_CONNECTION_REQUIRED`
- * `email-required` - if present, registrations require a valid email address
-   to process
+
  * `custom-account-name` - if present, the account name can be different
    from the user's current nickname
+
+ * `email-required` - if present, registrations require a valid email address
+   to process
+
+* `max-password-length=<num>` - if present, passwords are restricted to a
+  maximum length of `<num>` bytes. `<num>` MUST be a non-zero positive integer.
+
+* `min-password-length=<num>` - if present, passwords are restricted to a
+  minimum length of `<num>` bytes. `<num>` MUST be a non-zero positive integer.
 
 ### Commands
 
--- a/extensions/account-registration.md
+++ b/extensions/account-registration.md
@@ -76,8 +76,8 @@ The only defined capability keys so far are:
 
 ### Commands
 
-    REGISTER <account> {<email> | "*"} <password>
-    
+    REGISTER {<account> | "*"} {<email> | "*"} <password>
+
 The `REGISTER` command informs the server of a request to register
 an account named for the current nick of the requestor.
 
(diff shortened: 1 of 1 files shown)


slingamn (approved):

jwheare:
If all happy, will merge next week.

progval (approved):

extensions/account-registration.md
```diff
@@ -49,7 +49,10 @@ This specification uses [standard replies][] framework.
 ### Capability
 
 This specification adds the `draft/account-registration` capability, whose
-presence signifies that the server accepts the `REGISTER` command.
+presence signifies that the server supports account registration.
+
+The client MUST enable the `draft/account-registration` capability before using
+the `REGISTER` and `VERIFY` commands defined by this specification.
```
ValwareIRC, line 59: Why? I can understand for during the connection flow, this makes sense, but typing `/register` and its parameters are user-intuitive to type and it doesn't make sense to gate this behind the cap. Many servers already implement aliases that forward `/REGISTER` to NickServ automatically, too, existing by default in UnrealIRCd even.

extensions/account-registration.md
```diff
@@ -66,11 +69,19 @@ The only defined capability keys so far are:
...
- * `email-required` - if present, registrations require a valid email address
-   to process
+
  * `custom-account-name` - if present, the account name can be different
    from the user's current nickname
 
+ * `email-required` - if present, registrations require a valid email address
+   to process
+
+ * `max-password-length=<num>` - if present, passwords are restricted to a
```
ValwareIRC, line 83: What is the benefit of capping password lengths?

extensions/account-registration.md
0x5c: The command already has different semantics than existing nickserv `REGISTER` commands, both in the arguments it accepts and the replies from the server. A user sending `REGISTER` without client support will likely not see the server's repliees in their client, and that's with assuming they already guessed the proper argument order and values.

extensions/account-registration.md
sadiepowell: This exposes a feature of existing services implementations. The limit is generally large (e.g. Anope uses 50 characters) but there still has to be some kind of limit to prevent abuse.

extensions/account-registration.md
sadiepowell: This requirement only limits this specification, UnrealIRCd is free to still provide its existing `/REGISTER` behaviour if the user does not have the account registration capability.

PR #596, opened 2026-04-01
Ratify the channel-context client tag specification.

sadiepowell:
I don't think there are any objections relating to this and we have enough implementations so lets ratify and ship it!

---

### Client

* [x] gamja
* [x] Goguma
* [x] IRCCloud
* [x] ObsidianIRC
* [x] senpai

### Bots

* [x] Anope (entry messages on compatible IRCds)
* [x] Limnoria


jwheare:
Will merge next week.

PR #597, opened 2026-04-03
Add `draft/extoidc` and `draft/OBJECTSTORAGE`

reesericci:
This pull request provides a secure and standardized flow for authenticating external services using OpenID Connect over an IRC backchannel (no web redirects!), as well as secure object storage provider discovery and authentication using extoidc, as well as resumable uploads using tus.io.

Stemmed from:

#562
https://github.com/ircv3/ircv3-specifications/issues/562#issuecomment-4179169125
https://github.com/ircv3/ircv3-specifications/issues/562#issuecomment-4104704886

### DIFF
--- a/extensions/extoidc.md
+++ b/extensions/extoidc.md
@@ -1,0 +1,105 @@
+---
+title: OpenID Connect with the Client-Initiated Backchannel Authentication Flow (CIBA) for authentication to external services
+layout: spec
+work-in-progress: true
+copyrights:
+  -
+    name: "Reese Armstrong"
+    email: "me@reeseric.ci"
+    period: "2026"
+---
+
+# extoidc
+
+This is a work-in-progress specification.
+
+This specification introduces the `draft/EXTOIDC` CAP token for servers to indicate support for this specification.
+
+Software implementing this work-in-progress specification MUST NOT use the unprefixed `EXTOIDC` CAP name. Instead, implementations SHOULD use the `draft/EXTOIDC` CAP name to be interoperable with other software implementing a compatible work-in-progress version. The final version of the specification will use unprefixed CAP names.
+
+## Motivation
+
+This specification allows for end-users to authenticate to external services securely over IRC transport without a web-browser requirement. It additionally forms the core of the draft/OBJECTSTORAGE` authentication system.
+
+## Definitions
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED",  "MAY", and "OPTIONAL" in this document are to be interpreted as described in RFC 2119.
+
+“Server” refers to the IRC daemon
+
+“Client” refers to the end-user software interacting with the server
+
+“Relying Party” refers to the service requiring authentication from the client.
+
+## Flow
+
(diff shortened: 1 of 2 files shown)


extensions/extoidc.md
```diff
@@ -0,0 +1,105 @@
...
+    name: "Reese Armstrong"
+    email: "me@reeseric.ci"
+    period: "2026"
+---
+
+# extoidc
+
+This is a work-in-progress specification.
+
+This specification introduces the `draft/EXTOIDC` CAP token for servers to indicate support for this specification.
```
ValwareIRC: [Typically a CAP name is lowercase](https://ircv3.net/specs/extensions/capability-negotiation.html#rules-for-naming-capabilities), ISUPPORT tokens are uppercase. Both support the lowercase `draft/` prefix

ValwareIRC (commented):
Just curious, how comes you didn't go with a SASL mechanism? (I just woke up, sorry if dumb question)

reesericci:
> Just curious, how comes you didn't go with a SASL mechanism? (I just woke up, sorry if dumb question)

It implicitly uses SASL because the authentication backchannel is over a (presumably SASL-authenticated) IRC connection

extensions/extoidc.md
reesericci: Fixed

ValwareIRC:
it would be really nice if you could include some normative examples of how the flow might look to the client and server and third-party

anotherdoesnm:
well i think extoidc can be replaced with just SASL OAUTHBEARER?
more: https://emersion.fr/blog/2022/irc-and-oauth2/
and storage is... draft/FILEHOST?

reesericci:
> it would be really nice if you could include some normative examples of how the flow might look to the client and server and third-party

Can do, I'll make some diagrams.

> well i think extoidc can be replaced with just SASL OAUTHBEARER?

No, that standard is to allow for logging into an IRC server with OAuth, not for the IRC server to expose an OIDC directory. This is for logging into external services (ie a file host) securely with your IRC connection, not for establishing that IRC connection.

> and storage is... draft/FILEHOST?

Please read before commenting. This was created explicitly to address security issues in FILEHOST.

reesericci:
Ok I added some normative diagrams that should clear up the flow for folks. Let me know what you think.

emersion:
I've written my thoughts about this PR here: https://github.com/ircv3/ircv3-specifications/pull/562#issuecomment-4188668243

Celelibi:
I could get behind this if I get convinced it's the minimum complexity for the client protocol.
I'm not well versed in OAuth/OIDC and I'm not sure I understand the purpose of the whole dance to confirm the usage of the token.
Couldn't we at least send all the information to the client at once when asking for a confirmation? At best, could we remove this confirmation entirely? Why do we need to confirm on something we explicitly asked anyway?

Also, why does it need to be OIDC explicitly on the client side? Couldn't it be an opaque token?
I know you mentioned token signature to check for non-modification here: https://github.com/ircv3/ircv3-specifications/pull/562#issuecomment-4189421248
But does the client need to do that? It's very likely that the connection uses TLS anyway. So this seem to add very little security. And if it's not on TLS... well, you're probably already screwed anyway.

I see no mention of the service to request a token for. This information could allow the server to generate tokens differently depending on the external service. Even assuming every external service uses OIDC, bugs and implementation subtleties always come up.

Issue #437, opened 2021-02-24
ratify CHATHISTORY


slingamn:
Current implementations:

Servers:
1. Oragono

Clients:
1. Kiwi (modulo kiwiirc/kiwiirc#1250 and kiwiirc/kiwiirc#1397)
2. Gamja

Bouncers (as server):
1. Soju
2. Kiwibnc (probably with some bugs / incompatibilities)

Networks:
1. https://testnet.oragono.io

Where to: I'd love to see this in a desktop and a mobile client. It would also be useful to have this supported by a server implementation that currently buffers history for autoreplay on channel join: requesting the chathistory cap would suppress the auto-playback, and then CHATHISTORY queries could be answered from the buffer.

emersion:
> It would also be useful to have this supported by a server implementation that currently buffers history for autoreplay on channel join: requesting the chathistory cap would suppress the auto-playback, and then CHATHISTORY queries could be answered from the buffer.

FWIW, soju does exactly this. It's a bouncer though.

slingamn:
@kylef notes an incompatibility with the [chathistory batch spec](https://ircv3.net/specs/extensions/batch/chathistory-3.3), which seemingly forbids the use of the `*` target:

>This parameter contains the target of messages within the batch. The target MUST either be the nick of the remote client for private messages or the name of a channel which the local client is in for public messages.

So if possible, we should amend that spec?

jwheare:
A couple of questions that remain unanswered from the initial draft PR discussion (https://github.com/ircv3/ircv3-specifications/pull/393#issuecomment-721265459)

* Is there a reason TAGMSG is only included with event-playback?
* Would e.g. `chathistory-events` be a better name than `event-playback`?

slingamn:
I had some skepticism about this rationale:

>There's no risk of TAGMSG causing a state related desync, they are by design lossy, so can't communicate important consistent state.

How is it explicit in the design of TAGMSG that they are lossy? For example, naive implementations of `+draft/react` would cause a local state change.

As for the name --- what would be the rationale for changing it? To clarify that it's a sub-capability of `chathistory`?

jwheare:
They’re lossy because only some clients see them. And they carry no more state than a privmsg. It’s not like a quit or join that affects the channel list.

jwheare:
The rationale for changing the events is that the original one doesn’t match the base name. It’s not chat-playback. draft stage means me can change stuff to be better before it becomes permanent.

[...]

PR #598, opened 2026-04-09
chathistory: add draft/chathistory-end tag

emersion:
See https://github.com/ircv3/ircv3-specifications/issues/437#issuecomment-4137718984

Implementations:

- https://github.com/ergochat/ergo/pull/2379
- https://codeberg.org/emersion/soju/pulls/381
- https://codeberg.org/emersion/goguma/commit/e6b43928aaffdf32d838663d0ab8bb5c2475af78

### DIFF
--- a/extensions/chathistory.md
+++ b/extensions/chathistory.md
@@ -129,6 +129,10 @@ Servers SHOULD provide clients with a consistent message order that is valid acr
 
 The server may choose to return additional, related messages alongside regular chat history. Such messages MUST be tagged with the `draft/chathistory-context` tag and MUST immediately follow their parent message. Context messages MUST NOT be counted towards the message limit. Context messages are sent at the server's discretion and MAY include reacts, redacts, edits, etc.
 
+#### `draft/chathistory-end` tag
+
+To signal that no more messages are available, servers SHOULD attach a `draft/chathistory-end` message tag without a value to the first `BATCH` message in the `chathistory` or `draft/chathistory-targets` batch. This indicates to the client that trying to request the next page of results will return an empty batch.
+
 #### Errors and Warnings
 Errors are returned using the standard replies syntax.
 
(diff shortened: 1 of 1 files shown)


slingamn:
From #437, should we warn clients that a response with fewer than `<limit>` messages does not indicate the last page? (To avoid compatibility issues with existing clients, I wouldn't change the Ergo behavior to start sending short non-final pages until ratification.)

emersion:
I've pushed a suggestion to that effect. I've chosen to spec that servers SHOULD apply the exact limit, because a server completely ignoring the limit would cause havoc - is that too strong for your taste?

slingamn:
SHOULD does seem a little strong to me because there are conditions where Ergo will predictably violate the limit (for example, relaying stored multiline messages to non-multiline-capable clients). I expect other servers will run into similar conditions.

That said, I don't have very strong feelings, and the current language gives servers explicit permission to deviate.

emersion:
If you have ideas for an alternative wording, please feel free to suggest!

slingamn:
I think basically just reversing the capitalization:

>A `limit` parameter MUST be a positive integer. Servers should attempt to return exactly `limit` messages (or as many are available), but MAY return more or fewer due to implementation constraints. The `draft/chathistory-end` tag indicates whether more messages are available.

emersion:
Works for me, I've pushed this wording!

extensions/chathistory.md
```diff
@@ -66,6 +66,8 @@ The `target` parameter specifies a single buffer (channel or nickname) from whic
 
 A `timestamp` parameter MUST have the format `timestamp=YYYY-MM-DDThh:mm:ss.sssZ`, as in the [server-time][server-time] extension. A `msgid` parameter MUST have the format `msgid=foobar`, as in the [message-ids][message-ids] extension.
 
+A `limit` parameter MUST be a positive integer. Server should attempt to return exactly `limit` messages (or as many as available), but MAY return more or fewer due to implementation constraints. The `draft/chathistory-end` tag indicates whether more messages are available.
```
slingamn: "Servers" plural, I think

slingamn (approved):

extensions/chathistory.md
emersion: Oops, fixed!

emersion:
Any chance to get this merged? :)

delthas (approved):

emersion:
Thanks!

delthas:
Added to senpai: https://git.sr.ht/~delthas/senpai/commit/4b426b9c069b2567ae0bdcf9b9ad52a7c533aacb

delthas:
For future reference: this also applies to TARGETS.

PR #599, opened 2026-04-10
network-icon: Support varying icons sent to clients using HTTP headers

internet-catte:
Clients may want to be able to display an icon that fits better with a light or dark background/theme, especially if the icon has a transparent background.

Users may also wish to avoid animated icons (such as animated GIFs or APNG images), so clients could register that preference when requesting the icon.

This might not be necessary to put in the specification, but I think this is worth exposing in the spec so clients and servers implement support for it. I would be open to putting it in a non-normative implementation note, if not in the spec itself.

I expect to be implementing support for this on Libera, when [support for `network-icon`](https://github.com/solanum-ircd/solanum/compare/main...internet-catte:solanum:draft-network-icon) is deployed.

### DIFF
--- a/extensions/network-icon.md
+++ b/extensions/network-icon.md
@@ -7,6 +7,10 @@ copyrights:
     name: "Simon Ser"
     period: "2025"
     email: "contact@emersion.fr"
+  -
+    name: "internet-catte"
+    period: "2026"
+    email: "catte@libera.chat"
 ---
 
 ## Notes for implementing work-in-progress version
@@ -27,6 +31,33 @@ If a server publishes a `draft/ICON` ISUPPORT token, the value MUST be a URL to 
 
 Servers MAY implement [extended-isupport](extended-isupport.html) to allow clients to fetch the network icon before connection registration.
 
-## Example
+Servers MAY support requesting different icons using [HTTP client hints], including [`Sec-CH-Prefers-Color-Scheme`] for colour scheme-optimised variants and [`Sec-CH-Prefers-Reduced-Motion`] for non-animated variants.
+
+## Examples
 
     S: :irc.example.org 005 * NETWORK=Example draft/ICON=https://example.org/icon.svg :are supported by this server
+
+Server supporting light and dark colour scheme icons:
+
+    S: :irc.example.org 005 * NETWORK=Example draft/ICON=https://example.org/icon.png :are supported by this server
+
+    C: GET /icon.png HTTP/1.1
+    C: Host: example.org
+
+    S: HTTP/1.1 200 OK
+    S: Content-Type: image/png
+    S: Accept-CH: Sec-CH-Prefers-Color-Scheme
+    S: Vary: Sec-CH-Prefers-Color-Scheme
+    S: (default image)
+
+    C: GET /icon.png HTTP/1.1
+    C: Host: example.org
+    C: Sec-CH-Prefers-Color-Scheme: "dark"
+
+    S: HTTP/1.1 200 OK
+    S: Content-Type: image/png
+    S: (image optimised for dark backgrounds)
+
+[HTTP client hints]: https://datatracker.ietf.org/doc/html/rfc8942
+[`Sec-CH-Prefers-Color-Scheme`]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Sec-CH-Prefers-Color-Scheme
+[`Sec-CH-Prefers-Reduced-Motion`]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Sec-CH-Prefers-Reduced-Motion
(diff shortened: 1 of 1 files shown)


skizzerz:
We should specify a couple more things than this. There's other useful headers that would be good to inform implementations to be on the lookout for (for servers) or things they can send that might have an effect (for clients):

- `Accept`: For the client to list supported/preferred filetypes. This could be used to e.g. send a raster vs vector image depending on client preference as not all of them support SVG
- `Sec-CH-DPR`: Let the client indicate the pixel density of the user's display, so a higher-density image can be returned if supported. This more directly maps to `<img srcset` than the `{size}` attribute, although both can be used in concert ("I want a 64px image at 2x pixel density").

extensions/network-icon.md
```diff
@@ -33,6 +37,10 @@ The URL MAY contain the `{size}` template variable that clients MAY replace with
 
 Servers MAY implement [extended-isupport](extended-isupport.html) to allow clients to fetch the network icon before connection registration.
 
+Clients MAY send the [`Accept`] HTTP header to indicate their supported image formats. Servers MAY support interpreting the [`Accept`] HTTP header to send clients their preferred image format.
+
+Servers MAY support requesting different icons using [HTTP client hints], including [`Sec-CH-Prefers-Color-Scheme`] for colour scheme-optimised variants, [`Sec-CH-Prefers-Reduced-Motion`] for non-animated variants, [`Sec-CH-Prefers-Reduced-Transparency`] for variants without a transparent background, and [`Sec-CH-DPR`] for high-DPI-optimised variants.
```
jwheare, line 42: I hadn't realised the Sec-DH headers need to be opted into by servers via Accept-CH. This isn't amazingly obvious for those not already familiar with the concept. I realise the example shows this in action, but I wonder if we should make the wording be a bit more explicit about that.

```suggestion
Servers MAY support requesting different icons using [HTTP client hints] including [`Sec-CH-Prefers-Color-Scheme`] for colour scheme-optimised variants, [`Sec-CH-Prefers-Reduced-Motion`] for non-animated variants, [`Sec-CH-Prefers-Reduced-Transparency`] for variants without a transparent background, and [`Sec-CH-DPR`] for high-DPI-optimised variants. Clients MAY retry requests with these headers if indicated by the [`Accept-CH`] response header.
```

extensions/network-icon.md
```diff
@@ -44,3 +52,31 @@ Server using an SVG icon:
...
+    C: GET /icon.png HTTP/1.1
+    C: Host: example.org
+    C: Sec-CH-Prefers-Color-Scheme: "dark"
+
+    S: HTTP/1.1 200 OK
+    S: Content-Type: image/png
+    S: (image optimised for dark backgrounds)
+
+[`Accept`]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Accept
+[HTTP client hints]: https://datatracker.ietf.org/doc/html/rfc8942
```
jwheare: A friendlier client hints link and Accept-CH link.

```suggestion
[`Accept-CH`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Accept-CH)
[HTTP client hints]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/Client_hints
```

extensions/network-icon.md
emersion, line 42: `Sec-CH-Viewport-Height` and `Sec-CH-Width` are better, standard ways to communicate the desired icon size.

sadiepowell: If client developers agree I'd support replacing my `{size}` proposal with this.

jwheare: I prefer size as an explicit way for clients to choose the image dimensions rather than a multi request hint that doesn't even specify the image size. I don't foresee servers being able to adequately predict what image size I'd want based on a viewport size.

emersion: As far as I understand, the viewport size is the available space for the image, not the full client viewport. And `Sec-CH-Width` is an explicit request for an icon of that size.

jwheare: Viewport is for the full window, not just the image.

https://wicg.github.io/responsive-image-client-hints/#sec-ch-viewport-height

It's only defined within web contexts (which a web client fits in) and it's pretty explicit.

> For [fetches](https://fetch.spec.whatwg.org/#concept-fetch) within web contexts, its value SHOULD be the [Window](https://html.spec.whatwg.org/multipage/nav-history-apis.html#window)'s current [innerHeight](https://drafts.csswg.org/cssom-view-1/#dom-window-innerheight).

jwheare: But yes, Sec-CH-Width is more explicit.

emersion: Oh, my bad!

internet-catte: FWIW, clients could probably send these headers on the initial request and it wouldn't harm anything (I think this is for anti-fingerprinting but that's less of a concern for non-web browsing?), but yes the client hint header spec says the server needs to state the hints they accept before the client should send them

internet-catte: if this is for a square icon, `Sec-CH-Width` should be enough :)

PR #600, opened 2026-04-11
message-tags: forbid duplicate tags in CLIENTTAGDENY

emersion:
Main motivation is to disallow the following case:

    CLIENTTAGDENY=foo,-foo

### DIFF
--- a/extensions/message-tags.md
+++ b/extensions/message-tags.md
@@ -228,6 +228,8 @@ A hyphen `-` (0x2D) indicates that a block is negated, i.e. when certain client-
 
 An empty or missing `CLIENTTAGDENY` matches the default case and indicates that all client-only tags are allowed.
 
+Servers MUST NOT advertise the same tag twice in the list, negated or otherwise.
+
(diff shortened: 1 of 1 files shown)


ArsenArsen:
in similar vein, could we forbid `*` anywhere but in the first list item?  i.e. make `CLIENTTAGDENY=-foo,*` and `CLIENTTAGDENY=foo,*` invalid

emersion:
While I'm confident that no existing server sends `CLIENTTAGDENY=foo,-foo`, I wouldn't say the same about `*` being first. Additionally, `CLIENTTAGDENY=foo,-foo` is nonsensical while `*` position would be mainly about simplifying parsing.

ArsenArsen:
IMO it's not more sensible since, for any tag list T, and any tag list without star elements T2, `T,*,T2`is equivalent to `*,T2`

ArsenArsen:
ah, nevermind, that clause exists already; I just missed it

emersion:
Gentle ping: any objections?

progval (approved):

PR #602, opened 2026-04-20
Add draft authtoken spec

skizzerz:
This spec aims to be a replacement for the EXTJWT pull request by specifying a more general mechanism for generating and validating authentication tokens. Unlike EXTJWT, this does not limit the tokens to the JWT format, instead strongly preferring single-use OTPs (although JWTs are still permissible and supported by the spec).

Validation occurs through the IRC protocol as well, to reduce implementation burden on servers (it's just another command, rather than needing to spin up an httpd or other external service).

### DIFF
--- a/extensions/authtoken.md
+++ b/extensions/authtoken.md
@@ -1,0 +1,325 @@
+---
+title: "Authentication Tokens"
+layout: spec
+work-in-progress: true
+copyrights:
+  -
+    name: "Catherine 'whitequark'"
+    period: "2026"
+    email: "whitequark@whitequark.org"
+  -
+    name: "Ryan Schmidt"
+    period: "2026"
+    email: "moonmoon@libera.chat"
+---
+
+## Notes for implementing work-in-progress version
+
+This is a work-in-progress specification.
+
+Software implementing this work-in-progress specification MUST NOT use the unprefixed `authtoken` capability name, `AUTHTOKEN` ISUPPORT name, or `authtoken` batch type. Instead, implementations SHOULD use the `draft/authtoken` capability name, `draft/AUTHTOKEN` ISUPPORT name, and `draft/authtoken` batch type to be interoperable with other software implementing a compatible work-in-progress version.
+
+The final version of the specification will use an unprefixed capability name, ISUPPORT name, and batch type.
+
+## Introduction
+
+This specification introduces the ability for clients to request authentication tokens which may be passed to other services, and for those services to be able to validate said tokens and retrieve a list of pre-configured authorization claims for the token.
+
+Previous attempts have been made to specify authentication to external (non-IRC) services, however none of them provide robust security and leave the validation aspect implementation-defined. This specification aims to solve these challenges by defining both the means of generating and validating a token, as well as a workflow that works best with the most secure token type (single-use OTPs) but can be extended for stateless servers to support something else instead (such as JWT). Generated tokens are associated with any number of claims and are tied to one specific pre-defined (by network opers) external service to prevent token confusion attacks wherein a token generated for one service is passed to a different one to exploit differing sets of per-service claims. This is meant to be a framework that other specifications as well as custom per-network services can use to handle authentication and authorization decisions for external services in a well-defined and consistent manner.
+
+This specification makes no assumptions that the external service is run by the same people who run the network vs. a 3rd party. It is also not tied exclusively to HTTP, although generated tokens are certainly usable there via e.g. the Authorization header. Token validation runs through the same IRC endpoint that clients use to avoid the need for networks to deploy additional services specifically for this specification.
+
+## Dependencies
+
+This specification depends on the [`batch`][] capability which MUST be negotiated to use the `TOKEN GENERATE` or `TOKEN VALIDATE` commands. This specification additionally defines the optional `draft/authtoken` capability. The order of capability negotiation is not significant and MUST NOT be enforced.
+
+This specification additionally makes use of the [standard replies][] and [client-initiated batch][] frameworks.
+
+## Capability
+
+To allow for authentication tokens larger than one IRC protocol line, servers MAY support and advertise the `draft/authtoken` capability. If a server does not advertise this capability, it MUST NOT generate any tokens longer than 200 bytes in response to a `TOKEN GENERATE` command.
+
+Servers MUST NOT require that clients negotiate the `draft/authtoken` capability before making use of the `TOKEN` command.
+
+## ISUPPORT token
+
+This specification defines one ISUPPORT token. Servers MUST publish the `draft/AUTHTOKEN` ISUPPORT token to indicate they support this specification. The token does not have any value, but clients MUST NOT reject tokens published with a value, to allow for future revisions to this specification. Servers SHOULD additionally support [extended ISUPPORT][] since token verification can happen before connection registration is complete, so clients can determine the implementation status of this extension before sending any TOKEN commands.
+
+## `draft/authtoken` batch type
+
+The `draft/authtoken` batch type has two parameters: the service key and the URL corresponding to that service key.
+
+When sent by a server, this batch MUST contain only `TOKEN` messages.
+
+A client MUST negotiate the `draft/authtoken` capability before sending client-initiated `draft/authtoken` batches. A client-initiated `draft/authtoken` batch MUST contain only `TOKEN VALIDATE` commands, and these commands MUST omit the optional service and url parameters.
+
+## TOKEN command
+
+This specification defines one new command named TOKEN. Clients do not need to negotiate any capabilities in order to use this command. The command has three subcommands, which are defined as follows:
+
+```
+TOKEN SERVICELIST
+TOKEN GENERATE <service> [<scope>]
+TOKEN VALIDATE [<service> <url>] :<token>
+```
+
+The `TOKEN VALIDATE` subcommand MUST be usable before a client has completed connection registration. Servers MAY choose to allow the use of the other `TOKEN` subcommands before clients complete connection registration as well.
+
+If a client sends an unrecognized subcommand, the server MUST reply with the following `FAIL` message:
+
+```
+FAIL TOKEN UNKNOWN_COMMAND <command> :No such subcommand TOKEN <command>
+```
+
+The text of the final parameter MAY vary wildly between implementations.
+
+### SERVICELIST subcommand
+
+Syntax:
+
+```
+TOKEN SERVICELIST
+```
+
+The `TOKEN SERVICELIST` subcommand provides a list of all recognized service keys as well as their corresponding URLs. The description parameter provides some sort of description about the service.
+
+Each line of output is one service with the following syntax:
+
+```
+NOTE TOKEN SERVICE <service> <url> :<description>
+```
+
+After all lines of output are sent, the server MUST send the following `NOTE` message:
+
+```
+NOTE TOKEN END_OF_LIST :End of service list
+```
+
+The text of the final parameter MAY vary wildly between implementations.
+
+*[[Begin non-normative example--*
+
+A user sends `TOKEN SERVICELIST` to list all services and receives the following output. One line is for the draft FILEHOST service and the other is for a user-defined QDB service.
+
+```
+NOTE TOKEN SERVICE FILEHOST https://upload.example.com :file upload service
+NOTE TOKEN SERVICE QDB https://qdb.example.com :Quote Database
+NOTE TOKEN END_OF_LIST :End of service list
+```
+
+A user sends `TOKEN SERVICELIST` on a network where no services are configured. As such, they receive only the end-of-list marker:
+
+```
+NOTE TOKEN END_OF_LIST foobar :End of service list
+```
+
+*--End non-normative example]]*
+
+### GENERATE subcommand
+
+Syntax:
+
+```
+TOKEN GENERATE <service> [<scope>]
+```
+
+When a client sends the `TOKEN GENERATE` subcommand, the server responds with an opaque authentication token usable to authenticate to the provided external service. The scope parameter indicates the scope (e.g. user or channel) that the token SHOULD be scoped to, if relevant. If no scope is needed for a particular external service, the parameter MAY be omitted. See the "Server implementation considerations" section below for recommendations about security of generated tokens.
+
+Each line of output is a `TOKEN` message with the following syntax:
+
+```
+TOKEN GENERATE <service> <url> :<token>
+```
+
+If the reply contains multiple lines (due to IRC line length limitations), the server MUST batch the reply in a `draft/authtoken` batch. The batch MUST have two parameters: the service and URL. `TOKEN GENERATE` messages inside of the batch SHOULD use asterisks (`*`) for the service and url parameters instead of specifying them again for each batch line. When receiving a batched reply, clients MUST concatenate the token parameters together without any separators to form the complete token. If a server does not support the draft/authtoken capability, it MUST NOT generate tokens that require multiple lines.
+
+Clients MUST NOT assume that the returned token has any particular format. Clients SHOULD re-run `TOKEN GENERATE` each time they need to interact with the external service, as there is no guarantee that a token is usable more than once.
+
+If the server refuses to generate a token for any reason, it MUST send a `FAIL` message to the client indicating the reason a token could not be generated; see the standard replies section below for a list of potential responses. The command in the message is `TOKEN`.
+
+*[[Begin non-normative example--*
+
+Simple example when a client sends `TOKEN GENERATE FILEHOST #chat`:
+
+```
+TOKEN GENERATE FILEHOST https://upload.example.com :54a333f6e0c218382ab5f64c1135f07798d0255e87115c451ba67cc440f5ea84
+```
+
+Example of a token spanning multiple lines when a client sends `TOKEN GENERATE FILEHOST #chat`:
+
+```
+BATCH +12345 draft/authtoken FILEHOST https://upload.example.com
+@batch=12345 TOKEN GENERATE * * :eyJhbGciOiJIUzUxMiIsInR5cCI6IkpXVCJ9.eyJhdWQiOiJodHRwczovL3VwbG9hZC5leGFtc
+@batch=12345 TOKEN GENERATE * * :GxlLmNvbSIsInN1YiI6Im1vb25tb29uIiwiYWNjb3VudCI6Im1vb25tb29uIiwic2VydmljZSI
+@batch=12345 TOKEN GENERATE * * :6IkZJTEVIT1NUIiwidGFyZ2V0IjoiI2NoYXQiLCJtZW1iZXJfb2YiOlsiI2NoYXQiLCIjaGVsc
+@batch=12345 TOKEN GENERATE * * :CIsIiNzZWNyZXQtZnJpZW5kLWNoYW5uZWwiXSwib3BlcmF0b3Jfb2YiOlsiI2NoYXQiLCIjY2h
+@batch=12345 TOKEN GENERATE * * :hdC1vcHMiLCIjaGVscCJdLCJvcGVyIjpmYWxzZSwicXVvdGEiOnsicGVyX3VwbG9hZCI6MTA0O
+@batch=12345 TOKEN GENERATE * * :DU3NjAwLCJ0b3RhbCI6MTYxMDYxMjczNn0sImlhdCI6MTc3NjAxNjU3NCwiZXhwIjoxNzc2MDE
+@batch=12345 TOKEN GENERATE * * :3NDc0fQ.wZi8okF7t1Ghk1jHLFyqhYwRZU4T3zG1B_qcCOz8gesMCbLhntccDoxDUgsV4OKHfQ
+@batch=12345 TOKEN GENERATE * * :3vjB2dpyD2gHyXcE0JJg
+BATCH -12345
+```
+
+*--End non-normative example]]*
+
+### VALIDATE subcommand (unbatched)
+
+Syntax:
+
+```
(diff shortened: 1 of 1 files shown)


whitequark:
Ergochat has a PR implementing `draft/AUTHTOKEN` (https://github.com/ergochat/ergo/pull/2385), I'll let @slingamn comment on it though.

extensions/authtoken.md
```diff
@@ -0,0 +1,325 @@
...
+
+Previous attempts have been made to specify authentication to external (non-IRC) services, however none of them provide robust security and leave the validation aspect implementation-defined. This specification aims to solve these challenges by defining both the means of generating and validating a token, as well as a workflow that works best with the most secure token type (single-use OTPs) but can be extended for stateless servers to support something else instead (such as JWT). Generated tokens are associated with any number of claims and are tied to one specific pre-defined (by network opers) external service to prevent token confusion attacks wherein a token generated for one service is passed to a different one to exploit differing sets of per-service claims. This is meant to be a framework that other specifications as well as custom per-network services can use to handle authentication and authorization decisions for external services in a well-defined and consistent manner.
```
slingamn: Shade for previous specifications ("robust security") seems unwarranted :-) It seems clear enough how this spec differs from previous proposals.

extensions/authtoken.md
```diff
@@ -0,0 +1,325 @@
...
+
+## Dependencies
+
+This specification depends on the [`batch`][] capability which MUST be negotiated to use the `TOKEN GENERATE` or `TOKEN VALIDATE` commands. This specification additionally defines the optional `draft/authtoken` capability. The order of capability negotiation is not significant and MUST NOT be enforced.
+
+This specification additionally makes use of the [standard replies][] and [client-initiated batch][] frameworks.
+
+## Capability
+
+To allow for authentication tokens larger than one IRC protocol line, servers MAY support and advertise the `draft/authtoken` capability. If a server does not advertise this capability, it MUST NOT generate any tokens longer than 200 bytes in response to a `TOKEN GENERATE` command.
```
slingamn: Proposal: negotiating the cap pre-registration should give you the output of `TOKEN SERVICELIST` in the registration burst. (Modern states that the preferred place is between `004 RPL_MYINFO` and the `LUSERS` numerics.) This allows autodiscovery of available services without having to duplicate the information in 005.

extensions/authtoken.md
```diff
@@ -0,0 +1,325 @@
...
+
+The key is a claim key (see below) and the value is the value of that claim. If a claim key is repeated across multiple lines, clients MUST concatenate all values for a given key together without any separators and treat the combined result as the true value for the key.
+
+*[[Begin non-normative example--*
+
+Response to the client sending a single `TOKEN VALIDATE` command. The server produces a leading space in the second line of the member_of claim because the client must concatenate the lines together with no separators.
+
+```
+Client:
+TOKEN VALIDATE FILEHOST https://upload.example.com :54a333f6e0c218382ab5f64c1135f07798d0255e87115c451ba67cc440f5ea84
```
slingamn: Do we really need the URL as a parameter here? The client can validate it from the batch parameter in the reply batch. It may also slightly constrain the lengths of URLs that can be used (since a single-line token can be up to 200 bytes).

extensions/authtoken.md
```diff
@@ -0,0 +1,325 @@
...
+
+### SERVICELIST subcommand
+
+Syntax:
+
+```
+TOKEN SERVICELIST
+```
+
+The `TOKEN SERVICELIST` subcommand provides a list of all recognized service keys as well as their corresponding URLs. The description parameter provides some sort of description about the service.
```
slingamn: This is the outlier in that (1) it uses `NOTE` instead of `TOKEN` (2) the output is not batched. Both of those seem slightly suboptimal to me.

extensions/authtoken.md
```diff
@@ -0,0 +1,325 @@
...
+
+```
+NOTE TOKEN END_OF_LIST :End of service list
+```
+
+The text of the final parameter MAY vary wildly between implementations.
+
+*[[Begin non-normative example--*
+
+A user sends `TOKEN SERVICELIST` to list all services and receives the following output. One line is for the draft FILEHOST service and the other is for a user-defined QDB service.
```
slingamn: From discussion in #ircv3: service names are machine-readable data that SHOULD be in a registry, or in a vendor namespace. (Pedantically, there is no grammar for service names (although we implicitly assume they can be sent as non-final parameters) and no explicit mention of whether they are case-sensitive.)

extensions/authtoken.md
```diff
@@ -0,0 +1,325 @@
...
+
+If a service chooses to validate tokens itself rather than using `TOKEN VALIDATE`, it should keep all of the following in mind to prevent security issues:
+
+- The service must validate that the token is intended for that service, to avoid token confusion attacks. Using URL-based validation (i.e. checking that one of the claims of the token matches the URL of the service) is a good way to do this. For JWT, this would be the "aud" claim.
+- The service must validate that the token came from the IRC server it is expecting tokens to come from. For JWT, this would be the "iss" claim.
+- The service must validate that the token is signed by the IRC server with a secure algorithm and that the signature is valid. The signing keys (recommended) or secrets (not recommended) should be pre-shared out of band.
```
slingamn: Is "secrets" intended to refer to symmetric signing keys used by, e.g. HS256? Discouraging this seems unwarranted to me.

extensions/authtoken.md
```diff
@@ -0,0 +1,325 @@
...
+
+## External service implementation considerations
+
+This section is non-normative.
+
+This specification defines a validation flow wherein the external service receives a token via some means from a client, connects to an IRC port, and issues a `TOKEN VALIDATE` command to transform that token into a list of claims. This is robust and works with all token types potentially generated by `TOKEN GENERATE`, and allows the external service to delegate token validity checking to the IRC server (allowing for token revocation to occur and avoiding other potential causes of security issues).
+
+If a service chooses to validate tokens itself rather than using `TOKEN VALIDATE`, it should keep all of the following in mind to prevent security issues:
+
+- The service must validate that the token is intended for that service, to avoid token confusion attacks. Using URL-based validation (i.e. checking that one of the claims of the token matches the URL of the service) is a good way to do this. For JWT, this would be the "aud" claim.
```
slingamn, line 290: A straightforward hardening recommendation (and one highly compatible with this spec) is to use a different key for each service. This immediately eliminates most token confusion attacks and facilitates key rotation.

extensions/authtoken.md
```diff
@@ -0,0 +1,325 @@
...
+BATCH +12345 draft/authtoken FILEHOST https://upload.example.com
+@batch=12345 TOKEN CLAIM account :moonmoon
+@batch=12345 TOKEN CLAIM member_of :#chat #help #channel1 #channel2 #channel3
+@batch=12345 TOKEN CLAIM member_of : #channel4 #channel5
+@batch=12345 TOKEN CLAIM operator_of :#chat #chat-ops
+@batch=12345 TOKEN CLAIM scope :#chat
+BATCH -12345
+```
+
+### VALIDATE subcommand (batched)
```
slingamn, line 218: Ingesting C2S-batched VALIDATE was the hardest implementation step for me. I'm not sure there's a better option overall, but I wanted to air this issue: C2S batches might seem like they simplify server ingestion of multiline requests, but I don't think they do, because each type of C2S batch will have its own application-specific constraints, length limits, and abuse risks, so you can't really write a generic "ingest C2S batch" handler. (For example, in the Ergo implementation, we turn off fakelag when a `draft/multiline` batch is in progress, so we need to make sure we turn it back on on every fail path that ends the batch.)

slingamn (commented):
I updated my PR implementation. Overall I really like this spec!

extensions/authtoken.md
slingamn: Another pain point for me was splitting input validation and permissions checking logic between the `BATCH` and `VALIDATE` handlers.

This would actually be cleaner for me if it were just:

```
VALIDATE TOKEN <servicename> <single-line token under 300 bytes>
```

or else

```
VALIDATE TOKEN <servicename> <multi-line token chunk exactly 300 bytes>
[repeat zero or more times]
VALIDATE TOKEN <servicename> <final token chunk under 300 bytes, possibly 0 bytes>
```

I invite comment as to whether this is cleaner in general, or just an idiosyncratic constraint imposed by my implementation :-) I would note in passing that `VALIDATE TOKEN` is not intended for use in "normal" IRC traffic, so the full extensibility of `batch` (e.g. allowing nested batches, allowing interspersed messages that are not part of the batch) is not directly relevant.

skizzerz: My solanum C2S batch implementation involves a generic handler that works across every registered batch type to collect the batched lines together and fire off whatever function is supposed to handle the batch once the entire thing is received, so your inability to do that is certainly an ergo limitation rather than an inherent limitation of C2S batches (this is just an observation; I'm not saying that your argument lacks merit). For reference/comparison, solanum doesn't turn off fakelag for C2S batches. It might take a while to send it but once fully sent it gets burst out all at once.

I originally had it worded to be like AUTHENTICATE where each intermediate line was a fixed byte count and the final line was the leftovers (or blank), however I felt that C2S batches were a more elegant solution that requires fewer bespoke per-command multiline handlers on the client and server side -- AUTHENTICATE is 400 bytes, this could be 400 or might be 300, something else may choose some other limit, etc. I'd also like to drive more adoption of C2S batches in general as a useful framing tool for things that take up more than a line.

extensions/authtoken.md
skizzerz: Yeah I was debating whether to require batch for this or not, thinking that some user may want to run it manually even if their client lacks batch support. Overall that's probably not a good enough reason to make the shape of this command be vastly different from all of the other ones.

extensions/authtoken.md
skizzerz: Reason for the URL here was to prevent token confusion attacks. Say an ircd is migrating from an old filehost to a new one; once the migration happens, we don't want users to be able to upload to the old filehost anymore. If all that was passed was the service key, we have no real means of preventing this without trusting the client. If the service also passes its own URL, then the server can recognize that it's not meant to be interacting with that service and reject the validate attempt. I think it's important to keep, which is partly why the byte limit for single-line tokens is 200 in order to give more space for the URL.

Edit: your suggestion to recommend a different key per service is another way to avoid the above scenario while not requiring a URL parameter.

extensions/authtoken.md
skizzerz: Yes, this section was going over what claims should be present in tokens that are non-interactively validated (i.e. the service is *not* using TOKEN VALIDATE). In such an event, HS256 or similar symmetric algorithms means that a secret key is being shared between multiple parties, since the specification makes no assumption that the ircd and external service are being run by the same operators. This in turn means that the external service could spoof its own properly-signed tokens without the ircd's input, for whatever that might allow it to do (in theory not much, but the risk exists that there are now multiple parties able to issue valid signed tokens).

If the ircd is intending for interactive validation only (i.e. external services must use TOKEN VALIDATE to get the claims list), then HS256 is perfectly appropriate because the secret key only lives in one place: on the ircd. Doing public key in such scenarios would be overkill and unnecessary.

slingamn: Makes sense. I was thinking that if you're doing one key per service, there would be no real disadvantage to using a symmetric secret key. But actually, with a symmetric key, the service would be able to sign its own tokens and then submit them for online validation, possibly learning some user data via `TOKEN CLAIM` that it might not normally be able to access. (If alice uses QDB and but bob never does, QDB would have normally have access to alice's `TOKEN CLAIM` outputs, but not bob's.) Probably not a big deal, but a real consideration.

extensions/authtoken.md
slingamn: I'm not totally clear on whether the threat model here is accident or fraud, so to speak. If it's fraud, the new service's URL is not actually a secret, so nothing prevents the old service from passing it instead of its own URL. If it's accident, the service can guard against this itself by validating the URL returned in the batch parameters (the same way JWT consumers normally validate the `aud` claim).

Rotating the key would work too if the tokens are JWTs signed with a unique per-service key, or MAC'ed UUIDs with a key held only by the ircd. But I'm not sure that's even relevant given what we're trying to prevent.

It seems like maybe the real concern here is usability, not security, in that we could use some way to push updated service lists to clients?

furudean: I wonder if a CAP to subscribe to updates like `authtoken-notify` would be useful to combat the usability problem. If negotiated, the ircd could output the `TOKEN SERVICELIST` on changes or any other relevant notify to tell the client that things have changed.

Reconnection being the only way to receive external service updates seems undesirable. That encourages polling, which is just inefficient overall. Your client could track the URL parameter, but I don't like that as much as having an implied mechanism for it.

extensions/authtoken.md
slingamn: Hmm, from the n=2 sample I have, I'm not convinced that C2S batching is a savings on implementation complexity. From my point of view, the server implementation still needs (1) custom validation rules, per batch type, about what can or can't be in the batch (2) custom length limits (which also have to be batch-type-aware). And then there generally seems to be a conservation effect, where any validation or limiting logic you try to make generic at the ingestion stage has to be made up later. (For example, you could try to implement generic length limits based on total byte size of the messages being queued, but you have to apply the real length limit later after doing whatever application-specific aggregation is required.)

One of the things that makes C2S batching compelling for `draft/multiline` is the interaction with `labeled-response` and `echo-message` (the client labels the initial `BATCH` line, then gets a labeled echo message after the whole batch is complete). What are some of the other use cases you're thinking about for C2S batching?

extensions/authtoken.md
slingamn: This suggests another hardening recommendation for JWTs: if interactive validation is not needed, it can be administratively disabled. And then there's really no problem with HS256 with unique keys: the service can forge whatever token it wants, but it's the only system that will accept the token.

If interactive validation is required, I think a unique asymmetric keypair per service should work (the ircd has the private key, the service has the public key, the general public has nothing).

extensions/authtoken.md
skizzerz: > What are some of the other use cases you're thinking about for C2S batching?

Rich content (aka "embeds" in discord or "blocks" in slack) is the main one I have in mind. Basically, a message that contains some title, content, optional image, optional buttons along the bottom, etc. I don't have anything written down for that yet but C2S batch would be the natural way of sending that and collecting all of those disparate things together as a single message.

extensions/authtoken.md
0x5c: > This in turn means that the external service could spoof its own properly-signed tokens without the ircd's input

Is it realistic to expect there would be services that are hosted by a potentially unstrusted 3rd-party and yet require the IRCd as authorisation provider? That's a weird combo if the IRCd isn't a generic auth provider, and external services instead have to be manually added to the `SERVICELIST` by network operators.

What is the intent of this spec? Allowing networks to host additional services with SSO but without HTTP-dependent standards like OIDC, or is it turning networks into generic decentralised authentication providers, like a IRC-based [Open-ID](https://en.wikipedia.org/wiki/OpenID) clone?

skizzerz: The intent is for federated authentication and authorization with allowlisted services, which may or may not be run by 3rd parties. Partnering with a 3rd party for some service doesn't imply a full-trust relationship, and it makes sense to limit the blast radius of potential security issues as much as possible in the event that the 3rd party is compromised. HTTP-based mechanisms are not specified because IRC is not HTTP, and there are likely going to be non-HTTP-based services that wish to integrate with IRC for authn/authz decisions in this manner.

By specifying both the token generation and token validation layers, these services do not need to ship a validation implementation on a per-ircd basis. This increases interoperability by allowing network operators to choose services based on criteria other than "do they support the specific magic incantations needed to talk to the ircd I'm running"

As a note, while JWTs are popular (which is why there is a lot of space devoted to talking about them), the spec recommends single-use randomly generated OTPs and interactive validation. These are significantly more secure, more compact, and much easier to implement. The JWT discussion is primarily around how to avoid the numerous footguns that the JWT specification comes with and have resulted in security vulnerabilities in other services, without the implementer needing to be an expert in JWT security and/or having done all of that prior research. There are a _lot_ of really bad JWT implementations out there and I don't want this spec to be another one in that pile.

extensions/authtoken.md
jwheare: The other way of doing that is just a client tag. e.g. https://github.com/ircv3/ircv3-ideas/issues/22 although the fallback mentioned there could potentially be batched along with it.

slingamn: I guess also if we wanted to go over 4 KB we would have to batch it?

extensions/authtoken.md
```diff
@@ -0,0 +1,324 @@
...
+BATCH +a draft/authtoken * *
+@batch=a TOKEN SERVICE FILEHOST https://upload.example.com :file upload service
+@batch=a TOKEN SERVICE example.com/QDB https://qdb.example.com :Quote Database
+BATCH -a
+```
+
+A user sends `TOKEN SERVICELIST` on a network where no services are configured. As such, they receive a note that no services are defined.
+
+```
+NOTE TOKEN NO_SERVICES :No services are defined for this network.
```
slingamn, line 126: Would `TOKEN NO_SERVICES :No services are defined` (i.e. `NO_SERVICES` as a new subcommand of `TOKEN`) make it easier for clients to route this message to a handler that clears the service list?

ValwareIRC: It wouldn't be easier necessarily, but I agree that a `NOTE` might not be the right reply, since `NOTE` seems to generally be informational, but I do feel that using a Standard Reply here is right, perhaps `WARN` instead? Slightly more actionable than a `NOTE` in terms of client handling

extensions/authtoken.md
slingamn: If the threat model is "the old filehost has turned evil, and is trying to impersonate the user to the new filehost", that seems generally impossible to prevent in this design (with or without interactive validation), since it's equivalent to MITM of the bearer token.

extensions/authtoken.md
Celelibi: Is it really the right spec to list services?
Wouldn't it be cleaner to have a separate spec for discovering external services? It could also list services that do not require a token.

whitequark: This is necessary for a client to discover service capabilities.

slingamn: AIUI this is intended as the unique discovery mechanism for services that use tokens. Services that do not use tokens can be discovered via 005 (as with `draft/filehost` for example).

extensions/authtoken.md
skizzerz: Been thinking about this more and I can't think of scenarios where the additional URL validation would provide a meaningful benefit above and beyond validating the service key. As such, I will remove it in a subsequent commit.

The intent behind the requirement was not to protect against evil/malicious services. It was to provide conforming services another layer of defense by ensuring that the service was positively validated to be the expected one. However, given the current spec's restraints of one service URL per service key, this is not necessary at the moment. If we ever expand to allow for multiple service URLs per service key, then we would need this level of validation as the service key alone would not be sufficient for audience validation. If/when that happens we'll need to look into adding this or something similar back in.

extensions/authtoken.md
skizzerz: From a client handling perspective, I would imagine that they would clear the service list as part of sending the `TOKEN SERVICELIST` command, and therefore by not getting any batch as a response the list simply remains empty. This is further made easier if the server supports the `draft/authtoken` CAP as then the client will receive `TOKEN NEW` and `TOKEN DEL` notifications as the service list changes provided it enables that CAP.

The NOTE is purely informational and is the appropriate standard-reply (compared to WARN/FAIL) because a network having zero services isn't really an error condition (which is what WARN/FAIL both imply -- warn for non-fatal errors and FAIL for fatal errors).

If it would help client implementations, I could see us doing an empty batch instead or in addition to the NOTE, but I'd like client authors to chime in if that's the case.

extensions/authtoken.md
```diff
@@ -0,0 +1,334 @@
...
+
+When a client sends the `TOKEN GENERATE` subcommand, the server responds with an opaque authentication token usable to authenticate to the provided external service. The scope parameter indicates the scope (e.g. user or channel) that the token SHOULD be scoped to, if relevant. If no scope is needed for a particular external service, the parameter MAY be omitted. See the "Server implementation considerations" section below for recommendations about security of generated tokens.
+
+Each line of output is a `TOKEN` message with the following syntax:
+
+```
+TOKEN GENERATE <service> :<token>
+```
+
+If the reply contains multiple lines (due to IRC line length limitations), the server MUST batch the reply in a `draft/authtoken` batch. The batch MUST have the service key as its first parameter. `TOKEN GENERATE` messages inside of the batch SHOULD use asterisks (`*`) for the service parameter instead of specifying them again for each batch line. When receiving a batched reply, clients MUST concatenate the token parameters together without any separators to form the complete token.
```
thatcher-gaming, line 147: Would it not make more sense to always send the token in a batch, regardless of size? I don't really see the benefit in having two different methods for sending the same type of response, and combining them would make the implementation on the client side much simpler.

skizzerz: @thatcher-gaming Requiring client-initiated batches is a major server-side implementation burden and would significantly hamper adoption of this spec; the vast majority of ircds do not support them. The way it is worded now, these are optional and their support is advertised via the CAP value. Servers that only generate short tokens have no need to implement client-initiated batches and that unblocks any specs dependent on authtoken without requiring additional burden.

thatcher-gaming: Sure, but this is a server-initiated batch. It's entirely reasonable for a client to receive a token as a single-message batch, and then pass it to another application that verifies it using a non-batched message.

PR #605, opened 2026-04-29
METADATA CLEAR should send RPL_KEYNOTSET

slingamn:
From discussion with @progval ; this is easier for clients to process (unlike `RPL_KEYVALUE`, which out of context means that the key *is* set), and there's no real need to echo the cleared values to the client.


jwheare:
Agreed, the current behaviour is impossible to process without label response and tracking what command you sent.

jwheare:
This would need an errata though.

ValwareIRC (approved):

PR #606, opened 2026-04-29
Missing specific metadata erratas

jwheare:
This covers changes since the previous PR #339 that was never merged but was implemented for a long time.

Also clarifies RPL_KEYNOTSET usage.

See also #605


jwheare:
The 766 change is a bit unfortunate, but I can see why it was done so I don't have a strong suggestion to revert it.

However, it's a bit confusing that `RPL_KEYNOTSET` and `FAIL METADATA KEY_NOT_SET` mean different things and need to be interpreted differently. Maybe worth renaming the FAIL code?

slingamn (approved):

PR #607, opened 2026-05-08
Add the draft/oper-role tag to the oper-tag specification.

sadiepowell:
As mentioned in previous issues/purrs some networks may wish to expose the role (oper type in InspIRCd, oper class in UnrealIRCd et al) of an IRC operator as well as (or instead of) their account name.

Support for this is currently available on the InspIRCd test network.

### DIFF
--- a/extensions/oper-tag.md
+++ b/extensions/oper-tag.md
@@ -13,10 +13,7 @@ copyrights:
 
 This is a work-in-progress specification.
 
-Software implementing this work-in-progress specification MUST NOT use the
-unprefixed `oper` tag name. Instead, implementations SHOULD use the
-`draft/oper` tag name to be interoperable with other software
-implementing a compatible work-in-progress version.
+Software implementing this work-in-progress specification MUST NOT use the unprefixed `oper` tag name. Instead, implementations SHOULD use the `draft/oper` tag name to be interoperable with other software implementing a compatible work-in-progress version.
 
 ## Description
 
@@ -24,8 +21,7 @@ The `draft/oper-tag` capability causes the server to add a [message tag][] to me
 
 The tag MUST be named `draft/oper`. The value of the tag, if specified, MUST identify the operator.
 
-The tag MUST be added by the ircd to all commands sent by a user (e.g. PRIVMSG,
-MODE, NOTICE, and all others) and SHOULD be added to any numeric replies sent on behalf of the user.
+The tag MUST be added by the ircd to all commands sent by a user (e.g. PRIVMSG, MODE, NOTICE, and all others) and SHOULD be added to any numeric replies sent on behalf of the user.
 
 Servers supporting this capability MAY be configured to restrict visibility of this tag or its value.
 
--- a/extensions/oper-tag.md
+++ b/extensions/oper-tag.md
@@ -7,6 +7,10 @@ copyrights:
     name: "David Schultz"
     period: "2022"
     email: "me@zpld.me"
+  -
+    name: "Sadie Powell"
+    period: "2026"
+    email: "sadie@sadiepowell.dev"
 ---
 
 ## Notes for implementing work-in-progress version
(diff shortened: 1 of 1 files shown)


sadiepowell:
Feedback on this would be appreciated.

extensions/oper-tag.md
```diff
@@ -7,34 +7,35 @@ copyrights:
...
 
-The tag MUST be named `draft/oper`. The value of the tag, if specified, MUST identify the operator.
+The `draft/oper` tag marks a user as an IRC operator. The value of this tag, if specified, MUST uniquely identify the operator.
 
-The tag MUST be added by the ircd to all commands sent by a user (e.g. PRIVMSG,
-MODE, NOTICE, and all others) and SHOULD be added to any numeric replies sent on behalf of the user.
+The `draft/oper-role` tag shows the role that a IRC operator has on a network (e.g. NetAdmin). This tag MUST only be sent on messages that have the `draft/oper`tag.
 
-Servers supporting this capability MAY be configured to restrict visibility of this tag or its value.
+Servers supporting this capability MAY be configured to restrict visibility of the tags or their values. For users who can see the tags they MUST be added by the IRC server to all commands sent by a user (e.g. `PRIVMSG`, `MODE`, `NOTICE`, etc) and SHOULD be added to any numeric replies sent on behalf of the user (e.g. `RPL_WHOSPCRPL`).
```
skizzerz, line 30: "numeric replies sent on behalf of the user" is a bit vague. This could perhaps be expanded by including a numeric example in the section below, or the wording could be reworked for clarity.

ValwareIRC: while I don't currently see any vagueness in the meaning of "numeric replies sent on behalf of the user", I'm eager to see this enhancement pushed forward, so... bump!

ValwareIRC:
thinking for a bit on this, I think "role" sounds (possibly) too formal like as if a standardized value is expected. my only concern is that someone may think that they can hardcode the values they can see while adding it to their bot or whatever, and expect that it works on other servers. that said, I can't think of a more suitable name, so I wonder if you'd be okay to add some wording that explicitly mentions that any `oper-role` tag value SHOULD/MUST NOT be considered with consistent framing over different networks?

sadiepowell:
"Role" was chosen here as a neutral term because different IRCds use different names for this feature (classes, types, etc).

I'm fine with specifying that the value doesn't have any technical meaning across-networks.

extensions/oper-tag.md
```diff
@@ -7,34 +7,35 @@ copyrights:
...
-implementing a compatible work-in-progress version.
+Software implementing this work-in-progress specification MUST NOT use the unprefixed `oper` tag name. Instead, implementations SHOULD use the `draft/oper` tag name to be interoperable with other software implementing a compatible work-in-progress version.
 
 ## Description
 
-The `draft/oper-tag` capability causes the server to add a [message tag][] to messages sent by a user who is currently an IRC operator.
+The `draft/oper-tag` capability causes the server to add the `draft/oper` and `draft/oper-role` [message tags][] to messages sent by a user who is currently an IRC operator.
 
-The tag MUST be named `draft/oper`. The value of the tag, if specified, MUST identify the operator.
+The `draft/oper` tag marks a user as an IRC operator. The value of this tag, if specified, MUST uniquely identify the operator.
```
skizzerz: The new unique constraint is largely impossible to implement. Say I have two operators who /oper using the exact same oper block for whatever reason. How do I distinguish them?

sadiepowell: If two users are logged into the same account then that's the same operator.

skizzerz: In that case, I'm not sure what benefit specifying "uniquely" here has, if we're allowing multiple client connections to have the same tag value. That doesn't sound particularly unique to me, and seemingly has the same end result of omitting the word (where multiple clients can share the same tag value).

skizzerz:
```suggestion
The `draft/oper` tag marks a user as an IRC operator. The value of this tag, if specified, MUST be the operator's name (e.g. from the `<name>` parameter of the `OPER <name> <password>` command).
```

PR #610, opened 2026-05-11
2-Factor Authentication spec

ValwareIRC:
Implemented and in use by [ObsidianIRC](https://github.com/ObsidianIRC/ObsidianIRC) and [ObbyIRCd](https://github.com/ObsidianIRC/ObbyIRCd)

[vidya-2026-05-04_05.30.58.webm](https://github.com/user-attachments/assets/b7009236-c65d-4cd7-a408-9651a74af469)

### DIFF
--- a/extensions/account-2fa.md
+++ b/extensions/account-2fa.md
@@ -1,0 +1,548 @@
+---
+title: Account 2-Factor Auth Extension
+layout: spec
+work-in-progress: true
+extends:
+  - sasl-3.1
+  - sasl-3.2
+  - account-registration
+copyrights:
+  -
+    name: "Valerie Pond"
+    period: "2026"
+    email: "v.a.pond@outlook.com"
+---
+
+## Notes for implementing work-in-progress version
+
+This is a work-in-progress specification.
+
+Software implementing this work-in-progress specification MUST NOT use the
+unprefixed `account-2fa` capability name. Instead, implementations SHOULD
+use the `draft/account-2fa` capability name to be interoperable with other
+software implementing a compatible work-in-progress version.
+
+The final version of the specification will use an unprefixed capability
+name.
+
+---
+
+## Introduction
+
+This specification defines a standardised mechanism for managing two-factor
+authentication (2FA) on IRC accounts. It covers:
+
+1. **Registration** of a second-factor credential with an account.  Two
+   credential types are defined:
+   * `totp` - RFC 6238 Time-Based One-Time Passwords, supported by the
+     vast majority of authenticator apps.
+   * `external` - a TLS client-certificate fingerprint, used as the
+     second factor via the standard
+     [`EXTERNAL`](https://datatracker.ietf.org/doc/html/rfc4422#appendix-A)
+     SASL mechanism.
+2. **Listing** and **removal** of registered second-factor credentials.
+3. **Enforcement** -- how servers signal that a second-factor SASL step is
+   required and how that step is carried out in a normal login flow.
+
+The command and reply structure are extensible so additional credential
+types may be defined by future specifications.
+
+## Motivation
+
+IRC networks increasingly want to offer accounts the same level of security
+as modern web services.  Historically, 2FA on IRC has been implemented in
+ad-hoc, network-specific ways (e.g. services bot challenges, custom
+numerics).  This specification provides a consistent interface that allows
+clients to offer integrated 2FA management and authentication flows.
+
+## Architecture
+
+### Dependencies
+
+This specification uses the [standard replies](standard-replies.html)
+framework.
+
+Authentication enforcement described in this specification builds on
+[IRCv3.1 SASL Authentication](sasl-3.1.html), as updated by
+[IRCv3.2 SASL Authentication](sasl-3.2.html).
+
+### Capability
+
+This specification adds the `draft/account-2fa` capability.
+
+Servers advertising this capability accept the `2FA` command described
+below.
+
+The capability MUST be advertised with a comma-separated list of supported
+credential types as its CAP value, for example:
+
+```
+CAP * LS :draft/account-2fa=totp,external
+```
+
+A capability with an empty value MUST be treated by clients as advertising
+no supported types (servers SHOULD always include at least one type).
+
+### Identifiers
+
+`<id>` values returned by `2FA LIST` and consumed by `2FA REMOVE` are
+opaque ASCII strings, MUST contain only `A-Z a-z 0-9 - _`, MUST be at
+most 32 characters, MUST NOT contain whitespace, and MUST be unique per
+account.
+
+`<name>` values supplied by clients to `2FA ADD` are user-chosen labels.
+They MUST contain only printable non-whitespace characters and MUST be at
+most 64 bytes.  Clients that wish to display a name with embedded
+whitespace (e.g. "MacBook Touch ID") SHOULD substitute an underscore or
+hyphen (e.g. `MacBook_Touch_ID`).
+
+### Commands
+
+The `2FA` command SHALL fail with `FAIL 2FA NOT_AUTHENTICATED` when
+issued by a client that has not completed a SASL exchange.
+
+#### `2FA STATUS`
+
+    2FA STATUS
+
+Reports whether 2FA enforcement is currently enabled on the authenticated
+account.  Replies with one of:
+
+    NOTE 2FA ENABLED  :<description>
+    NOTE 2FA DISABLED :<description>
+
+#### `2FA LIST`
+
+    2FA LIST
+
+Returns the list of registered credentials for the authenticated account,
+one per server reply.  See [Responses](#responses) for the reply format.
+
+If no credentials are registered the server MUST reply with:
+
+    NOTE 2FA NO_CREDENTIALS :<description>
+
+#### `2FA CHALLENGE`
+
+    2FA CHALLENGE <type>
+
+Requests an enrolment challenge for credential type `<type>`.  The server
+replies with `NOTE 2FA REGISTRATION_CHALLENGE` (see
+[Responses](#responses)).  The challenge SHOULD be valid for at most 120
+seconds and SHOULD be invalidated after a successful `2FA ADD` of the
+same type.
+
+#### `2FA ADD`
+
+    2FA ADD <type> <name> <data>
+
+Registers a new second-factor credential of `<type>` for the authenticated
+account.  `<data>` is a type-specific payload as defined in
+[Credential Types](#credential-types).
+
+#### `2FA REMOVE`
+
+    2FA REMOVE <id>
+
+Removes the registered credential identified by `<id>`.  When 2FA
+enforcement is enabled and the credential being removed is the last one
+registered, the server MUST reply with
+`FAIL 2FA REMOVE_LAST_CREDENTIAL`; the client must first call `2FA
+DISABLE`.
+
+#### `2FA ENABLE`
+
+    2FA ENABLE
+
+Enables 2FA enforcement on the account.  The server MUST reply with
+`FAIL 2FA NO_CREDENTIALS` if no credentials are registered.  Once enabled,
+every future SASL login MUST go through the
+[Two-Factor SASL Flow](#two-factor-sasl-flow).
+
+#### `2FA DISABLE`
+
+    2FA DISABLE <type> <data>
+
+Disables 2FA enforcement on the account.  `<type>` MUST be one of the
+credential types currently registered for the account, and `<data>` MUST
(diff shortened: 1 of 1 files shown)


extensions/account-2fa.md
```diff
@@ -0,0 +1,548 @@
...
+
+Authentication enforcement described in this specification builds on
+[IRCv3.1 SASL Authentication](sasl-3.1.html), as updated by
+[IRCv3.2 SASL Authentication](sasl-3.2.html).
+
+### Capability
+
+This specification adds the `draft/account-2fa` capability.
+
+Servers advertising this capability accept the `2FA` command described
```
skizzerz, line 73: IRC grammar does not allow numbers inside of command names (with exception of numerics, which must be exactly 3 numbers). You need to pick a different command name. I suggest `MFA` since this could in theory be used for more than 2 factors as well with the introduction of biometric authentication types (e.g. passkeys).

Other aspects of this spec should be similarly edited to update "2fa" with the other chosen string for consistency.

extensions/account-2fa.md
```diff
@@ -0,0 +1,548 @@
...
+```
+
+A capability with an empty value MUST be treated by clients as advertising
+no supported types (servers SHOULD always include at least one type).
+
+### Identifiers
+
+`<id>` values returned by `2FA LIST` and consumed by `2FA REMOVE` are
+opaque ASCII strings, MUST contain only `A-Z a-z 0-9 - _`, MUST be at
+most 32 characters, MUST NOT contain whitespace, and MUST be unique per
```
skizzerz, line 90: "MUST NOT contain whitespace" is superfluous as that is already forbidden by the specified set of allowed characters.

extensions/account-2fa.md
```diff
@@ -0,0 +1,548 @@
...
+account.
+
+`<name>` values supplied by clients to `2FA ADD` are user-chosen labels.
+They MUST contain only printable non-whitespace characters and MUST be at
```
skizzerz, line 94: "printable" is vague and there is no solid definition of that outside of the ASCII space. Specify precisely which characters are allowed here (or, more easily, which characters are forbidden and allow everything else)

extensions/account-2fa.md
```diff
@@ -0,0 +1,548 @@
...
+
+    2FA DISABLE <type> <data>
+
+Disables 2FA enforcement on the account.  `<type>` MUST be one of the
+credential types currently registered for the account, and `<data>` MUST
+be a valid second-factor proof of that type using the same on-the-wire
+format as the corresponding step-up SASL mechanism (see [Two-Factor SASL
+Flow](#two-factor-sasl-flow)).  In particular:
+
+* For `totp`, `<data>` is the 6-digit one-time password (no Base64).
```
skizzerz, line 172: "No base 64" is superfluous and should be removed.

extensions/account-2fa.md
```diff
@@ -0,0 +1,548 @@
...
+* For `external`, `<data>` is the lowercase hex SHA-256 fingerprint of
+  the client's TLS certificate, exactly as it would be authenticated by
+  the `EXTERNAL` SASL mechanism on the current connection.  The server
+  MUST reject the request unless the connection is TLS-secured and the
+  presented certificate's fingerprint matches `<data>`.
+
+The server MUST reject the request with
+`FAIL 2FA INVALID_CHALLENGE_RESPONSE` if the proof does not verify.  This
+prevents disabling 2FA with only the password.
```
skizzerz, line 181: This is already prevented because if 2-factor auth is enabled, they won't be able to fully authenticate without passing it. This entire section should be reworked tbh as it seems more like a footgun than anything else. A server may choose to not allow a user to remove their final second authentication factor (that is, require 2FA in some fashion). But if a server doesn't choose to allow that, we shouldn't be requiring the user to validate that they have precisely the thing they want to remove.

If you want to have an extra challenge on removal for additional assurance, allow them to authenticate using *any* of their registered secondary factors in order to remove them from the list.

extensions/account-2fa.md
```diff
@@ -0,0 +1,548 @@
...
+An `external` credential is a TLS client-certificate fingerprint
+associated with the account.  Step-up authentication uses the standard
+[`EXTERNAL`](https://datatracker.ietf.org/doc/html/rfc4422#appendix-A)
+SASL mechanism: the server simply checks that the TLS connection
+presented a certificate whose fingerprint matches one of the registered
+credentials.
+
+The fingerprint algorithm is SHA-256, and the fingerprint is encoded as
+a lowercase hex string (no separators) of the DER-encoded certificate.
+Implementations MUST NOT rely on an alternative algorithm or encoding.
```
skizzerz, line 239: Don't specify to this level. Some implementations use sha512, others may use quantum-proof hashes, and others hash the SPKI rather than the cert. Leave the hashing mechanism and data hashed implementation-defined.

Enforcing sha256 is bad and will cause this specification to ossify.

extensions/account-2fa.md
```diff
@@ -0,0 +1,548 @@
...
+
+| Field      | Type    | Description |
+| ---------- | ------- | ----------- |
+| `type`     | string  | MUST be `"totp"`. |
+| `secret`   | string  | Base32-encoded shared secret (≥ 16 bytes of entropy). |
+| `issuer`   | string  | Human-readable issuer (typically the network name). |
+| `account`  | string  | The account name. |
+| `algorithm`| string  | MUST be `"SHA1"`. |
+| `digits`   | integer | MUST be `6`. |
+| `period`   | integer | MUST be `30`. |
```
skizzerz, line 209: Why enforce all these parameters? Every compliant TOTP implementation is able to parse alternative values for these and handle them properly.

extensions/account-2fa.md
```diff
@@ -0,0 +1,548 @@
...
+
+    NOTE 2FA REGISTRATION_CHALLENGE external <challenge-data> :<description>
+
+`<challenge-data>` is a Base64-encoded JSON object with these fields:
+
+| Field         | Type   | Description |
+| ------------- | ------ | ----------- |
+| `type`        | string | MUST be `"external"`. |
+| `algorithm`   | string | MUST be `"sha-256"`. |
+| `fingerprint` | string | The lowercase hex SHA-256 fingerprint of the certificate the client presented on the current TLS connection, or the empty string if no client certificate was presented. |
```
skizzerz, line 254: As mentioned previously, this is pointless and bad. The specific fingerprint algorithm is an implementation detail that should not be exposed to the client. The server already knows the fingerprint it wants to use, so it should just use that by pulling the out-of-band data. There should be no handshake here, the server simply replies with "yep you're enrolled" if a certificate is presented and an error if the client is not providing a certificate.

extensions/account-2fa.md
```diff
@@ -0,0 +1,548 @@
...
+
+    2FA ADD external <name> <fingerprint>
+
+`<fingerprint>` is the lowercase hex SHA-256 fingerprint to register.
+If `<fingerprint>` is the empty string the server MUST reject the
+request with `FAIL 2FA INVALID_CREDENTIAL_DATA`.  Otherwise the server
+MUST persist `<fingerprint>` as a registered credential of type
+`external` and reply with `2FA ADD SUCCESS`.
+
+A server MAY require that `<fingerprint>` matches the fingerprint of
```
skizzerz, line 272: I would remove the capability to enroll a fingerprint that was not provided for the connection (i.e. make this a MUST and remove the ability for the client to pass data), as there is no way to verify that the client actually controls whatever certificate hashes to that fingerprint. Furthermore, given my strong recommendation above that the specific hash method be implementation-defined, there is no reliable way for a client to derive this fingerprint anyway.

Furthermore, this attempt should be rejected if the client performed primary authentication via SASL EXTERNAL, because at that point this isn't enrolling a second factor. They need to demonstrate a different factor before enrolling a certificate as a secondary one.

extensions/account-2fa.md
```diff
@@ -0,0 +1,548 @@
...
+
+    AUTHENTICATE 2FA-REQUIRED
+
+This is a synthetic `AUTHENTICATE` message issued by the server (not the
+client) and is the cue for the client to begin a second-factor SASL
+exchange.  The server MUST NOT send `900` or `903` until the
+second-factor exchange has succeeded.
+
+The client then begins a new SASL exchange using one of the
+second-factor mechanisms below by sending `AUTHENTICATE <mech>` exactly
```
skizzerz, line 353: Use the `2FA CHALLENGE` command for this instead of defining custom SASL types that don't actually perform primary authentication. This is a gross abuse of the SASL specification and is not what it is designed for.

Add a note that if SASL EXTERNAL was used for primary authentication, then external MUST NOT be allowed for secondary authentication.

Add some text for the case where the authn and authz ids differ (i.e. performing impersonation) -- this must still use the authn id's registered MFA methods rather than the authz id's.

extensions/account-2fa.md
```diff
@@ -0,0 +1,548 @@
...
+
+    FAIL 2FA NO_CREDENTIALS :<description>
+    :irc.example.com 904 <nick> :SASL authentication failed
+
+Recovery requires operator intervention to clear the 2FA enforcement
+flag on the account.  Servers MUST NOT silently fall back to
+single-factor login in this state, as that would defeat the purpose of
+2FA when the credential store has been tampered with.
+
+### Step-up SASL Mechanisms
```
skizzerz, line 373: This entire section should be removed, since we shouldn't be using SASL for this. The previous registry can be used for both enrollment and authentication.

extensions/account-2fa.md
```diff
@@ -0,0 +1,548 @@
...
+
+#### `2FA CHALLENGE`
+
+    2FA CHALLENGE <type>
+
+Requests an enrolment challenge for credential type `<type>`.  The server
+replies with `NOTE 2FA REGISTRATION_CHALLENGE` (see
+[Responses](#responses)).  The challenge SHOULD be valid for at most 120
+seconds and SHOULD be invalidated after a successful `2FA ADD` of the
+same type.
```
skizzerz, line 133: I touch more on this below, but this command should be used for actual authentication. For enrollment, just use ADD for the entire process (even if it requires multiple back and forth messages).

extensions/account-2fa.md
```diff
@@ -0,0 +1,548 @@
...
+* Verification of 6-digit OTPs SHOULD use a constant-time comparison.
+* Clients SHOULD display the registered credential list to the user
+  before confirming removal and SHOULD warn when the user is about to
+  remove their last enrolled credential.
+* If 2FA enforcement is enabled but no credentials remain registered,
+  the server MUST lock the account out of SASL login (see [Account With
+  No Registered Credentials](#account-with-no-registered-credentials))
+  rather than degrading to single-factor.  Treating credential loss as
+  "2FA effectively off" would let an attacker who can write to the
+  credential store bypass 2FA.
```
skizzerz, line 547: If an attacker can write to the credential store, they can either disable the 2FA flag entirely or inject their own credentials that they are in possession of, so this isn't a particularly compelling reason. Come up with something better or omit the explanation.

extensions/account-2fa.md
```diff
@@ -0,0 +1,548 @@
...
+
+```
+Client: AUTHENTICATE TOTP
+Server: AUTHENTICATE +
+Client: AUTHENTICATE MDAwMDAw
+Server: :irc.example.com 904 alice :SASL authentication failed
+```
+
+## Security Considerations
+
```
skizzerz, line 520: This section appears to be normative given the proliferation of MUST. I recommend that all of these notes be rolled into the actual specification text instead of living at the bottom of the document. This section should instead be moved into non-normative prose explaining security considerations in a human-understandable manner.

extensions/account-2fa.md
```diff
@@ -0,0 +1,548 @@
...
+authentication (2FA) on IRC accounts. It covers:
+
+1. **Registration** of a second-factor credential with an account.  Two
+   credential types are defined:
+   * `totp` - RFC 6238 Time-Based One-Time Passwords, supported by the
+     vast majority of authenticator apps.
+   * `external` - a TLS client-certificate fingerprint, used as the
+     second factor via the standard
+     [`EXTERNAL`](https://datatracker.ietf.org/doc/html/rfc4422#appendix-A)
+     SASL mechanism.
```
skizzerz, line 42: I would remove this listing from this section as it's just an additional place to (forget to) update as other mechs are added. The introduction section is more for human-readable prose to let people understand why the spec exists and what problems it solves. Diving into implementation details this early mostly serves to raise more questions than it answers.

skizzerz (changes requested):

slingamn:
I don't think IRCv3 is ready for 2FA in the absence of an analogue of cookies (client-local secrets that allow reconnection without user intervention or presence).

PR #612, opened 2026-05-17
draft/extended-isupport: Introduce value concatenation semantics

ValwareIRC:
This goes towards solving problems that exist with ISUPPORT token values that have the potential to overflow the line limit. This came after a conversation on #ircv3 on Libera regarding `CLIENTTAGDENY`. While I will be recommending we take `CLIENTTAGDENY` another direction in another spec addendum, I still think this goes towards solving other problems that may come up in future, like with `draft/FILEHOST` which expects a URL which technically can be thousands of bytes long, and just makes 005 a bit unsuitable for that kind of stuff.

I'd like to both try to solve this with concatenation logic as well as still suggest moving some things out of ISUPPORT altogether in a separate PR, but it seemed this WIP was the right place for this enhancement.

### DIFF
--- a/extensions/extended-isupport.md
+++ b/extensions/extended-isupport.md
@@ -91,8 +91,6 @@ When the `draft/extended-isupport` capability is enabled, the syntax of an
                 /  ( key )                      ; flag form
                 /  ( "-" key )                  ; deletion form
 
-The `+=` delimiter is treated atomically and MUST NOT be split across two
-tokens.
 
 The append form is defined as byte-wise concatenation. The client
 performs no interpretation of either the current value or the appended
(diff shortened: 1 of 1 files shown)


emersion:
This is a breaking change. It cannot be added to the existing spec without breaking clients.

ValwareIRC:
If anyone wants to test, I've implemented the 0.2 on ObbyIRCd (UnrealIRCd fork) at https://github.com/ObsidianIRC/ObbyIRCd/commit/d28ea326908c78c91690350c3633b5dfd85483df with the PR for ObsidianIRC client here:
https://github.com/ObsidianIRC/ObsidianIRC/pull/222

sadiepowell:
I'm happy to see this proposal (with the version bump) but I'm unsure as to how we should handle the fallback for tokens which don't fit?

sadiepowell:
@ValwareIRC Would you be okay with specifying that over-sized tokens MUST NOT be sent to legacy clients and that specs which rely on large tokens SHOULD require this cap?

It should probably be also specified that a token MUST be split in a way that does not create a malformed text in the encoding that the server uses (this will mostly be a problem for UTF-8).

ValwareIRC:
@sadiepowell yes that seems like a good addition, I'll add it in later today

Issue #588, opened 2026-03-10
Some cosmetic and clarity issues with METADATA


Herringway:
A handful of minor problems found while implementing METADATA last night:
- A few typos: "subscriping" instead of "subscribing", "responsds" instead of "responds"
- `METADATA GET`'s syntax implies it is legal to send 0 key parameters, which is inconsistent with `METADATA SUB` and co.
- `METADATA CLEAR`'s description mentions a `KEY_NO_PERMISSION` reply, but the standard replies table below only shows `TARGET_INVALID`
- `TARGET_INVALID` is listed in the same table as a possible response to all current subcommands, but is mentioned in only one of them
- `METADATA SUBS`'s description mentions the response is a batch, but only one of the examples has a BATCH message in it (19 out of the 20 examples are batchless)
- `METADATA SET`'s syntax includes a colon, implying a value with a space in it would be sent as `METADATA target SET key ::value with spaces`
- The "batch types" section doesn't mention that standard replies may be included in the batches

skizzerz:
More issues:

- Many subcommands specify that `FAIL` codes should be returned in batches alongside other replies. This violates the standard-replies specification, which states that `FAIL` is a complete failure to process the command/function. These should be specified as `WARN` instead.
- `METADATA CLEAR` should not return any values. This makes implementation a nightmare for large metadata lists and sending current values is unprecedented with any other specifications/commands for item deletion. Clients should simply assume that if the command didn't fail, everything they can delete got deleted. If they need to know exactly what went away, they can issue a `METADATA SYNC` afterwards.
- `RPL_METADATASUBOK`, `RPL_METADATAUNSUBOK`, and `RPL_METADATASUBS` are missing a `:` in front of the key list. We shouldn't be sending numerics that have a variable number of parameters, and by making each key a separate parameter we suddenly also need to care about the 15-parameter limit specified by RFC 1459. Making the key lists a trailing parameter (which each key separated by space) mitigates all of these issues and follows precedent for all other list-type numerics that give multiple pieces of data in one reply.
- `METADATA * SYNC` should exist to sync all channels/users in one go, to allow for server-side deduplication of user metadata (otherwise impossible if you sync each channel individually). If some server implementations find this too burdensome, we can expose this via an additional key in the CAP value like "syncall" as an optional feature rather than a required one.
- It is unclear/unspecified how we notify subscribed clients that a key was deleted via a `METADATA SET` which lacks a value.

slingamn:
[...]

PR #613, opened 2026-06-06
Update metadata spec

skizzerz:
The errata section at the bottom lists all normative changes. Additionally, the specification was rearranged and copyedited to flow more nicely when read from top-to-bottom. It should (hopefully) be possible to now follow the specification without needing to jump around, and all concepts are (hopefully) introduced before being used in the specification. Some examples were moved inline to help illustrate, although the rather large examples section at the end was maintained in case it was helpful.

Addresses issues and discussion from #588 

## Summary of changes between earlier versions of this PR and the current version

* The `LIMIT_REACHED` FAIL code has gained a new parameter to indicate the subcommand for which the limit was reached, allowing clients to statelessly track whether the limit was reached for SUB vs SET
* The `RATE_LIMITED` FAIL code was removed and the `ERR_METADATARATELIMIT` error numeric (775) was reintroduced to indicate commands are rate-limited; this allows clients to retry the command without needing stateful tracking as the numeric can pass a useful trailing parameter while standard reply trailing parameters must be for human consumption
* `RPL_METADATASYNCLATER` was renamed to `ERR_METADATASYNCLATER`; unlike the changes above this is not a breaking change but rather reflects that the command was not successfully executed

### Standard Replies

* Standard reply codes have been adjusted to use `WARN` and `NOTE` when and where appropriate. `FAIL` is reserved for when the entire command cannot proceed due to an error. `WARN` is used when a portion of a command cannot proceed but other parts may still proceed. `NOTE` is used when a portion of a command is fully successful but needs ancillary data.
* The `SUBCOMMAND_INVALID` standard reply code has been removed. The more generic `INVALID_PARAMS` code should be used in its place.
* The `KEY_NOT_SET` standard reply code has been removed. With its removal, `METADATA SET` is fully idempotent with regards to its replies.
* The `TOO_MANY_SUBS` standard reply code has been removed. The `LIMIT_REACHED` standard reply code should be used in its place.
* The `VALUE_INVALID` standard reply code has been renamed to `INVALID_VALUE` for better consistency between this specification and other specifications. It additionally now takes a `<Key>` parameter.
* The `KEY_INVALID` standard reply code has been renamed to `INVALID_KEY` for better consistency between this specification and other specifications.
* ~~The `RATE_LIMITED` standard reply code is now specified for every subcommand.~~ (now removed)
* The `INVALID_PARAMS` standard reply code has been added for all other errors with `METADATA` command parameters.

### Server Messages / Numerics

* The `METADATA` server message was removed; `761 RPL_KEYVALUE` and `766 RPL_KEYNOTSET` are used in all cases to advertise key values or the lack thereof.
* Servers sending either a `metadata` batch or `774 RPL_METADATASYNCLATER` upon `JOIN`, `730 RPL_MONONLINE`, or when a client subscribes to a new key post-registration is no longer optional.
* The `<RetryAfter>` parameter of `774 RPL_METADATASYNCLATER` and `RATE_LIMITED` is no longer optional and must be a non-negative integer.

### CAP changes

* The CAP token is now `draft/metadata-3`.
* Clients removing either the metadata or batch CAPs from themselves are now automatically unsubscribed from all keys.
* A `max-key-bytes` token was added to the CAP value to limit the maximum number of bytes a metadata key may consume.
* The `before-connect` token in the CAP value list was clarified to not have its own value and that clients must ignore any value it carries.
* CAP value tokens which specify maximum lengths or numbers of entries (i.e. `max-key-bytes`, `max-value-bytes`, `max-keys`, and `max-subs`) are clarified to be unlimited if those tokens are missing from the `CAP LS 302` response.

### Other

* The `*ALL` target has been specified for `METADATA SYNC` to synchronize all visible targets.
* Empty strings are no longer valid metadata values; setting a metadata key to an empty string will now remove it (in addition to removal by omitting the value parameter entirely).

### DIFF
--- a/extensions/metadata.md
+++ b/extensions/metadata.md
@@ -29,16 +29,17 @@ copyrights:
     name: "Val Lorentz"
     period: "2022-2023"
     email: "progval+ircv3@progval.net"
+  -
+    name: "Ryan Schmidt"
+    period: "2026"
+    email: "moonmoon@libera.chat"
 ---
 
 ## Notes for implementing work-in-progress version
 
 This is a work-in-progress specification.
 
-Software implementing this work-in-progress specification MUST NOT use the
-unprefixed `metadata-2` capability name.
-Instead, implementations SHOULD use the `draft/metadata-2` capability name to be interoperable with other
-software implementing a compatible work-in-progress version.
+Software implementing this work-in-progress specification MUST NOT use the unprefixed `metadata` capability name. Instead, implementations SHOULD use the `draft/metadata-2` capability name to be interoperable with other software implementing a compatible work-in-progress version.
 
 The final version of the specification will use an unprefixed capability name.
 
@@ -47,27 +48,27 @@ The final version of the specification will use an unprefixed capability name.
 It is useful to associate metadata with one's IRC presence, e.g. to make one's homepage or non-IRC contact details more discoverable.
 There are several mechanisms for doing this, but they normally rely on the presence of services and aren't really suitable for transient metadata like a user's current location.
 
-This feature lays out a command that can be used to set metadata, and a message that can be used to receive metadata updates from the server.
+This feature lays out a command that can be used to manage metadata associated with clients or channels and a notification framework to receive updates on metadata changes.
 
 ## Concepts
 
 Clients and channels can both contain metadata. Metadata acts as a key:value store (for example, the `display-name` key on a user, or the `url` key on a channel).
 
-Clients can [set/delete](#metadata-set) metadata on themselves, and on channels they administer. Privileged users (e.g. network admins) may be able to set certain metadata on other users, and set special keys on themselves or channels.
+Clients can set or delete metadata on themselves and on channels they administer. Privileged users (e.g. network admins) may be able to set certain metadata on other users, and set special keys on themselves or channels.
 
-Server administrators can setup lists of allowed or blocked keys, and may also restrict the setting/viewing of keys depending on whether the user is an admin, what kind of admin they are, etc (see the `Visibility` field).
+On joining a server, clients have to configure their 'metadata key subscriptions'. This is a list of which keys the client understands and wants to get updates about. For example, they may subscribe to a `status` key if they support user-set statuses. By default, this subscription list is empty. When metadata is updated or removed, users who have subscribed to the changed key will receive messages notifying them of the change.
 
-On joining a server, clients have to configure their 'metadata key [subscriptions](#metadata-sub)'. This is a list of which keys the client understands and wants to get updates about (for example, they may subscribe to a `status` key if they support user-set statuses). By default, this subscription list is empty.
+## Definitions
 
-When metadata is updated, users who have subscribed to the changed key will receive a [`METADATA` message](#metadata-server-message) notifying them of the change.
+*Visible channels* or *visible users* (generally, *visible target*) mean channels or users that a client would receive [metadata notifications](#notifications) about.
 
-For channel metadata, this applies only to subscribers who are members of the updated channel. For user metadata, this applies only to subscribers who either share a channel with the updated user or are [monitoring](https://ircv3.net/specs/extensions/monitor.html#monitor-command) them.
+Servers may additionally restrict the visibility of metadata keys to certain subsets of users (e.g. based on operator status). A *visible key* is a metadata key that the user has permissions to view, according to the server's implementation-defined means.
 
-On joining a channel, users will get the channel's current metadata sent to them with `METADATA` messages, and get the same information for all users who are in the channel. Specifically, they get that information for the keys they are subscribed to. The server may also tell them to request that information [at a later time](#metadata-sync).
+A *metadata-aware client* is a client who has negotiated both the `draft/metadata-2` capability and the `batch` capability.
 
 ## Relation with other specifications
 
-This specification depends on the [`batch`](../extensions/batch.html) capability which MUST be negotiated to use ``draft/metadata-2``. The order of capability negotiation is not significant and MUST NOT be enforced.
+This specification depends on the [`batch`](../extensions/batch.html) capability which MUST be negotiated to use `draft/metadata-2`. The order of capability negotiation is not significant and MUST NOT be enforced.
 
 This specification also uses the [standard replies](../extensions/standard-replies.html) framework.
 
@@ -77,9 +78,9 @@ Clients MUST NOT request both `metadata-notify` and `draft/metadata-2`. Servers 
 
 The `draft/metadata-2` capability indicates that a server supports metadata, and provides any limits and information about the system that clients must be aware of. Clients MUST request this capability in order to receive [`METADATA` notifications](#notifications).
 
-The ABNF format of the `metadata` capability is:
+The ABNF format of the `draft/metadata-2` capability is:
 
-    capability ::= 'metadata' ['=' tokens]
+    capability ::= 'draft/metadata-2' ['=' tokens]
     tokens     ::= token [',' token]*
     token      ::= key ['=' value]
     key        ::= <sequence of one or more a-z0-9_./->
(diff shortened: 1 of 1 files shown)


extensions/metadata.md
```diff
@@ -723,55 +835,51 @@ Twice:
...
+* The `766 RPL_KEYNOTSET` numeric replaces the `766 ERR_NOMATCHINGKEY` numeric in previous versions. It is no longer an error response and has a different meaning.
+* The numerics `770 RPL_METADATASUBOK`, `771 RPL_METADATAUNSUBOK`, and `772 RPL_METADATASUBS` changed from using a space separated final parameter `:<Key 1> [<Key 2> ... [<Key n>]]` to a list of individual parameters `<Key1> [<Key2> ... [<Key n>]]`.
+* Use of batch instead of `762 RPL_METADATAEND` in situations where more than one `761 RPL_KEYVALUE` is sent.
+* Non-standard keys should now use a vendor prefix.
 
 ## Errata
 
-* A previous version of this specification sent `RPL_KEYVALUE` as the response to `METADATA CLEAR`. This was changed to `RPL_KEYNOTSET` to simplify client processsing.
+* A previous version of this specification sent `761 RPL_KEYVALUE` as the response to `METADATA CLEAR`. This was changed to `766 RPL_KEYNOTSET` to simplify client processsing.
+* The `METADATA` server message was removed; `761 RPL_KEYVALUE` and `766 RPL_KEYNOTSET` are used in all cases to advertise key values or the lack thereof.
```
slingamn, line 871: This breaks the longstanding norm (albeit violated by `monitor`) that numerics are for replies and verbs are for asynchronous messages. I don't see a strong reason to uphold the norm at this time, but I wanted to flag it in case anyone else has opinions (I think I remember @syzop commenting on this in the past?)

extensions/metadata.md
```diff
@@ -29,16 +29,17 @@ copyrights:
...
 
 ## Notes for implementing work-in-progress version
 
 This is a work-in-progress specification.
 
-Software implementing this work-in-progress specification MUST NOT use the
-unprefixed `metadata-2` capability name.
-Instead, implementations SHOULD use the `draft/metadata-2` capability name to be interoperable with other
-software implementing a compatible work-in-progress version.
+Software implementing this work-in-progress specification MUST NOT use the unprefixed `metadata` capability name. Instead, implementations SHOULD use the `draft/metadata-2` capability name to be interoperable with other software implementing a compatible work-in-progress version.
```
slingamn: Thoughts on going up to `draft/metadata-3`?

extensions/metadata.md
skizzerz: `extended-isupport` also breaks this norm. Largely, I see no reason why we need two parallel notification mechanisms. Either the spec should be 100% `RPL_KEYVALUE/RPL_KEYNOTSET` or 100% `METADATA` notifications. I chose the numerics because I see value in separating "this key has a value" vs "this key does not exist" and the previous specification wording of `METADATA` notification didn't cover the "does not exist" case at all.

extensions/metadata.md
skizzerz: I feel this is mostly backwards compatible despite the large number of changes: command syntax is identical and numeric formats are identical. The changes in standard reply codes are not backwards compatible, except for `RATE_LIMITED`, which is the only standard reply code that required/suggested the client to automatically act on it. Every other code was informative to alert the user to issues, so changing the code doesn't really change how client software can do this (although it may break GUI notifications looking for previous codes that no longer exist).

Something missing from this list that's in my next set of changes to push up after gathering more feedback is that I made empty strings no longer valid for metadata values, which is another backwards incompatible change. This was originally done before I decided to get rid of the `METADATA` notification in favor of numerics, as a way of specifying that a notification without a value means the key is not set, but I left the language in there after moving to numerics-only because I couldn't think of any good reasons why a piece of metadata MUST have an empty value instead of the client specifying some random byte to just fill something there.

All that being said, if people would rather this be draft/metadata-3 then I can do that. But in existing client implementations things won't break horribly much if we keep this at draft/metadata-2, I think (this needs additional testing but I'm working on fixing some bugs with my server-side implementation of these changes so I won't be able to test clients for a couple weeks).

extensions/metadata.md
```diff
@@ -723,55 +835,51 @@ Twice:
...
+* The `TOO_MANY_SUBS` standard reply code has been removed. The `LIMIT_REACHED` standard reply code should be used in its place.
+* The `VALUE_INVALID` standard reply code has been renamed to `INVALID_VALUE` for better consistency between this specification and other specifications. It additionally now takes a `<Key>` parameter.
+* The `KEY_INVALID` standard reply code has been renamed to `INVALID_KEY` for better consistency between this specification and other specifications.
+* The `RATE_LIMITED` standard reply code is now specified for every subcommand.
+* The `INVALID_PARAMS` standard reply code has been added for all other errors with `METADATA` command parameters.
+* Standard reply codes have been adjusted to use `WARN` and `NOTE` when and where appropriate. `FAIL` is reserved for when the entire command cannot proceed due to an error. `WARN` is used when a portion of a command cannot proceed but other parts may still proceed. `NOTE` is used for ancillary data when the command is fully successful.
+* The `*ALL` target has been specified for `METADATA SYNC` to synchronize all visible targets.
```
skizzerz, line 881: Someone on `#ircv3` brought up that
> *ALL seems weird, I'd proably [*sic*] go for 0 or % or @ (e.g. '%' in IMAP is a wildcard, and '@' in IRC is an invalid nickname)

There wasn't much in-channel discussion although I noted my distaste for `0` since that historically signals "nothing" instead of "everything". I suggested `$*` as another potential option, borrowing from the `$` prefix used in some ircds for "broadcast" messages. Because of this prefix in that context, modern.ircdocs forbids `$` at the start of nicknames, and it is highly unlikely that the IRCv3 WG would standardize a `$` channel prefix as that would collide with the broadcast message prefix on those ircds.

extensions/metadata.md
slingamn: What are the current client implementations?

extensions/metadata.md
slingamn: Ergo sends `METADATA` with no final parameter when a key is unset (although this was not explicit in the specification).

That's a good point about 005. It's actually not just `extended-isupport` that sends it asynchronously, it's Modern:

>If the value of a parameter changes, the server SHOULD re-advertise the parameter with the new value in an RPL_ISUPPORT reply. An example of this is a client becoming an IRC operator and their CHANLIMIT changing.

I think I'm fine with consolidating everything into the numerics.

extensions/metadata.md
slingamn: I personally like the idea of `*` as a non-colliding prefix (perhaps namespace) for special entities / targets, but I'm certainly open to `$*` here if other people have strong feelings.

extensions/metadata.md
jwheare: Halloy, IRCCloud, Obsidian, Goguma

extensions/metadata.md
slingamn: Actually, thoughts on consolidating everything into `METADATA` instead? This would avoid having to echo the user's own nick on everything, as is conventional for numerics.

slingamn: (Never mind, this has slightly more problematic compatibility implications and doesn't seem worth it)

extensions/metadata.md
```diff
@@ -175,224 +262,193 @@ The format of the `METADATA` server message is:
...
 
 ### METADATA SET
 
-    METADATA <Target> SET <Key> [:Value]
-
-This subcommand sets the key on the target to the given value. If no value is given, the key is removed.
+    METADATA <Target> SET <Key> [:<Value>]
 
-If the key is invalid, the server responsds with `FAIL METADATA KEY_INVALID` and fails the request.
+This subcommand sets the key on the target to the given value. If the Value parameter is omitted or is an empty string, the key is removed. If the request is successful, the server carries out the requested change and responds with one `761 RPL_KEYVALUE` event, representing the new value if any, or `766 RPL_KEYNOTSET` if no value is set. This new value MAY differ from the one sent by the client.
```
slingamn, line 307: I get the case for disallowing the empty value, but then what's the case for accepting it here and treating it as a deletion, instead of rejecting it with a `FAIL`? It seems like a bit of a footgun.

skizzerz: I originally was going the other direction: making the value param mandatory instead of having the number of parameters vary based on set vs delete. I ended up keeping the syntax as-is for backwards compatibility because I didn't feel there was a strong enough reason to break it.

The advantage of allowing empty string = delete is that it makes it simpler client-side to delete a key. They can just "stupidly" append the key value to "METADATA foo SET key :" and it'll Just Work(tm). No need to detect deletion client-side and send it on the wire in any special format.

jwheare:
FAIL METADATA LIMIT_REACHED isn't documented in METADATA SET subcommand section.

---

I'm not going to block it but for the record, as mentioned in IRC, I would still prefer TOO_MANY_SUBS retained for SUB and LIMIT_REACHED dedicated for SET.

The overloaded meaning of the `Item` param and the need to include a `Command` param are API smells to me. It's also a bit of a headache to support both this and the previous version for compatibility, and that's somewhat expected for drafts, but seems unnecessary.

I understand the stylistic desire to abstract these two similar seeming error codes, but I disagree that this is an elegant solution.

PR #615, opened 2026-09-13
Point batch types to the IRCv3 registry

vanosg:
batch types in the batch spec currently only list netjoin/split and chathistory. The registry appears to be more complete and maintained. Updating with a link pointing to the single record of truth will prevent the spec from needing to be maintained separately and becoming outdated.


progval:
LGTM but we need to add links from https://ircv3.net/registry#batches to the respective specs first

